PR #12, opened 2022-03-29
merge koord-controller and koor-webhook into koord-manager

eahydra:
Signed-off-by: Tao Li <joseph.t.lee@outlook.com>

Simplified deployment of components


cmd/koord-manager/main.go
```diff
@@ -36,8 +36,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
```
jasonliu747, line 37: I think we need to keep `_ "k8s.io/client-go/plugin/pkg/client/auth"`. Since by default, plugins that use [client-go](https://godoc.org/k8s.io/client-go) cannot authenticate to Kubernetes clusters on many cloud providers.

jasonliu747 (commented):
@hormes WDYT

PR #42, opened 2022-04-06
add some error handlers

Mr-Linus:
Signed-off-by: mr-linus <mrfunky001@gmail.com>

### DIFF
--- a/pkg/koordlet/koordlet.go
+++ b/pkg/koordlet/koordlet.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
 	clientsetbeta1 "github.com/koordinator-sh/koordinator/pkg/client/clientset/versioned"
@@ -145,12 +144,6 @@ func (d *daemon) Run(stopCh <-chan struct{}) {
 			os.Exit(1)
 		}
 	}()
-
-	// wait informer synced
-	if cache.WaitForCacheSync(stopCh, d.statesInformer.HasSynced) {
-		klog.Error("stateInformer not synced! ")
-		os.Exit(1)
-	}
 
 	// start collector
 	go func() {
--- a/pkg/koordlet/koordlet.go
+++ b/pkg/koordlet/koordlet.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


jasonliu747:
@Mr-Linus Thanks for your contribution. Would you like to implement a HasSync function for collector?

Mr-Linus:
> @Mr-Linus Thanks for your contribution. Would you like to implement a HasSync function for collector?

@jasonliu747 OK, but it will introduce a new structure for checking synchronization state because there are many goroutines collecting state that need to be checked if they are executed. If this works, I will implement it in another PR.

pkg/koordlet/metricsadvisor/collector.go
```diff
@@ -232,8 +236,8 @@ func (c *collector) collectPodResUsed() {
 		}
 		klog.V(6).Infof("collect pod %s/%s, uid %s finished, metric %+v",
 			meta.Pod.Namespace, meta.Pod.Name, meta.Pod.UID, podMetric)
-		err := c.metricCache.InsertPodResourceMetric(collectTime, &podMetric)
-		if err != nil {
+
+		if err := c.metricCache.InsertPodResourceMetric(collectTime, &podMetric); err != nil {
 			klog.Infof("insert pod %s/%s, uid %s resource metric failed, metric %v, err %v",
```
eahydra, line 241: here should replace klog.Infof with klog.Errorf

Mr-Linus: Fixed.

pkg/koordlet/koordlet.go
```diff
@@ -145,6 +146,12 @@ func (d *daemon) Run(stopCh <-chan struct{}) {
 		}
 	}()
 
+	// wait informer synced
+	if cache.WaitForCacheSync(stopCh, d.statesInformer.HasSynced) {
```
jasonliu747, line 150: In this case, HasSynced check in the following sections could be removed.
https://github.com/koordinator-sh/koordinator/blob/33c2c34102bb19ddf7c691d7c7caed6b2795437b/pkg/koordlet/resmanager/resmanager.go#L240-L242
https://github.com/koordinator-sh/koordinator/blob/33c2c34102bb19ddf7c691d7c7caed6b2795437b/pkg/koordlet/reporter/reporter.go#L138-L140

pkg/koordlet/koordlet.go
```diff
@@ -145,6 +146,12 @@ func (d *daemon) Run(stopCh <-chan struct{}) {
...
+		klog.Error("stateInformer not synced! ")
```
jasonliu747, line 151: *statesInformer

pkg/koordlet/koordlet.go
zwzhang0107: “WaitForCacheSync” is not necessary since all related modules have already added.

Mr-Linus: > “WaitForCacheSync” is not necessary since all related modules have already added.

OK.I removed it in koordlet started.

jasonliu747 (approved):
/lgtm

Issue #52, opened 2022-04-08
[feature request] add cpu burst strategy


zwzhang0107:
**What would you like to be added**:

**Why is this needed**:

PR #70, opened 2022-04-14
New feature: cpu burst strategy

stormgbs:

New feature: cpu burst strategy #52


pkg/tools/cache/expiration_cache.go
```diff
@@ -0,0 +1,118 @@
...
+	DefaultExpiration = 2 * time.Minute
+	DefaultGCInterval = time.Minute
+)
+
+type item struct {
+	object         interface{}
+	expirationTime time.Time
+}
+
+type Cache struct {
```
jasonliu747, line 39: looks like this cache is not needed yet. maybe we could implement it in another PR.

jasonliu747 (changes requested):

jasonliu747 (changes requested):
there are a huge amount of unused codes, please purge them all 🧹🧹🧹
image

PR #76, opened 2022-04-15
Introduce protocol between RuntimeManager and RuntimeHookServer

honpey:


apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+*/
+
+// To regenerate api.pb.go run scripts/generate_runtime.sh
+syntax = "proto3";
+
+package runtime.v1alpha1;
+option go_package = "github.com/koordinator-sh/koordinator/apis/runtime/v1alpha1";
+
+
+// PodSandboxMetadata holds all necessary information for sandbox name.
```
jasonliu747, line 24: PodSandboxMetadata holds all necessary information for sandbox.

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+  LinuxContainerResources overhead = 6;
+  // Optional resources represents the sum of container resources for this sandbox
+  // same as LinuxPodSandboxConfig.resources under CRI scenario
+  LinuxContainerResources resources = 7;
+}
+
+// PreRunPodSandboxHookResponse is RuntimeHookServer's response to PreRunPodSandboxHookRequest.
+// RuntimeManager will merge PreRunPodSandboxHookResponse and RunPodSandboxRequest to generate a
+// RunPodSandboxRequest to containerd(dockerd).
+message PreRunPodSandboxHookResponse{
```
jasonliu747, line 61: Space between `PreRunPodSandboxHookResponse` and `{`

scripts/generate_runtime.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
jasonliu747, line 1: add this script into Makefile like `generate_client.sh`

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+// PodSandboxMetadata holds all necessary information for sandbox name.
+message PodSandboxMetadata {
+  // Pod name of the sandbox. Same as the pod name in the Pod ObjectMeta.
+  string name = 1;
+  // Pod UID of the sandbox. Same as the pod UID in the Pod ObjectMeta.
+  string uid = 2;
+  // Pod namespace of the sandbox. Same as the pod namespace in the Pod ObjectMeta.
+  string namespace = 3;
+  // Attempt number of creating the sandbox. Default: 0.
+  uint32 attempt = 4;
```
ZYecho, line 33: remove this filed

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// To regenerate api.pb.go run scripts/generate_runtime.sh
+syntax = "proto3";
```
ZYecho, line 18: for the docker proxy, we need to transfer the http request to grpc request?  I think it's a little complicated.

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+  string namespace = 3;
+  // Attempt number of creating the sandbox. Default: 0.
+  uint32 attempt = 4;
+}
+
+// PreRunPodSandboxHookRequest is sent to RuntimeHookServer before pod creating request transferred to
+// backend containerd or dockerd. This Request is generated basing on CRI's RunPodSandboxRequest, including pod
+// Meta, Resources, Annotations .e.g and RuntimeHookServer should ensure correct operations basing on
+// this request.
+message PreRunPodSandboxHookRequest {
```
zwzhang0107, line 40: PreRunPodSandboxHookRequest -> PodSandboxHookRequest
which can be used for PostRunPodSandboxHook, PreStopPodSandboxHook in the future

apis/runtime/v1alpha1/api.proto
honpey: the request body has no difference between grpc and http when talking to RuntimeHookServer,  so  it should NOT be more complicated under docker scenario?

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+// RuntimeManager will merge PreRunPodSandboxHookResponse and RunPodSandboxRequest to generate a
+// RunPodSandboxRequest to containerd(dockerd).
+message PreRunPodSandboxHookResponse{
+  // RuntimeHookServer may inject additional labels/annotations to PodSandboxRequest to be send to
+  // backend contaienrd/dockerd
+  map<string, string> labels = 1;
+  map<string, string> annotations = 2;
+  // RuntimeHookSever may modify cgroup_parent to construct customized cgroup topology.
+  string cgroup_parent = 3;
+  // RuntimeHookServer may modify the linux resource config.
```
hormes, line 68: overhead&resources wont be modified,

apis/runtime/v1alpha1/api.proto
honpey: fixed

apis/runtime/v1alpha1/api.proto
honpey: fixed

honpey: > 

fixed

apis/runtime/v1alpha1/api.proto
honpey: > PodSandboxMetadata holds all necessary information for sandbox.
fixed

apis/runtime/v1alpha1/api.proto
honpey: fixed

apis/runtime/v1alpha1/api.proto
honpey: > 

fixed

apis/runtime/v1alpha1/api.proto
```diff
@@ -0,0 +1,158 @@
...
+
+// PreRunPodSandboxHookRequest is sent to RuntimeHookServer before pod creating request transferred to
+// backend containerd or dockerd. This Request is generated basing on CRI's RunPodSandboxRequest, including pod
+// Meta, Resources, Annotations .e.g and RuntimeHookServer should ensure correct operations basing on
+// this request.
+message PreRunPodSandboxHookRequest {
+  // Metadata of the sandbox. This information will uniquely identify the sandbox.
+  PodSandboxMetadata pod_meta = 1;
+  // Named runtime to use for podSandbox.
+  string runtime_handler = 2;
```
hormes, line 44: What scenarios need to reveal the internal runtime type? If this is the case, maybe we should leave an extension field, such as a map, to avoid frequent modification of the protocol

honpey: > What scenarios need to reveal the internal runtime type? If this is the case, maybe we should leave an extension field, such as a map, to avoid frequent modification of the protocol

for container info sampling example, under which kata-containers cgroups are inside guest kernel while runc-containers cgroups exist on host.
runtime type exists in PodSandboxHookRequest only and is readonly, so it seems nowhere comes frequent modification?

apis/runtime/v1alpha1/api.proto
honpey: resources info would be assigned to pod-level cgroup.
so it makes sense to allow HookServert to change pod-level cgroup config here.

overhead can be removed.

honpey: > overhead&resources wont be modified, so remove those fields from response ?

resources info would be assigned to pod-level cgroup.
so it makes sense to allow HookServert to change pod-level cgroup config here.

overhead can be removed.

scripts/generate_runtime.sh
honpey: > add this script into Makefile like `generate_client.sh`
would solve in subsequent patch

honpey: > add this script into Makefile like `generate_client.sh`

would solve in subsequent patch

jasonliu747: I think you should include it in test pipeline in this PR. Check diff or anything else.

apis/runtime/v1alpha1/api.proto
honpey: > the request body has no difference between grpc and http when talking to RuntimeHookServer, so it should NOT be more complicated under docker scenario?

the request body has no difference between grpc and http when talking to RuntimeHookServer, so it should NOT be more complicated under docker scenario?

apis/runtime/v1alpha1/api.proto
honpey: reverve for sequence features, such as ingoring core adjustment when attempt > 1

honpey: > remove this filed

reverve for sequence features, such as ingoring core adjustment when attempt > 1

apis/runtime/v1alpha1/api.proto
ZYecho: ok

ZYecho:
LGTM

apis/runtime/v1alpha1/api.proto
hormes: when do you need to change this resource, to make response.resources different than request.resources?

apis/runtime/v1alpha1/api.proto
hormes: is this a conventional concept in container community, as it is called runtimeclass in kubernetes?

apis/runtime/v1alpha1/api.proto
honpey: > when do you need to change this resource, to make response.resources different than request.resources?

for batch pod,  lower cpu preemption priority by setting pod's pod-level cgroup cpu.shares to 2

apis/runtime/v1alpha1/api.proto
honpey: > kubernetes

yup. the 'runtime_handler' comes from CRI, 
See https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L413

jasonliu747 (commented):
/lgtm

PR #86, opened 2022-04-20
🐛 fix nodeslo syncing issue

### LINKED ISSUE #84: [BUG] not syncing NodeSLO from configMap slo-controller-config
**What happened**:
1. Not syncing to NodeSLO after `memoryEvictThresholdPercent` of `resource-threshold-config` in configMap `slo-controller-config`  

Expect syncing to NodeSLO with `memoryEvictThresholdPercent` of `resource-threshold-config`

1. update configMap `slo-controller-config`
```yaml
apiVersion: v1
data:
  colocation-config: |
    {
      "enable": true
    }
  resource-threshold-config: |
    {
      "clusterStrategy": {
        "enable": true,
        "memoryEvictThresholdPercent": 70
      }
    }
kind: ConfigMap
metadata:
  name: slo-controller-config
  namespace: koordinator-system
```
2.  check NodeSLO
```shell
$ kubectl get nodeslo -o yaml | grep memoryEvictThresholdPercent # get nothing
```

**Environment**:
- App version:
- Others:

jasonliu747:

fixes #84

### DIFF
--- a/pkg/util/config.go
+++ b/pkg/util/config.go
@@ -34,5 +34,6 @@ func DefaultResourceThresholdStrategy() *slov1alpha1.ResourceThresholdStrategy {
 		Enable:                      pointer.BoolPtr(false),
 		CPUSuppressThresholdPercent: pointer.Int64Ptr(65),
 		CPUSuppressPolicy:           slov1alpha1.CPUSetPolicy,
+		MemoryEvictThresholdPercent: pointer.Int64Ptr(70),
 	}
 }
--- a/pkg/slo-controller/nodeslo/resource_strategy_test.go
+++ b/pkg/slo-controller/nodeslo/resource_strategy_test.go
@@ -37,6 +37,7 @@ func Test_getResourceThresholdSpec(t *testing.T) {
 		ClusterStrategy: &slov1alpha1.ResourceThresholdStrategy{
 			Enable:                      pointer.BoolPtr(true),
 			CPUSuppressThresholdPercent: pointer.Int64Ptr(60),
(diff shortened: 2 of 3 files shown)


pkg/util/config.go
```diff
@@ -34,5 +34,6 @@ func DefaultResourceThresholdStrategy() *slov1alpha1.ResourceThresholdStrategy {
 		Enable:                      pointer.BoolPtr(false),
 		CPUSuppressThresholdPercent: pointer.Int64Ptr(65),
 		CPUSuppressPolicy:           slov1alpha1.CPUSetPolicy,
+		MemoryEvictThresholdPercent: pointer.Int64Ptr(70),
```
hormes, line 37: add a default value here, but this field should be overridden in the nodeslo controller according to the configmap configuration.

jasonliu747: it will be overridden by `getResourceThresholdSpec`:
https://github.com/koordinator-sh/koordinator/blob/cb24c0804ec7ab53becef318ddf07513cfc974e4/pkg/slo-controller/nodeslo/resource_strategy.go#L33-L52

also it's verified by unit test here:
https://github.com/koordinator-sh/koordinator/blob/48f0ccc88cf917f55bec30f8f8110dadbaf275ba/pkg/slo-controller/nodeslo/resource_strategy_test.go#L36-L42
https://github.com/koordinator-sh/koordinator/blob/48f0ccc88cf917f55bec30f8f8110dadbaf275ba/pkg/slo-controller/nodeslo/resource_strategy_test.go#L143-L148

PR #134, opened 2022-05-12
test(controller): add unit test for `config/config.go`

jasonliu747:

increase code coverage of `pkg/slo-controller/config/config.go`

### DIFF
--- a/pkg/slo-controller/config/config_test.go
+++ b/pkg/slo-controller/config/config_test.go
@@ -1,0 +1,422 @@
+/*
+ Copyright 2022 The Koordinator Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


pkg/slo-controller/config/config.go
```diff
@@ -36,7 +36,7 @@ type ColocationCfg struct {
 
 type NodeColocationCfg struct {
 	NodeSelector *metav1.LabelSelector
-	ColocationCfg
+	ColocationStrategy
```
hormes, line 39: Why change this api？

jasonliu747: Because instead of `ColocationCfg`, NodeColocationCfg shoud contains `ColocationStrategy`.

PR #140, opened 2022-05-16
feat(nodeslo-ctrl): use DefaultCPUBurstStrategy when config map not exist

jasonliu747:

User might not specify a value in config map. In that case, we use `DefaultCPUBurstStrategy` to create default value for CPU Burst Strategy.


pkg/slo-controller/nodeslo/nodeslo_controller.go
```diff
@@ -78,6 +78,7 @@ func (r *NodeSLOReconciler) getNodeSLOSpec(node *corev1.Node, oldSpec *slov1alph
 	nodeSLOSpec := &slov1alpha1.NodeSLOSpec{
 		ResourceUsedThresholdWithBE: util.DefaultResourceThresholdStrategy(),
 		ResourceQoSStrategy:         &slov1alpha1.ResourceQoSStrategy{},
+		CPUBurstStrategy:            util.DefaultCPUBurstStrategy(),
```
saintube, line 81: maybe we can let `nodeSLO.spec.cpuBurstStrategy` be empty since the config will be merged with the default again in Koordlet.

hormes: please keep it empty, we don't want to do a refresh for all crs when we update the default values.

PR #151, opened 2022-05-18
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

NONE

All tests should continue to pass. No leftover temporary directories when the tests finish.

### DIFF
--- a/pkg/koordlet/resmanager/cgroup_reconcile_test.go
+++ b/pkg/koordlet/resmanager/cgroup_reconcile_test.go
@@ -804,7 +804,7 @@ func TestCgroupResourceReconcile_calculateResources(t *testing.T) {
 			assert.NoError(t, err)
 			defer func() { stop <- struct{}{} }()
 
-			_ = system.NewFileTestUtil(t)
+			system.NewFileTestUtil(t)
 
 			got, got1, got2 := m.calculateResources(tt.args.nodeCfg, tt.args.node, tt.args.podMetas)
 			assertCgroupResourceEqual(t, tt.want, got)
--- a/pkg/koordlet/resmanager/cgroup_reconcile_test.go
+++ b/pkg/koordlet/resmanager/cgroup_reconcile_test.go
@@ -425,7 +425,6 @@ func Test_calculateAndUpdateResources(t *testing.T) {
 			defer func() { stop <- struct{}{} }()
 
 			helper := system.NewFileTestUtil(t)
-			defer helper.Cleanup()
 
 			initQoSStrategy := defaultQoSStrategy()
 			initQoSCgroupFile(initQoSStrategy, helper)
@@ -805,8 +804,7 @@ func TestCgroupResourceReconcile_calculateResources(t *testing.T) {
 			assert.NoError(t, err)
 			defer func() { stop <- struct{}{} }()
 
-			helper := system.NewFileTestUtil(t)
-			defer helper.Cleanup()
+			_ = system.NewFileTestUtil(t)
 
 			got, got1, got2 := m.calculateResources(tt.args.nodeCfg, tt.args.node, tt.args.podMetas)
 			assertCgroupResourceEqual(t, tt.want, got)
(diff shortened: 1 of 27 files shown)


jasonliu747 (approved):
This is actually a very good suggestion!
/lgtm
/approve

pkg/koordlet/resmanager/cgroup_reconcile_test.go
```diff
@@ -805,8 +804,7 @@ func TestCgroupResourceReconcile_calculateResources(t *testing.T) {
 			assert.NoError(t, err)
 			defer func() { stop <- struct{}{} }()
 
-			helper := system.NewFileTestUtil(t)
-			defer helper.Cleanup()
+			_ = system.NewFileTestUtil(t)
```
saintube, line 807: can we remove the equation?

pkg/koordlet/resmanager/resctrl_reconcile_test.go
```diff
@@ -400,9 +399,7 @@ func Test_getPodCgroupNewTaskIds(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-
-			helper := system.NewFileTestUtil(t)
-			defer helper.Cleanup()
+			_ = system.NewFileTestUtil(t)
```
saintube, line 402: can we remove the equation

pkg/util/system/util_test_tool.go
```diff
@@ -30,15 +30,11 @@ type FileTestUtil struct {
 	t *testing.T
 }
 
-// Creates a new test util for the specified subsystem
+// NewFileTestUtil creates a new test util for the specified subsystem
 func NewFileTestUtil(t *testing.T) *FileTestUtil {
-	tempDir, err := ioutil.TempDir("/tmp", "koordlet_test")
+	tempDir := t.TempDir()
```
saintube, line 35: It seems `(*FileTestUtil).Cleanup()` is useless with this patch and its references are removed. 
Can we remove the `Cleanup()` since its work would be done implicitly?

saintube (changes requested):

pkg/util/system/util_test_tool.go
jasonliu747: +1

jasonliu747 (commented):
/lgtm

saintube (approved):
/lgtm

jasonliu747:
@Juneezee Hi, when I ran `runtime_test.go` locally on my Mac, it would throw an error: `Unix socket path "/var/folders/t3/fm5gy2413t304ksnq2_jd91h0000gn/T/Test_GetRuntimeHandlertest_have_dockerRuntime_but_need_containerd3848019481/001/var/run/docker.sock" is too long`. 

Did it ever happen to you?

Juneezee:
@jasonliu747 No, I don't get this error on my Linux machine. This is probably a Mac specific issue, see https://unix.stackexchange.com/a/367012/376279

jasonliu747:
> @jasonliu747 No, I don't get this error on my Linux machine. This is probably a Mac specific issue, see https://unix.stackexchange.com/a/367012/376279

Could you try to fix this? I believe majority of programmer are using Mac for development.

Juneezee:
@jasonliu747 PR #156 fixes the issue. 

I think the only way to tell `t.TempDir()` to use `/tmp` on MacOS is to set the environment variable `$TMPDIR` to empty so that it uses `/tmp` (https://github.com/golang/go/blob/128279e5034ca29bad4320eef81a8abd5b40ea7e/src/os/file.go#L379-L390). This might do more harm than good and break other tests.

Another way is to use `ioutil.TempDir` and specify `/tmp` explicitly, just like in #156.

PR #160, opened 2022-05-20
fix build errors

hormes:
1. fix build errors when install gnu commands in MacOS
2. add a script to fix license-header

Signed-off-by: Fansong Zeng <fanster.z@gmail.com>


hack/update-license-header.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
jasonliu747, line 1: this is included in `Makefile`

hack/update-license-header.sh
```diff
@@ -0,0 +1,6 @@
...
+
+cd $(dirname $0)
+
+cd ../
+bin/license-header-checker -r -a -v -i vendor "hack/boilerplate.go.txt" . go
```
eahydra, line 6: maybe we need update `hack/boilerplate.go.txt` to remove the space prefix before License.

eahydra (changes requested):

hack/update-license-header.sh
hormes: CI is suitable for inspection, because the CI phase replacement does not affect the user's code, and the update of the code is handed over to the user.

hack/update-license-header.sh
hormes: it is another story, we can change it in another PR.

hack/update-license-header.sh
jasonliu747: if a user sees there is an error in CI report, he/she will re-run `make lint-license` and push the code again.

hormes: can CI report errors directly in license-header-checker instead of relying on git diff?

jasonliu747: no, but it will show which file got replaced.

hack/generate_client.sh
```diff
@@ -20,11 +20,7 @@ rm -rf ./pkg/client/{clientset,informers,listers}
 mv "${TMP_DIR}"/src/github.com/koordinator-sh/koordinator/pkg/client/* ./pkg/client
 
 function custom_sed(){
-    if [[ "$OSTYPE" == "darwin"* ]]; then
-      sed -i '' $@
-    else
-      sed -i $@
-    fi
+  sed -i -e $@
```
jasonliu747, line 23: it's not working :(
image

hormes: awkward

hack/update-license-header.sh
jasonliu747, line 1: Since you updated Makefile, this file is no longer needed.

jasonliu747 (changes requested):

hack/generate_client.sh
```diff
@@ -21,7 +21,11 @@ mv "${TMP_DIR}"/src/github.com/koordinator-sh/koordinator/pkg/client/* ./pkg/cli
 
 function custom_sed(){
     if [[ "$OSTYPE" == "darwin"* ]]; then
-      sed -i '' $@
+      if sed --v 2>/dev/null | grep 'GNU sed'>/dev/null; then
```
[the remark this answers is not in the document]
hormes: too many changes in different versions of sed, use perl instead

jasonliu747:
let's move `generate_client.sh`, `fix_crd_plural.sh`, and `update-license-header.sh` into ./scripts. wdyt?

hack/generate_client.sh
```diff
@@ -20,12 +20,9 @@ rm -rf ./pkg/client/{clientset,informers,listers}
...
-      sed -i '' $@
-    else
-      sed -i $@
-    fi
+    perl -i -pe $@
 }
 
 custom_sed 's#\"config\"#\"config.koordinator.sh\"#g' ./pkg/client/clientset/versioned/typed/config/v1alpha1/fake/fake_*.go
 custom_sed 's#\"slo\"#\"slo.koordinator.sh\"#g' ./pkg/client/clientset/versioned/typed/slo/v1alpha1/fake/fake_*.go
+
```
jasonliu747, line 28: let's move `generate_client.sh`, `fix_crd_plural.sh`, and `update-license-header.sh` into `./scripts`. wdyt?

hormes: These two directories need to be merged to keep only one, considered in another pr. Because of the existence of tools.go, etc., the directory name scripts may not be appropriate, anyway, do it later.

PR #162, opened 2022-05-21
ci: support running unit test on multiple os

### LINKED ISSUE #163: [support request] Run GitHub Action on multiple environments
**What is your problem or scenario?**
Since majority of programmer are using Mac for development, and deploy Koordinator in a Linux environment. So after #151 , I think it's necessary to run github action on multiple environments like MacOS and Linux.

jasonliu747:

Since majority of programmer are using Mac for development, and deploy Koordinator in a Linux environment. So after https://github.com/koordinator-sh/koordinator/pull/151 , I think it's necessary to run github action on multiple environments like MacOS and Linux.

fixes #163

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,25 +9,13 @@ on:
   workflow_dispatch: {}
 
 jobs:
-  get-go-version:
-    runs-on: ubuntu-latest
-    outputs:
-      go-version: ${{ steps.read-go-version.outputs.version }}
-    steps:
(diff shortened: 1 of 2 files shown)


hormes:
It might be too slow to run on Mac, can it be made optional？

jasonliu747:
> It might be too slow to run on Mac, can it be made optional？

@hormes I've tried almost every possible solution but without success. Let's merge this PR firtst, and I'll open an issue in [here](https://github.com/actions/virtual-environments).

PR #165, opened 2022-05-23
Add pod annotations/labels for container level hook

honpey:


apis/runtime/v1alpha1/api.proto
```diff
@@ -124,6 +124,9 @@ message ContainerResourceHookRequest {
   map<string, string> container_annotations = 3;
   LinuxContainerResources container_resources = 4;
   LinuxContainerResources pod_resources = 5;
+  // pod related annotations and labels
+  map<string, string> pod_annotations = 6;
+  map<string, string> pod_labels = 7;
```
hormes, line 129: Why consider putting it at the interface level instead of letting the plugin get it by itself?

honpey: > Why consider putting it at the interface level instead of letting the plugin get it by itself?
where to get pod info for plugin? talk with runtime-manager or containerd?

honpey: > where to get pod info for plugin? talk with runtime-manager or containerd?

where to get pod info for plugin? talk with runtime-manager or containerd?

honpey: > et. koordlet can get it from it local cache.

there may be only PreStartContainerHook registered, so koordlet could not catch pod level annotations/levels during RunSandbox stage without  PreRunPodSandboxHook hook.

yet, koordlet can register PreRunPodSandboxHook explicitly to solve, but we had better do not import dependency between two hooks(e.g PreRunPodSandboxHook and PreStartContainerHook).

jasonliu747 (commented):
/lgtm

apis/runtime/v1alpha1/api.proto
honpey: > > et. koordlet can get it from it local cache.
> 
> there may be only PreStartContainerHook registered, so koordlet could not catch pod level annotations/levels during RunSandbox stage without PreRunPodSandboxHook hook.
> 
> yet, koordlet can register PreRunPodSandboxHook explicitly to solve, but we had better do not import dependency between two hooks(e.g PreRunPodSandboxHook and PreStartContainerHook).

then， koordlet would be stateless, since no need to store the pod info

zwzhang0107: 1、pod syncing is for metric analysis and dynamic QoS policies, which is asynchronous with hook injections. So the injections will be ineffective if the hook wait

hormes: Such as communication, keep these two fields, but let's do some deductions in the early stage(restricted to koordinator.sh)

zwzhang0107: memo:
1、priority class for pod/container req
2、filter with DomainPrefix

PR #170, opened 2022-05-25
runtime-manager: refactor codes about store and resource-exectutor

honpey:
1. set metaManager as inner class inside store package. other package call store.WritePodSandboxInfo
   directly instead of getting a new Metamanager first.
2. resource-executor adapts for the new structed PodSandboxInfo/ContianerInfo

Signed-off-by: honpey <honpey@gmail.com>


honpey:
cc @jasonliu747

PR #180, opened 2022-05-27
feat: add kubelet http2 support

### LINKED ISSUE #150: [BUG]I install koordinator on huaweicloud cce,but I can't get the right result for the spark-pi job.
**What happened**:
when I use koordinator on huaweiCloud cce.
image

I describe the pod:
image
image

I can't find batch-cpu and batch-memory information from the node:
image

 I fetch logs from koordlet use kubectl logs ds:
kubectl logs --tail=10 koordlet-cd9xg
image

So I modify the kubelet config for the node:
image

I get the error info when the koordlet send request to kubelet:
this is the logs on the koordlet pod:
image
this is the thumbnail on the node:
image
image

kubelet config:
`kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
address: 0.0.0.0
#address: 7.183.231.172
authentication:
  anonymous:
    enabled: false
  x509:
    clientCAFile: /opt/cloud/cce/srv/kubernetes/cluster-ca.crt
authorization:
  mode: Webhook
clusterDNS:

\- 10.247.3.10
clusterDomain: cluster.local
enableControllerAttachDetach: true
evictionHard:
  memory.available: 100Mi
  nodefs.available: 10%
  nodefs.inodesFree: 5%
  imagefs.available: 10%
featureGates:
  DevicePlugins: true
  MultiGPUScheduling: true
  ExpandCSIVolumes: true
  CSIInlineVolume: true
  CSIMigrationFlexVolumeFuxi: true
  CSIMigrationFlexVolumeFuxiComplete: true
  CSIMigration: true
  IPv6DualStack: false
  ReserveMemoryCgroupForPageCache: false
  SizeMemoryBackedVolumes: true
  OversubscriptionResource: false

maxPods: 110
podPidsLimit: -1
port: 10250
readOnlyPort: 10255
staticPodPath: "/opt/cloud/cce/kubernetes/manifests"
serverTLSBootstrap: true
imageGCHighThresholdPercent: 80
imageGCLowThresholdPercent: 70
kubeAPIQPS: 100
kubeAPIBurst: 100
cpuManagerPolicy: none
tlsCipherSuites:
\- TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
\- TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
\- TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
\- TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
kubeletCgroups: "/kubelet"
eventRecordQPS: 5
`

kubernetes version: 
image

kubelet response the pod information to the koordlet

use huawei cloud cce and helm install koordinator.

**Environment**:
- App version:
- Others:

LambdaHJ:
support cce kubernetes cluster kubelet read-only port http2 scheme

fixes #150


jasonliu747 (changes requested):
Thanks for your contribution. Please follow the [instruction](https://github.com/koordinator-sh/koordinator/pull/180/checks?check_run_id=6628091763) to sign off your commit.

pkg/koordlet/statesinformer/config.go
```diff
@@ -18,25 +18,41 @@ package statesinformer
 
 import "flag"
 
+const (
+	SOURCE_HTTP  = "http"
```
saintube, line 22: Camel Case is recommended (no underscore here)

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -77,9 +78,21 @@ func NewStatesInformer(config *Config, kubeClient clientset.Interface, crdClient
...
+	var stub kubelet.KubeletStub
+	switch config.ReadPodScheme {
+	case SOURCE_HTTP:
+		stub = kubelet.NewKubeletStub(config.KubeletIPAddr, config.KubeletHTTPPort, config.KubeletSyncTimeoutSeconds)
+	case SOURCE_HTTP2:
+		stub = kubelet.NewKubeletHTTP2Stub(config.KubeletIPAddr, config.KubeletHTTPPort,
+			config.KubeletClientCertificate, config.KubeletClientKey,
+			config.KubeletCertificateAuthority, config.KubeletInsecureTls, config.KubeletSyncTimeoutSeconds)
+	default:
+		panic("not support read pod scheme.")
```
saintube, line 90: we prefer `klog.Fatal` than a raw `panic`

pkg/util/http2.go
```diff
@@ -0,0 +1,82 @@
...
+	defer res.Body.Close()
+
+	result, err = ioutil.ReadAll(res.Body)
+	if err != nil {
+		return result, fmt.Errorf("failed to read from response, url: %v, http code: %v, err: %v",
+			httpURL, res.StatusCode, err)
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return result, fmt.Errorf("http response statue code %v, body %v", res.StatusCode, string(result))
```
saintube, line 79: typo: `statue` -> `status`

saintube (commented):
please don't forget to add unit tests :)

jasonliu747:
@LambdaHJ came up with another soultion. Have you ever tried using token in `/var/run/secrets/kubernetes.io/serviceaccount/token` as BearerToken to make an HTTP request?

pkg/util/http2.go
```diff
@@ -0,0 +1,82 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"golang.org/x/net/http2"
+)
+
+func DoHTTP2Get(methodName string, ip string, port int, certFile, keyFile, caFile string, insecure bool, timeout int) ([]byte, error) {
+
+	httpClient := &http.Client{
```
eahydra, line 32: The initialization logic of the client should be independent, and it is not necessary to do repeated things every time an HTTP request is made.

eahydra (changes requested):

pkg/util/http2.go
eahydra: We can hide the difference between http/1.x and http/2

LambdaHJ: I think cce is not standard implementation, maybe there are another implementation in the future. In order to make the code logic easy to manage, we'd better implement it separately.

eahydra: There is indeed a problem with the implementation in `pkg/util/http.go` in the existing code, but I think you can directly encapsulate a general HTTP Client in this PR to facilitate refactoring. In addition, I understand that this PR has nothing to do with whether the CCE implementation is standard or not.

LambdaHJ: OK, I will rewrite this part.

pkg/koordlet/statesinformer/config.go
```diff
@@ -18,25 +18,41 @@ package statesinformer
...
+)
+
 type Config struct {
-	KubeletIPAddr              string
-	KubeletHTTPPort            int
-	KubeletSyncIntervalSeconds int
-	KubeletSyncTimeoutSeconds  int
+	ReadPodScheme               string
+	KubeletIPAddr               string
+	KubeletHTTPPort             int
```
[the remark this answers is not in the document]
LambdaHJ: Modifying  `KubeletHTTPPort` may result in incompatibility.

jasonliu747:
As discussed earlier, let's use service account as token to access kubelet. And use apiserver as a failover **only** when kubelet doesn't work. Also assume `node.status.daemonEndpoints.kubeletEndpoint` is the default port for kubelet.

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -108,6 +111,19 @@ func (s *statesInformer) Run(stopCh <-chan struct{}) error {
...
 	go s.nodeSLOInformer.Run(stopCh)
 
+	// waiting for node synced.
+	waitInformersSynced := []cache.InformerSynced{
+		s.nodeInformer.HasSynced, s.nodeSLOInformer.HasSynced}
+	if !cache.WaitForCacheSync(stopCh, waitInformersSynced...) {
+		return fmt.Errorf("timed out waiting for states informer caches to sync")
+	}
+
+	stud, err := newKubeletStub(s.GetNode(), s.config.KubeletPreferredAddressType, s.config.KubeletSyncTimeoutSeconds)
```
jasonliu747, line 121: stud -> stub

pkg/koordlet/statesinformer/states_pod.go
```diff
@@ -42,10 +49,22 @@ func (s *statesInformer) syncKubelet() error {
...
 	return nil
 }
 
+func (s *statesInformer) getPodsFromApiserver() (corev1.PodList, error) {
+	timeoutInSecond := time.Second * time.Duration(s.config.ApiServerSyncTimeoutSeconds)
+	ctx, cancel := context.WithTimeout(context.Background(), timeoutInSecond)
+	defer cancel()
+	selector := fields.SelectorFromSet(fields.Set{"spec.nodeName": s.GetNode().Name})
+	opt := metav1.ListOptions{
+		FieldSelector: selector.String(),
```
jasonliu747, line 62: Consider adding `resourceVersion` and `resourceVersionMatch` to achieve better performance and scalability. See https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions for details.

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -209,3 +225,19 @@ func (s *statesInformer) setupInformers() {
 	s.setupNodeInformer()
 	s.setupNodeSLOInformer()
 }
+
+func newKubeletStub(node *corev1.Node, addressPreferred string, timeout int) (KubeletStub, error) {
+	var address string
+	var err error
+	address = util.GetNodeAddress(node, corev1.NodeAddressType(addressPreferred))
```
jasonliu747, line 232: you should check `addressPreferred` first

pkg/util/node.go
```diff
@@ -75,3 +75,12 @@ func GetRootCgroupCurCFSPeriod(qosClass corev1.PodQOSClass) (int64, error) {
...
 	return strconv.ParseInt(strings.TrimSpace(rawContent), 10, 64)
 }
+
+func GetNodeAddress(node *corev1.Node, addrType corev1.NodeAddressType) string {
+	for _, address := range node.Status.Addresses {
+		if address.Type == addrType {
+			return address.Address
+		}
+	}
+	return ""
```
jasonliu747, line 85: consider returing an error here if there is no match.
`return "", fmt.Errorf("no address matched types %v", addrType)`

jasonliu747, line 85: Return an error here if there is no match. `return "", fmt.Errorf("no address matched types %v", addrType)`

Or you can validate the commad line input in the first place. Personally, I preffer the latter option.

pkg/koordlet/statesinformer/kubelet_stub.go
```diff
@@ -19,38 +19,81 @@ package statesinformer
...
+		TLS: transport.TLSConfig{
+			Insecure: true,
+		},
 	}
+	return tlsConfig
 }
 
 func (k *kubeletStub) GetAllPods() (corev1.PodList, error) {
 	podList := corev1.PodList{}
-	result, err := util.DoHTTPGet("pods", k.ipAddr, k.httpPort, k.timeoutSeconds)
```
jasonliu747, line 48: I guess after this PR, `DoHTTPGet` is no longer needed. Feel free to delete that file.

pkg/koordlet/statesinformer/kubelet_stub.go
```diff
@@ -19,38 +19,81 @@ package statesinformer
...
 	}
+	if rsp.StatusCode != http.StatusOK {
+		return podList, fmt.Errorf("request http failed, code %d", rsp.StatusCode)
+	}
+
+	body, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		return podList, err
+	}
+	defer rsp.Body.Close()
```
jasonliu747, line 93: I would move this `Close()` call before checking status code. After L84.

pkg/util/node.go
jasonliu747, line 85: Return an error here if there is no match. `return "", fmt.Errorf("no address matched types %v", addrType)`

Or you can validate the commad line input in the first place. Personally, I preffer the latter option.

pkg/koordlet/statesinformer/kubelet_stub.go
jasonliu747, line 48: I assume after this PR, `DoHTTPGet` is no longer needed. Feel free to delete `pkg/util/http.go`.

pkg/koordlet/statesinformer/kubelet_stub.go
jasonliu747, line 93: I would move this `Close()` before checking status code and after error checking of HTTP Get call.

jasonliu747: https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body

pkg/koordlet/statesinformer/states_pod.go
```diff
@@ -17,20 +17,29 @@ limitations under the License.
...
 )
 
-func (s *statesInformer) syncKubelet() error {
+func (s *statesInformer) syncPods() error {
 	podList, err := s.kubelet.GetAllPods()
-	if err != nil {
-		klog.Warningf("get pods from kubelet failed, err: %v", err)
-		return err
+
+	// when kubelet recovery form crash, podList maybe a emtpy struct.
```
jasonliu747, line 35: when kubelet recovers form crash, podList may be empty.

jasonliu747 (commented):
Nice job, look good to me overall!. 
If possible, I would like to see more tests. The diff coverage is only 38% right not.

image

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -209,3 +226,29 @@ func (s *statesInformer) setupInformers() {
 	s.setupNodeInformer()
 	s.setupNodeSLOInformer()
 }
+
+func newKubeletStub(node *corev1.Node, addressPreferred string, timeout int) (KubeletStub, error) {
+	var address string
+	var err error
+	addressPreferredType := corev1.NodeAddressType(addressPreferred)
+	// if the address of the specified type has not been set or error type, InternalIP will be used.
+	if addressPreferredType == "" ||
```
saintube, line 235: could you please improve the readability of the statement? e.g. `if IsAddressTypeSupported()`

pkg/koordlet/statesinformer/states_pod.go
```diff
@@ -17,20 +17,29 @@ limitations under the License.
...
 )
 
-func (s *statesInformer) syncKubelet() error {
+func (s *statesInformer) syncPods() error {
 	podList, err := s.kubelet.GetAllPods()
-	if err != nil {
-		klog.Warningf("get pods from kubelet failed, err: %v", err)
-		return err
+
+	// when kubelet recovers form crash, podList may be empty.
```
saintube, line 35: type: `form` -> `from`

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -52,6 +54,7 @@ type StatesInformer interface {
 
 type statesInformer struct {
 	config       *Config
+	kubeClinet   clientset.Interface
```
saintube, line 57: type: `kubeClinet` -> `kubeClient`. same below

pkg/koordlet/statesinformer/states_pod.go
```diff
@@ -17,20 +17,29 @@ limitations under the License.
...
+	if err != nil || len(podList.Items) == 0 {
+		klog.Warningf("get pods from kubelet failed back to apiserver, err: %v", err)
+		podList, err = s.getPodsFromApiserver()
```
saintube, line 38: Should the getting apiserver be the default behavior for the fallback, rather than an option of `syncPods`? WDYT @LambdaHJ @jasonliu747 @zwzhang0107 @hormes

jasonliu747: Koordlet is a DeamonSet, so it seems to me instead of letting each daemon query APIServer, accessing Kubelet on each node might be a best choice to achieve better performance and scalability.
I'd love to hear your view on this PR as well @eahydra.

pkg/koordlet/statesinformer/config.go
```diff
@@ -16,27 +16,31 @@ limitations under the License.
...
+		KubeletSyncIntervalSeconds:  30,
+		KubeletSyncTimeoutSeconds:   3,
+		ApiServerSyncTimeoutSeconds: 10,
 	}
 }
 
 func (c *Config) InitFlags(fs *flag.FlagSet) {
-	fs.StringVar(&c.KubeletIPAddr, "KubeletIPAddr", c.KubeletIPAddr, "Kubelet IP Address.")
-	fs.IntVar(&c.KubeletHTTPPort, "KubeletHTTPPort", c.KubeletHTTPPort, "Kubelet HTTP httpPort.")
+	fs.StringVar(&c.KubeletPreferredAddressType, "KubeletPreferredAddressType", c.KubeletPreferredAddressType, "Kubelet endpoint preferred address type, options are Hostname|InternalIP|ExternalIP|InternalDNS|ExternalDNS")
```
jasonliu747, line 42: Usage: The node address types to use when determining which address to use to connect to a particular node.

pkg/koordlet/statesinformer/config.go
```diff
@@ -16,27 +16,31 @@ limitations under the License.
...
 	fs.IntVar(&c.KubeletSyncIntervalSeconds, "KubeletSyncIntervalSeconds", c.KubeletSyncIntervalSeconds, "Kubelet sync interval by seconds.")
 	fs.IntVar(&c.KubeletSyncTimeoutSeconds, "KubeletSyncTimeoutSeconds", c.KubeletSyncTimeoutSeconds, "Kubelet sync timeout by seconds.")
+	fs.IntVar(&c.ApiServerSyncTimeoutSeconds, "ApiServerSyncTimeoutSeconds", c.ApiServerSyncTimeoutSeconds, "Apiserver sync timeout by seconds.")
```
jasonliu747, line 45: Usage: The length of time to wait before giving up on a single request to APIServer.

pkg/koordlet/statesinformer/config.go
```diff
@@ -16,27 +16,31 @@ limitations under the License.
...
+		ApiServerSyncTimeoutSeconds: 10,
 	}
 }
 
 func (c *Config) InitFlags(fs *flag.FlagSet) {
-	fs.StringVar(&c.KubeletIPAddr, "KubeletIPAddr", c.KubeletIPAddr, "Kubelet IP Address.")
-	fs.IntVar(&c.KubeletHTTPPort, "KubeletHTTPPort", c.KubeletHTTPPort, "Kubelet HTTP httpPort.")
+	fs.StringVar(&c.KubeletPreferredAddressType, "KubeletPreferredAddressType", c.KubeletPreferredAddressType, "Kubelet endpoint preferred address type, options are Hostname|InternalIP|ExternalIP|InternalDNS|ExternalDNS")
 	fs.IntVar(&c.KubeletSyncIntervalSeconds, "KubeletSyncIntervalSeconds", c.KubeletSyncIntervalSeconds, "Kubelet sync interval by seconds.")
 	fs.IntVar(&c.KubeletSyncTimeoutSeconds, "KubeletSyncTimeoutSeconds", c.KubeletSyncTimeoutSeconds, "Kubelet sync timeout by seconds.")
```
jasonliu747, line 44: Usage: The length of time to wait before giving up on a single request to Kubelet.

pkg/koordlet/statesinformer/config.go
```diff
@@ -16,27 +16,31 @@ limitations under the License.
...
+		KubeletSyncTimeoutSeconds:   3,
+		ApiServerSyncTimeoutSeconds: 10,
 	}
 }
 
 func (c *Config) InitFlags(fs *flag.FlagSet) {
-	fs.StringVar(&c.KubeletIPAddr, "KubeletIPAddr", c.KubeletIPAddr, "Kubelet IP Address.")
-	fs.IntVar(&c.KubeletHTTPPort, "KubeletHTTPPort", c.KubeletHTTPPort, "Kubelet HTTP httpPort.")
+	fs.StringVar(&c.KubeletPreferredAddressType, "KubeletPreferredAddressType", c.KubeletPreferredAddressType, "Kubelet endpoint preferred address type, options are Hostname|InternalIP|ExternalIP|InternalDNS|ExternalDNS")
 	fs.IntVar(&c.KubeletSyncIntervalSeconds, "KubeletSyncIntervalSeconds", c.KubeletSyncIntervalSeconds, "Kubelet sync interval by seconds.")
```
jasonliu747, line 43: Usage: The interval at which Koordlet will retain datas from Kubelet.

pkg/koordlet/statesinformer/states_informer.go
```diff
@@ -209,3 +226,29 @@ func (s *statesInformer) setupInformers() {
...
+			addressPreferredType != corev1.NodeInternalIP &&
+			addressPreferredType != corev1.NodeInternalDNS) {
+		klog.Warningf("Wrong address type or empty type, InternalIP will be used, error: (%+v).", addressPreferredType)
+		addressPreferredType = corev1.NodeInternalIP
+	}
+	address, err = util.GetNodeAddress(node, addressPreferredType)
+	if err != nil {
+		klog.Fatalf("Get node address error: %v type(%s) ", err, addressPreferred)
+	}
+	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
```
jasonliu747, line 248: let's declare `/var/run/secrets/kubernetes.io/serviceaccount/token` in a const

pkg/koordlet/statesinformer/states_pod.go
eahydra: According to the implementation of apiserver in the community version, requests received when APIServer itself failover will penetrate to etcd. Considering that koordlet itself runs in the form of a daemonset, it may generate a large number of requests to the APIServer under a large cluster scale, which affects the reliability of the cluster dimension. Therefore, we should consider protecting the APIServer. In addition, I am wondering if there are users who expect to go directly to the apiserver to synchronize data?

hormes: Because the frequency of this call is very high, it falls back to calling the apiserver when it fails, which may lead to excessive pressure on the apiserver. A solution that can be considered is to use a list-watch as one of data source Of course, this can be implemented in another PR.

pkg/koordlet/statesinformer/states_informer.go
hormes: the serviceaccount  can used to access kubelet https port？

pkg/koordlet/statesinformer/config.go
```diff
@@ -16,27 +16,31 @@ limitations under the License.
...
 
 type Config struct {
-	KubeletIPAddr              string
-	KubeletHTTPPort            int
-	KubeletSyncIntervalSeconds int
-	KubeletSyncTimeoutSeconds  int
+	KubeletPreferredAddressType string
+	KubeletSyncIntervalSeconds  int
+	KubeletSyncTimeoutSeconds   int
+	ApiServerSyncTimeoutSeconds int
```
hormes, line 29: Is this field useless?

jasonliu747 (commented):
/lgtm
Very nice PR, thanks for your contribution. @LambdaHJ

PR #220, opened 2022-06-07
Fix bug where runtime proxy cannot decode annotations in Docker config

cheimu:
Signed-off-by: cheimu yimo@xiaohongshu.com

In runtime proxy, for Docker backend, the splitLablesAndAnnotations cannot successfully decode annotations.


cheimu:
cc @ZYecho

PR #239, opened 2022-06-09
Implement cri scenario PreCreateContainerHook and PostStopPodSandboxHook

cheimu:
Signed-off-by: yimo <yimo@xiaohongshu.com>

implement cri version new runtime api

### Ⅰ. Describe what this PR does
This pr implements cri scenario PreCreateContainerHook and PostStopPodSnadboxHook hook client behavior

### Ⅱ. Does this pull request fix one issue?

#218

PR #236, opened 2022-06-09
docker-proxy: support createContainer and stopSandbox hook

ZYecho:
Signed-off-by: Yue Zhang <huaihuan.zy@alibaba-inc.com>

### Ⅰ. Describe what this PR does

fix #217 #218

Issue #218, opened 2022-06-07
[support request] create container hook


ZYecho:
**What is your problem or scenario?**
implement create container hook

PR #243, opened 2022-06-10
Add container id to ContainerResourceHookRequest

cheimu:
Signed-off-by: Cheimu <xerhoneyfc@gmail.com>

Add container id to ContainerResourceHookRequest and Response.

Currently there is no `container id` field within `ContainerResourceHookRequest`, and there is only `pod_meta.uid`. So, in hook servers' hook plugins, for those rules which will modify cgroup files directly, they only know which pod-level cgroup they should work on, but don't know exactly which container-level cgroup they should work on (for multi-containers pod scenario). 

Add this field which solve this case. 

#218 #236 #239


jasonliu747 (commented):
/lgtm

apis/runtime/v1alpha1/api.proto
```diff
@@ -111,8 +111,10 @@ message HugepageLimit {
 message ContainerMetadata {
   // Name of the container. Same as the container name in the PodSpec.
   string name = 1;
+  // Id of the container.
+  string id = 2;
   // Attempt number of creating the container. Default: 0.
-  uint32 attempt = 2;
+  uint32 attempt = 3;
```
hormes, line 117: Should we keep the number to keep compatibility?

cheimu: Done

jasonliu747 (commented):
/lgtm

PR #250, opened 2022-06-13
refactor: prune socket before launching runtime-proxy

jasonliu747:

There might be a situation where runtime-proxy was killed by `SIGKILL` signal. In this case, 
```
defer os.Remove(options.RuntimeProxyEndpoint)
``` 
won't have a chance to be executed. So it's better to re-create socket file before launching runtime-proxy.

### II. Special notes for reviews
None

### DIFF
--- a/cmd/koord-runtime-proxy/main.go
+++ b/cmd/koord-runtime-proxy/main.go
@@ -43,12 +43,13 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	dir, _ := filepath.Split(options.RuntimeProxyEndpoint)
-	err := os.MkdirAll(dir, 0777)
-	if err != nil {
-		klog.Fatalf("Failed to create socket dir, err: %v", err)
+	if err := os.Remove(options.RuntimeProxyEndpoint); err != nil && !os.IsNotExist(err) {
+		klog.Fatalf("failed to unlink %v: %v", options.RuntimeProxyEndpoint, err)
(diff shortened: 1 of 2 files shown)


cmd/koord-runtime-proxy/main.go
```diff
@@ -43,12 +43,13 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	dir, _ := filepath.Split(options.RuntimeProxyEndpoint)
-	err := os.MkdirAll(dir, 0777)
-	if err != nil {
-		klog.Fatalf("Failed to create socket dir, err: %v", err)
+	if err := os.Remove(options.RuntimeProxyEndpoint); err != nil && !os.IsNotExist(err) {
```
cheimu, line 46: Could we use syscall.Unlink() to achieve it? like https://github.com/koordinator-sh/koordinator/blob/0072d6bf8c2edee509340e644e5808590bff3c4f/pkg/koordlet/runtimehooks/server/server.go#L74 ?

jasonliu747: > Could we use syscall.Unlink() to achieve it?

`os.Remove` will call either `syscall.Unlink` or `syscall.Rmdir` under different circumstances, so I would like to leave the cleanup job to it.

> what if user wrongly used an already used socket file by mistake, doing so could delete the socket file by force right? Is this allowed?

It's fine.

cheimu (commented):
/lgtm

PR #252, opened 2022-06-16
docker-proxy: update podInfo and containerInfo with hook resp cgroupParent

cheimu:
Signed-off-by: cheimu <yimo@xiaohongshu.com>

Current docker proxy forgot to update hook response's cgroup parent field, so there is no updated information within cache. This pr fix this issue.


jasonliu747 (commented):
/lgtm

PR #259, opened 2022-06-18
refactor: remove useless field in NodeSLOReconciler

jasonliu747:

`logr` is no longer needed since we have `klog`

This PR should close #185

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.17
 require (
 	github.com/docker/docker v20.10.2+incompatible
 	github.com/fsnotify/fsnotify v1.5.4
-	github.com/go-logr/logr v0.4.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -60,6 +59,7 @@ require (
 	github.com/emicklei/go-restful v2.9.5+incompatible // indirect
 	github.com/evanphx/json-patch v4.11.0+incompatible // indirect
 	github.com/felixge/httpsnoop v1.0.1 // indirect
+	github.com/go-logr/logr v0.4.0 // indirect
```
saintube, line 62: should we need the indirect reference

jasonliu747: unfortunately, we can't. looks like an imported package is using this, so `go mod tidy` will automatically add it.

PR #266, opened 2022-06-20
refactor: move runtime module under util package

jasonliu747:

users might get easily confused with `runtime` and `runtime-proxy`, so this RP moves `runtime` under `util`.


eahydra:
It may be better to put the package in pkg/koordlet, the code layout is being refactored but it will take some time, and it is appropriate to adjust the location temporarily.

/lgtm

PR #271, opened 2022-06-20
Runtime proxy docker dev

honpey:


pkg/runtimeproxy/server/docker/resexecutor.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	if resp.ContainerResources != nil {
+		p.ContainerResources = resp.ContainerResources
+		p.UpdateConfig = UpdateUpdateConfigByResource(p.UpdateConfig, resp.ContainerResources)
+	}
+}
+
+// ReConstructRequestByHookResponse update original http request by hook server's response
+// and update the
+func (p *PodResourceExecutorDocker) ReConstructRequestByHookResponse(response interface{}, request interface{}) error {
```
ZYecho, line 280: why the request is a interface {} ?

pkg/runtimeproxy/server/docker/handler.go
```diff
@@ -18,278 +18,46 @@ package docker
...
+	if resp, err := d.dispatcher.Dispatch(ctx, resExecutor.RuntimeRequestPath, config.PreHook, resExecutor.GenerateHookRequest()); err != nil {
+		d.generateErrorInfo(wr, "failed to call pre start container hook server %v", err)
 		return
-	}
-	containerConfig := &container.UpdateConfig{}
-	if err := json.Unmarshal(reqBytes, containerConfig); err != nil {
-		klog.Errorf("Failed to Unmarshal req body to docker config, err: %v", err)
-		http.Error(wr, err.Error(), http.StatusInternalServerError)
+	} else if err = resExecutor.ReConstructRequestByHookResponse(resp, req); err != nil {
+		d.generateErrorInfo(wr, "failed to reconstruct request by hook server response %v", err)
```
cheimu, line 50: I had a conversation with @ZYecho last week about the following situation where if hook gives an error response, whether proxy should directly return or simply bypass it. The reason we discuss it because we found that the docker will directly return but cri will bypass as shown by the following code: 
https://github.com/koordinator-sh/koordinator/blob/main/pkg/runtimeproxy/server/cri/criserver.go#L108 

Currently the hook failure policy is global https://github.com/koordinator-sh/koordinator/blob/main/examples/runtime-hook-server/koordlet.json#L3, so do you think it is a good idea to make the failure policy to a hook granularity? so we can check if we should bypass the specific hook when hook return an error to proxy? What do you think?

pkg/runtimeproxy/server/docker/resexecutor.go
```diff
@@ -0,0 +1,293 @@
...
+	*container.UpdateConfig
+
+	*store.PodSandboxInfo
+	*store.ContainerInfo
+
+	// key in
+	id string
+}
+
+func (p *PodResourceExecutorDocker) GetMetaInfo() string {
```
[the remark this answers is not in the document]
honpey: removed

pkg/runtimeproxy/server/docker/resexecutor.go
honpey: fixed

hormes:
please add some unittests, testability is an important part of refactoring.

PR #282, opened 2022-06-21
ci: use matrix and cache to speed up the build

jasonliu747:

save some time from building docker images.


jasonliu747:
Before optimization:
image
After:
image

PR #317, opened 2022-06-27
koord-scheduler: loadAwareScheduling skip node without NodeMetric in …

### LINKED ISSUE #316: [BUG] koord-scheduler failed to schedule Pod without NodeMetric not found
**What happened**:

Some nodes in the cluster are used for co-location. When creating Pod scheduling, it is required to schedule to non-co-location nodes. At this time, Pod cannot be scheduled, and the related error message is nodemetrics not found.

koord-scheduler scheduled  the new created Pod  successfully even if there is not NodeMetrics in the node.

**Environment**:
- App version:
- Others:

eahydra:
…filter/score phase

fix #316 

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/loadaware/load_aware.go
+++ b/pkg/scheduler/plugins/loadaware/load_aware.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


jasonliu747 (commented):
/lgtm

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -102,6 +103,9 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 
 	nodeMetric, err := p.nodeMetricLister.Get(node.Name)
 	if err != nil {
+		if errors.IsNotFound(err) {
```
hormes, line 106: please add some comments like:

For nodes that lack load information, fall back to the situation where there is no load-aware scheduling, and deal with the scenario where the user installs the koordlet component according to the grayscale of the nodes in the cluster.

eahydra: done

PR #209, opened 2022-06-02
Add fine-grained cpu orchestration proposal

eahydra:

PR #322, opened 2022-06-29
add fine-grained device scheduling proposal

buptcozy:
Signed-off-by: yangzhang <bupt_cozy@126.com>

add schedule-device-in-card-level.md

### V. Checklist


docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,143 @@
...
+type DeviceLink struct {
+	Type      string      `json:"LinkType,omitempty"`
+	Bandwidth int64       `json:"bandwidth,omitempty"`
+}
+```
+
+### RuntimeProxy
+our target is to work well with origin k8s kubelet and k8s device plugins, so:
+1. we still allow kubelet and device plugin to allocate concrete device card.
+2. in runtimeproxy, we will use pod's devive-allocation-result in annotation to replace the step1's result of container's args and envs
```
shinytang6, line 143: If doing this way, l think in runtime layer you need to know the exact env name like `NVIDIA_VISIBLE_DEVICES `, then you can overwrite the allocation, but this is different for different devices, how should you know that?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,143 @@
...
+A100 gpu card support mig function to split a card into different size of ci\gi, which provide isolate ability.
+in some scenarios, we need create ci\gi dynamically with pod-scheduling, for example, some pods need 1c.4g.40gb, some pods
+need 1c.2g.20gb, we can't expect the pods requirements in advance, so there is a demand for scheduler to make the 
+best placement for ci\gi-create and schedule pod in it.
+
+#### Story Oversold
+in some cases, we find although some cards free is zero in scheduler account, but we observed that the physical
+utilization of gpu-mem\gpu-core still has buffer, so we may schedule some low-priority jobs to use this buffer, and
+when buffer disappear, we should kill these jobs to avoid affecting normal jobs. because the utilization is different
+bewteen cards, so scheduler must schedule based card level.
```
shinytang6, line 77: Will you plan to cooperate with some gpu isolation way such like cgpu when oversold?

shinytang6 (changes requested):
That's a great proposal, thanks for the work!
I commented some of my questions, please help to answer, thanks

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,143 @@
...
+    * [RuntimeProxy](#runtimeProxy)
+
+<!--te-->
+
+## Summary
+support schedule devices like gpu\rdma\fpga in card level for better schedule placement. 
+
+## Motivation
+there is lots of cases that need scheduler make better scheule placement with devices like gpu\rdma etc.
+like gpu-nvlink in A100\V100、gpu Mig in A100.
```
hormes, line 46: scheule  -> schedule

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Goals
+1、a new program on can report card level device crd, which will not conflict with kubelet and existed device plugin
```
hormes, line 49: a new program on node ?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: sure, it's future work

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: fixed

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: different devices have relatively fixed pattern and less likely to change due to compatibility reason

jasonliu747 (commented):
Let's focus on GPU in this proposal first. And more details would be appreciated.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
shinytang6: Thanks. But in addition to these environment variables, some device plugins also set the mount and devices of the container, modifying them at a lower level may need to do some dirty work 🤔

eahydra: We don't have to consider supporting the existing DevicePlugin. As we discussed before, we can report device capacity through the koordlet.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Summary
+support schedule devices like gpu\rdma\fpga in card level for better schedule placement. 
+
+## Motivation
+there is lots of cases that need scheduler make better schedule placement with devices like gpu\rdma etc.
+like gpu-nvlink in A100\V100、gpu Mig in A100.
+
+### Goals
+1、a device reporter per node will report card level device crd, which will not conflict with kubelet and existed device plugin
```
eahydra, line 50: Recommend use markdown style. e.g. 1. 2.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
shinytang6: yeah l know you can collect device info through koordlet and make allocation decision in scheduler, but you have to know the exact ways to set correct device envs, mounts and devices in runtime layer, just as what device plugin does

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+5. schedule-result will be written on pod annotations.
+
+#### device reporter
+we need create a new device reporter on node to update a new device crd to report each device card resource and health status
+
+```go
+type DeviceType string
+
+const (
+	GPU  DeviceType = "gpu-device"
```
eahydra, line 97: Is it necessary to consider FPGA equipment? Or should we define a unified CRD for these device types?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+	NPU  DeviceType = "npu-device"
+	FPGA DeviceType = "fpga-device"
+	RDMA DeviceType = "rdma-device"
+	POV  DeviceType = "pov-device"
```
eahydra, line 101: hmm…what is the pov？

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items  []Device `json:"items"`
+}
+
+type Device struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	NodeName     string        `json:"nodeName"`
```
eahydra, line 114: When creating cr, directly use the node name as the name of cr, and there is no need to customise a separate nodeName field. This is also a common practice in the community.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+	DeviceType   DeviceType    `json:"deviceType"`
```
eahydra, line 115: The deviceType field can be moved into the spec.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,144 @@
...
+type DeviceInfo struct {
+	ID        string                       `json:"id"`       // Device UUID
+	Resources map[string]resource.Quantity `json:"resource"` // map of sub-resources
+	Health    bool                         `json:"health"`
+	Minor     int32                        `json:"minor"` // index starting from 0
+}
+
+type DeviceTopology struct {
+	// 2D array of p2p linkMeta, indexed by device minor
+	Links [][]GpuLink `json:"links,omitempty"`
```
eahydra, line 133: Typo? GpuLink -> DeviceLink?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
hormes: In order to add devices without changing the code, we may need another device plugins framework.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,141 @@
...
+	Spec         DeviceSpec    `json:"spec,omitempty"`
+}
+
+type DeviceSpec struct {
+	DeviceInfos     []DeviceInfo    `json:"deviceInfos"`
+    deviceTopology  DeviceTopology  `json:"deviceTopology"`
+}
+
+type DeviceInfo struct {
+	ID        string                       `json:"id"`       // Device UUID
```
tzzcfrank, line 123: Can we add NUMA information here? Excluding the GPU, same NUMA Node is a better choice when allocating multiple devices in a container.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,141 @@
...
+     
+
+## Implementation Details
+
+#### Scheduler
+1. we need a new data structure to describe each device card resources and healthy status on one node
+2. we need new plugins to execute filter\reserve\bind stage
+3. support user to apply any ratio of one card, like 50%/25%
+4. automatically recognize different kind devices. when a new device added, we don't need modify any code
+5. schedule-result will be written on pod annotations.
```
tzzcfrank, line 88: The schedule-result data structure needs to be defined in this proposal

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
eahydra: > In order to add devices without changing the code, we may need another device plugins framework.

+1

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
eahydra: > 1.Can we add NUMA information here? Excluding the GPU, same NUMA Node is a better choice when allocating multiple devices in a container.

There is already a proposal(#209) that uses NodeResourceTopology to describe NUMA Topology information, including such as memory, CPU and device in NUMA Node. NUMA Topology Alignment policies have  different scheduling behaviors, But It may not be appropriate to consider a certain resource individually.

tzzcfrank: I know that proposal, as you mentioned, NUMA Topology information only has resource quantity information, not resource individual information, there will be certain restrictions.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+
+2.user apply device resource:
+
+   a. if user apply {nvidia.com/gpu:1}, we translate to {gpu-core-ratio:100, gpu-mem-ratio:100， gpu-mem:0GB}. gpu-mem 0GB
+is because different gpu model has different gpu-memory. so when we filter different nodes, we will determine the 
+real gpu-mem value by node gpu memory * pod's gpu-mem-ratio in real time.
+
+   b. if user apply {koordinator/gpu:100}, we translate to {gpu-core-ratio:100, gpu-mem-ratio:100， gpu-mem:0GB}.
+
+   c. if user apply {koordinator/gpu-core:50, koordinator/gpu-mem-ratio:50}, we translate to {gpu-core-ratio:50,
```
hormes, line 113: Memory supports ratio or absolute value allocation, but (cuda) core supports ratio allocation only, which can be confusing. Keep a consistent description, so that it can be satisfied in the future when it needs to be scheduled according to the absolute cuda core number? user apply resource like this:
{koordinator/gpu-ratio:100}
{koordinator/gpu-core-ratio:20, koordinator/gpu-mem-ratio:50}

{koordinator/gpu-core:1024, koordinator/gpu-mem:5GB} // reserved for future consideration

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+
+type DeviceTopology struct {
+    Links [][]DeviceLink `json:"links,omitempty"`
+}
+type DeviceLink struct {
+    Type      string      `json:"LinkType,omitempty"`
+    Bandwidth int64       `json:"bandwidth,omitempty"`
+}
+
+this version we only provide gpu device reporter. it will execution nvidia-smi commands to get each minor resource just
```
eahydra, line 238: Is the `gpu device reporter` a daemonset? Or implement the reporter in koordlet? Need to clarify this detail.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+like k8s-gpu-device-plugins and supply some health check logic.
+```
+
+### RuntimeProxy
+our target is to work well with origin k8s kubelet and k8s device plugins, so:
+1. we still allow kubelet and device plugin to allocate concrete device card.
+2. in runtimeproxy, we will use pod's devive-allocation-result in annotation to replace the step1's result of container's
```
eahydra, line 245: Since runtimeproxy will modify the environment variables of the container, why do you need kubelet to interact with device plugin? And this proposal adds some new resource types that will not be supported by existing device plugins. Are you considering reporting resource information through the link that kubelet interacts with device plugin? If yes, why not consider implementing this reporting logic in koordlet? If not, need add alternative details.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
+---
+title: schedule device(gpu\rdma) in card level
```
eahydra, line 2: This title is confusing. Kubernetes itself supports card granularity allocation. It is recommended to change to Fine-grained GPU Management

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+
+![image](/docs/images/nvlink.jpg)
+
+through the picture we can find that although there's 8 card in A100\V100 node, the data-transport-speed between cards 
+are not same. when a pod need multi-gpu-cards, we have chance to select a better placement with largest data-transport-speed. 
+
+#### Story MIG
+![image](/docs/images/mig.jpg)
+
+A100 gpu card support mig function to split a card into different size of ci\gi, which provide isolate ability.
```
eahydra, line 68: No design details about the MIG are seen in this Proposal. If missing, please add details; if not, it is recommended to submit a new proposal separately.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+   e. when dimension's value > 100, means pod need multi-devices. now we only allow value can be divided by 100.
+
+3.device resource data structure:
+for each device, we use DeviceResource to present resources. DeviceResource's DeviceKey2ValueMap describe each minor's resources
+ ```go
+// "0": {gpu-core-ratio:100, gpu-mem-ratio:100, gpu-mem: 16GB}
+// "1": {gpu-core-ratio:100, gpu-mem-ratio:100, gpu-mem: 16GB}
+type DeviceResource struct {
+    // key is minor 
+    DeviceKey2ValueMap map[int32]map[string]resource.Quantity
```
eahydra, line 128: Do not use this XXX2YYY format for naming. And this data structure seems to be only used by scheduling plugins, it is recommended to describe in plugin related details.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+best placement for ci\gi-create and schedule pod in it.
+
+#### Story Oversold
+in some cases, we find although some cards free is zero in scheduler account, but we observed that the physical
+utilization of gpu-mem\gpu-core still has buffer, so we may schedule some low-priority jobs to use this buffer, and
+when buffer disappear, we should kill these jobs to avoid affecting normal jobs. because the utilization is different
+bewteen cards, so scheduler must schedule based card level.
+     
+
+## Implementation Details
```
eahydra, line 80: I don't see details about compatibility with already existing pods that use GPUs. This detail needs to be added, otherwise the user cannot use the Koordinator.

eahydra (changes requested):

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
eahydra: I Get it. Really need NUMA Node related information. I also commented on this question separately.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
jasonliu747: why do you need a new reporter when there is koordlet?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,340 @@
...
+  * [Unsolved Problems](#unsolved-problems)
+  * [Alternatives](#alternatives)
+  * [Implementation History](#implementation-history)
+  * [References](#references)
+  
+
+<!--te-->
+
+## Summary
+support schedule devices like gpu\rdma\fpga in card level for better schedule placement.
```
jasonliu747, line 49: the summary of this proposal is not about better schedule placement, it's about allowing scheduler to specify which card should be allocated.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,340 @@
...
+
+##### Oversold
+in some cases, we find although some cards free is zero in scheduler account, but we observed that the physical
+utilization of gpu-mem\gpu-core still has buffer, so we may schedule some low-priority jobs to use this buffer, and
+when buffer disappear, we should kill these jobs to avoid affecting normal jobs. because the utilization is different
+bewteen cards, so scheduler must schedule based card level.
+
+these cases all need scheduler support fine-gained gpu scheduling first. 
+
+#### Goals
```
jasonliu747, line 79: `#### Goals` -> `### Goals`

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,340 @@
...
+type NodeDevicePlugin struct {
+    frameworkHandler     framework.Handle
+    deviceClient         deviceClient.Interface
+    deviceLister         devicelister.DeviceLister
+    nodeLister           listerv1.NodeLister
+    nodeDeviceCache      *NodeDeviceCache
+}
+
+type NodeDevicePlugin interface{
+    MonotorLoop()
```
jasonliu747, line 231: I assume you mean MonitorLoop()?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+type DeviceTopology struct {
+    Links [][]DeviceLink `json:"links,omitempty"`
+}
+type DeviceLink struct {
+    Type      string      `json:"LinkType,omitempty"`
+    Bandwidth int64       `json:"bandwidth,omitempty"`
+}
+
+this version we only provide gpu device reporter. it will execution nvidia-smi commands to get each minor resource just
+like k8s-gpu-device-plugins and supply some health check logic.
```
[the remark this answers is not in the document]
buptcozy: at the beginning we only use k8s default health check logic.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,340 @@
...
+  
+
+<!--te-->
+
+## Summary
+support schedule devices like gpu\rdma\fpga in card level for better schedule placement. 
+
+## Motivation
+
+#### Why Need This Proposal
```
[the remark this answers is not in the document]
buptcozy: fixed

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: fixed

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
eahydra: In this case, it actually has nothing to do with this PR and can be removed directly.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,339 @@
...
+type Device struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    DeviceType   DeviceType    `json:"deviceType"`
+    Spec         DeviceSpec    `json:"spec,omitempty"`
+}
+
+type DeviceSpec struct {
+    DeviceInfos     []DeviceInfo    `json:"deviceInfos"`
```
eahydra, line 277: why removed the DeviceTopology?

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+    metav1.ListMeta `json:"metadata,omitempty"`
+    Items  []Device `json:"items"`
+}
+
+type Device struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    NodeName     string        `json:"nodeName"`
+    DeviceType   DeviceType    `json:"deviceType"`
```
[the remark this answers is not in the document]
hormes: This choice is very important, do we want multiple devices to share one cr , or one cr for each type of device. For example, in Tencent Cloud's k8s cluster, there is an upper limit on the number of CRs under a user's fixed cluster. Too many CRs may affect it.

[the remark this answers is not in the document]
buptcozy: all devices in one cr

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,268 @@
...
+}
+
+type DeviceInfo struct {
+    ID        string                       `json:"id"`       // Device UUID
+    Resources map[string]resource.Quantity `json:"resource"` // map of sub-resources
+    Health    bool                         `json:"health"`
+    Minor     int32                        `json:"minor"` // index starting from 0
+}
+
+type DeviceTopology struct {
```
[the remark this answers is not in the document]
buptcozy: add DeviceTopology back

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: deleted

jasonliu747 (changes requested):
`docs/images/mig.jpg` could be deleted, btw.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,357 @@
...
+
+### API
+
+#### device resource dimension:
+
+For GPU, we have 3 dimensions:koordinator.sh/gpu-core, koordinator.sh/gpu-memory-ratio, koordinator.sh/gpu-memory, 
+this design is for future node isolation feature. 
+
+```go
+koordinator.sh/gpu-core means calculation percentage of one card
```
yihuifeng, line 92: as described, is it better to name gpu-core as gpu-core-ratio, then add a new dimension gpu-core.
that we should has the same name-style for gpu-core and gpu-memory.

eahydra: In this proposal, it is expected to abstract the computing unit of a card into 100 copies and support the GPU Share scenario. This abstraction is similar to the concept of K8s MilliCPU. I don’t think it needs to be represented by the concept of percentage, and use `gpu-core` simple enough.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,355 @@
...
+Our design has no conflict with the above process. our device reporter will report koordinator gpu resources for kubelet
+updating node resources. Then we schedule device request in our new plugin with new device resource account. In pre-bind 
+stage, we will update container resources with koordinator gpu resources, this is for kubelet to check resource limitation.
+We will also add device allocation information to pod's annotation. In node side, the k8s device plugin will first patch
+container env, but we will overwrite these envs in runtimeproxy by allocation result in pod's annotation.
+
+If there is no gpu device plugin, the only different is in pre-bind stage, we will erase the origin-gpu-resource-dimension
+in container to pass kubelet resource limitation check. This will be a flag in scheduler side.
+
+When we hot-update, how can we handle the scheduled-pod without device allocation in pod's annotation? We can get
```
zwzhang0107, line 346: For those pods whose GPUs are already allocated by kubelet(before koordinator installation), koordlet also needs to report the allocs in CRD, so that scheduler can use them for making allocations.

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: updated

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,357 @@
...
+    Filter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status
+    Reserve(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeName string) *framework.Status
+    UnReserve(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string)
+    PreBind(ctx context.Context, state *CycleState, p *v1.Pod, nodeName string) *Status
+}
+```
+
+We will register node and device event handler to maintain device account.
+
+In Filter, we will make-up each device request by a node(the gpu-memory example), and try compare each device free
```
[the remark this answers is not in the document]
buptcozy: updated

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
buptcozy: as we discuss offline, we can get the allocation information for container's env to help recovering device cache usage

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,358 @@
...
+##### Oversold
+In some cases, we find although some cards free is zero in scheduler account, but we observed that the physical
+utilization of gpu-mem\gpu-core still has buffer, so we may schedule some low-priority jobs to use this buffer.
+When buffer disappear, we should kill these jobs to avoid affecting normal jobs. because the utilization is different
+between cards, the scheduler must have ability to schedule based on card level.
+
+These cases all need scheduler have ability to schedule based on card level.
+
+### Goals
+1.Device reporter will report card level device crd, which will not conflict with kubelet and existed device-plugin.
```
eahydra, line 73: Please add Definition Device CRD API and the Resource Definition and Specification in the goals

docs/proposals/scheduling/20220629-schedule-device-in-card-level.md
```diff
@@ -0,0 +1,358 @@
...
+  * [Unsolved Problems](#unsolved-problems)
+  * [Alternatives](#alternatives)
+  * [Implementation History](#implementation-history)
+  * [References](#references)
+  
+
+<!--te-->
+
+## Summary
+Support schedule devices like gpu\rdma\fpga in card level for better schedule placement.
```
eahydra, line 48: Need to clarify what exactly "card level" means.

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
```diff
@@ -0,0 +1,419 @@
...
+    cpu: "4"
+    memory: "8Gi"
+```
+
+The scheduler translates the `nvida.com/gpu: 1` to the following spec:
+
+```yaml
+resources:
+  requests:
+    koordinator.sh/gpu-core: "200"
```
jasonliu747, line 145: A little bit confused here. Please correct me if I'm wrong, doesn't `koordinator.sh/gpu-core: "200"` mean 2 GPU cards?

buptcozy: yes.

eahydra (approved):
/lgtm

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
```diff
@@ -0,0 +1,419 @@
...
+half or quarter percentage of memory, so user can request the GPU memory with `koordinator.sh/gpu-memory-ratio`. 
+When scheduler assigned Pod on concrete node, scheduler will translate the `koordinator.sh/gpu-memory-ratio` to `koordinator.sh/gpu-memory` by the formulas:  ***allocatedMemory = totalMemoryOf(GPU)  * `koordinator.sh/gpu-memory-ratio`***, so that the GPU isolation can work.
+
+During the scheduling filter phase, the scheduler will do special processing for `koordinator.sh/gpu-memory` and `koordinator.sh/gpu-memory-ratio`. When a Pod specifies `koordinator.sh/gpu-memory-ratio`, the scheduler checks each GPU instance on each node for unallocated or remaining resources to ensure that the remaining memory on each GPU instance meets the ratio requirement.
+
+If the user knows exactly or can roughly estimate the specific memory consumption of the workload, he can apply for GPU memory through `koordinator.sh/gpu-memory`. All details can be seen below.
+
+Besides, when dimension's value > 100, means Pod need multi-devices. now only allow the value can be divided by 100.
```
caohe, line 126: If the value of a container's gpu-core is greater than 100 and cannot be divided by 100 (e.g. 101), will the pod be rejected?

jasonliu747: yes, should be rejected.

caohe: is it rejected by the webhook, or by the scheduler?

jasonliu747: by scheduler in Prefilter step.

jasonliu747: koordinator.sh/gpu-core should be something like: 25, 50, 75, 100, 200, 300. otherwise it will be rejected by scheduler in Prefilter step.

caohe: if a pod is rejected by the scheduler, it will go into `Pending` phase, and the scheduler will keep retrying to schedule the pod. I think this retry may be useless, and may increase the load on the scheduler. would it be better to reject the pod in the webhook?

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
```diff
@@ -0,0 +1,419 @@
...
+    koordinator.sh/fpga: 100
+  allocatable:
+    koordinator.sh/rdma: 100
+    koordinator.sh/fpga: 100
+```
+
+Why do we need `koordinator.sh/gpu-memory-ratio` and `koordinator.sh/gpu-memory` ? 
+When user apply 0.5/0.25 GPU, the user don't know the exact memory total bytes per GPU, only wants to use 
+half or quarter percentage of memory, so user can request the GPU memory with `koordinator.sh/gpu-memory-ratio`. 
+When scheduler assigned Pod on concrete node, scheduler will translate the `koordinator.sh/gpu-memory-ratio` to `koordinator.sh/gpu-memory` by the formulas:  ***allocatedMemory = totalMemoryOf(GPU)  * `koordinator.sh/gpu-memory-ratio`***, so that the GPU isolation can work.
```
caohe, line 120: > scheduler will translate the `koordinator.sh/gpu-memory-ratio` to `koordinator.sh/gpu-memory`

Does this mean that scheduler will call kube-apiserver to update the pod's spec?

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
```diff
@@ -0,0 +1,419 @@
...
+```yaml
+resources:
+  requests:
+    koordinator.sh/gpu-core: "50"
+    koordinator.sh/gpu-memory-ratio: "60"
+    cpu: "4"
+    memory: "8Gi"
+```
+
+##### Apply `koordinator.sh/gpu-core` and `koordinator.sh/gpu-memory` separately
```
caohe, line 185: if a container only requests `gpu-core` and `gpu-memory`, will the amount of `gpu-memory-ratio` resources for the node in the scheduler cache be incorrect? because `gpu-memory-ratio` resources may be not assumed.

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
```diff
@@ -0,0 +1,419 @@
...
+Besides, the koordlet should report GPU model to node labels same as device plugin, this is in-case koordinator working without device-plugin.
+
+Finally, we should modify `ContainerResourceExecutor`'s `UpdateRequest` function in koord-runtime-proxy, and let new GPU env covering old GPU env.
+
+### Compatibility
+
+As we know, the GPU scheduling in kube-scheduler side has no any different with other scalar resources. The concrete 
+device-level assigning is done by kubelet and GPU device plugin, which will generate container's GPU env. 
+
+Our design has no conflict with the above process. our device reporter will report koordinator GPU resources for kubelet
```
caohe, line 402: > our device reporter will report koordinator GPU resources for kubelet updating node resources

how does device reporter report koordinator GPU resources to kubelet? 

does this mean that device reporter still implements a device plugin for koordinator GPU resources?

caohe: resolved after reading https://github.com/koordinator-sh/koordinator/pull/410

caohe:
@jasonliu747 @zwzhang0107 hello, I have figured out some of these questions after reading #410. However, I still have some doubts about the scheduler:

1. Could you provide more details about the resource translation mechanism?
2. How to ensure that the the quantities of node-level extended resources in the core cache is correct, if a pod only requests some kinds of these resources?

Thanks for any reply!

jasonliu747:
@caohe if it's ok for you, let's discuss this on WeChat or DingTalk. WDYT?

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
buptcozy: would it would be better to reject the pod in the webhook.

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
buptcozy: it will update container resources.

docs/proposals/scheduling/20220629-fine-grained-device-scheduling.md
buptcozy: we will translate gpu-memory-ratio based on gpu-memory on concrete node.

caohe:
@jasonliu747 sure, happy to discuss this on WeChat or DingTalk.

jasonliu747:
> @jasonliu747 sure, happy to discuss this on WeChat or DingTalk.

@caohe You can find our DingTalk QR Code in [README](https://github.com/koordinator-sh/koordinator#community). Please PM me once you join the group, below is my DingTalk avatar. Thanks.
image

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

JasonRD:
Preparing PR description...

PR #325, opened 2022-06-29
Fix wrong cgroup path for PLEG

cheimu:
When watch pod lifecycle, the pleg manager uses a wrong cgroup path. The expected cgroup path is `/sys/fs/cgroup/<subsystem>/kubepods/<qosclass>/...`, but now it only gets `/sys/fs/cgroup/kubepods/<qosclass>/...` which will return error. This pr fixes this issue.

Since one container will have records in all subsystem cgroup hierarchy, while we only need to watch one of them, since cpu subsystem must and always exist, we chould use cpu cgroup for monitoring.

### V. Checklist


jasonliu747 (approved):
/lgtm

jasonliu747 (approved):
/lgtm

PR #333, opened 2022-07-01
add proposal for gang scheduling

buptcozy:
### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20220701-schedule-gang.md
+++ b/docs/proposals/scheduling/20220701-schedule-gang.md
@@ -1,0 +1,435 @@
+---
+title: Gang scheduling
+authors:
+  - "@buptcozy"
+reviewers:
+  - "@eahydra"
+  - "@hormes"
+  - "@yihuifeng"
+  - "@honpey"
+  - "@zwzhang0107"
+  - "@jasonliu747"
+creation-date: 2022-07-01
+last-updated: 2022-07-01
+
+---
+
+# Gang scheduling
+
+## Table of Contents
+
+<!--ts-->
+
+* [Gang scheduling](#Gang-scheduling)
+    * [Table of Contents](#table-of-contents)
+    * [Summary](#summary)
+    * [Motivation](#motivation)
+      * [Compared with competitors](#Compared-with-competitors)
+        * [Coescheduling](#Coescheduling)
+      * [Goals](#goals)
+      * [Non Goals and Future Work](#Non-Goals-and-Future-Work)
+    * [Proposal](#Proposal)
+      * [Key concept](#key-concept)
+        * [Strict-mode and non-strict-mode](#strict-mode-and-non-strict-mode)
+        * [GangGroup](#ganggroup)
+        * [After gang](#after-gang)
+      * [API](#API)
+        * [Definition](#definition)
+          * [CRD way](#crd-way)
+            * [Example](#example)  
+          * [Annotation way](#annotation-way)
+            * [Example](#example)         
+      * [Implementation Details](#Implementation-Details)
+        * [QueueSortPlugin](#QueueSortPlugin)
+        * [Data-Structure](#data-structure)
+        * [GangPlugin](#gang-plugin)
+    * [Unsolved Problems](#Unsolved-Problems)
+    * [Alternatives](#Alternatives)
+    * [Implementation History](#Implementation-History)
+    * [References](#References)
+<!--te-->
+
+## Summary
+This proposal provides Gang mechanism for the scheduler to control pods binding opportunity. User can declare a resource-collection-minimum number, 
+only when assigned-resources reach the given limitation can trigger the binding. We provide `strict-mode` and `non-strict-mode` to 
+control the resource-accumulation-process by a configuration. We also provide a two-level Gang description for better matching 
+the real scenario, which is different from community.
+
+## Motivation
+In AI scenarios, lots of jobs need Gang scheduling. The community have lots of related implements such as `coescheduling` or `vocalno`.
+We received lots of inspirations in the design process from them.
+
+### Compared with competitors
+
+#### Coescheduling
+1. `coescheduling` implement a new queue-sort interface and other methods to let one Gang's pods get out of the queue in order as much as possible.
+If a pod failed to be scheduled, the requests that have been successfully scheduled in this round of Gang scheduling cycle will be rolled back,
+and the remaining pods waiting for scheduling will be rejected in PreFilter check until this scheduling cycle passed. 
+For example, there is a Gang requires 10 tasks to be scheduled, if first 5 tasks allocated, the 6th task failed to be scheduled,
+`coescheduling` will roll-back first 5 tasks and ignore the remaining 4 tasks in this Gang scheduling cycle. `coescheduling` simply use a 
+global time interval to control the Gang scheduling cycle. The first defect is that the uniform time interval will cause 
+some problems. If the time configuration is too long, it will lead to useless waiting; If the time configuration is too short, 
+it will lead to useless scheduling. Secondly, it is very difficult for a large job to meet all resource requests at one time. 
+This mechanism will lead to a very low probability of full resources, and eventually make the job starve to death. We call this process as `strict-mode`.
+
+2. Some jobs have complex Gang requirements. For example, a job has several roles. Each role will have several pods 
+and its own Gang conditions. Jobs also need different roles to form different GangGroups. All pods in a GangGroup can 
+trigger the bind process only after all roles in a GangGroup meet their Gang conditions. The `coescheduling` can't meet
+this requirement.
+
+### Goals
+1. Define API to announce Gang scheduling configuration.
+
+2. Provides a scheduler plugin to achieve Gang scheduling ability.
+
+### Non Goals and Future Work
+1. Provide ability to solve Gang resource deadlock problems with `non-strict-mode`.
+
+## Proposal
+
+### Key concept
+
+#### strict-mode and non-strict-mode
+
+As mentioned above, in `strict-mode`, if a pod failed to be scheduled, the pods that have been successfully scheduled in 
+this scheduling cycle will be rolled back, and the remaining pods waiting for scheduling will be rejected in 
+PreFilter check util this scheduling cycle passed. We call this mode is `strict-mode`.
+
+In `non-strict-mode`, if a pod failed to be scheduled, it has no impact on any other pod. We will continue to accumulate 
+the allocated pod until the condition of Gang is met. This process is friendly to Gangs with large number of pods, but it 
+will increase the risk of resource deadlock between Gangs. For example, the quota of the quota group is 10(quota will be proposed later), 
+and the user submits three Gangs with 5 pods. Due to various plugin constraints, Gang1\2\3 may allocate resources of 3\3\4 respectively. 
+Since the quota group's quota is full, there will be no new resource scheduling. We call this is resource deadlock of resource Gang.
+In future proposal, we will try to fix this problem.
+
+#### GangGroup
+As mentioned above, Some jobs have complex Gang requirements. For example, a job has several roles. Each role will have several pods 
+and its own Gang conditions. Jobs also need different roles to form different GangGroups. All pods in a GangGroup can 
+trigger the bind process only after all roles in a GangGroup meet their Gang conditions. So we introduce `GangGroup` concept,
+which allow user to bundle different Gangs together.
+
+#### After Gang
+It should be noted that, if the resource accumulation conditions of Gang are met, then some pods failed in the process of binding,
+or some bound pods are preempted\rescheduled, should the constraints of Gang still be effective in the process of resource reallocation? 
+Because the initial purpose of Gang is to require pods to be pulled up at the same time, if some pods have been pulled up, 
+then the subsequent Gang behavior is meaningless. Therefore, when once Gang has been satisfied, all subsequent resource allocations 
+are no longer constrained by Gang rules, and their performance is similar to ordinary pod.
+
+As mentioned above, `WaitTime` is the max wait time since first pod comes to permit stage. If `WaitTime` is timeout, 
+scheduler will roll back all assumed pods, update each pod's annotation with `gang.scheduling.koordinator.sh/timeout=true`, and
+won't schedule these pods anymore. User should pay attention to this status and delete pods timely.
+
+### API
+#### Definition
+
+Our original intention is to improve and enhance the ability of the community's original `PodGroup`, so we will be 
+compatible with the way the community declares the `PodGroup`. We also provide a lighting way to just use annotations to 
+use Gang feature.
+
+#### CRD way
+User can use `PodGroup` CRD in community to declare a gang:
+```go
+type PodGroup struct {
+    metav1.TypeMeta `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    Spec PodGroupSpec `json:"spec,omitempty"`
+    Status PodGroupStatus `json:"status,omitempty"`
+}
+type PodGroupSpec struct {
+    MinMember int32 `json:"minMember,omitempty"`
+    MinResources *v1.ResourceList `json:"minResources,omitempty"`
+    
+    ScheduleTimeoutSeconds *int32 `json:"scheduleTimeoutSeconds,omitempty"`
+}
+```
+Pod should use `pod-group.scheduling.sigs.k8s.io` in label to associate with `PodGroup`.
+
+Also, we introduce some optional definitions as below:
+```yaml
+gang.scheduling.koordinator.sh/total-number
+gang.scheduling.koordinator.sh/gang-mode        
+gang.scheduling.koordinator.sh/groups
+```
+
+- `gang.scheduling.koordinator.sh/total-number` helps to calculate Gang scheduling cycle in `strict mode`, you can 
+find more detail in `Data-Structure` chapter. Default equals to `gang.scheduling.koordinator.sh/min-available`.
+
+- `gang.scheduling.koordinator.sh/gang-mode` determines `strict-mode` or `non-strict-mode`. Default is `strict-mode`.
+
+- `gang.scheduling.koordinator.sh/groups` describes GangGroups. Default is empty, which means don't need to form a `GangGroup` with others.
+
+`gang.scheduling.koordinator.sh/total-number`, `gang.scheduling.koordinator.sh/gang-mode`, `gang.scheduling.koordinator.sh/gang-groups` should be found in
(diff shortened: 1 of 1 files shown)


docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,156 @@
...
+3. scheduler should provide gang data-structure and schedule-plugin 
+
+### User Stories
+
+#### Story AI
+lots of AI jobs such as tf\pytorch need gang-scheduling
+     
+## Implementation Details
+### pod annotation to describe gang
+we choose to create gang-related by pod's annotations instead of gang crd. This is because we don't want high level
```
saintube, line 60: Could you elaborate more about why we choose annotations but not a CRD?
For instance, what is the case we need a high-level operator to maintain CRD lifecycles, what are the functional differences between the upstream implementation...
We may need more details about the comparisons with the upstream gang scheduling ([co-scheduling](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/coscheduling)).

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,156 @@
...
+<!--te-->
+
+## Summary
+gang scheduling
+
+## Motivation
+in AI scenario, we may want worker launch at same time.
+
+### Goals
+1. propose some pod's annotations to announce gang properties
```
jasonliu747, line 47: 1. What's the difference between this proposl and [coscheduling](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/coscheduling/README.md)
2. What's the difference between this proposal and [volcano](https://github.com/volcano-sh/volcano)
3. Why do we need a CRD here? What's the pros and cons?
4. Why do we use annotations here? What's the pros and cons?
5. Could you elaborate each field in your CRD? So ppl would get a better idea.
6. What's the responsibilities of each stage in scheduling framework. PreFilter\Filter\Allocate\Permit.
7. A sequence diagram could explain a lot. Better have one.
8. What kind of YAML shoud user submit when he needs this feature? Could you give us a simple example?
9. Is there any limitations/alternatives of this design?

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: updated

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: updated

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,280 @@
...
+##### detail
+We introduce Gang and BundleInfo for record gang status in scheduler memory. We especially explain ScheduleCycle and ChildrenScheduleRoundMap.
+These variables control bundle's scheduling cycle. at the beginning, ScheduleCycle is 1. when each pod comes to pre-filter, we will check pod's value in 
+ChildrenScheduleRoundMap is euqal to the  ScheduleCycle which means they are scheduled  at the same peace. If so, we continue to do the preFilter logic. If they are  not euqal, we set the pod's cycle in ChildrenScheduleRoundMap equal with ScheduleCycle. Finally,  when the last pod comes to 
+make all ChildrenScheduleRoundMap's value equal to ScheduleCycle, ScheduleCycle added by 1, which means a new schedule cycle.
+
+We continue to explain ScheduleCycleValid. in strict-mode, when a pod failed for scheduling, we will set ScheduleCycleValid to false in post-filter, which means in 
+this schedule cycle, the remaining pods should be rejected in pre-filter stage. when ScheduleCycle added by 1, we will reset the ScheduleCycleValid to true.
+```go
+type Gang struct {
```
hormes, line 168: It is necessary to go in the same direction with upstream, because the open source project ecology must be compatible with the concept of the business community, so that the TF Job ecology is easier to connect to the Koordinator.

Some general suggestions:
1、Use the atomic concept of PodGroup
2、Make enhancements on top of PodGroup, such as adding multiple PodGroup gangScheduling semantics (similar to the Bundle concept here)
3、For users without complex requirements, no modifications from upstream to koordinator

buptcozy: I think podgroup and gang are different,  podgroup means a batch of pods have same scheduling-policy.  gang's only limit the bind condition,  so in some features like `nvlink` or `gpu-pods-run-in-same-gpu-model-nodes`,  I think podgroup may be quoted.  Also as you mentioned,   if we introduce the bundle concept, the data-structure will be a little bigger different than PodGroup in community.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+2. Provides a scheduler plugin to achieve gang-scheduling ability.
+
+### Non Goals and Future Work
+1. Provide ability to solve gang-resource-deadlock problems with `non-strict-mode`.
+
+## Proposal
+### API
+We recommend users declaring gang-scheduling-configuration by pod's annotation. First Reason, high level operator has no need 
+to maintain gang-crd's life circle like handling `update/create/delete` events. Second Reason, from a Scheduler perspective, 
+it's inconvenient to maintain receive-order-issue's between gang-crd and pod. According to practical experience, pod's annotation
```
eahydra, line 89: Can you give more details about `receive-order-issue`? Why use the annotation form to declare Gang without this problem?

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+1. Provide ability to solve gang-resource-deadlock problems with `non-strict-mode`.
+
+## Proposal
+### API
+We recommend users declaring gang-scheduling-configuration by pod's annotation. First Reason, high level operator has no need 
+to maintain gang-crd's life circle like handling `update/create/delete` events. Second Reason, from a Scheduler perspective, 
+it's inconvenient to maintain receive-order-issue's between gang-crd and pod. According to practical experience, pod's annotation 
+is enough for declaring gang-scheduling-configuration,so we build the gang's configuration according to the first pod which declared the gang.
+
+- `koordinater.io.gang/name`                gang name.
```
eahydra, line 92: The domain name prefix API generally follows the DNS specification. Taking this as an example, it is recommended to use `scheduling.koordinator.sh/gangName` Or `gang.scheduling.koordinator.sh/name`

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+- `koordinater.io.gang/bundleName`          equals to roleName.
```
eahydra, line 93: Please give more details about `bundleName`.   I support @hormes ' suggestion. We should provide atomic gang capabilities. For the concept of bundle, it can be managed at the upper level. As far as I know, when a complex job requires multiple roles, there is a dependency order between roles. Simply declaring bundleName to separate roles does not solve this dependency order problem. And according to this bundle concept, the final object of Gang Scheduling is to process {GangName, BundleName}

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+- `koordinater.io.gang/waitingTime`         max wait time since first pod comes to permit stage.
+- `koordinater.io.gang/minRequiredNumber`   the minimum pod's number satisfying bundle condition.
+- `koordinater.io.gang/totalChildrenNumber` the total pod's number in a bundle.
```
eahydra, line 96: Why do we need `totalChildrenNumber`?

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+- `koordinater.io.gang/isStrictMode`        work in `strict-mode` or `non-strict-mode`.
```
eahydra, line 97: It's better to rename `isStrictMode` to `mode`,  and define the values including `StrictMode` and `NonStrictMode`.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+- `koordinater.io.gang/ganggroups`          describe gang groups.
```
eahydra, line 98: Hmm... What's the ganggroups??

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+#### QueueSortPlugin
+
+We design an independent plugin to implement the `QueueSort` extension point separately, so that we can integrate 
+queue sort logic of all plugins, and register them at one time.
+
+In this proposal, we implement the Less function to gather pods belong to same gang. The specific queuing rule is:
+
+1. Firstly, compare the priorities of the two pods, the higher priority is at the front of the queue.
+
+2. Secondly, compare create-time-stamp of two pods, if pod belongs to a gang, then we compare create-time-stamp of the gang,
```
eahydra, line 166: `create-time-stamp` -> `creationTimestamp`

docs/proposals/scheduling/20220701-schedule-gang.md
eahydra: `PodGroup` is a bad name. But it is indeed the API that the community currently uses to support Gang Scheduling. I also suggest we do enhancements on PodGroup.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,341 @@
...
+
+It should be noted that, if the resource accumulation conditions of gang are met, then some pods failed in the process of binding;
+Or some bound pods are preempted\rescheduled, should the constraints of gang still be effective in the process of resource reallocation? 
+Because the initial purpose of gang is to require pods to be pulled up at the same time, if some pods have been pulled up, 
+then the subsequent gang behavior is meaningless. Therefore, when `resourcequalified=true`, all subsequent resource allocations 
+are no longer constrained by gang rules, and their performance is similar to ordinary pod.
+
+As mentioned above, `WaitTime` is the max wait time since first pod comes to permit stage. if `WaitTime` is timeout, 
+scheduler will roll back all assumed pods, update each pod's annotation with `koordinater.io.gang/timeout=true`, and
+won't schedule these pods anymore. User should pay attention to this status and delete pods timely.
```
eahydra, line 206: "User should pay attention to this status and delete pods timely", but `We design the gang to record gang status in scheduler memory `. So is the gang status just in memory? so how to the user pay attention to the status? It's confusing here.

docs/proposals/scheduling/20220701-schedule-gang.md
eahydra: We alos should consider [lightweight-coscheduling](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/2-lightweight-coscheduling) compatible with the community. We should consider the cooperation with the surrounding ecology to reduce the complexity of integrating Koordinator with the surrounding ecology.

hormes: > podgroup means a batch of pods have same scheduling-policy

The PodGroup of the community is just a group and does not carry a scheduling policy.

hormes: https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/42-podgroup-coscheduling/README.md

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: through monitor pod's annotation of `koordinater.io.gang/timeout=true`,  we just want to have a way to notify user gang is timeout.

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: I modified as you wish.

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: I modified as your wish.

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: explained in proposal.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,384 @@
...
+- If `strict-mode`, we will set `scheduleCycleValid` to false and release all assumed pods.
+
+- If `non-strict mode`, we will do nothing.
+
+###### **Permit**
+
+Any pod passes Filter stage will come to this stage. Scheduler will calculate all Gangs in GangGroup whether the current 
+number of assumed-pods in each Gang meets the Gang's minimum requirement.
+
+- If Gang don't meet the bind-condition, we will give the pod a "Wait" Status with a timeout duration, and the bind
```
saintube, line 353: Please notice the maxTimeout is `15 minutes`, so we may need to hook `RunPermitPlugins` to enlarge the timeout when `15 minutes` is not enough.

buptcozy: thanks! updated

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,471 @@
...
+apiVersion: v1alpha1
+kind: PodGroup
+metadata:
+  creationTimestamp: "2022-07-11T18:26:33Z"
+  name: test
+  namespace: default
+  annotations:
+    gang.scheduling.koordinator.sh/total-number: 5,
+    gang.scheduling.koordinator.sh/gang-mode: strict-mode,
+    gang.scheduling.koordinator.sh/gang-groups": [[gangA, gangB]],
```
hormes, line 308: gangA should be the name 'test'

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,471 @@
...
+or
+
+pod-group.scheduling.sigs.k8s.io         
+pod-group.scheduling.sigs.k8s.io/min-available
+
+or
+
+gang.scheduling.koordinator.sh/name           
+gang.scheduling.koordinator.sh/min-available
+```
```
hormes, line 150: lightweight interface can be customized, use gang.scheduling.koordinator.sh/name to be consistent with other fields.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,471 @@
...
+
+Our original intention is to improve and enhance the ability of the community's original `PodGroup`, so we will be 
+compatible with the way the community declares the `PodGroup`.
+
+We recommend users declaring Gang scheduling configuration by pod's annotation. First Reason, high level operator has no need 
+to maintain Gang crd life circle like handling `update/create/delete` events. Second Reason, from a Scheduler perspective, 
+it's inconvenient to maintain receive-order-issue's between Gang crd and pod. For example, if pods arrive to scheduler before Gang crd,
+we have to build a fake Gang data structure temporarily to collect all related pods, and need to suspend the scheduling of pods 
+until parse the configuration from real Gang crd. According to practical experience, pod's annotation is enough for declaring Gang 
+scheduling configuration, so we build the Gang's configuration according to the first pod which declared the Gang.
```
hormes, line 134: It is recommended to adjust the order here:
1、first describe the user story
2、describe the extension of annotation on the PodGroup to meet the needs of these scenarios
3、finally describe the way of annotation and what its benefits are. But don't make recommendations in proposals, this tendency is too obvious.

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Summary
+This proposal provides Gang mechanism for the scheduler to control pods binding opportunity. User can declare a resource-collection-minimum number, 
+only when assigned-resources reach the given limitation can trigger the binding. We provide `strict-mode` and `non-strict-mode` to 
+control the resource-accumulation-process by a configuration. We also provide a two-level Gang description for better matching 
+the real scenario, which is different from community.
+
+## Motivation
+In AI scenarios, lots of jobs need Gang scheduling. The community have lots of related implements such as `coescheduling` or `vocalno`.
+However, in practice we find that the above solutions do not meet the needs of the business in some places.
```
hormes, line 60: Don't diss other solutions, praise them and enhance it.

docs/proposals/scheduling/20220701-schedule-gang.md
```diff
@@ -0,0 +1,421 @@
...
+won't schedule these pods anymore. User should pay attention to this status and delete pods timely.
+
+### API
+#### Definition
+
+Our original intention is to improve and enhance the ability of the community's original `PodGroup`, so we will be 
+compatible with the way the community declares the `PodGroup`. We also provide a lighting way to just use annotations to 
+use Gang feature.
+
+#### Annotation way
```
hormes, line 129: Move the annotation way under the PodGroup? This will make it easier for users to understand why this mode is supported. In addition, when the annotation mode is used, you can ignore the expression of the compatible community (because the community has migrated to PodGroup).

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

buptcozy: fixed

docs/proposals/scheduling/20220701-schedule-gang.md
buptcozy: fixed

PR #328, opened 2022-06-29
chore: refactor the structure of logging

rudeigerc:
Signed-off-by: Yuchen Cheng <rudeigerc@gmail.com>

### Ⅰ. Describe what this PR does

This pull request refactors the structure of logging according to the [documentation](https://github.com/koordinator-sh/koordinator/blob/main/docs/contributors/logging.md).

### Ⅱ. Does this pull request fix one issue?

Related to #222 

### Ⅲ. Describe how to verify it

N/A

### Ⅳ. Special notes for reviews

I follow some practical rules to decide which logging level the log should be. Please feel free to leave reviews if you consider that they are not suitable.

- `klog.V(1)`
  - The starting and shutting down of the main components.
  - Information related to configurations.
- `klog.V(2)`
  - The changing of state.
- `klog.V(4)`
  - The logging of values for debugging.
- `klog.V(5)`
  - Some tracing info.

### V. Checklist

- [ ] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [x] All checks passed in `make test`

PR #338, opened 2022-07-04
improve koordlet log verbosity

saintube:

improve the log verbosity of koordlet.

For example, when we use Centos 7, we don't expect an error-like log for the cgroup file `cpu.burst_us`, since this file may not exist on OSs other than Anolis OS.

reduce misunderstandings when koordlet is deployed with the default log level

some refactoring may conflict with #328 

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/cgroup_reconcile.go
+++ b/pkg/koordlet/resmanager/cgroup_reconcile.go
@@ -467,6 +467,8 @@ func makeCgroupResourcesForAnolis(owner *OwnerRef, parentDir string, summary *cg
 	var resources []MergeableResourceUpdater
 
 	if !system.HostSystemInfo.IsAnolisOS {
+		klog.V(5).Infof("ignored cgroup resources which required non Anolis OS, owner: %v, parentDir: %v",
+			owner, parentDir)
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/koordlet/resmanager/memory_evict.go
```diff
@@ -51,11 +51,11 @@ func NewMemoryEvictor(mgr *resmanager) *MemoryEvictor {
 }
 
 func (m *MemoryEvictor) memoryEvict() {
-	klog.Infof("starting memory evict process")
-	defer klog.Infof("memory evict process completed")
+	klog.V(4).Infof("starting memory evict process")
+	defer klog.V(5).Infof("memory evict process completed")
```
jasonliu747, line 55: shouldn't these two logs have the same level?

pkg/koordlet/resmanager/cgroup_reconcile.go
```diff
@@ -467,6 +467,8 @@ func makeCgroupResourcesForAnolis(owner *OwnerRef, parentDir string, summary *cg
 	var resources []MergeableResourceUpdater
 
 	if !system.HostSystemInfo.IsAnolisOS {
+		klog.V(6).Infof("ignored cgroup resources which required non Anolis OS, owner: %v, parentDir: %v",
```
jasonliu747, line 470: according to our convention, the highest level should be 5.

saintube: done

PR #344, opened 2022-07-06
Add pod uid to pod meta when failover

cheimu:
Currently, when proxy restart, the pod uids are not updated to cache, so hooks won't get any information about the id. This PR will add uids when proxy failover

### V. Checklist


ZYecho:
good catch!
LGTM

PR #353, opened 2022-07-08
koord-scheduler: add reservation plugin

### LINKED ISSUE #188: [proposal] Support resource reservation mechanism
**What is your proposal**:

Support resource reservation mechanism.

**Why is this needed**:

There are some scenarios where this mechanism needs to be used:
- Preemption. Through the reservation mechanism, it can be guaranteed that Pods with the same priority must have resources available after preempting other Pods
- Descheduling. When descheduling, it is necessary to ensure that the descheduled Pod has resources available, which can better ensure the reliability of the application.
- Scaling. When the PaaS layer scaling replicas, it hopes to have a definite result, that is, whether there is enough capacity in the current cluster to expand new Pods.
- Reserve resources in advance. In some resource operation scenarios, some resources need to be reserved in advance to ensure future resource demands.

The support of resource reservation can be split into issues as follow:

1. [X] Propose the design of the reservation mechanism: #241 
2. [x] Define the Reservation API: #276 
3. [ ] Implement the reservation scheduling plugin.

saintube:

implement scheduling plugin and extend framework for resource reservation

fixes #188 

submit reservation CR and check the scheduling result

Pre-Allocation is not ready in this version.

### V. Checklist


apis/scheduling/config/types.go
```diff
@@ -113,3 +113,13 @@ const (
 	// NUMADistributeEvenly indicates that evenly distribute CPUs across NUMA Nodes.
 	NUMADistributeEvenly NUMAAllocateStrategy = "DistributeEvenly"
 )
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ReservationArgs holds arguments used to configure the Reservation plugin.
+type ReservationArgs struct {
```
eahydra, line 120: A type with the same name also needs to be defined in `config/v1beta2`.

apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -32,20 +32,26 @@ type ReservationSpec struct {
...
 	Owners []ReservationOwner `json:"owners,omitempty"`
 	// By default, the resources requirements of reservation (specified in `template.spec`) is filtered by whether the
 	// node has sufficient free resources (i.e. ReservationRequest <  NodeFree).
 	// When `preAllocation` is set, the scheduler will skip this validation and allow overcommitment. The scheduled
 	// reservation would be waiting to be available until free resources are sufficient.
+	// +optional
 	PreAllocation bool `json:"preAllocation,omitempty"`
 	// Time-to-Live period for the reservation.
 	// `expires` and `ttl` are mutually exclusive. Defaults to 24h.
+	// +kubebuilder:default='24h'
```
eahydra, line 49: It is recommended to turn the changes of the Reservation CRD into a separate PR.

eahydra, line 49: Hmm.. There is a scenario where some components don't want to set the expiration time, and the components that create the Reservation is responsible for destroying it. So is it reasonable to set a default time?

apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -123,9 +149,8 @@ type ReservationCondition struct {
 // Reservation is the Schema for the reservation API
 type Reservation struct {
 	metav1.TypeMeta `json:",inline"`
-	// A Reservation object is non-namespaced.
-	// It can reserve resources for pods of any namespace. Any affinity/anti-affinity of reservation scheduling can be
-	// specified in the pod template.
+	// A Reservation object is namespaced. But it can reserve resources for pods of any namespace. Any
```
eahydra, line 152: it's better to move the comment to the head of Reservation definition.

eahydra (changes requested):

pkg/scheduler/plugins/reservation/rpod.go
```diff
@@ -0,0 +1,74 @@
...
+var (
+	// AnnotationReservePod indicates whether the pod is a reserved pod.
+	AnnotationReservePod = apiext.SchedulingDomainPrefix + "/reserve-pod"
+	// AnnotationReservationNode indicates the node name if the reservation specifies a node.
+	AnnotationReservationNode = apiext.SchedulingDomainPrefix + "/reservation-node"
+)
+
+// GetReservePod returns a fake pod set as the reservation's specifications.
+// The reserve pod is only visible for the scheduler and does not make actual creation on nodes.
+func GetReservePod(r *schedulingv1alpha1.Reservation) *corev1.Pod {
```
eahydra, line 35: The name `GetReservePod` is ambiguous, NewReservationPod may be better.

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func addReservation(sched *scheduler.Scheduler, r *schedulingv1alpha1.Reservation) {
+	if r == nil || r.Spec.Template == nil {
+		klog.Errorf("failed to add reservation for invalid reservation %v", r)
+		return
+	}
+	reservePod := reservation.GetReservePod(r)
+	// TBD: should the reservation with nodeName specified enqueue?
+	if reservePod.Status.Phase != schedulingv1alpha1.ReasonReservationExpired {
```
eahydra, line 76: check the reservePod.Status.Phase whether is ReasonReservationExpired? is it a typo?
I guess you want to check reservation.status.phase.

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,132 @@
...
+		err := sched.SchedulingQueue.Add(reservePod)
+		if err != nil {
+			klog.Errorf("failed to add reserve pod into scheduling queue, reservation %v, err: %v", reservation.GetReservationNamespacedName(r), err)
+		}
+		klog.V(4).Infof("add reservation %v", reservation.GetReservationNamespacedName(r))
```
eahydra, line 81: if successfully add reservePod into SchedulingQueue, log it..otherwise...

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func updateReservation(sched *scheduler.Scheduler, oldR, newR *schedulingv1alpha1.Reservation) {
+	if oldR == nil || newR == nil {
+		klog.Errorf("failed to update reservation for invalid reservation old %v, new %v", oldR, newR)
+		return
+	}
+	// only react to status.phase change
+	if oldR.Status.Phase == newR.Status.Phase {
```
eahydra, line 91: If the users modify the ReservationSpec, what should be done?  Or do we allow users to modify the ReservationSpec?

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,132 @@
...
+		klog.V(5).Infof("skip for status.phase of reservation %v not change", reservation.GetReservationNamespacedName(newR))
+		return
+	}
+	klog.V(4).Infof("got reservation %v phase update, old %v, new %v", reservation.GetReservationNamespacedName(newR), oldR.Status.Phase, newR.Status.Phase)
+
+	if oldR.Status.Phase == schedulingv1alpha1.ReservationAvailable &&
```
eahydra, line 97: There may be a scenario: when the Reservation expires, the GC is delayed for a certain period of time.

apis/scheduling/v1alpha1/reservation_types.go
saintube: So it means we can disable the expiration of the Reservation?
It's fine if we can manually clean up them.

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,132 @@
...
+func deleteReservation(sched *scheduler.Scheduler, r *schedulingv1alpha1.Reservation) {
+	if r == nil || r.Spec.Template == nil {
+		klog.Errorf("failed to delete reservation for invalid reservation %v", r)
+		return
+	}
+
+	reservePod := reservation.GetReservePod(r)
+	// if reservation is scheduled, delete it from the schedule cache;
+	// otherwise, remove it from the scheduling queue.
+	if reservation.IsReservationScheduled(r) {
```
eahydra, line 112: Could there be a concurrency race condition problem? When deleting the Reservation, the ReservePod may be scheduled exactly.

pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -0,0 +1,209 @@
...
+
+type Hook struct{}
+
+func NewHook() *Hook {
+	return &Hook{}
+}
+
+func (h *Hook) Name() string { return Name }
+
+func (h *Hook) PreFilterHook(handle frameworkext.ExtendedHandle, cycleState *framework.CycleState, pod *corev1.Pod) (*corev1.Pod, bool) {
```
eahydra, line 47: As discussed with @allwmh before, do we need PreFilter Hook? Or we can consider implementing this ability by adjusting the order of the PreFilter plugins.

apis/scheduling/config/types.go
saintube: fixed

apis/scheduling/v1alpha1/reservation_types.go
saintube: done

apis/scheduling/v1alpha1/reservation_types.go
eahydra: it does not matter. It is acceptable to change this CRD definition at this stage.

eahydra: @hormes WDYT?

pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -0,0 +1,209 @@
...
+	if len(rOnNode) <= 0 {
+		return nil, nil, false
+	}
+
+	// fix-up reserved resources and ports
+	return pod, prepareFilterNodeInfo(pod, nodeInfo, rOnNode), true
+}
+
+func prepareMatchReservationState(handle frameworkext.ExtendedHandle, pod *corev1.Pod) (*AvailableCache, error) {
+	rList, err := handle.KoordinatorSharedInformerFactory().Scheduling().V1alpha1().Reservations().Lister().List(labels.Everything())
```
eahydra, line 95: The cost of this function is quite large. Maybe we can consider some optimizations. 
For example, we establish the index relationship between NodeName and Reservation through the indexer mechanism provided by the shared informer for the Reservation that has been successfully scheduled. Then query whether the Node has Reservations through indexer. This avoids rebuilding newAvailableCache. In addition, we can also add an atomic counter to count how many reservation instances have been scheduled in the current cluster when creating an index. If this counter is 0, return directly.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+// Also do validations in this phase.
+func (p *Plugin) PreFilter(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod) *framework.Status {
+	// if the pod is a reserve pod
+	if IsReservePod(pod) {
+		// validate reserve pod and reservation
+		klog.V(3).InfoS("Attempting to pre-filter reserve pod", "pod", klog.KObj(pod))
+		r, err := p.lister.Reservations(pod.Namespace).Get(pod.Name)
+		if errors.IsNotFound(err) {
+			klog.Warningf("abort the pre-filter for reservation %s since it is not found", util.GetPodKey(pod))
+			return nil
```
eahydra, line 196: return nil cannot abort the pre-filter

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+	if IsReservePod(pod) {
+		return framework.MinNodeScore, nil
+	}
+
+	matchedCache := getPreFilterState(cycleState)
+	if matchedCache == nil { // expected matchedCache is prepared
+		return framework.MinNodeScore, framework.NewStatus(framework.Error, ErrReasonUnexpectedState)
+	}
+	rOnNode := matchedCache.GetOnNode(nodeName)
+	klog.V(4).InfoS("Attempting to score pod for reservation state", "pod", klog.KObj(pod),
```
eahydra, line 255: need adjust the the log level to higher, 5 or 6.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+		"node", nodeName, "matched count", len(rOnNode))
+
+	// prefer the nodes which have reservation matched
+	if len(rOnNode) <= 0 {
```
eahydra, line 259: Here you can consider implementing a more refined scoring mechanism. If this version does not support it, it is recommended to add a TODO.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+		// TODO: handle pre-allocation cases
+
+		return nil
+	}
+
+	// 2. check if the pod match a reservation
+	matchedCache := getPreFilterState(cycleState)
+	// expected matchedCache is prepared
+	if matchedCache == nil {
```
eahydra, line 238: As we discussed, if the normal Pod hasn't preFilterState, we should passthrough the Pod continuing scheduled to find other free resources.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+
+		if !IsReservationScheduled(curR) {
+			klog.Warningf("failed to allocate resources on a non-scheduled reservation %v, phase %v",
+				GetReservationNamespacedName(curR), curR.Status.Phase)
+			// TBD: may try with the next candidate instead of return error
+			return fmt.Errorf(ErrReasonReservationExpired)
+		}
+
+		setReservationAllocated(curR, pod)
```
eahydra, line 316: before updating curR, please make DeepCopy... otherwise pollutes the data in the informer

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+		return framework.MinNodeScore, nil
+	}
+	return framework.MaxNodeScore, nil
+}
+
+func (p *Plugin) ScoreExtensions() framework.ScoreExtensions {
+	return nil
+}
+
+func (p *Plugin) PreBind(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeName string) *framework.Status {
```
eahydra, line 269: We need to implement the `Reserve` extension point to prevent a Reservation from being allocated repeatedly.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+	return nil
+}
+
+// Bind fake binds reserve pod and mark corresponding reservation as Available.
+// NOTE: This Bind plugin should get called before DefaultBinder; plugin order should be configured.
+func (p *Plugin) Bind(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeName string) *framework.Status {
+	// 1. if the pod is a reserve pod
+	if IsReservePod(pod) {
+		klog.V(3).InfoS("Attempting to fake bind reserve pod to node", "pod", klog.KObj(pod), "node", nodeName)
+		err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
eahydra, line 361: also need to support TooManyRequests

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+				return err1
+			}
+
+			// check if the reservation has been expired
+			if r.Status.Phase == schedulingv1alpha1.ReservationExpired || p.reservationCache.IsExpired(GetReservationKey(r)) {
+				return fmt.Errorf("reservation is expired")
+			}
+
+			// mark reservation as available
+			setReservationAvailable(r, nodeName)
```
eahydra, line 378: deepcopy reservation.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+		if err != nil {
+			return framework.NewStatus(framework.Error, "failed to bind reservation, err: "+err.Error())
+		}
+		return nil
+	}
+
+	return framework.NewStatus(framework.Skip, SkipReasonNotReservation)
+}
+
+func (p *Plugin) gcReservations() {
```
eahydra, line 401: It is recommended to separate these GC logic into a separate go file.

pkg/scheduler/plugins/reservation/rpod.go
saintube: done

pkg/scheduler/eventhandlers/reservation_handler.go
saintube: nice catch

pkg/scheduler/eventhandlers/reservation_handler.go
saintube: fixed

pkg/scheduler/eventhandlers/reservation_handler.go
saintube: fixed

pkg/scheduler/plugins/reservation/plugin.go
saintube: done

pkg/scheduler/plugins/reservation/plugin.go
saintube: done

pkg/scheduler/plugins/reservation/hook.go
saintube: I think PreFilter Hook is needed.  
To support reservation, we alter the pod copy and pass it to PreFilter plugins, while these plugins use the Pod object from PodInformer originally. If we turned to re-order the execution of PreFilter plugins and change the pod object inside Pre-Filter, the initial object would get polluted.

pkg/scheduler/eventhandlers/reservation_handler.go
eahydra: Please update the design proposal.

apis/scheduling/v1alpha1/reservation_types.go
eahydra: As discussed with @hormes,  If `TTL` is nil, use the default value, else if `TTL` is 0, disable the expiration, if `TTL` > 0, use the value as expiration.

pkg/scheduler/plugins/reservation/hook.go
eahydra: OK. got it!

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,150 @@
...
+			sched.SchedulingQueue.MoveAllToActiveOrBackoffQueue(assignedPodDelete, nil)
+		}
+
+		err = sched.SchedulingQueue.Delete(reservePod)
+		if err != nil {
+			klog.Errorf("failed to delete reserve pod in scheduling queue, reservation %v, err: %v", reservation.GetReservationNamespacedName(r), err)
+		}
+		fwk, ok := sched.Profiles[reservePod.Spec.SchedulerName]
+		if ok {
+			fwk.RejectWaitingPod(reservePod.UID)
```
eahydra, line 144: According to the current implementation, there is no need to deal with the waiting state.

pkg/scheduler/eventhandlers/reservation_handler.go
```diff
@@ -0,0 +1,150 @@
...
+		klog.Errorf("failed to delete reservation for invalid reservation %v", r)
+		return
+	}
+
+	reservePod := reservation.NewReservePod(r)
+	// if reservation is scheduled, delete it from the schedule cache;
+	// otherwise, remove it from the scheduling queue.
+
+	if reservation.IsReservationScheduled(r) {
+		err := sched.SchedulerCache.RemovePod(reservePod)
```
eahydra, line 120: We also need to be careful with Assume Pods. The current process also lacks this.

eahydra (changes requested):

saintube:
API change is split into #384

pkg/scheduler/plugins/reservation/plugin.go
saintube: done

pkg/scheduler/plugins/reservation/plugin.go
saintube: fixed

pkg/scheduler/plugins/reservation/plugin.go
saintube: done

pkg/scheduler/eventhandlers/reservation_handler.go
saintube: fixed

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -0,0 +1,530 @@
...
+	for i := range rOnNode {
+		rOnNode[i].ScoreForPod(pod)
+	}
+	sort.Slice(rOnNode, func(i, j int) bool {
+		return rOnNode[i].Score >= rOnNode[j].Score
+	})
+	target := rOnNode[0].R
+
+	// update: update current owner and allocated resources info
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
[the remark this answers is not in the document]
saintube: done

pkg/scheduler/plugins/reservation/hook.go
saintube: fixed. use index with `status.nodeName` here

pkg/scheduler/plugins/reservation/plugin.go
saintube: done

pkg/scheduler/plugins/reservation/plugin.go
saintube: fixed

pkg/scheduler/eventhandlers/reservation_handler.go
saintube: fixed. now we process the expired reservation in event handler

saintube:
### Demo reservation

> will be added in user manuals

```bash
$ # create CRD
$ kubectl create -f config/crd/bases/scheduling.koordinator.sh_reservations.yaml
$ # create a reservation
$ kubectl create -f test-reservation-big.yaml
reservation.scheduling.koordinator.sh/reservation-sample-big created
$ # check if the reservation get scheduled
$ kubectl get reservation -A -o wide
NAMESPACE   NAME                        PHASE       AGE    NODE             TTL   EXPIRES
test        reservation-sample-big     Available   4s       xxx-node-0   24h
...
$ kubectl create -f test-pod-big.yaml
pod/reserve-demo-big-1 created
$ kubectl get pod reserve-demo-big-1 -o wide
NAME                 READY   STATUS    RESTARTS   AGE   IP            NODE            NOMINATED NODE   READINESS GATES
reserve-demo-big-1   1/1     Running   0          73s   10.17.0.123   xxx-node-0   <none>           <none>
$ kubectl get reservation -n test reservation-sample-big -oyaml
kind: Reservation
metadata:
  creationTimestamp: "2022-07-27T11:31:22Z"
  generation: 1
  name: reservation-sample-big
  namespace: test
  ...
spec:
  owners:
  - object:
      name: reserve-demo-0
      namespace: default
  - labelSelector:
      matchLabels:
        app: reserve-demo
  template:
    spec:
      containers:
      - args:
        - -c
        - "1"
        command:
        - stress
        image: polinux/stress
        imagePullPolicy: Always
        name: stress
        resources:
          limits:
            cpu: "8"
            memory: 30Gi
          requests:
            cpu: 4000m
            memory: 10Gi
      restartPolicy: Always
      schedulerName: koord-scheduler-test
  ttl: 24h
status:
  allocatable:
    cpu: "4"
    memory: 10Gi
  allocated:
    cpu: "3"
    memory: 8Gi
  conditions:
  - lastProbeTime: "2022-07-27T11:31:22Z"
    lastTransitionTime: "2022-07-27T11:31:22Z"
    reason: Scheduled
    status: "True"
    type: Scheduled
  - lastProbeTime: "2022-07-27T11:31:22Z"
    lastTransitionTime: "2022-07-27T11:31:22Z"
    reason: Available
    status: "True"
    type: Ready
  currentOwners:
  - name: reserve-demo-big-1
    namespace: default
    uid: xxxxxx
  nodeName: xxx-node-0
  phase: Available
```

eahydra (commented):
The overall implementation looks good for me. Regarding the change of the scope of the Reservation CRD from Cluster to Namespaced, I think it should be discussed.

eahydra (approved):
/lgtm

PR #360, opened 2022-07-11
Support node cpu orchestration api

eahydra:
### V. Checklist

### DIFF
--- a/apis/scheduling/config/v1beta2/defaults.go
+++ b/apis/scheduling/config/v1beta2/defaults.go
@@ -41,7 +41,6 @@ var (
 	}
 
 	defaultPreferredCPUBindPolicy          = CPUBindPolicyFullPCPUs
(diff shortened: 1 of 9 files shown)


apis/scheduling/config/v1beta2/defaults.go
```diff
@@ -77,9 +77,6 @@ func SetDefaults_NodeNUMAResourceArgs(obj *NodeNUMAResourceArgs) {
 	if obj.DefaultCPUBindPolicy == "" {
 		obj.DefaultCPUBindPolicy = defaultPreferredCPUBindPolicy
 	}
-	if obj.NUMAAllocateStrategy == "" {
-		obj.NUMAAllocateStrategy = defaultNUMAAllocateStrategy
-	}
```
hormes, line 82: remove the definition?

eahydra: done.

PR #372, opened 2022-07-15
fix: consider lse/lsr when cpu suppress (#234)

### LINKED ISSUE #234: [proposal] cpu suppress policy should take fine-grined cpu orchestration into consideration
**What is your proposal**:
`cpu suppress` policy change the cpuset of BE pod dynamically. fine-grined cpu orchestration rule should be considers. see https://github.com/koordinator-sh/koordinator/issues/96 for more details.

**Why is this needed**:
BE pod should not share CPU with LSE.
fairness of the BE cpuset should be considered, in case the BE cpuset are shared to only few LSR/LS pods. 

take the cpu core number into account during the share of cpuset between BE and LSR/LS.
1. calculate the target number of suppressed cpu.
2. each LSR/LS contribute some cpu to share with BE by their request proportion.
3. apply cpuset for BE pods

ZYecho:

fix #234 

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/cpu_suppress.go
+++ b/pkg/koordlet/resmanager/cpu_suppress.go
@@ -27,6 +27,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
 	slov1alpha1 "github.com/koordinator-sh/koordinator/apis/slo/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,16 +159,9 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
 
 // calculateBESuppressPolicy calculates the be cpu suppress policy with cpuset cpus number and node cpu info
-func calculateBESuppressCPUSetPolicy(cpusetQuantity *resource.Quantity, oldCPUSetNum int, nodeCPUInfo *metriccache.NodeCPUInfo) []int32 {
-	// set the number of cpuset cpus no less than 2
-	cpus := int32(math.Ceil(float64(cpusetQuantity.MilliValue()) / 1000))
-	if cpus < 2 {
-		cpus = 2
-	}
-
+func calculateBESuppressCPUSetPolicy(cpus int32, oldCPUSetNum int, processorInfos []util.ProcessorInfo) []int32 {
```
zwzhang0107, line 162: split the processors into several partitions(socket+node), try to get 1 core(2 processors) from each partition for each iteration

zwzhang0107:
recoverCPUSetIfNeed needs to reset as be-share-pool(node-lse)

zwzhang0107 (commented):
recoverCPUSetIfNeed need to reset cpu as be-share-pool(node-LSE)

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -182,29 +176,37 @@ func calculateBESuppressCPUSetPolicy(cpusetQuantity *resource.Quantity, oldCPUSe
...
+		nodeIndex := getNodeIndex(processor)
+		if _, ok := coreIdSetofNode[nodeIndex]; !ok {
+			coreIdSetofNode[nodeIndex] = map[int32]int32{}
 		}
+		coreIdSetofNode[nodeIndex][processor.CoreID]++
 	}
 
-	// first we prefer the scattered node/socket, secondly the same core
+	for _, processor := range prioritizedCPUs {
+		coreIndexOfNode[processor.CoreID] = int32(len(coreIdSetofNode[getNodeIndex(processor)]))
```
eahydra, line 202: In the past, I have encountered a special machine with the same Core ID under different Sockets. Should not directly take the CoreID to participate in the calculation.

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -182,29 +176,37 @@ func calculateBESuppressCPUSetPolicy(cpusetQuantity *resource.Quantity, oldCPUSe
...
-	indexOfNode := map[int32]int32{}
-	numProcessorsOfNode := map[int32]int32{}
+	coreIndexOfNode := map[int32]int32{}
+	coreIdSetofNode := map[int32]map[int32]int32{}
 	for _, processor := range prioritizedCPUs {
-		numProcessorsOfNode[getNodeIndex(processor)]++
-		if indexOfNode[processor.CoreID] <= 0 {
-			indexOfNode[processor.CoreID] = numProcessorsOfNode[getNodeIndex(processor)]
+		nodeIndex := getNodeIndex(processor)
+		if _, ok := coreIdSetofNode[nodeIndex]; !ok {
```
saintube, line 195: `coreIdSetofNode` may be incorrect when two CPUs are on different Sockets but have the same Core ID.

pkg/koordlet/resmanager/cpu_suppress.go
eahydra: @hormes reminded that the data of `lscpu -e` and `/proc/cpuinfo` are inconsistent. After investigation and analysis, the problems I encountered in the past should be caused by the internal version not processing the data of cadvisor correctly. Taking the X86 architecture as an example, the so-called `CoreID` we see through `lscpu -e` is actually the minimum ID among the two logical cores under the same physical core. For example, Core A has two logical cores 26 and 80, then the CoreID is 26.

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -341,20 +332,68 @@ func (r *CPUSuppress) suppressBECPU() {
...
+				continue
+			}
+			for _, cpuID := range set.ToSliceNoSort() {
+				cpuIdToPool[int32(cpuID)] = apiext.GetPodQoSClass(podMeta.Pod)
+			}
+		}
+	}
+	lsrCpus := []util.ProcessorInfo{}
+	lsCpus := []util.ProcessorInfo{}
+	for _, processor := range nodeCPUInfo.ProcessorInfos {
```
zwzhang0107, line 369: please add a FIXME comments here: "be pods might be starved since lse pods can run out of all cpus"

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,95 +159,85 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
+			}
+			if cpuBucket[bucketIndex][i].CoreID == cpuBucket[bucketIndex][i+1].CoreID {
+				selectdIndex = i
+				break
+			}
 		}
-		CPUSets = append(CPUSets, prioritizedCPUs[i].CPUID)
-		needCPUs--
+		CPUSets = append(CPUSets, cpuBucket[bucketIndex][selectdIndex].CPUID, cpuBucket[bucketIndex][selectdIndex+1].CPUID)
+		cpuBucket[bucketIndex][selectdIndex].CPUID = -1
```
zwzhang0107, line 237: mark "-1" to cpu id seems tricky, consider using a bitmap to record allocation result

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,95 +159,85 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
+			}
+			CPUSets = append(CPUSets, cpuBucket[bucketIndex][selectdIndex].CPUID)
+			cpuBucket[bucketIndex][selectdIndex].CPUID = -1
+			needCPUs--
+			bucketIndex = (bucketIndex + 1) % len(cpuBucket)
+			continue
+		}
+		// prefer the same core
+		selectdIndex := 0
+		for i := 0; i < len(cpuBucket[bucketIndex])-1; i++ {
```
zwzhang0107, line 227: one bucket might runs out early, cpu[0-1] will be reallocated

pkg/util/system/util_test_tool.go
```diff
@@ -36,7 +36,7 @@ type FileTestUtil struct {
 func NewFileTestUtil(t *testing.T) *FileTestUtil {
 	// NOTE: When $TMPDIR is not set, `t.TempDir()` can use different base directory on Mac OS X and Linux, which may
 	// generates too long paths to test unix socket.
-	t.Setenv("TMPDIR", "/tmp")
+	// t.Setenv("TMPDIR", "/tmp")
```
saintube, line 39: why comment this line?

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,95 +159,85 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
-		return a.CPUID > b.CPUID
+		return len(cpuBucket[i]) > len(cpuBucket[j])
 	})
 
 	needCPUs := cpus
-	for i := range prioritizedCPUs {
-		if needCPUs <= 0 {
-			break
+	bucketIndex := 0
+	for needCPUs > 0 {
```
zwzhang0107, line 206: add a protection in case of looping forever

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,95 +159,85 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
+		if needCPUs == 1 || len(cpuBucket[bucketIndex]) == 1 {
+			// prefer the single core cpu id
+			selectdIndex := 0
+			for i := 0; i < len(cpuBucket[bucketIndex]); i++ {
+				if cpuBucket[bucketIndex][i].CPUID < 0 {
+					continue
+				}
+				if (i == 0 || cpuBucket[bucketIndex][i].CoreID != cpuBucket[bucketIndex][i-1].CoreID) && (i+1 == len(cpuBucket[bucketIndex]) || cpuBucket[bucketIndex][i].CoreID != cpuBucket[bucketIndex][i+1].CoreID) {
```
zwzhang0107, line 214: this condition seems too complex, consider set the allocation into two stages: 1. allocate cpu pair of same physical core. 2. allocate from other independent cpu-HT.

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,95 +159,85 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
-
 	// getNodeIndex is a function to calculate an index for every numa node or socket
 	getNodeIndex := func(info util.ProcessorInfo) int32 {
+		// (nodeId, socketId) => nodeIndex
 		return (info.NodeID + numProcessors) * (info.SocketID + 1)
 	}
+	cpuBucketOfNode := map[int32][]util.ProcessorInfo{}
+	for _, p := range processorInfos {
+		nodeIndex := getNodeIndex(p)
+		cpuBucketOfNode[nodeIndex] = append(cpuBucketOfNode[nodeIndex], p)
```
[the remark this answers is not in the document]
ZYecho: append(nil, p) will work

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,98 +159,102 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
-	for i := range prioritizedCPUs {
+	usedCpu := map[int32]bool{}
+	// select same core cpu id
+	preNeedCpus := int32(-1)
+	i := 0
+	for ; i < len(cpuBucket); i = (i + 1) % len(cpuBucket) {
+		if needCPUs <= 1 {
+			break
+		}
+		if i == len(cpuBucket) {
```
zwzhang0107, line 213: this never happens due to `(i + 1) % len(cpuBucket)`
consider check then current and pre length of usedCpu

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,98 +159,102 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
+			usedCpu[cpuBucket[i][selectdIndex+1].CPUID] = true
+			needCPUs = needCPUs - 2
+		}
+	}
+
+	// select single cpu id
+	for ; i < len(cpuBucket); i = (i + 1) % len(cpuBucket) {
 		if needCPUs <= 0 {
 			break
 		}
```
zwzhang0107, line 242: add a protection break by checking current and pre length of usedCpu

pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -158,98 +159,111 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
...
-	// it will get sorted as below:
-	//	CPUID	CoreID	SocketID	NodeID
-	//	7		3		1			1
-	//	3		3		1			1
-	//	5		1		0			0
-	//	1		1		0			0
-	//	6		2		1			1
-	//	2		2		1			1
-	//	4		0		0			0
-	//	0		0		0			0
```
jasonliu747, line 228: can you keep this kind of figure in comment, so ppl could understand what we're trying to do more easily.

PR #373, opened 2022-07-15
chore: supply two UT for pkg/util

ZiMengSheng:
### V. Checklist


pkg/util/utils_test.go
```diff
@@ -279,3 +284,91 @@ func Test_UtilCgroupCPUSet(t *testing.T) {
...
+		{
+			name: "i < j",
+			args: args{
+				i: 0,
+				j: 1,
+			},
+			want: 1,
+		},
+		{
+			name: "i < j",
```
jasonliu747, line 351: you probably mean `i > j`?

PR #379, opened 2022-07-18
fix auditor test in MacOS

hormes:
Signed-off-by: Fansong Zeng <fanster.z@gmail.com>

curl http://:%d seems to fail on MacOS, set it to http://127.0.0.1:%d

http://localhost:%d will panic in go test -race which is hard to explain.

### V. Checklist


jasonliu747 (commented):
this test was working until this week, any idea?

jasonliu747 (commented):
/lgtm

PR #386, opened 2022-07-20
feature: report pod alloc of Guaranteed pod and cpu manager policy

ZYecho:

### V. Checklist

### DIFF
--- a/apis/extension/node.go
+++ b/apis/extension/node.go
@@ -27,11 +27,13 @@ import (
 const (
 	// AnnotationNodeCPUTopology describes the detailed CPU topology.
 	AnnotationNodeCPUTopology = NodeDomainPrefix + "/cpu-topology"
-	// AnnotationNodeCPUAllocs describes the CPUs allocated by Koordinator LSE/LSR and K8s Guaranteed Pods.
+	// AnnotationNodeCPUAllocs describes K8s Guaranteed Pods.
 	AnnotationNodeCPUAllocs = NodeDomainPrefix + "/pod-cpu-allocs"
 	// AnnotationNodeCPUSharedPools describes the CPU Shared Pool defined by Koordinator.
 	// The shared pool is mainly used by Koordinator LS Pods or K8s Burstable Pods.
 	AnnotationNodeCPUSharedPools = NodeDomainPrefix + "/cpu-shared-pools"
+	// AnnotationNodeCPUManagerPolicy describes the cpu manager policy options of kubelet
+	AnnotationNodeCPUManagerPolicy = NodeDomainPrefix + "/cpu-manager-policy"
 
 	// LabelNodeCPUBindPolicy constrains how to bind CPU logical CPUs when scheduling.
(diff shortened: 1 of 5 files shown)


pkg/util/system/cpu_manager_linux.go
```diff
@@ -0,0 +1,54 @@
...
+
+// return cpu policy, cpu state file path and cpu manager opt
+func GuessCPUManagerOptFromKubelet() (string, string, map[string]string, error) {
+	pids, err := PidOf(Conf.ProcRootDir, "kubelet")
+	if err != nil || len(pids) == 0 {
+		return "", "", nil, fmt.Errorf("failed to find kubelet's pid, kubelet may stop: %v", err)
+	}
+	kubeletPid := pids[0]
+
+	kubeletArgs, err := ProcCmdLine(Conf.ProcRootDir, kubeletPid)
```
eahydra, line 38: I remember that kubelet also supports configuring CPU management policies via configuration file.  FYI: https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/

apis/extension/node.go
```diff
@@ -27,11 +27,13 @@ import (
 const (
 	// AnnotationNodeCPUTopology describes the detailed CPU topology.
 	AnnotationNodeCPUTopology = NodeDomainPrefix + "/cpu-topology"
-	// AnnotationNodeCPUAllocs describes the CPUs allocated by Koordinator LSE/LSR and K8s Guaranteed Pods.
+	// AnnotationNodeCPUAllocs describes K8s Guaranteed Pods.
```
eahydra, line 30: According to the previous design, the koordlet needs to report the CPU information of the Pod scheduled by the Koordinator. This CPU information is obtained from the cgroup. With this information, we can analyze changes and problems on the node side.

pkg/util/system/cpu_manager_linux.go
ZYecho: add TODO

apis/extension/node.go
ZYecho: add TODO

pkg/koordlet/statesinformer/states_noderesourcetopology.go
```diff
@@ -74,16 +78,82 @@ func (s *statesInformer) syncNodeResourceTopology(node *corev1.Node) {
...
+		klog.Errorf("failed to guess kubelet cpu manager opt, err: %v", err)
+	}
+	cpuManagerPolicy := extension.CPUManagerPolicy{
+		Policy:  cpuPolicy,
+		Options: cpuManagerOpt,
+	}
+	cpuManagerPolicyJson, err := json.Marshal(cpuManagerPolicy)
+	if err != nil {
+		klog.Errorf("failed to marshal cpu manager policy, err: %v", err)
+		return
```
zwzhang0107, line 134: consider remove the `return` because other fields are still need report, so as the following `return`.

ZYecho: ok

PR #402, opened 2022-07-25
add gangPlugin and gangCache for gang scheduling

Wenshiqi222:
1. add gangPlugin to do the scheduling at some extension points 
2. add gangCache to stroe the gang info

### V. Checklist


pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,383 @@
...
+		lock:      new(sync.RWMutex),
+	}
+}
+
+func (gangCache *gangCache) onPodAdd(obj interface{}) {
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		return
+	}
+	gangCache.AddPod(pod)
```
buptcozy, line 30: function should be: first get gang, then execute gang's function. Don't put all logic to gangCache, gangCache only maintain gang's add\update\delete

Wenshiqi222: done

apis/scheduling/config/v1beta2/defaults.go
```diff
@@ -52,6 +52,8 @@ var (
 	}
 
 	defaultEnablePreemption = pointer.Bool(false)
+
+	defaultGangTimeOutSeconds int64 = 120
```
buptcozy, line 56: in commuity,  the timeout type is *time.Duration, can we follow it?  And I think 120 as default may be a little small,  10 min may be better.

Wenshiqi222: 1. Got it,10 mins should be fine;
2. The int64 is followed by other Timeout args in Koordinator, this value will be converted into time.Duration when initing the Gang

cmd/koord-scheduler/app/server.go
```diff
@@ -191,11 +191,13 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
 	cc.InformerFactory.Start(ctx.Done())
 	cc.KoordinatorSharedInformerFactory.Start(ctx.Done())
 	cc.NRTSharedInformerFactory.Start(ctx.Done())
+	cc.PodGroupInformerFactory.Start(ctx.Done())
 
 	// Wait for all caches to sync before scheduling.
 	cc.InformerFactory.WaitForCacheSync(ctx.Done())
 	cc.KoordinatorSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	cc.NRTSharedInformerFactory.WaitForCacheSync(ctx.Done())
+	cc.PodGroupInformerFactory.WaitForCacheSync(ctx.Done())
```
buptcozy, line 200: if cluster not import PodGroup, this step will hang, so should we add a flag to active by need?

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+		klog.Errorf("pod's annotation TotalNumber illegal, gangName:%v, value:%v",
+			gang.Name, pod.Annotations[extension.AnnotationGangTotalNum])
+		totalChildrenNum = minRequiredNumber
+	} else if totalChildrenNum != 0 && totalChildrenNum < minRequiredNumber {
+		klog.Errorf("pod's annotation TotalNumber cannot less than minRequiredNumber, gangName:%v, TotalNumber:%v,minRequiredNumber:%v",
+			gang.Name, pod.Annotations[extension.AnnotationGangTotalNum], minRequiredNumber)
+		totalChildrenNum = minRequiredNumber
+	}
+	gang.TotalChildrenNum = totalChildrenNum
+
```
buptcozy, line 104: gang.CreateTime = pod.CreationTimestamp.Time

cmd/koord-scheduler/app/server.go
Wenshiqi222: Ok,i will check how NRT CRD in Koordinator solve this problem

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+	schedulingconfig "github.com/koordinator-sh/koordinator/apis/scheduling/config"
+)
+
+var (
+	timeNowFn = time.Now
+)
+
+const (
+	GangFromCrd        string = "GangFromCrd"
+	GangFromAnnotation string = "GangFromAnnotation"
```
buptcozy, line 26: GangFromAnnotation-> GangFromPodAnnotation; GangFromCrd-> GangFromPodGroupCrd

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+		}
+	}
+
+	if num == gang.TotalChildrenNum {
+		gang.ScheduleCycleValid = true
+		klog.Infof("SetScheduleCycleValid, gangName:%v, valid:%v", gang.Name, true)
+		gang.ScheduleCycle += 1
+	}
+
+	klog.Infof("TryUpdateScheduleCycle, gangName:%v, ScheduleCycle:%v",
```
buptcozy, line 346: this log is useless, please merge in "if num == gang.TotalChildrenNum {}"

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+	num := 0
+	for _, childScheduleCycle := range gang.ChildrenScheduleRoundMap {
+		if childScheduleCycle == gang.ScheduleCycle {
+			num++
+		}
+	}
+
+	if num == gang.TotalChildrenNum {
+		gang.ScheduleCycleValid = true
+		klog.Infof("SetScheduleCycleValid, gangName:%v, valid:%v", gang.Name, true)
```
buptcozy, line 342: this log should be in the end of the "if", and add "ScheduleCycle"

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+func ParsePgTimeoutSeconds(timeoutSeconds int32) (time.Duration, error) {
+	if timeoutSeconds <= 0 {
+		return 0, fmt.Errorf("podGroup timeout value is illegal,timeout Value:%v", timeoutSeconds)
+	}
+	return time.Duration(timeoutSeconds) * time.Second, nil
+}
+
+// GetSubPriority
+// Get pod's sub-priority in Koordinator from label
+func GetSubPriority(pod *v1.Pod) (int32, error) {
```
buptcozy, line 402: GetSubPriority-> GetPodSubPriority

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/scheduler-plugins/apis/scheduling/v1alpha1"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+	schedulingconfig "github.com/koordinator-sh/koordinator/apis/scheduling/config"
+)
+
+type gangCache struct {
```
buptcozy, line 16: gangCache->GangCache

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	if !ok {
+		return
+	}
+
+	if !IsValidPod(pod) {
+		return
+	}
+	var gang *Gang
+	//the gang is created in either CRD way or Annotation way
+	if _, exist := pod.Labels[v1alpha1.PodGroupLabel]; exist {
```
buptcozy, line 72: please use GetGangNameByPod of Gang

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/******************GANG Event END******************/
+
+func IsValidPod(pod *v1.Pod) bool {
+	if pod.Annotations == nil && pod.Labels == nil {
+		return false
+	}
+
+	_, annotationWay := pod.Annotations[extension.AnnotationGangName]
```
buptcozy, line 175: please use 'GetGangNameByPod() == ""'

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	schedulingconfig "github.com/koordinator-sh/koordinator/apis/scheduling/config"
+)
+
+type gangCache struct {
+	lock       *sync.RWMutex
+	gangItems  map[string]*Gang
+	pluginArgs *schedulingconfig.GangSchedulingArgs
+}
+
+func NewGangCache() *gangCache {
```
buptcozy, line 22: please add pluginArgs to NewGangCache's parameter

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	if !IsValidPod(pod) {
+		return
+	}
+	var gang *Gang
+	//the gang is created in either CRD way or Annotation way
+	if _, exist := pod.Labels[v1alpha1.PodGroupLabel]; exist {
+		gangName := pod.Labels[v1alpha1.PodGroupLabel]
+		gangNamespace := pod.Namespace
+		gangId := GetNamespaceSplicingName(gangNamespace, gangName)
+		gang = gangCache.GetGangFromCache(gangId)
```
buptcozy, line 76: consider a time order, first receive pod, then receive pg; we should create gang first, and then init gang, to avoid lost pod event. please add a ut to cover this case

apis/scheduling/config/v1beta2/defaults.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	gangId := GetNamespaceSplicingName(gangNamespace, gangName)
+	gang := gangCache.GetGangFromCache(gangId)
+	if gang == nil {
+		return
+	}
+
+	shouldDeleteGang := gang.deletePod(pod)
+	if shouldDeleteGang {
+		gangCache.deleteGangFromCache(gangId)
+		klog.Infof("Delete gang from gang cache, gangName:%v", gang.Name)
```
buptcozy, line 121: gang.Name->gangId

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+}
+
+/******************Pod Event END******************/
+
+/******************GANG Event BEG******************/
+func (gangCache *gangCache) OnGangAdd(obj interface{}) {
```
buptcozy, line 128: OnGangAdd->OnPodGroupAdd

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+	pg, ok := obj.(*v1alpha1.PodGroup)
+	if !ok {
+		return
+	}
+	if pg.Status.Phase == v1alpha1.PodGroupFinished || pg.Status.Phase == v1alpha1.PodGroupFailed {
```
buptcozy, line 133: if pg.Status.Phase == v1alpha1.PodGroupFinished || pg.Status.Phase == v1alpha1.PodGroupFailed{}.    I think this can be deleted

pkg/scheduler/plugins/gang/gang_cache.go
```diff
@@ -0,0 +1,198 @@
...
+		return false
+	}
+	return true
+}
+
+// StringToGangGroupSlice
+// Parse gang group's annotation like :"[nsA/gangA,nsB/gangB]"  => goLang slice : []string{"nsA/gangA"."nsB/gangB"}
+// todo:need to be more graceful
+func StringToGangGroupSlice(s string) ([]string, error) {
+	defaultSlice := make([]string, 0)
```
buptcozy, line 187: gangGroup := make([]string, 0)
	json.Unmarshal([]byte(s), &gangGroup)

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+	gang := &GangPlugin{
+		frameworkHandler: handle,
+		podLister:        podLister,
+		pgLister:         pgInformer.Lister(),
+		gangCache:        gangCache,
+		pluginArgs:       pluginArgs,
+	}
+
+	//recover the gangCache
+	gang.RecoverGangCache()
```
buptcozy, line 103: RecoverGangCache should not be in new stage. GangCache should also be in framework, after waitForCacheSync Done,  execute framework.GangCache.Recover()

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// PreFilter
+//if non-strict-mode, we only do step1 and step2:
+// i.Check whether childes in Gang has met the requirements of minimum number under each Gang, and reject the pod if negative.
+// ii.Check whether the Gang has been timeout(check the pod's annotation,later introduced at Permit section), and reject the pod if positive.
+// iii.Check whether the Gang has met the scheduleCycleValid check, and reject the pod if negative.
+// iv.Try update scheduleCycle, scheduleCycleValid, childrenScheduleRoundMap as mentioned above.
+func (p *GangPlugin) PreFilter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod) *framework.Status {
+	gangId := GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 147: first judge whether pod need gang, if no need, return nil

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+		gangScheduleCycle := gang.GetGangScheduleCycle()
+		if podScheduleCycle >= gangScheduleCycle {
+			klog.Errorf("pod's schedule cycle too large, gangName:%v, podName:%v, podCycle:%v, gangCycle:%v",
+				gangId, GetNamespaceSplicingName(pod.Namespace, pod.Name), podScheduleCycle, gangScheduleCycle)
+			return framework.NewStatus(framework.UnschedulableAndUnresolvable, "pod's schedule cycle too large")
+		}
+
+		gang.SetChildScheduleCycle(pod, gangScheduleCycle)
+
+		if !gang.IsGangScheduleCycleValid() {
```
buptcozy, line 189: !gang.IsGangScheduleCycleValid() should first check. then check podScheduleCycle;  SetChildScheduleCycle and TryUpdateScheduleCycle both in defer

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+func (p *GangPlugin) PreFilterExtensions() framework.PreFilterExtensions {
+	return nil
+}
+
+// PostFilter
+//i. If strict-mode, we will set scheduleCycleValid to false and release all assumed pods.
+//ii. If non-strict mode, we will do nothing.
+func (p *GangPlugin) PostFilter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	gangId := GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 208: first judge whether pod need gang, if no need, return nil

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
+	}
+
+	if gang.GetGangMode() == extension.GangModeStrict {
+		p.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+			waitingGangId := GetNamespaceSplicingName(waitingPod.GetPod().Namespace,
+				GetGangNameByPod(waitingPod.GetPod()))
+			if waitingGangId == gangId {
+				klog.Errorf("postFilter rejects the pod, gangName:%v, podName:%v",
+					GetNamespaceSplicingName(pod.Namespace, pod.Name), gangId)
```
buptcozy, line 227: klog content is reverse

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+			gang := p.gangCache.GetGangFromCache(gangId)
+			gang.AddAssumedPod(pod)
+			gang.SetResourceSatisfied()
+		}
+	}
+
+}
+
+func (p *GangPlugin) PermitCheck(pod *corev1.Pod) (time.Duration, Status) {
+	gangId := GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 384: first judge whether pod need gang, if no need, return nil; PermitCheck-> permitCheck

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+		GetGangNameByPod(pod))
+	gang := p.gangCache.GetGangFromCache(gangId)
+	if gang == nil {
+		klog.Errorf("can't find gang, gangName:%v, podName:%v", gangId,
+			GetNamespaceSplicingName(pod.Namespace, pod.Name))
+		return 0, NotFoundInCache
+	}
+
+	//first we need to add the pod to assumedMap of gang
+	gang.AddAssumedPod(pod)
```
buptcozy, line 394: gang.AddAssumedPod(pod) should not in XXXCheck func

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+func (p *GangPlugin) Permit(ctx context.Context, state *framework.CycleState, pod *corev1.Pod, nodeName string) (*framework.Status, time.Duration) {
+	waitTime, s := p.PermitCheck(pod)
+	var retStatus *framework.Status
+	switch s {
+	case NotFoundInCache:
+		return framework.NewStatus(framework.Unschedulable, "Gang not found in gangCache"), 0
+	case Wait:
+		klog.Infof("Pod %v from gang %v is waiting to be scheduled at Permit stage",
+			GetNamespaceSplicingName(pod.Namespace, pod.Name),
+			GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 252: klog content is dup

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// Unreserve
+//(1) handle the timeout gang
+//(2) do nothing when bound failed
+func (p *GangPlugin) Unreserve(ctx context.Context, state *framework.CycleState, pod *corev1.Pod, nodeName string) {
+	gangId := GetNamespaceSplicingName(pod.Namespace,
+		GetGangNameByPod(pod))
+	gang := p.gangCache.GetGangFromCache(gangId)
+	if gang == nil {
```
buptcozy, line 276: Wait

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+		klog.Errorf("can't find gang, gangName:%v, podName:%v", gangId,
+			GetNamespaceSplicingName(pod.Namespace, pod.Name))
+		return
+	}
+
+	// gang time out
```
buptcozy, line 282: "gang time out" not correct,  may be preempted.

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+	if !gang.IsGangResourceSatisfied() {
+		klog.Infof("gang is time out,start to release the assumed resource and add annotations to the "+
+			"gang's children, gangName:%v, podName:%v", gangId, GetNamespaceSplicingName(pod.Namespace, pod.Name))
+		pods, err := p.podLister.List(labels.NewSelector())
+		if err != nil {
+			klog.Errorf("unReserve list  pod err:%v", err.Error())
+			return
+		}
+		for _, podOriginal := range pods {
```
buptcozy, line 291: if pod is preempted, not timeout,  we should not update timeout annotation; how can we distinguish timeout or preemtion? may be we can record a firstPremit timestamp on gang, and when unreserve, we judge curtime -  firstPremit  >= waitTime, is yes, no matter is preempted or timeout, we should update timeout annotation.

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+					continue
+				}
+				err = retry.OnError(
+					retry.DefaultRetry,
+					errors.IsTooManyRequests,
+					func() error {
+						_, err = p.frameworkHandler.ClientSet().CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name,
+							types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
+						if err != nil {
+							klog.Errorf("unReserve when patch gang timeout annotation to pod:%v, err:%v",
```
buptcozy, line 314: should add "Failed to"

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+								GetNamespaceSplicingName(pod.Namespace, pod.Name), err.Error())
+						}
+						return err
+					})
+				if err != nil {
```
buptcozy, line 319: if err == nil, add "success" log

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+						GetNamespaceSplicingName(pod.Namespace, pod.Name), err.Error())
+				}
+			}
+		}
+
+		// release resource of all assumed children of the gang
+		p.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+			if GetNamespaceSplicingName(waitingPod.GetPod().Namespace,
+				GetGangNameByPod(waitingPod.GetPod())) == gangId {
+				klog.Errorf("unReserve rejects the pod name:%v from Gang %s due to timeout",
```
buptcozy, line 330: unReserve rejects the pod name:%v from Gang %s due to timeout ->. unReserve rejects the pod:%v from Gang:%s due to timeout

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+					GetNamespaceSplicingName(pod.Namespace, pod.Name), gangId)
+				waitingPod.Reject(p.Name(), "optimistic rejection in unReserve due to timeout")
+			}
+		})
+	}
+}
+
+// PostBind just update the gang's BoundChildren
+func (p *GangPlugin) PostBind(ctx context.Context, _ *framework.CycleState, pod *corev1.Pod, nodeName string) {
+	gangId := GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 340: first judge whether pod need gang, if no need, return nil

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+	podLister := p.podLister
+	podsList, err := podLister.List(labels.NewSelector())
+	if err != nil {
+		klog.Errorf("RecoverGangCache podsList List error %v", err.Error())
+	}
+	for _, pod := range podsList {
+		p.gangCache.OnPodAdd(pod)
+
+		if pod.Spec.NodeName != "" {
+			gangId := GetNamespaceSplicingName(pod.Namespace,
```
buptcozy, line 373: judge whether pod need gang

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+				GetGangNameByPod(pod))
+			gang := p.gangCache.GetGangFromCache(gangId)
+			gang.AddAssumedPod(pod)
```
buptcozy, line 376: AddAssumedPod is error; shoud like AddBoundPod

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+			GetNamespaceSplicingName(pod.Namespace, pod.Name))
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.UnschedulableAndUnresolvable, "can't find gang")
+	}
+
+	if gang.IsGangResourceSatisfied() {
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
+	}
+
+	if gang.GetGangMode() == extension.GangModeStrict {
+		p.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
```
buptcozy, line 222: please make every line has ut covered;  please think more e2e ut cases

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: yes, podGroup will not be handled during the whole scheduling process

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: this log's second %v  is gangName

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: yes, we should also create a gang here rather than return

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_cache.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved, add gangCache to extendHandler, after running the informer, recover the gang

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: yes, it can make the Permit stage more clear

Wenshiqi222: solved

hormes:
@Wenshiqi222  use `git commit -s ` to fix DCO check

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

apis/scheduling/config/types.go
```diff
@@ -123,3 +123,13 @@ type ReservationArgs struct {
...
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// GangSchedulingArgs defines the parameters for GangScheduling plugin.
+type GangSchedulingArgs struct {
+	metav1.TypeMeta
+
+	//DefaultTimeOutSeconds is the default gang's waiting time in Permit stage.
+	DefaultTimeOutSeconds *int64 `json:"defaultTimeOutSeconds,omitempty"`
```
eahydra, line 134: It is recommended to use *metav1.Duration, which is easier to use

apis/scheduling/config/validation/validation_pluginargs.go
```diff
@@ -91,3 +91,10 @@ func validateEstimatedResourceThresholds(thresholds map[corev1.ResourceName]int6
 	}
 	return nil
 }
+
+func ValidateGangSchedulingArgs(gangSchedulingArgs *config.GangSchedulingArgs) error {
+	if gangSchedulingArgs.DefaultTimeOutSeconds != nil && *gangSchedulingArgs.DefaultTimeOutSeconds <= 0 {
+		return fmt.Errorf("gangSchedulingArgs  DefaultTimeOutSeconds should be a positive value")
```
eahydra, line 97: Please make the error message simpler

cmd/koord-scheduler/app/config/config.go
```diff
@@ -30,6 +31,7 @@ type Config struct {
 	KoordinatorClient                koordinatorclientset.Interface
 	KoordinatorSharedInformerFactory koordinatorinformers.SharedInformerFactory
 	NRTSharedInformerFactory         nrtinformers.SharedInformerFactory
+	PodGroupInformerFactory          pginformers.SharedInformerFactory
```
eahydra, line 34: This implementation is not recommended anymore. In the past I tried a new implementation to pass CRD Client and Informer Factory that can be reused frequently, but I noticed that most of the CRD Client and Informer Factory are actually not reused. It is recommended to initialize the CRD Client and Informer Factory inside the plugin at first. Also start the Informer Factory inside the plugin and wait for the synchronization to complete.

cmd/koord-scheduler/app/server.go
eahydra: When installing koordinator via helm charts it will import the PodGroup CRD's

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
+package gang
+
+import (
+	"fmt"
+	"strconv"
+	"sync"
+	"time"
```
eahydra, line 7: please keep the import order like the following:

```go
import (
     "encoding/json"
     "fmt"
    "strconv"

	"k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/util/strategicpatch"
	"k8s.io/klog/v2"
	"sigs.k8s.io/scheduler-plugins/apis/scheduling/v1alpha1"

	"github.com/koordinator-sh/koordinator/apis/extension"
	schedulingconfig "github.com/koordinator-sh/koordinator/apis/scheduling/config"
)

```

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/scheduler-plugins/apis/scheduling/v1alpha1"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+	schedulingconfig "github.com/koordinator-sh/koordinator/apis/scheduling/config"
+)
+
+var (
+	timeNowFn = time.Now
```
eahydra, line 21: you can use  clock.Clock instead.

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,437 @@
...
+		BoundChildren:            make(map[string]*v1.Pod),
+		ScheduleCycleValid:       true,
+		ScheduleCycle:            1,
+		ChildrenScheduleRoundMap: make(map[string]int),
+		GangFrom:                 GangFromAnnotation,
+		HasGangInit:              false,
+	}
+}
+
+/******************Pod\PodGroup Event BEG******************/
```
eahydra, line 76: Please do not use this style of comments.

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+	_ framework.QueueSortPlugin  = &GangPlugin{}
+	_ framework.PreFilterPlugin  = &GangPlugin{}
+	_ framework.PostFilterPlugin = &GangPlugin{}
+	_ framework.ReservePlugin    = &GangPlugin{}
+	_ framework.PostBindPlugin   = &GangPlugin{}
+	_ framework.PostBindPlugin   = &GangPlugin{}
+)
+
+type GangPlugin struct {
+	frameworkHandler framework.Handle
```
eahydra, line 49: it's handle not handler..

pkg/scheduler/plugins/gang/gang_plugin.go
```diff
@@ -0,0 +1,507 @@
...
+	podLister        v1.PodLister
+	pgLister         pgLister.PodGroupLister
```
eahydra, line 51: podGroupLister is already short...

apis/scheduling/config/types.go
```diff
@@ -123,3 +123,13 @@ type ReservationArgs struct {
 	// EnablePreemption indicates whether to enable preemption for reservations.
 	EnablePreemption *bool `json:"enablePreemption,omitempty"`
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// GangSchedulingArgs defines the parameters for GangScheduling plugin.
+type GangSchedulingArgs struct {
```
eahydra, line 130: the plugin called `Gang` but the args called `GangSchedulingArgs`.   This implementation is invalid as required by the framework. It is necessary to keep the prefix of Args consistent with the plugin name.

pkg/scheduler/plugins/gang/gang.go
eahydra: What I'm not sure about is whether it's possible to set the creation time on first initialization from a pod that is relatively new in time dimension. Should we use the oldest time within the same gang or the newest time? Or that it doesn't matter?

apis/scheduling/config/types.go
Wenshiqi222: right,solved

apis/scheduling/config/validation/validation_pluginargs.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: ok

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang_plugin.go
Wenshiqi222: ok😂

apis/scheduling/config/types.go
Wenshiqi222: yes, just  call the prefix "Gang"

cmd/koord-scheduler/app/config/config.go
Wenshiqi222: ok,solved

Wenshiqi222:
> @Wenshiqi222 use `git commit -s ` to fix DCO check

ok

apis/scheduling/config/types.go
```diff
@@ -123,3 +123,14 @@ type ReservationArgs struct {
 	// EnablePreemption indicates whether to enable preemption for reservations.
 	EnablePreemption *bool `json:"enablePreemption,omitempty"`
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// GangArgs defines the parameters for Gang Scheduling plugin.
+type GangArgs struct {
```
eahydra, line 130: The new plugin args must be register in register.go include `apis/scheduling/config` and `apis/scheduling/config/v1beta2`

cmd/koord-scheduler/app/server.go
```diff
@@ -196,6 +197,15 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
...
 	cc.KoordinatorSharedInformerFactory.WaitForCacheSync(ctx.Done())
 	cc.NRTSharedInformerFactory.WaitForCacheSync(ctx.Done())
+	// after all caches to sync
+	// we start to recover GangCache
+	gangCacheRaw := extHandle.GetGangCache()
+	gangCache, ok := gangCacheRaw.(*gang.GangCache)
+	if !ok {
+		klog.Errorf("Run scheduler get gangCache assert failed")
+	} else {
+		gangCache.RecoverGangCache()
```
eahydra, line 207: I think there should be no need to do this. You can generate and start the Informer in the plugin. This is reliable even when relying on PodInformer.

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -203,3 +218,19 @@ func PluginFactoryProxy(extendHandle ExtendedHandle, factoryFn frameworkruntime.
 		return factoryFn(args, extendHandle)
 	}
 }
+
+// gang test used
+type PodGroupClientSetAndHandle struct {
```
eahydra, line 223: Hey, If the type is just used to testing, please move the codes to the testing package.

eahydra (changes requested):

apis/scheduling/config/types.go
Wenshiqi222: done

go.mod
```diff
@@ -35,7 +35,8 @@ require (
 	k8s.io/kubernetes v1.22.6
 	k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a
 	sigs.k8s.io/controller-runtime v0.10.3
-	sigs.k8s.io/yaml v1.3.0
+	sigs.k8s.io/scheduler-plugins v0.22.6
```
jasonliu747, line 38: go.mod 应该只有 `sigs.k8s.io/scheduler-plugins v0.22.6` 需要更新，其他不用修改版本，可以 rebase 一下最新的代码

Wenshiqi222: 好的

cmd/koord-scheduler/app/server.go
Wenshiqi222: got it, solved

pkg/scheduler/frameworkext/framework_extender.go
Wenshiqi222: ok,done

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: we set gang a field called "TimeoutStartTime",which means from this time we start coutning the timeout,it depends on the first pod coming to the Permit stage and decided to Wait at this stage,we set the TimeoutStartTime of that moment.

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	return 0, Success
+}
+
+// ActivateGang
+// Put all the pods belong to the Gang which in UnSchedulableQueue or backoffQueue back to activeQueue,
+func (p *GangPlugin) ActivateGang(pod *corev1.Pod, state *framework.CycleState) {
+	gangId := getNamespaceSplicingName(pod.Namespace,
+		getGangNameByPod(pod))
+
+	pods, err := p.podLister.Pods(pod.Namespace).List(labels.NewSelector())
```
eahydra, line 359: List ALL Pods ???

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+	gangId := getNamespaceSplicingName(pod.Namespace,
+		gangName)
+	gang := p.gangCache.getGangFromCache(gangId)
+	if gang == nil {
+		klog.Errorf("can't find gang, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "can't find gang")
+	}
+	gang.addAssumedPod(pod)
```
eahydra, line 287: why not add the waiting Pod in permit phase??

eahydra (changes requested):

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	Success         Status = "Success"
+	Wait            Status = "Wait"
+)
+
+var (
+	_ framework.QueueSortPlugin  = &GangPlugin{}
+	_ framework.PreFilterPlugin  = &GangPlugin{}
+	_ framework.PostFilterPlugin = &GangPlugin{}
+	_ framework.ReservePlugin    = &GangPlugin{}
+	_ framework.PostBindPlugin   = &GangPlugin{}
```
eahydra, line 61: duplicate.

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+				waitingPod.Reject(p.Name(), "optimistic rejection in unReserve due to timeout")
+			}
+		})
+	}
+}
+
+// PostBind just update the gang's BoundChildren
+func (p *GangPlugin) PostBind(ctx context.Context, _ *framework.CycleState, pod *corev1.Pod, nodeName string) {
+	gangName := getGangNameByPod(pod)
+	if gangName == "" {
```
eahydra, line 502: if the gang is PodGroup, should update the PodGroup status..

eahydra (changes requested):

eahydra:
After reading the entire PR, there are a few questions that confuse me.
1. Why not extend it based on coscheduling in scheduler-plugins? The difference from coscheduling is that we need to build Gang according to Annotation, and we need to support GangGroup, and the others are no different. And the current implementation does have a lot in common with coscheduling.
2. Compared with DeniedPodGroup in coscheduling, the concept of ScheduleCycle is more complicated to understand.

I think we should consider from a longer-term development perspective, we must maintain compatibility with the community PodGroup, but also need to provide some expansion capabilities that we think are needed, although @Wenshiqi222 has done a lot of hard work (expressed here My personal sincere respects), but we should considering coscheduling based implementations.

@buptcozy @hormes @yihuifeng

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	gangId := getNamespaceSplicingName(pod.Namespace,
+		gangName)
+	gang := p.gangCache.getGangFromCache(gangId)
+	if gang == nil {
+		klog.Errorf("can't find gang, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "can't find gang")
+	}
+	// check if gang is inited
+	if !gang.HasGangInit {
```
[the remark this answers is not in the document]
Wenshiqi222: this check is used to deal with such the case：If a pod belongs to the podGroup，but user doesn‘t create the podGroup，or the informer received the onPodAdd event before the onPodGroupAdd event，so at that time we need to hold a gang which hasn’t been initialized until the podGroup has been created.

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "gang has timeout")
+	}
+
+	if gang.getChildrenNum() < gang.getGangMinNum() {
+		klog.Errorf("gang child pod not collect enough, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "gang child pod not collect enough")
+	}
+
+	if pod.Annotations[extension.AnnotationGangTimeout] == "true" {
```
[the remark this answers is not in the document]
Wenshiqi222: solved

pkg/scheduler/plugins/gang/plugin.go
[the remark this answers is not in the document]
eahydra: So why not initialize Gang during the PreFilter stage?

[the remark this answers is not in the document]
Wenshiqi222: > So why not initialize Gang during the PreFilter stage?

There are only two kinds of Gang, one is announced by the Annotations, which will be initialized once the pod is created; The other is announced by podGroup which will be initialized once the podGroup is created; 
But from the case i replied above ,if the podGroup is still not created during the pod's PreFilter stage, we cannot initialize the Gang cause we don't have enough gang's information.

Makefile
```diff
@@ -4,10 +4,10 @@ GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 GIT_COMMIT_ID ?= $(shell git rev-parse --short HEAD)
 
 # Image URL to use all building/pushing image targets
-REG ?= ghcr.io
-REG_NS ?= koordinator-sh
-REG_USER ?= ""
-REG_PWD ?= ""
+REG ?= docker.io
```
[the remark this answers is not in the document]
Wenshiqi222: done

cmd/koord-scheduler/app/server.go
```diff
@@ -123,16 +123,16 @@ func runCommand(cmd *cobra.Command, opts *options.Options, schedulingHooks []fra
 		cancel()
 	}()
 
-	cc, sched, err := Setup(ctx, opts, schedulingHooks, registryOptions...)
+	cc, sched, extHandle, err := Setup(ctx, opts, schedulingHooks, registryOptions...)
```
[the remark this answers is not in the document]
Wenshiqi222: my fault , done

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+	// check if gang is timeout by pod's Annotation
+	if pod.Annotations[extension.AnnotationGangTimeout] == "true" {
+		klog.Errorf("gang has timeout, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "gang has timeout")
+	}
+
+	if gang.getChildrenNum() < gang.getGangMinNum() {
+		klog.Errorf("gang child pod not collect enough, gangName: %v, podName: %v", gangId,
```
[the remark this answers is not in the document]
Wenshiqi222: got it, solved

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+}
+
+// PostFilter
+// i. If strict-mode, we will set scheduleCycleValid to false and release all assumed pods.
+// ii. If non-strict mode, we will do nothing.
+func (p *GangPlugin) PostFilter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	gangName := getGangNameByPod(pod)
+	if gangName == "" {
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
```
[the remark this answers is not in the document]
Wenshiqi222: yeah, didn't notice this before, we should always return Unscheduable so that the pod can trigger Preemt

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+	gangId := getNamespaceSplicingName(pod.Namespace,
+		gangName)
+	gang := p.gangCache.getGangFromCache(gangId)
+	if gang == nil {
+		klog.Errorf("can't find gang, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.UnschedulableAndUnresolvable, "can't find gang")
```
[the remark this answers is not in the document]
Wenshiqi222: solved

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: 😂, done

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,534 @@
...
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "can't find gang")
+	}
+	// check if gang is inited
+	if !gang.HasGangInit {
+		klog.Errorf("gang has not init, gangName: %v, podName: %v", gangId,
+			getNamespaceSplicingName(pod.Namespace, pod.Name))
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, "gang has not init")
+	}
+	// check if gang is timeout by pod's Annotation
+	if pod.Annotations[extension.AnnotationGangTimeout] == "true" {
```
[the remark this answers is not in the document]
Wenshiqi222: i think once the gang is timeout, it shows that the cluster's condition is not that good to hold the gang, so we use the annotations to aware users about this

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: In my implement, Permit is used to check the Gang's assuming condition, as for adding the waitingPod to Gang, this action should happen just after the pod being assumed during the Kubernetes ScheduleOne, which is the Reserve stage

buptcozy:
> After reading the entire PR, there are a few questions that confuse me.
> 
> 1. Why not extend it based on coscheduling in scheduler-plugins? The difference from coscheduling is that we need to build Gang according to Annotation, and we need to support GangGroup, and the others are no different. And the current implementation does have a lot in common with coscheduling.
> 2. Compared with DeniedPodGroup in coscheduling, the concept of ScheduleCycle is more complicated to understand.
> 
> I think we should consider from a longer-term development perspective, we must maintain compatibility with the community PodGroup, but also need to provide some expansion capabilities that we think are needed, although @Wenshiqi222 has done a lot of hard work (expressed here My personal sincere respects), but we should considering coscheduling based implementations.
> 
> @buptcozy @hormes @yihuifeng

Besides "annotation", We offer a new "non-strict" mode to help large job get resources more easily.   Also, 'Scheduling cycle' is used to solve coscheduling's incorrect calcucation in "strict mode".  I think these are neccesory to imporve considering.

eahydra:
> 

We can copy the `coscheduling` into koordinator, and add these enhancements including nonStrictMode.

eahydra:
> 

We can copy the coscheduling into koordinator, and add these enhancements including nonStrictMode.

pkg/scheduler/plugins/gang/cache.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+			gangId := getNamespaceSplicingName(pod.Namespace,
+				gangName)
+			gang := gangCache.getGangFromCache(gangId)
+			if gang == nil {
+				klog.Errorf("Didn't find gang when recover by pod, pod:%v, gang: %v", pod.Name, gangId)
+				continue
+			}
+			gang.addBoundPod(pod)
+			gang.setResourceSatisfied()
```
yihuifeng, line 117: set gang as resource satisfied status if only a child pod is bound?

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,519 @@
...
+	creationTime2 := p.getCreatTime(podInfo2)
+	if creationTime1.Equal(creationTime2) {
+		return getNamespaceSplicingName(podInfo1.Pod.Name, podInfo1.Pod.Namespace) < getNamespaceSplicingName(podInfo2.Pod.Name, podInfo2.Pod.Namespace)
+	}
+	return creationTime1.Before(creationTime2)
+}
+
+// PreFilter
+// if non-strict-mode, we only do step1 and step2:
+// i.Check whether childes in Gang has met the requirements of minimum number under each Gang, and reject the pod if negative.
```
yihuifeng, line 162: childes -> children

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,428 @@
...
+	Children          map[string]*v1.Pod
+	// pods that have already assumed(waiting in Permit stage)
+	WaitingForBindChildren map[string]*v1.Pod
+	// pods that have already bound
+	BoundChildren map[string]*v1.Pod
+	// if assumed  pods number has reached to MinRequiredNumber
+	ResourceSatisfied bool
+
+	// if the gang should be passed at PreFilter stage(Strict-Mode)
+	ScheduleCycleValid bool
```
yihuifeng, line 64: please give some instruction for the following three members, ScheduleCycleValid, ScheduleCycle and ChildrenScheduleRoundMap.

pkg/scheduler/plugins/gang/plugin.go
```diff
@@ -0,0 +1,519 @@
...
+					namespacedName := getNamespaceSplicingName(pod.Namespace, pod.Name)
+					s.Map[namespacedName] = pod
+				}
+				s.Unlock()
+			}
+		}
+	}
+}
+
+func (p *GangPlugin) AllowGangGroup(pod *corev1.Pod) {
```
yihuifeng, line 375: it is better to use gang as the parameter, rather than pod.

pkg/scheduler/plugins/gang/cache.go
Wenshiqi222: if a pod is bound, which means the gang has already passed the Permit stage check, so the gang must be  resouceSatisfied

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,428 @@
...
+	// these fields used to count the cycle
+	ScheduleCycle            int
+	ChildrenScheduleRoundMap map[string]int
+
+	GangFrom    string
```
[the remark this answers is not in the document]
Wenshiqi222: it indicates the gang is created from either the Annotations way or PodGroup crd way

pkg/scheduler/plugins/gang/cache.go
```diff
@@ -0,0 +1,222 @@
...
+	gangNamespace := pg.Namespace
+	gangName := pg.Name
+
+	gangId := getNamespaceSplicingName(gangNamespace, gangName)
+	gang := gangCache.getGangFromCache(gangId)
+	if gang == nil {
+		gang = gangCache.createGangToCache(gangId)
+		klog.Infof("Create gang by podGroup on add, gangName: %v", gangId)
+	}
+	gang.tryInitByPodGroup(pg, gangCache.pluginArgs)
```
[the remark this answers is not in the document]
Wenshiqi222: we got the HasGangInit field to prevent the gang's duplicated initializations

pkg/scheduler/plugins/gang/gang_helper.go
```diff
@@ -0,0 +1,110 @@
...
+	return strategicpatch.CreateTwoWayMergePatch(oldData, newData, &v1.Pod{})
+}
+
+// StringToGangGroupSlice
+// Parse gang group's annotation like :"["nsA/gangA","nsB/gangB"]"  => goLang slice : []string{"nsA/gangA"."nsB/gangB"}
+func stringToGangGroupSlice(s string) ([]string, error) {
+	gangGroup := make([]string, 0)
+	err := json.Unmarshal([]byte(s), &gangGroup)
+	if err != nil {
+
```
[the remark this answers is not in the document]
Wenshiqi222: > > 
> 
> We can copy the coscheduling into koordinator, and add these enhancements including nonStrictMode.

Ok，I will copy the Coscheduling Plugin and it's Controller code of Scheduler-Plugins to my implement and  enhance it  according our previous Gang's Proposal.  So next week, I will rewrite all the code and adopt all the comments you @eahydra @buptcozy @yihuifeng @jasonliu747  mentioned before, really thanks for your help.

Wenshiqi222:
> > 
> 
> We can copy the coscheduling into koordinator, and add these enhancements including nonStrictMode.

Ok，I will copy the Coscheduling Plugin and it's Controller code of Scheduler-Plugins to my implement and enhance it according our previous Gang's Proposal. So next week, I will rewrite all the code and adopt all the comments you @eahydra @buptcozy @yihuifeng @jasonliu747 mentioned before, really thanks for your help.

Wenshiqi222:
Should i close this Pr and open another one? @eahydra

jasonliu747:
> Should i close this Pr and open another one? @eahydra

yes, you should. and please make sure to split your change into logical chunks.

Wenshiqi222:
> > Should i close this Pr and open another one? @eahydra
> 
> yes, you should. and please make sure to split your change into logical chunks.

got it, thanks

jasonliu747:
> Should i close this Pr and open another one? @eahydra

我认为这个 PR 可以暂时不关，然后同时提交新 PR 做对比

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: solved, by get pods from gang directly

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: yes

pkg/scheduler/plugins/gang/gang_helper.go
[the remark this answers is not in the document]
Wenshiqi222: solved

pkg/scheduler/plugins/gang/gang_helper.go
```diff
@@ -0,0 +1,110 @@
...
+func parsePgTimeoutSeconds(timeoutSeconds int32) (time.Duration, error) {
+	if timeoutSeconds <= 0 {
+		return 0, fmt.Errorf("podGroup timeout value is illegal,timeout Value:%v", timeoutSeconds)
+	}
+	return time.Duration(timeoutSeconds) * time.Second, nil
+}
+
+// GetPodSubPriority
+// Get pod's sub-priority in Koordinator from label
+func getPodSubPriority(pod *v1.Pod) (int32, error) {
```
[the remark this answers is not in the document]
Wenshiqi222: subPriority is named by Koordinator , should i call subPriority? @eahydra

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,428 @@
...
+	defer gang.lock.Unlock()
+
+	podId := getNamespaceSplicingName(pod.Namespace, pod.Name)
+	klog.Infof("Delete pod from gang: %v, podName: %v", gang.Name, podId)
+
+	delete(gang.Children, podId)
+	delete(gang.WaitingForBindChildren, podId)
+	delete(gang.BoundChildren, podId)
+	delete(gang.ChildrenScheduleRoundMap, podId)
+
```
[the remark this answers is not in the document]
Wenshiqi222: yeah, we should update the resourceSatisfied field here

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -0,0 +1,428 @@
...
+	klog.Infof("TryInitByPodConfig done, gangName: %v, minRequiredNumber: %v, totalChildrenNum: %v, "+
+		"mode: %v, waitTime: %v, groupSlice: %v", gang.Name, gang.MinRequiredNumber, gang.TotalChildrenNum,
+		gang.Mode, gang.WaitTime, gang.GangGroup)
+}
+
+func (gang *Gang) tryInitByPodGroup(pg *v1alpha1.PodGroup, args *schedulingconfig.GangArgs) {
+	gang.lock.Lock()
+	defer gang.lock.Unlock()
+
+	if gang.HasGangInit {
```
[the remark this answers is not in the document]
Wenshiqi222: done

pkg/scheduler/plugins/gang/plugin.go
Wenshiqi222: during the new implement [510](https://github.com/koordinator-sh/koordinator/pull/510), AlloGangGroup is the interface by PodGroupMagager, and it can only be start by the Coescheduling Plugin outside, which only has the parameter pod, so i think it's better to pass pod to the func

pkg/scheduler/plugins/gang/gang.go
Wenshiqi222: done

pkg/scheduler/plugins/gang/plugin.go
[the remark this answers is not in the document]
Wenshiqi222: ok, as my new implement, this action is removed, timeout will do as society does. [510](https://github.com/koordinator-sh/koordinator/pull/510)

pkg/scheduler/plugins/gang/plugin.go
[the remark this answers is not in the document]
Wenshiqi222: ok, as my new implement, this action is removed, timeout will do as society does. [510](https://github.com/koordinator-sh/koordinator/pull/510)

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

JasonRD:
Preparing review...

PR #405, opened 2022-07-26
add QueueSortPlugin

Wenshiqi222:
We design the QueueSortPlugin plugin to implement the QueueSort extension point separately,
so that we can integrate queue sort logic of all plugins, and register them at one time.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/queueSort/queuePlugin.go
+++ b/pkg/scheduler/plugins/queueSort/queuePlugin.go
@@ -1,0 +1,49 @@
+package queueSort
+
+import (
+	"github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/gang"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/component-helpers/scheduling/corev1"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/queueSort/queuePlugin.go
```diff
@@ -0,0 +1,49 @@
...
+	}, nil
+}
+func (qs QueueSortPlugin) Name() string { return Name }
+
+//Less is used to sort pods in the scheduling queue in the following order.
+//Firstly, compare the priorities of the two pods, the higher priority is at the front of the queue.
+//Secondly, compare creationTimestamp of two pods, if pod belongs to a Gang, then we compare creationTimestamp of the Gang, the one created first will be at the front of the queue.
+//Finally, compare pod's namespace, if pod belongs to a Gang, then we compare Gang name.
+func (qs *QueueSortPlugin) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
+	prio1 := corev1.PodPriority(podInfo1.Pod)
```
hormes, line 37: how about k8s priority and subpriority of koordinator ？

PR #421, opened 2022-08-01
chore: execute staticcheck instead of github action by running golang…

eahydra:
…ci-lint

### V. Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -80,6 +80,7 @@ linters:
     - misspell
     - vet
     - unconvert
+    - staticcheck
 issues:
   exclude:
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -80,6 +80,7 @@ linters:
     - misspell
     - vet
     - unconvert
+    - staticcheck
```
jasonliu747, line 83: Is it the same as the deleted action?

eahydra: Yes same.

jasonliu747 (approved):
/lgtm

PR #431, opened 2022-08-02
fix reservation on mutil-scheduler

saintube:

Fix the event handler for unscheduled reservations in the mutil-scheduler scenario:

1. When multiple schedulers run in the same cluster, unscheduled reservations should only get enqueued for the specified scheduler (according to field `spec.template.spec.schedulerName`).
2. Ignore the deletion of expired reservations which have been existing before the scheduler started.

1. Create reservation objects with different scheduler names.
2. Check the log of the koord-scheduler about event handlers' processing.
3. Unscheduled reservations which specify other schedulers should not get enqueued in the koord-scheduler.
4. When the scheduler starts, no ERROR logs for removing expired reservations.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/rpod.go
+++ b/pkg/scheduler/plugins/reservation/rpod.go
@@ -87,3 +87,10 @@ func GetReservePodNodeName(pod *corev1.Pod) string {
 func GetReservationNameFromReservePod(pod *corev1.Pod) string {
 	return pod.Annotations[AnnotationReservationName]
 }
+
+func GetReservationSchedulerName(r *schedulingv1alpha1.Reservation) string {
+	if r == nil || r.Spec.Template == nil || len(r.Spec.Template.Spec.SchedulerName) <= 0 {
+		return corev1.DefaultSchedulerName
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/reservation/rpod.go
```diff
@@ -87,3 +87,10 @@ func GetReservePodNodeName(pod *corev1.Pod) string {
 func GetReservationNameFromReservePod(pod *corev1.Pod) string {
 	return pod.Annotations[AnnotationReservationName]
 }
+
+func GetReservationSchedulerName(r *schedulingv1alpha1.Reservation) string {
+	if r == nil || r.Spec.Template == nil {
+		return corev1.DefaultSchedulerName
+	}
+	return r.Spec.Template.Spec.SchedulerName
```
hormes, line 95: this field may be ""

saintube: fixed

eahydra (commented):
/lgtm

PR #445, opened 2022-08-05
bugfix: avoid pod terminating in docker

ZYecho:

### V. Checklist


cheimu:
Note: this pr is very important, since current version, v0.6, is still has problem for kubelet to terminate pods under some circumstance.

PR #446, opened 2022-08-08
bugfix: fix exec exit with 126

### LINKED ISSUE #437: [BUG]koord-runtime-proxy failed to proxy exec request without -it option
**What happened**:
Any scenario involving with exec request, including `ExecSync` for docker, `PostStartHook` for pod Lifecycle, and readiness probe etc., the exec will be failed as shown below.
image

They should be executed successfully.

It is very easy to reproduce. As long as echo something when exec. For example, the following pod manifest has a `postStart` hook and `echo 1` . It will fail to start pods and the error is the screenshot from the above.
```
apiVersion: v1
kind: Pod
metadata:
  name: sample
spec:
  containers:
  - env:
    - name: TEST
      value: value0
    image: nginx:1.20.1-alpine
    lifecycle:
      postStart:
        exec:
          command:
          - /bin/sh
          - -c
          - echo 1 && sleep 2s
    name: main
    readinessProbe:
      failureThreshold: 3
      periodSeconds: 10
      successThreshold: 1
      tcpSocket:
        port: 80
      timeoutSeconds: 1
```

We found a very interesting fact: 
When try to manually trigger it using docker cli
```
docker --host=unix:///var/run/runtimeproxy/runtimeproxy.sock exec -it 246ab5790f80 /bin/sh -c "echo 1 && sleep 2s"
```
Everything is doing well. However, when remove `-it` flag 
```
docker --host=unix:///var/run/runtimeproxy/runtimeproxy.sock exec 246ab5790f80 /bin/sh -c "echo 1 && sleep 2s"
```
The error reproduced.
It turns out that, kubelet will set `in` to `nil` and `tty` to `false` when doing `exec`. 
image

And koord-runtime-proxy only proxies `exec` request without intercepting. We highly doubt the problem is the reverse proxy.
image

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18

ZYecho:

fix #437 

### V. Checklist

### DIFF
--- a/pkg/util/httputil/reverseproxy.go
+++ b/pkg/util/httputil/reverseproxy.go
@@ -1,0 +1,654 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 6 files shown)


pkg/util/httputil/reverseproxy.go
```diff
@@ -0,0 +1,650 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
```
hormes, line 3: should add the original license here

Issue #440, opened 2022-08-04
[proposal] Reservation support allocate once semantics


eahydra:
**What is your proposal**:

Reservation support allocate once semantics.

**Why is this needed**:

The Reservation currently supported by koord-scheduler is valid until the TTL expires.  

However, in some scenarios, it is expected that the Reservation will fail after being consumed, so as to avoid the problem that the Reservation will continue to occupy and waste resources after the Pod associated with the Reservation is reduced in size. 

For example, PaaS creates reservation resources before initiating expansion. The newly created Pods use these reservations during expansion. After a period of time, the expansion is performed, and these expanded Pods are destroyed, but these reservations are still retained in the cluster. In the implementation of the current scheduler, these reservations continue to occupy resources and can only be deleted by the upper-layer PaaS. This method is not flexible enough.

**Is there a suggested solution, if so, please add it**:

PR #447, opened 2022-08-08
apis: add allocateOnce in reservation

saintube:

Update with `allocateOnce` API in Reservation.

#440 

### V. Checklist

### DIFF
--- a/apis/scheduling/v1alpha1/reservation_types.go
+++ b/apis/scheduling/v1alpha1/reservation_types.go
@@ -41,12 +41,6 @@ type ReservationSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
 	Owners []ReservationOwner `json:"owners"`
-	// By default, the resources requirements of reservation (specified in `template.spec`) is filtered by whether the
-	// node has sufficient free resources (i.e. ReservationRequest <  NodeFree).
-	// When `preAllocation` is set, the scheduler will skip this validation and allow overcommitment. The scheduled
-	// reservation would be waiting to be available until free resources are sufficient.
-	// +optional
-	PreAllocation bool `json:"preAllocation,omitempty"`
 	// Time-to-Live period for the reservation.
 	// `expires` and `ttl` are mutually exclusive. Defaults to 24h. Set 0 to disable expiration.
 	// +kubebuilder:default="24h"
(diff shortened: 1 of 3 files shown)


apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -57,6 +51,17 @@ type ReservationSpec struct {
...
+	// node has sufficient free resources (i.e. Reservation Request <  Node Free).
+	// When `preAllocation` is set, the scheduler will skip this validation and allow overcommitment. The scheduled
+	// reservation would be waiting to be available until free resources are sufficient.
+	// +optional
+	PreAllocation bool `json:"preAllocation,omitempty"`
+	// By default, reserved resources are always allocatable as long as the reservation phase is Available. When
+	// `AllocateOnce` is set, the reserved resources are only available for the first owner who allocates successfully
+	// and are not allocatable to other owners anymore.
+	// +optional
+	AllocateOnce bool `json:"allocateOnce,omitempty"`
```
saintube, line 64: Alternative:

- wrap `AllocateOnce` and `PreAllocation` into a new type `AllocationOption`:

```go
type AllocationOption struct {
    PreAllocation bool
    Once bool
}
```

hormes: How to deal with the concurrency problem of bind and changing reservation state?

saintube: Currently, the reservation plugin simply uses optimistic concurrency control for these cases.

e.g. When multiple pods allocate(Bind) one reservation almost simultaneously, the plugin

1. gets the latest version of the reservation to bind
2. double-checks if the reservation still matches the pod
3. updates the reservation's status

If a reservation is allocated and becomes `Succeeded`, other pods which match the old version will fail in step 2.

This may introduce some updating conflicts. We can further reduce the probability of concurrency conflicts by the following:

- Only the scheduler which specified by the reserve pod (via `spec.schedulerName`) can change the phase of the reservation.
- The scheduling pod only can allocate the available reservations which specify the same scheduler.

hormes: The update of reservation and the bind phase of the pod cannot maintain atomicity, thus the semantics of one cannot be guaranteed?

et. Pod A use reservation B, bind Pod A succeed but update reservation B failed, and after scheduler restart, reservation B will continue to exist.

saintube: > The update of reservation and the bind phase of the pod cannot maintain atomicity, thus the semantics of one cannot be guaranteed?
> 
> et. Pod A use reservation B, bind Pod A succeed but update reservation B failed, and after scheduler restart, reservation B will continue to exist.

In the current design, the Pod A can succeed the binding only if it has updated reservation B successfully. Otherwise, it would fail the binding cycle.

eahydra (commented):
/lgtm

PR #458, opened 2022-08-10
koord-descheduler: init evictor plugin first

eahydra:

Other plugins may depend on the evictor plugin, so we should initialize the evictor plugin first.

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/runtime/framework.go
+++ b/pkg/descheduler/framework/runtime/framework.go
@@ -115,9 +115,6 @@ func NewFramework(r Registry, profile *deschedulerconfig.DeschedulerProfile, opt
 		getPodsAssignedToNodeFunc: options.getPodsAssignedToNodeFunc,
 	}
 
-	// get needed plugins from config
-	pg := f.pluginsNeeded(profile.Plugins)
-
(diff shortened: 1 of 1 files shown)


jasonliu747:
why do we need this change? add a brief summary in PR description 😉

eahydra:
> why do we need this change? add a brief summary in PR description 😉

There is a comment in code "Other plugins may depend on the evictor plugin, so we should initialize the evictor plugin first."

pkg/descheduler/framework/runtime/framework.go
```diff
@@ -219,32 +244,33 @@ type extensionPoint struct {
...
 	}
 }
 
-func (f *frameworkImpl) pluginsNeeded(plugins *deschedulerconfig.Plugins) sets.String {
-	pgSet := sets.String{}
-
+func (f *frameworkImpl) getOtherExtensionPoints(plugins *deschedulerconfig.Plugins) []extensionPoint {
 	if plugins == nil {
-		return pgSet
+		return []extensionPoint{}
```
saintube, line 259: how about `return nil`

PR #470, opened 2022-08-11
feature: support read args from KubeletConfiguration

### LINKED ISSUE #457: [BUG] koordlet is not compatible with kubelet using --config to set static policy
**What happened**:

When kubelet uses --config to set static policy, koordlet cannot perceive the corresponding configuration, which will cause compatibility problems.

ZYecho:

fix #457 

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/states_noderesourcetopology.go
+++ b/pkg/koordlet/statesinformer/states_noderesourcetopology.go
@@ -29,8 +29,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
+	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpumanager/state"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+	"k8s.io/kubernetes/pkg/kubelet/kubeletconfig/configfiles"
+	utilfs "k8s.io/kubernetes/pkg/util/filesystem"
 
 	"github.com/koordinator-sh/koordinator/apis/extension"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/metriccache"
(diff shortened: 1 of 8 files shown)


pkg/koordlet/statesinformer/states_noderesourcetopology.go
```diff
@@ -165,6 +168,25 @@ func (s *statesInformer) reportNodeTopology() {
...
+	filePath, err := system.GuessConfigFilePathFromKubelet()
+	if err != nil {
+		klog.Error("failed to read config file path from kubelet args")
+	}
+	if filePath != "" {
+		kubeletConfig, err := loadConfigFile(filePath)
+		if err != nil {
+			klog.Errorf("failed to load kubelet config, err: %v", err)
+		} else {
+			if len(cpuManagerPolicy.Options) == 0 {
```
zwzhang0107, line 181: kubelet use params in `--config ${file}` first?

pkg/util/system/cpu_manager_linux.go
```diff
@@ -52,3 +53,39 @@ func GuessCPUManagerOptFromKubelet() (string, string, map[string]string, error)
...
+	}
+	var argsFilePath string
+	fs := pflag.NewFlagSet("GuessTest", pflag.ContinueOnError)
+	fs.ParseErrorsWhitelist.UnknownFlags = true
+	fs.StringVar(&argsFilePath, "config", argsFilePath, "")
+	if err := fs.Parse(kubeletArgs[1:]); err != nil {
+		return "", fmt.Errorf("failed to parse kubelet's args, kubelet version may not support: %v", err)
+	}
+	if argsFilePath == "" {
+		return "", fmt.Errorf("failed to get config file path, --config is empty")
```
zwzhang0107, line 77: if kubelet not specifies `--config`, error logs will be overwhelming

pkg/koordlet/statesinformer/states_noderesourcetopology.go
zwzhang0107: add a comment here: kubelet use args first

pkg/koordlet/statesinformer/states_noderesourcetopology.go
```diff
@@ -165,6 +168,25 @@ func (s *statesInformer) reportNodeTopology() {
 		Policy:  cpuPolicy,
 		Options: cpuManagerOpt,
 	}
+
```
hormes, line 171: Only one of GuessCPUManagerOptFromKubelet and GuessConfigFilePathFromKubelet will be succeed, but we should not log error in the failed path.

PR #478, opened 2022-08-12
koord-scheduler: reservation support allocateOnce

### LINKED ISSUE #440: [proposal] Reservation support allocate once semantics

saintube:

Reservation supports `allocateOnce`.

Once a pod allocates an available reservation with `allocateOnce` set to true, the reservation will become `Succeeded` and no longer allocatable to other owners.

fixes #440 

1. Submit an available reservation with `allocateOnce` set.
2. Submit a pod A that matches the reservation.
3. Watch the reservation is allocated, and its status becomes `Succeeded`.
4. Submit another pod B that also matches the reservation (if it is available).
5. Check if the reservation is not allocatable to pod B.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/hook.go
+++ b/pkg/scheduler/plugins/reservation/hook.go
@@ -159,7 +159,8 @@ func (h *Hook) prepareMatchReservationState(handle frameworkext.ExtendedHandle, 
 				klog.V(5).Infof("unable to convert to *schedulingv1alpha1.Reservation, obj %T", obj)
 				continue
 			}
-			if !IsReservationScheduled(r) { // only count scheduled reservations
+			// only count available reservations, ignore succeeded ones
+			if !IsReservationAvailable(r) {
 				continue
 			}
 			if matchReservation(pod, newReservationInfo(r)) {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -159,7 +159,8 @@ func (h *Hook) prepareMatchReservationState(handle frameworkext.ExtendedHandle,
 				klog.V(5).Infof("unable to convert to *schedulingv1alpha1.Reservation, obj %T", obj)
 				continue
 			}
-			if !IsReservationScheduled(r) { // only count scheduled reservations
+			// only count scheduled & available reservations, ignore succeeded ones
+			if !IsReservationScheduled(r) {
```
eahydra, line 163: The name `IsReservationScheduled` is confusing,`Scheduled` and `Available` are two different states. `Available` requires that the Reservation MUST BE successfully scheduled and there are remaining resources to allocate, but if there are no resources to allocate, it is also successfully scheduled. Why not just call it `IsReservationAllocatable` or `IsReservationAvailable`?

saintube: OK. It is renamed `IsReservationAvailable`.

eahydra (commented):
/lgtm

PR #551, opened 2022-08-29
koord-scheduler: CPU allocate algorithm supports maxRefCount 

ZiMengSheng:

In some over-commitment scenario, a logic cpu can be allocated more than once and less than a upper bound. We call the uppder bound “MaxRefCount” and call the number of allocations of a logic cpu “RefCount” of that cpu.
This PR noticed RefCount logic has been partial supported in NodeNUMAResourcePlugin. This PR supports MaxRefCount and makes RefCount logic complete in CPU allocate algorithm.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/cpu_allocation.go
+++ b/pkg/scheduler/plugins/nodenumaresource/cpu_allocation.go
@@ -82,8 +82,11 @@ func (n *cpuAllocation) releaseCPUs(podUID types.UID) {
 	}
 }
 
-func (n *cpuAllocation) getAvailableCPUs(cpuTopology *CPUTopology, reservedCPUs CPUSet) (availableCPUs CPUSet, allocated CPUDetails) {
-	allocated = n.allocatedCPUs.Clone()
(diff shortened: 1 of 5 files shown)


pkg/scheduler/plugins/nodenumaresource/cpu_allocation.go
```diff
@@ -82,8 +82,18 @@ func (n *cpuAllocation) releaseCPUs(podUID types.UID) {
 	}
 }
 
-func (n *cpuAllocation) getAvailableCPUs(cpuTopology *CPUTopology, reservedCPUs CPUSet) (availableCPUs CPUSet, allocated CPUDetails) {
+func (n *cpuAllocation) getAvailableCPUs(cpuTopology *CPUTopology, maxRefCount int, reservedCPUs CPUSet) (availableCPUs CPUSet, allocated CPUDetails) {
 	allocated = n.allocatedCPUs.Clone()
+	if maxRefCount > 1 {
```
eahydra, line 87: if maxRefCount > 1, the line 86 no need to execute.

eahydra (changes requested):

eahydra:
as we talked, we should add some UT cases for maxRefCount

eahydra (approved):
/lgtm

PR #556, opened 2022-08-30
cpu suppress: only limits cpuset at besteffot container level if kube…

### LINKED ISSUE #546: [BUG] failed to use cpu suppress for BE pods on K8s 1.18 with kubelet cpu-static policy
**What happened**:
if kubelet specifies `static` option for `cpu-policy`, when enable cpu suppress policy, BE pod will be failed created with error:
```
    state:
      terminated:
        containerID: containerd://aabc281f6798f88cf6b5eb2e35e997121a01386e23396ea466ce0c628c3e313b
        exitCode: 128
        finishedAt: "2022-08-29T07:28:40Z"
        message: 'failed to create containerd task: OCI runtime create failed: container_linux.go:367:
          starting container process caused: process_linux.go:340: applying cgroup
          configuration for process caused: failed to write "0,3-52,55-103": write
          /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod80eeefbd_e18a_4493_a617_c9634dd46bdb.slice/cri-containerd-aabc281f6798f88cf6b5eb2e35e997121a01386e23396ea466ce0c628c3e313b.scope/cpuset.cpus:
          permission denied: unknown'
        reason: StartError
        startedAt: "1970-01-01T00:00:00Z"
```

**Environment**:
- App version:
- Kubernetes version (use `kubectl version`): kubelet 1.18
- Others:

**How to fix**:
- [ ] koordlet runtime hook plugin `cpusetAllocator` should support cpuset injection during `UpdateContainer`
- [ ] runtime-proxy should support `PreUpdateContainer` injection, which is already defined but not implemented yet.
- [ ] cpu suppress strategy should keep cpuset of besteffort dir and pod level as `allCPUSets`, only change the cpuset on container level within  `be-share-pool` (aka. allCPU - CPUSet(LSE). so that cpu suppress can still work well under `cpu-static` policy even without runtime-proxy.

zwzhang0107:
…let use static cpu policy

for kubelet none policy: keep cpu suppress at besteffort root dir level, make sure that new pod can be throttled in time.
for kubelet static policy: keep cpu suppress at container level, so that we can avoid the conflicts during container creating.

after we finish this, we can make cpu suppress policy work well without runtime-proxy.

fixes #546

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/cpu_suppress.go
+++ b/pkg/koordlet/resmanager/cpu_suppress.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -38,6 +39,11 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
 	"github.com/koordinator-sh/koordinator/pkg/util"
 	"github.com/koordinator-sh/koordinator/pkg/util/system"
+)
+
+const (
+	podCgroupPathRelativeDepth       = 1
+	containerCgroupPathRelativeDepth = 2
 )
 
 var (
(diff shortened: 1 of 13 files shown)


pkg/koordlet/resmanager/cpu_suppress.go
```diff
@@ -161,6 +191,24 @@ func (r *CPUSuppress) calculateBESuppressCPU(node *corev1.Node, nodeMetric *metr
 	return nodeBESuppressCPU
 }
 
+func (r *CPUSuppress) applyBESuppressCPUSet(beCPUSet []int32, oldCPUSet []int32) error {
+	nodeTopo := r.resmanager.statesInformer.GetNodeTopo()
+	kubeletPolicy, err := apiext.GetKubeletCPUManagerPolicy(nodeTopo.Annotations)
+	if err != nil {
+		klog.Warningf("failed to get kubelet cpu manager policy, error %v", err)
+	}
```
ZYecho, line 199: return err?

eahydra (commented):
/lgtm

PR #559, opened 2022-08-30
koord-descheduler: fix failed evict with StatefulSet Pod

eahydra:

1. When evict StatefulSet pod via PodMigrationJob,  koord-descheduler cannot confirm whether the Pod has been migrated successfully since the newly created Pod has the same name. So koord-descheduler should check if the evicting Pod's UID is different from the previous Pod. 
2. Also in the scene,  the logic of check whether the Reservation reserved on same on with Pod is also invalid.

### V. Checklist


jasonliu747:
/assign
I'll take a look later today~

PR #572, opened 2022-09-01
cpu suppress check whether nodeslo enable

huiwq1990:
Signed-off-by: huiwq1990 <huiwq1990@163.com>

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/be_reconcile.go
+++ b/pkg/koordlet/resmanager/be_reconcile.go
@@ -38,7 +38,7 @@ func (r *resmanager) reconcileBECgroup() {
 		if extension.GetPodQoSClass(podMeta.Pod) != extension.QoSBE {
 			continue
 		}
-		if getCPUSuppressPolicy(nodeSLO) != v1alpha1.CPUCfsQuotaPolicy {
+		if enable, policy := getCPUSuppressPolicy(nodeSLO); enable || policy != v1alpha1.CPUCfsQuotaPolicy {
 			reconcileBECPULimit(podMeta)
 		}
 		reconcileBECPUShare(podMeta)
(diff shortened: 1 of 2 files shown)


jasonliu747 (commented):
/lgtm

pkg/koordlet/resmanager/be_reconcile.go
```diff
@@ -38,7 +38,7 @@ func (r *resmanager) reconcileBECgroup() {
 		if extension.GetPodQoSClass(podMeta.Pod) != extension.QoSBE {
 			continue
 		}
-		if getCPUSuppressPolicy(nodeSLO) != v1alpha1.CPUCfsQuotaPolicy {
+		if enable, policy := getCPUSuppressPolicy(nodeSLO); enable && policy != v1alpha1.CPUCfsQuotaPolicy {
```
saintube, line 41: We should check `!enable || policy != CPUCfsQuotaPolicy`.

> `ResourceUsedThresholdWithBE.Enable` is a switch for the *CPUSuppress* feature. And we generally expect the BEReconcile is enabled by default, and if needed to disable it, we use the feature gate BEReconcile. However, CPUSuppress may conflict with BEReconcile when we set SuppressPolicy as `CPUCfsQuotaPolicy` (both modify the cgroups `cpu.cfs_quota_us`), so the policy is checked in BEReconcile.  
> You got part of it right. `CPUCfsQuotaPolicy` does nothing when `ResourceUsedThresholdWithBE.Enable=false`.

zwzhang0107: +1

pkg/koordlet/resmanager/be_reconcile.go
```diff
@@ -38,7 +38,7 @@ func (r *resmanager) reconcileBECgroup() {
 		if extension.GetPodQoSClass(podMeta.Pod) != extension.QoSBE {
 			continue
 		}
-		if getCPUSuppressPolicy(nodeSLO) != v1alpha1.CPUCfsQuotaPolicy {
+		if enable, policy := getCPUSuppressPolicy(nodeSLO); enable || policy != v1alpha1.CPUCfsQuotaPolicy {
```
saintube, line 41: `enable || policy != v1alpha1.CPUCfsQuotaPolicy` or `!enable || policy != v1alpha1.CPUCfsQuotaPolicy` ?

PR #579, opened 2022-09-02
add extension plugin framework for nodeslo

zwzhang0107:

### V. Checklist


pkg/util/system/util_test_tool.go
```diff
@@ -48,6 +48,10 @@ func NewFileTestUtil(t *testing.T) *FileTestUtil {
 	return &FileTestUtil{TempDir: tempDir, t: t}
 }
 
+func (c *FileTestUtil) Cleanup() {
+	os.RemoveAll(c.TempDir)
```
saintube, line 52: Do we still need the `Cleanup()`? Directories created by `t.TempDir()` are cleaned automatically when the unit test finished.

zwzhang0107: keep this function only for ut

jasonliu747 (commented):
The diff coverage is a little bit low. Would you mind adding more UTs for this PR?

PR #588, opened 2022-09-06
koord-scheduler: fix Reserve Pod concurrent map write fatal error

eahydra:

If multiple goroutines call NewReservePod to build a Reserve Pod for the same Reservation object at the same time, a concurrent map write error will be encountered.

### V. Checklist


jasonliu747 (commented):
/lgtm

PR #602, opened 2022-09-08
Qos manager framework

stormgbs:
### V. Checklist


pkg/koordlet/common/utils/containers.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/pkg/util"
+	"github.com/koordinator-sh/koordinator/pkg/util/runtime"
+)
+
+// KillContainers kills containers inside the pod
+func KillContainers(pod *corev1.Pod, message string) {
```
jasonliu747, line 28: should we move `KillContainers` under `pkg/utils/container`?

pkg/koordlet/common/utils/pod.go
```diff
@@ -0,0 +1,36 @@
...
+package utils
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
+	"github.com/koordinator-sh/koordinator/pkg/util"
+)
+
+func GetPodMetas(pods []*corev1.Pod) []*statesinformer.PodMeta {
```
jasonliu747, line 26: Should we move `GetPodMetas` under `pkg/utils/pod`?

pkg/koordlet/common/utils/containers.go
hormes: It's better to move when needed

Issue #407, opened 2022-07-26
[proposal]cgroup v2 support of koordlet


zwzhang0107:
**What is your proposal**:

**Why is this needed**:
for details, please check https://github.com/koordinator-sh/koordinator/issues/369#issuecomment-1192098467

**Is there a suggested solution, if so, please add it**:

saintube:
link #369

saintube:
The Plan of Cgroups-v2 Support:

1. [ ] Refactor Resource Executor module
  - [ ] Refactor CgroupResource
  - [ ] Refactor ResourceUpdater
2. [ ] Compatibility of Koordlet Features
 - [ ] NodeMetric Report
 - [ ] Dynamic Resource Overcommitment
 - [ ] BECgroupReconcile
 - [ ] CgroupReconcile
 - [ ] CPUBurst
 - [ ] CPUSuppress
 - [ ] RdtResctrl
 - [ ] MemoryEvict

zwzhang0107:
> * [ ]  CgroupReconcile

also performance_collector for @songtao98

zwzhang0107:
milestone v1.2

zwzhang0107:
- [ ] remove the former `executor` module

songtao98:
- [ ] PerformanceCollector @Re-Grh

zwzhang0107:
checking public methods for cgoups r/w should be set as priviate @saintube @Re-Grh

PR #631, opened 2022-09-16
fix: panic on node nil

guilhem:


jasonliu747 (commented):
/lgtm

pkg/koordlet/koordlet.go
```diff
@@ -105,7 +105,13 @@ func NewDaemon(config *config.Configuration) (Daemon, error) {
 		}
 		klog.Infof("can not detect cgroup driver from 'kubepods' cgroup name")
 
-		port := int(statesInformer.GetNode().Status.DaemonEndpoints.KubeletEndpoint.Port)
+		node := statesInformer.GetNode()
```
eahydra, line 108: thanks for your fix the issue, but why can't get Node, have you found the root cause?

guilhem: Root cause seems to be #407

PR #641, opened 2022-09-20
koord-scheduler: fix schedulingPhaseHook registerd log nil pointer bug

ZiMengSheng:

When schedulingPhase hook is registered, we print the registerd hook type info in one log before. But if hook doesn't implement all type hooks, log will lead to nil pointer err.
This pr fix this bug.

### V. Checklist


jasonliu747 (commented):
/lgtm

eahydra (approved):
/lgtm
/approve

PR #643, opened 2022-09-20
fix CoScheduling ut:wg count negative bug

Wenshiqi222:
Signed-off-by: wenshiqi <wen470366661@163.com>
fix wg count negative bug
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/coscheduling_test.go
+++ b/pkg/scheduler/plugins/coscheduling/coscheduling_test.go
@@ -199,18 +199,18 @@ func TestLess(t *testing.T) {
 	gp := suit.plugin.(*Coscheduling)
 
 	var lowPriority, highPriority = int32(10), int32(100)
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/coscheduling/coscheduling_test.go
```diff
@@ -221,11 +221,11 @@ func TestLess(t *testing.T) {
 			},
 		},
 	}
-	//GangB by PodGroup
+	// GangB by PodGroup
 	gangBCreatTime := now.Add(5 * time.Second)
 	pg := makePg("gangB", gangB_ns, 2, &gangBCreatTime, nil)
 	suit.start()
-	//creat gangA and gangB
+	// creat gangA and gangB
```
jasonliu747, line 228: create

Wenshiqi222: yes done

PR #689, opened 2022-10-12
bugfix: append origin env when inject gpu env

ZYecho:

### V. Checklist


pkg/koordlet/runtimehooks/proxyserver/service.go
```diff
@@ -68,6 +68,7 @@ func (s *server) PreCreateContainerHook(ctx context.Context,
 		ContainerAnnotations: req.GetContainerAnnotations(),
 		ContainerResources:   req.GetContainerResources(),
 		PodCgroupParent:      req.GetPodCgroupParent(),
+		ContainerEnvs:        req.GetContainerEnvs(),
```
zwzhang0107, line 71: other hooks also should be added

ZYecho: 只有create能mutate env args，其他阶段没有机会

zwzhang0107: 统一对齐吧，就像stop阶段也改不了PodCgroupParent一样

PR #698, opened 2022-10-14
koord-manager: debug engine mode

xulinfei1996:

### V. Checklist

### DIFF
--- a/pkg/webhook/add_quota.go
+++ b/pkg/webhook/add_quota.go
@@ -21,6 +21,7 @@ import (
 	utilfeature "github.com/koordinator-sh/koordinator/pkg/util/feature"
 	"github.com/koordinator-sh/koordinator/pkg/webhook/elasticquota/mutating"
 	"github.com/koordinator-sh/koordinator/pkg/webhook/elasticquota/validating"
+	"github.com/koordinator-sh/koordinator/pkg/webhook/services"
 )
 
 func init() {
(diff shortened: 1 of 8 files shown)


pkg/webhook/add_quota.go
```diff
@@ -31,4 +32,6 @@ func init() {
 	addHandlersWithGate(validating.HandlerMap, func() (enabled bool) {
 		return utilfeature.DefaultFeatureGate.Enabled(features.ElasticQuotaValidatingWebhook)
 	})
+
+	services.RegisterWebhookService("elasticQuota", &validating.ElasticQuotaValidatingHandler{})
```
FillZpp, line 36: I don't recommend to create the new webhook service provider, that's a little weird and seems defined with `/apis/v1/` base for no reason.

How about add a new `HTTPHandlerMap map[string]http.Handler` in `server.go`, the handlers in which will all be registered by `server.Register`. And then you can add your custom URL path and HTTP handler in `add_quota.go` from any other webhook plugins.

eahydra: > I don't recommend to create the new webhook service provider, that's a little weird and seems defined with `/apis/v1/` base for no reason.
> 
> How about add a new `HTTPHandlerMap map[string]http.Handler` in `server.go`, the handlers in which will all be registered by `server.Register`. And then you can add your custom URL path and HTTP handler in `add_quota.go` from any other webhook plugins.

If it is just to query the internal status of the ElasticQuota webhook through HTTP to help diagnose and analyze the problem, it can indeed be achieved by adding an http.Handler, which is indeed relatively simple. 

There are similar implementations in the scheduler, and there are many plugins in the scheduler that expect to provide an HTTP API to expose internal state. Therefore, a new interface `APIServiceProvider` is defined in the scheduler, each plugin implements it as needed, and the interface is automatically called to register the API when the plugin is initialized, and the framework also constrains the API path `/apis/v1/plugins`. 

But I'm not sure if webhook needs to do this now, because there is really no similar demand. Providing a webhookServiceProvider is also a good abstraction if there is one. For now, perhaps this is an over-engineering.

Issue #657, opened 2022-09-23
[question] in elasticquota plugin，is the method syncHandler()  update quota cr too frequently?


fjding:
**What happened:**
if the patch operation takes 20ms,  the apiserver need to add 50 times/s overhead？

**What you expected to happen:**

**Environment:**

- Koordinator version: - v0.6.2
- Kubernetes version (use kubectl version): v1.22.5
- docker/containerd version: containerd 1.5.0
- OS (e.g: cat /etc/os-release): Ubuntu 20.04.4 LTS
- Kernel (e.g. uname -a): Linux 5.10.112-11.al8.x86_64 #1 SMP Tue May 24 16:05:50 CST 2022 x86_64 x86_64 x86_64 GNU/Linux

**Anything else we need to know:**

xulinfei1996:
https://github.com/koordinator-sh/koordinator/pull/681/files this pr is to optimize some logics of the syncHandler, and SyncHandlerCycle can be a configuration if in need in the future.

PR #703, opened 2022-10-17
koord-scheduler: make elasticQuota controller's syncHandler interval a configuration

xulinfei1996:

 make elasticQuota controller's syncHandler interval a configuration to allow users configure the interval in need.

fixes https://github.com/koordinator-sh/koordinator/issues/657
### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -138,6 +138,9 @@ type ElasticQuotaArgs struct {
 	// RevokePodInterval is the interval to check quotaGroup's used and runtime
 	RevokePodInterval *metav1.Duration `json:"revokePodInterval,omitempty"`
 
+	// SyncHandlerInterval is the interval to check whether elasticQuota's crd need to update.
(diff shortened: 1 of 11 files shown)


jasonliu747 (commented):
/retest

eahydra:
For the #657 issue, we'll see if frequent updates to ElasticQuota are reasonable and necessary. If it is necessary, then we have to see if each update is reasonable. If the updates are very frequent, we can reduce some unnecessary updates, such as checking whether there are changes before updating, and do not update if there are no changes. And we should also add some metrics to observe the RPS updated by ElasticQuota, and judge whether it is reasonable through the data.

As an API, ElasticQuota CRD may be depended on by other components, and they may expect ElasticQuota to synchronize the latest Quota changes in real time. It may not be a good idea to set the cycle time, as setting how long is an empirical value and may not be very common in Quota scenarios.

hormes:
It seems that this parameter affects the timeliness of runtime quota calculation, and the configuration is reasonable, but the name of the parameter is easy to misunderstand.

PR #718, opened 2022-10-19
RuntimeProxy: Support hook server deployed by k8s pod

honpey:
If hookserver is deployed as k8s pod, there may exist cycled dependency: If hookserver is updated, the new hookserver pod create CRI request would come to kubelet, runtime-proxy and then hookserver, however, at this time, the previous pod may have exited, which results in calling hookserver timeout. If RuntimeHookConfig.FailurePolicy is configed as FailurePolicyType, such pod hookserver would always be rejected by RuntimeProxy to re-create.

To fix this issue, we introduce annotations to tag pod as hook server. For such tagged pods, runtime-proxy would transfer cri request to backend runtime engine transparently to solve the cycled dependency.

Signed-off-by: honpey <honpey@gmail.com>

### V. Checklist

### DIFF
--- a/pkg/runtimeproxy/resexecutor/cri/container_test.go
+++ b/pkg/runtimeproxy/resexecutor/cri/container_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/koordinator-sh/koordinator/apis/runtime/v1alpha1"
 	"github.com/koordinator-sh/koordinator/pkg/runtimeproxy/store"
+	"github.com/koordinator-sh/koordinator/pkg/runtimeproxy/utils"
 )
 
 func TestContainerResourceExecutor_UpdateRequestForCreateContainerRequest(t *testing.T) {
@@ -277,6 +278,7 @@ func TestContainerResourceExecutor_ParseRequest_CreateContainerRequest(t *testin
 		name                  string
 		args                  args
 		wantContainerExecutor store.ContainerInfo
+		expectedOperation     utils.CallHookPluginOperation
 	}{
 		{
 			name: "normal case",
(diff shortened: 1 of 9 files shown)


pkg/runtimeproxy/resexecutor/cri/container_test.go
```diff
@@ -367,12 +367,12 @@ func TestContainerResourceExecutor_ParseRequest_CreateContainerRequest(t *testin
 	for _, tt := range tests {
 		// mock pod cache
 		p := NewPodResourceExecutor()
-		_ = p.ParseRequest(tt.args.podReq)
+		_, _ = p.ParseRequest(tt.args.podReq)
```
zwzhang0107, line 370: how about add more ut?

FillZpp:
So how about expose the hook server in Koordlet instead of a new daemonset on all nodes?

hormes:
> So how about expose the hook server in Koordlet instead of a new daemonset on all nodes?

In fact, this is the case, the implementation of hooks is in koordlet, and koordlet is the process of the hooks server of support workers.

pkg/runtimeproxy/server/cri/criserver.go
```diff
@@ -94,45 +95,47 @@ func (c *RuntimeManagerCriServer) getRuntimeHookInfo(serviceType RuntimeServiceT
...
-			klog.Errorf("failed to update cri request %v", err)
+	switch callHookOperation {
+	case utils.ShouldCallHookPlugin:
+		// TODO deal with the Dispatch response
+		response, err := c.hookDispatcher.Dispatch(ctx, runtimeHookPath, config.PreHook, resourceExecutor.GenerateHookRequest())
+		if err != nil {
+			klog.Errorf("fail to call hook server %v", err)
+		} else if response == nil {
+			// when hook is not registered, the response will become nil
+			klog.Warningf("runtime hook path %s does not register any PreHooks", string(runtimeHookPath))
```
zwzhang0107, line 115: maybe level klog.V(4).Info is enough for no hook registered

honpey: > maybe level klog.V(4).Info is enough for no hook registered

this log would be removed in next patch

PR #724, opened 2022-10-20
fix: delete unused return args

lucming:
### V. Checklist


pkg/koordlet/executor/resource_update_executor.go
```diff
@@ -219,6 +219,8 @@ func (e *LeveledResourceUpdateExecutor) LeveledUpdateBatchByCache(resources [][]
 			}
 		}
 	}
+
+	updated = err == nil
```
zwzhang0107, line 223: i suggest we can remove the return value since the meaning is alreay fuzzy.

lucming: > i suggest we can remove the return value since the meaning is alreay fuzzy.

done

jasonliu747 (commented):
/lgtm
@lucming 辛苦更新一下 PR 标题

PR #728, opened 2022-10-20
fix:delete unuse logic

lucming:
containers of bepod is controlled by kubelet, so container will restart according to pod's restartPolicy when stop container,and in after logic will do evict,so delete container is unuseless in there.

### V. Checklist


zwzhang0107:
besides, BE containers are not recommended with `restartPolicy=Always`.
we must killing BE containers ASAP if interference happened, calling api-server to evict maybe not in time.
i suggest we should keep the killing container operation.

lucming:
> besides, BE containers are not recommended with `restartPolicy=Always`. we must killing BE containers ASAP if interference happened, calling api-server to evict maybe not in time. i suggest we should keep the killing container operation.

en,agree with you,thanks for review.

Issue #743, opened 2022-10-26
[BUG]failed to find kubelet pid when there is not tcp6 in node


lucming:
**What happened**:
koordlet will report error when tcp6 is not find in node,and the logs look like this:
`
I1026 16:16:23.894144  292078 shared_informer.go:270] caches populated
I1026 16:16:23.894875  292078 cpu_suppress.go:322] calculated BE suppress policy: cpuset [0 12 6 18 1 13 7 19 2 14 8]
W1026 16:16:23.894902  292078 cpu_suppress.go:502] suppressBECPU failed to apply be cpu suppress policy, err: NodeTopo is nil
W1026 16:16:23.993471  292078 metrics_query.go:77] get pod 1641ddd2-9f42-407f-a238-4bf06e3d65ad resource metric failed, error get pod 1641ddd2-9f42-407f-a238-4bf06e3d65ad resource metric not exist, query params &{last 2022-10-26 16:16:21.589697149 +0800 CST m=+3384.158647217 2022-10-26 16:16:23.589697149 +0800 CST m=+3386.158647217}
E1026 16:16:24.076557  292078 states_noderesourcetopology.go:171] Failed to GetKubeletCommandline with kubeletPort 10250, err: failed to get pid, err is open /proc/net/tcp6: no such file or directory
`

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

PR #744, opened 2022-10-26
fix:failed to start nodeTopoInformer when tcp6 is not find in node

lucming:

fix: https://github.com/koordinator-sh/koordinator/issues/743

### V. Checklist

### DIFF
--- a/pkg/util/system/common_linux.go
+++ b/pkg/util/system/common_linux.go
@@ -36,6 +36,7 @@ import (
 	"unicode"
 
 	"github.com/cakturk/go-netstat/netstat"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/util/system/common_linux.go
```diff
@@ -52,14 +52,16 @@ type portAndPid struct {
...
-	if err != nil {
-		return nil, err
+
+	socks, err = netstat.TCP6Socks(fn)
+	if err == nil && socks != nil {
+		return socks, nil
 	}
-	return append(socks, v6Socks...), nil
+
+	return nil, err
```
eahydra, line 64: Nice!  But both TCPSocks and TCP6Socks may return errors. It is recommended to return the aggregate  errors.

lucming: > Nice! But both TCPSocks and TCP6Socks may return errors. It is recommended to return the aggregate errors.

thanks for reviewing.
done.

eahydra (approved):
/lgtm

jasonliu747 (approved):
/lgtm

PR #747, opened 2022-10-26
koordlet: add pprof flag

zwzhang0107:
### V. Checklist

### DIFF
--- a/cmd/koordlet/main.go
+++ b/cmd/koordlet/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/koordlet/main.go
```diff
@@ -46,6 +46,15 @@ func main() {
 	go wait.Forever(klog.Flush, 5*time.Second)
 	defer klog.Flush()
 
+	if *options.EnablePprof {
+		go func() {
+			klog.V(4).Infof("Starting pprof on %v", *options.PprofAddr)
+			if err := http.ListenAndServe(*options.PprofAddr, nil); err != nil {
```
FillZpp, line 52: You sure this will provide pprof server without `_ "net/http/pprof"` imported?

zwzhang0107: it happens to works because other related pkg imports this

PR #750, opened 2022-10-27
fix: error to init option(SysRootDir)

lucming:
option(SysRootDir) initialization error, but this variable is also useless, should we keep it?

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #754, opened 2022-10-28
RuntimeProxy: return err when FailurePolicyType set to Fail

honpey:
### V. Checklist


pkg/runtimeproxy/dispatcher/dispatcher.go
```diff
@@ -84,8 +84,9 @@ func (rd *RuntimeHookDispatcher) Dispatch(ctx context.Context, runtimeRequestPat
 			}
 			// currently, only one hook be called during one runtime
 			// TODO: multi hook server to merge response
-			return rd.dispatchInternal(ctx, hookType, client, request)
+			rsp, err := rd.dispatchInternal(ctx, hookType, client, request)
+			return rsp, err, hookServer.FailurePolicy
```
zwzhang0107, line 88: `hookServer.FailurePolicy` is this a required field or optional(with default)

PR #761, opened 2022-10-28
koordlet: fix kubelet pid not found in tcp4 socks

saintube:

Fix when the kubelet uses tcp6 socks stats_noderesourcetopology fails to calculate node topo.

Error msg: `states_noderesourcetopology.go:302] failed to calculate node topology, err: failed to get Kubeletcommand linee with kubeletPort 10250, err: failed to getPIDd, the port is not used`

After #744, when retrieving kubelet's PID via netstat, koordlet always uses tcp4 socks instead of tcp6 socks, since even if there is no kubelet port set in tcp4, `netstat.TCP6Socks()` can return an empty result with non-error. It makes the koordlet ignores the kubelet port in tcp6 unexpectedly.

### V. Checklist

### DIFF
--- a/pkg/util/system/common_linux.go
+++ b/pkg/util/system/common_linux.go
@@ -52,14 +52,14 @@ type portAndPid struct {
 }
 
 func TCPSocks(fn netstat.AcceptFn) ([]netstat.SockTabEntry, error) {
+	v6Socks, v6Err := netstat.TCP6Socks(fn)
+	if v6Err == nil && v6Socks != nil && len(v6Socks) > 0 {
+		return v6Socks, nil
+	}
+
 	socks, err := netstat.TCPSocks(fn)
 	if err == nil && socks != nil {
 		return socks, nil
-	}
-
-	v6Socks, v6Err := netstat.TCP6Socks(fn)
-	if v6Err == nil && v6Socks != nil {
-		return v6Socks, nil
 	}
 
 	return nil, utilerrors.NewAggregate([]error{err, v6Err})


jasonliu747 (approved):
/lgtm

saintube:
@lucming Please check if this patch does not let #743 return.

lucming:
> @lucming Please check if this patch does not let #743 return.

it can return,but is it not necessary to adjust the order of finding processes? Just add judgement(len(socks)>0) will be fine,preferably for both tcp and tcp6 cases .

saintube:
> > @lucming Please check if this patch does not let #743 return.
> 
> it can return,but is it not necessary to adjust the order of finding processes? Just add judgement(len(socks)>0) will be fine,preferably for both tcp and tcp6 cases .

Do you mean this patch does not make regression of #743? We change the order of judgment just for tcp6 preference cc @zwzhang0107 .

lucming:
> > > @lucming Please check if this patch does not let #743 return.
> > 
> > 
> > it can return,but is it not necessary to adjust the order of finding processes? Just add judgement(len(socks)>0) will be fine,preferably for both tcp and tcp6 cases .
> 
> Do you mean this patch does not make regression of #743? We change the order of judgment just for tcp6 preference cc @zwzhang0107 .

I have verified that the current logic works, this pr should solve the case that if the node goes tcp6 there will also be a file /proc/net/tcp, so if we go through the tcp logic first to find the socks we will not find them but err=nil, if this is the case adding the judgement (len(socks)>0) should solve it. In fact we should ignore the priority of the two modes tcp or tcp6 and return as soon as we find socks, so just look for them one by one and return when you find them, it's easier to read this way too.

PR #763, opened 2022-10-31
koord-manager: remove the quota webhook's dependency on koord-scheduler

xulinfei1996:

Removed webhook's dependency on scheduler.

### V. Checklist

### DIFF
--- a/pkg/webhook/elasticquota/quota_info.go
+++ b/pkg/webhook/elasticquota/quota_info.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


jasonliu747 (commented):
/lgtm

pkg/webhook/elasticquota/quota_info.go
```diff
@@ -0,0 +1,101 @@
...
+package elasticquota
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/scheduler-plugins/pkg/apis/scheduling/v1alpha1"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+)
+
+type QuotaInfo struct {
```
hormes, line 26: we define a new struct, and if you want to add fields in the future, do you need to modify both places at the same time?

xulinfei1996: There are two reasons to define a new struct, first remove the dependency on scheduler; second, webhook no need to check some members of the quotaInfo, such as Used, Runtime, Request and so on. To this question, if the fields we want to add need to be checked in webhook, then we should add the fields both places.

PR #765, opened 2022-10-31
no processing for successfully exited pods

lucming:

The pod has successfully exited, and koordlet still keeps trying to modify the pod and container cgroups, which is not necessary
截屏2022-10-31 17 46 55
koordlet will keep writing the cgroup file even though the pod has exited successfully.However, as the pod exits successfully, the associated cgroup file is deleted, so the following error log is always reported.
```shell
I1031 17:42:38.810404 3754286 resource_update_executor.go:137] manager: CPUBurstExecutor, currentResource: &{0xc000ecfcc0 0 kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a/989ab9f8388c97628682d96662ab0573c319ddcb2743d7e49765ed8d6baa79ce {cpu.cfs_burst_us cpu/ true 0x3965fa0} {0 0 <nil>} 0x1cbe080 <nil> false}, preResource: <nil>, need update

I1031 17:42:38.810432 3754286 cpu_burst.go:518] update container kube-system/node-shell-44e1d62f-cc8e-468e-a966-e453e58afc02/shell cpu burst failed, dir kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a/989ab9f8388c97628682d96662ab0573c319ddcb2743d7e49765ed8d6baa79ce, updated true, error open /host-cgroup/cpu/kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a/989ab9f8388c97628682d96662ab0573c319ddcb2743d7e49765ed8d6baa79ce/cpu.cfs_burst_us: no such file or directory

I1031 17:42:38.810445 3754286 resource_update_executor.go:137] manager: CPUBurstExecutor, currentResource: &{0xc000ecfd40 0 kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a {cpu.cfs_burst_us cpu/ true 0x3965fa0} {0 0 <nil>} 0x1cbe080 <nil> false}, preResource: <nil>, need update

I1031 17:42:38.810474 3754286 cpu_burst.go:538] update pod kube-system/node-shell-44e1d62f-cc8e-468e-a966-e453e58afc02 cpu burst failed, dir kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a, updated true, error open /host-cgroup/cpu/kubepods/besteffort/podd7f19c1d-c555-476f-a237-ad4bf3154e2a/cpu.cfs_burst_us: no such file or directory
```

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/cpu_burst.go
+++ b/pkg/koordlet/resmanager/cpu_burst.go
@@ -198,6 +198,14 @@ func (b *CPUBurst) start() {
 			// ignore non-burstable pod, e.g. LSR, BE pods
 			continue
 		}
+		if podMeta.Pod.Status.Phase != corev1.PodPending && podMeta.Pod.Status.Phase != corev1.PodRunning {
+			// ignore pods that status.phase is not pending or running,
+			// because the other pods(include succeed,failed and unknown) do not have any containers running
+			// and therefore do not have a cgroup file,
+			// so there is no need to deal with it
+			continue
+		}
+
 		// merge burst config from pod and node
 		cpuBurstCfg := genPodBurstConfig(podMeta.Pod, &b.nodeCPUBurstStrategy.CPUBurstConfig)
 		if cpuBurstCfg == nil {
--- a/pkg/koordlet/resmanager/cpu_burst_test.go
+++ b/pkg/koordlet/resmanager/cpu_burst_test.go
@@ -140,6 +140,7 @@ func newTestPodWithQOS(name string, qos apiext.QoSClass, cpuMilli, memoryBytes i
 					ContainerID: genTestContainerIDByName(containerName),
 				},
 			},
+			Phase: corev1.PodRunning,
 		},
 	}
 }


pkg/koordlet/resmanager/cpu_burst.go
```diff
@@ -198,6 +198,12 @@ func (b *CPUBurst) start() {
 			// ignore non-burstable pod, e.g. LSR, BE pods
 			continue
 		}
+		if podMeta.Pod.Status.Phase == corev1.PodSucceeded {
```
saintube, line 201: How about other phases like `corev1.PodFailed`? We may check if the phase is not Running or Pending.

lucming: > How about other phases like `corev1.PodFailed`? We may check if the phase is not Running or Pending.

Because only successfully exited pods are sure never to be restarted.
kubernetes-pod-life-cycle

jasonliu747: What if it's restart policy is set to never?

saintube: 'PodFailed' means all containers have terminated, so the koordlet would fail to update any of the container-level cgroups without correct cgroup paths.

lucming: In fact, the container cgroup file is only available as long as the container exists.

saintube: The comments indicate that `Pending` pods may have some containers running.  
e.g. a pod with part of the init containers started, a pod with part of regular containers terminated.

lucming: okay, I'll change the code then.

pkg/koordlet/resmanager/cpu_burst.go
```diff
@@ -198,6 +198,14 @@ func (b *CPUBurst) start() {
 			// ignore non-burstable pod, e.g. LSR, BE pods
 			continue
 		}
+		if podMeta.Pod.Status.Phase != corev1.PodPending && podMeta.Pod.Status.Phase != corev1.PodRunning {
```
saintube, line 201: /lgtm

PR #773, opened 2022-11-01
koordlet: runtime hook support failpolicy

huiwq1990:
### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/config.go
+++ b/pkg/koordlet/runtimehooks/config.go
@@ -77,24 +77,26 @@ var (
 )
 
 type Config struct {
-	RuntimeHooksNetwork       string
-	RuntimeHooksAddr          string
-	RuntimeHooksFailurePolicy string
-	RuntimeHookConfigFilePath string
-	RuntimeHookHostEndpoint   string
-	RuntimeHookDisableStages  []string
-	FeatureGates              map[string]bool // Deprecated
+	RuntimeHooksNetwork             string
+	RuntimeHooksAddr                string
+	RuntimeHooksFailurePolicy       string
+	RuntimeHooksPluginFailurePolicy string
+	RuntimeHookConfigFilePath       string
+	RuntimeHookHostEndpoint         string
+	RuntimeHookDisableStages        []string
+	FeatureGates                    map[string]bool // Deprecated
 }
 
 func NewDefaultConfig() *Config {
 	return &Config{
-		RuntimeHooksNetwork:       "unix",
-		RuntimeHooksAddr:          "/host-var-run-koordlet/koordlet.sock",
-		RuntimeHooksFailurePolicy: "Ignore",
-		RuntimeHookConfigFilePath: system.Conf.RuntimeHooksConfigDir,
-		RuntimeHookHostEndpoint:   "/var/run/koordlet/koordlet.sock",
-		RuntimeHookDisableStages:  []string{},
-		FeatureGates:              map[string]bool{},
+		RuntimeHooksNetwork:             "unix",
+		RuntimeHooksAddr:                "/host-var-run-koordlet/koordlet.sock",
+		RuntimeHooksFailurePolicy:       "Ignore",
+		RuntimeHooksPluginFailurePolicy: "Ignore",
+		RuntimeHookConfigFilePath:       system.Conf.RuntimeHooksConfigDir,
+		RuntimeHookHostEndpoint:         "/var/run/koordlet/koordlet.sock",
+		RuntimeHookDisableStages:        []string{},
+		FeatureGates:                    map[string]bool{},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/koordlet/runtimehooks/hooks/hooks.go
```diff
@@ -72,15 +72,21 @@ func getHooksByStage(stage rmconfig.RuntimeHookType) []*Hook {
...
-func RunHooks(stage rmconfig.RuntimeHookType, protocol protocol.HooksProtocol) {
+func RunHooks(failPolicy rmconfig.FailurePolicyType, stage rmconfig.RuntimeHookType, protocol protocol.HooksProtocol) error {
 	hooks := getHooksByStage(stage)
 	klog.V(5).Infof("start run %v hooks at %s", len(hooks), stage)
 	for _, hook := range hooks {
 		klog.V(5).Infof("call hook %v", hook.name)
 		if err := hook.fn(protocol); err != nil {
-			klog.Warningf("failed to run hook %s in stage %s, reason: %v", hook.name, stage, err)
+			if failPolicy == rmconfig.PolicyFail {
+				klog.Errorf("failed to run hook %s in stage %s, reason: %v", hook.name, stage, err)
```
honpey, line 82: rmconfig has been registered to RuntimeProxy,  so there is no need to filter errors basing on  rmconfig.PolicyFail duplicately. 
'return err' is enough.

honpey: emm.. , for koordlet, under PolicyFail scenario, return error immediately when error happens, while PolicyIgnore, all hooks should be called even if errors happen for one or more hook?
if it is, it makes sense to check rmconfig.PolicyFail here.
but there had better be more comments here to avoid confusing.

zwzhang0107: +1, if we really need to support `stop running other hooks once someone failed`, it is better to add another option choice to runtime hook in koordlet, since the `rmconfig.Policy` is only used for `runtime-proxy`.

hormes:
@huiwq1990 What do you think about the above suggestion?

huiwq1990:
@hormes I just fix them.

pkg/koordlet/runtimehooks/hooks/hooks.go
zwzhang0107: hi @huiwq1990, do you have any other ideas about these comments?

huiwq1990: @zwzhang0107 I agress the comments, and define a new options `RuntimeHooksCallFailurePolicy`. And just reuse the  `runtimeproxy.FailurePolicyType`.

pkg/koordlet/runtimehooks/config.go
```diff
@@ -77,31 +77,34 @@ var (
...
+		RuntimeHookDisableStages:      []string{},
+		FeatureGates:                  map[string]bool{},
 	}
 }
 
 func (c *Config) InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.RuntimeHooksNetwork, "runtime-hooks-network", c.RuntimeHooksNetwork, "rpc server network type for runtime hooks")
 	fs.StringVar(&c.RuntimeHooksAddr, "runtime-hooks-addr", c.RuntimeHooksAddr, "rpc server address for runtime hooks")
 	fs.StringVar(&c.RuntimeHooksFailurePolicy, "runtime-hooks-failure-policy", c.RuntimeHooksFailurePolicy, "failure policy for runtime hooks")
+	fs.StringVar(&c.RuntimeHooksCallFailurePolicy, "runtime-hooks-call-failure-policy", c.RuntimeHooksCallFailurePolicy, "stop running other hooks once someone failed")
```
zwzhang0107, line 107: how about naming as `RuntimeHooksPluginFailurePolicy`, since it represents "stop running other plugins once someone failed"

pkg/koordlet/runtimehooks/hooks/hooks.go
```diff
@@ -72,15 +72,21 @@ func getHooksByStage(stage rmconfig.RuntimeHookType) []*Hook {
...
 
-func RunHooks(stage rmconfig.RuntimeHookType, protocol protocol.HooksProtocol) {
+func RunHooks(failPolicy rmconfig.FailurePolicyType, stage rmconfig.RuntimeHookType, protocol protocol.HooksProtocol) error {
 	hooks := getHooksByStage(stage)
 	klog.V(5).Infof("start run %v hooks at %s", len(hooks), stage)
 	for _, hook := range hooks {
 		klog.V(5).Infof("call hook %v", hook.name)
 		if err := hook.fn(protocol); err != nil {
-			klog.Warningf("failed to run hook %s in stage %s, reason: %v", hook.name, stage, err)
+			if failPolicy == rmconfig.PolicyFail {
```
zwzhang0107, line 81: if err := hook.fn(protocol); err != nil && failPolicy == rmconfig.PolicyFail

huiwq1990: err != nil && failPolicy != rmconfig.PolicyFail, log warning message

pkg/koordlet/runtimehooks/config.go
huiwq1990: fixed

huiwq1990:
@zwzhang0107 Why upload coverage to codecov failed?

zwzhang0107:
seems the network problem, can be fixed by retry

Issue #717, opened 2022-10-19
[proposal] Support kata container to use koordinator resources


czhenly:
**What is your proposal**:
  Support kata container to use koordinator resources

**Why is this needed**:
For isolation requirements, I want to use kata container to run offline tasks, which is co-located with online tasks in a cluster.

**Current status**:
Pod yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: test-kata
spec:
  runtimeClassName: kata
  containers:
  - name: anolisos
    image: openanolis/anolisos:8.6-x86_64
    command: ["sleep", "1d"]
    resources:
      limits:
        koordinator.sh/batch-cpu: "4000"
        koordinator.sh/batch-memory: 4Gi
      requests:
        koordinator.sh/batch-cpu: "4000"
        koordinator.sh/batch-memory: 4Gi
    securityContext:
      privileged: true
  nodeName: host1
```

resource info in container:
```
# lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              77
On-line CPU(s) list: 0-76
Thread(s) per core:  1
Core(s) per socket:  77
Socket(s):           1
....

# free -h
              total        used        free      shared  buff/cache   available
Mem:          1.9Gi       144Mi       1.8Gi       4.0Mi        22Mi       1.8Gi
Swap:            0B          0B          0B
```

kata config: configuration.toml
```
default_vcpus = 1
default_memory = 2048
```
expect result: cpu: 5core `(4c+1c)`，memory：6g `(4g+2g)`

zwzhang0107:
consider mv be cgroup reconcile into runtime-hook-plugins

PR #778, opened 2022-11-02
api: pass extended resource spec via annotation

saintube:

Add an annotation protocol for passing extended resource spec throughout CRI.

Currently, the CRI server cannot get the original pod spec but a calculated linux resource configuration derived from native resources (i.e. `CPU` and `Memory`). So the koordlet runtime hooks have no way to calculate Batch pods' `cpu.cfs_quota_us`, `cpu.shares` `memory.limit_in_bytes`, since Batch pods only specify extended resources.

With this patch, the extended cpu and memory resource requirements of Batch pods will be passed via the pod annotation, allowing the runtime hooks and other components to calculate correct `cpu.cfs_quota_us`, `cpu.shares` `memory.limit_in_bytes` for Batch pods.

Partially for #717.

### V. Checklist

### DIFF
--- a/apis/extension/resource.go
+++ b/apis/extension/resource.go
@@ -52,6 +52,11 @@ const (
 	// AnnotationResourceStatus represents resource allocation result.
 	// koord-scheduler patch Pod with the annotation before binding to node.
 	AnnotationResourceStatus = SchedulingDomainPrefix + "/resource-status"
+
+	// AnnotationExtendedResourceSpec specifies the resource requirements of extended resources for internal usage.
+	// It annotates the requests/limits of extended resources and can be used by runtime proxy and koordlet that
+	// cannot get the original pod spec in CRI requests.
+	AnnotationExtendedResourceSpec = NodeDomainPrefix + "/extended-resource-spec"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


jasonliu747 (approved):
/lgtm

apis/extension/resource.go
```diff
@@ -177,3 +182,41 @@ func TranslateResourceNameByPriorityClass(priorityClass PriorityClass, defaultRe
...
+		return spec, nil
+	}
+	err := json.Unmarshal([]byte(data), spec)
+	if err != nil {
+		return nil, err
+	}
+	return spec, nil
+}
+
+func SetExtendedResourceSpec(pod *corev1.Pod, status *ExtendedResourceSpec) error {
```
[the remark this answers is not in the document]
saintube: typo fixed

apis/extension/resource.go
```diff
@@ -177,3 +182,41 @@ func TranslateResourceNameByPriorityClass(priorityClass PriorityClass, defaultRe
...
+
+type ExtendedResourceContainerSpec struct {
+	Limits   corev1.ResourceList `json:"limits,omitempty"`
+	Requests corev1.ResourceList `json:"requests,omitempty"`
+}
+
+// GetExtendedResourceSpec parses ExtendedResourceSpec from annotations
+func GetExtendedResourceSpec(annotations map[string]string) (*ExtendedResourceSpec, error) {
+	spec := &ExtendedResourceSpec{}
+	data, ok := annotations[AnnotationResourceSpec]
```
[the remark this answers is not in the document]
saintube: done

apis/extension/resource.go
```diff
@@ -177,3 +182,44 @@ func TranslateResourceNameByPriorityClass(priorityClass PriorityClass, defaultRe
...
+	Containers map[string]ExtendedResourceContainerSpec `json:"containers,omitempty"`
+}
+
+type ExtendedResourceContainerSpec struct {
+	Limits   corev1.ResourceList `json:"limits,omitempty"`
+	Requests corev1.ResourceList `json:"requests,omitempty"`
+}
+
+// GetExtendedResourceSpec parses ExtendedResourceSpec from annotations
+func GetExtendedResourceSpec(annotations map[string]string) (*ExtendedResourceSpec, error) {
```
jasonliu747, line 196: Could you also add a UT for this?

saintube: fixed

jasonliu747 (approved):
/lgtm

apis/extension/resource.go
```diff
@@ -52,6 +52,11 @@ const (
 	// AnnotationResourceStatus represents resource allocation result.
 	// koord-scheduler patch Pod with the annotation before binding to node.
 	AnnotationResourceStatus = SchedulingDomainPrefix + "/resource-status"
+
+	// AnnotationExtendedResourceSpec specifies the resource requirements of extended resources.
+	// It annotates the requests/limits of extended resources and can be used for CRI hooks which cannot get the
+	// original pod spec.
+	AnnotationExtendedResourceSpec = DomainPrefix + "/extended-resource-spec"
```
hormes, line 59: I'm wondering which domain is more appropriate, because we need to consider the scope of this protocol, for example, should the scheduler support this extended protocol? From what I have seen, it may be more appropriate to put it on node domain, only as an internal API.

saintube: fixed

FillZpp (approved):

PR #782, opened 2022-11-03
koordlet: fix cpi collector bad fd err

songtao98:
Signed-off-by: songtao98 <songtao2603060@gmail.com>

This PR fixes the error occurs at koordlet CPI collector when closing container cgroup files after profiling container CPI. Related error information is shown as below:
> E1102 21:10:28.757191  435643 performance_collector_linux.go:99] close CgroupFd 36, err : bad file descriptor

Previously we need to get one container's cgroup directory file descriptor to call `perf_event_open()` as param `pid`, thus we use struct `PerfCollector` to save that fd after open the cgroup path instead of saving a reference to the `*os.File` it self. It turns out that with golang gc mechanism the `*os.File` gets garbage collected, then the fd gets closed. But at `performance_collector_linux.go` we close that file explicitly again and that leads to the above error. 

A solution to the problem is to change `PerfCollector` to record the `*os.File` instead of Fd int and call `File.Fd()` when you need it. For the purpose of minimum change, this PR store both of them. 

### V. Checklist

### DIFF
--- a/pkg/koordlet/metricsadvisor/performance_collector_linux.go
+++ b/pkg/koordlet/metricsadvisor/performance_collector_linux.go
@@ -23,7 +23,6 @@ import (
 	"sync"
 	"time"
 
-	"golang.org/x/sys/unix"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
@@ -92,11 +91,16 @@ func (c *performanceCollector) collectContainerCPI() {
 			if !ok {
 				return
 			}
-			c.profilePerfOnSingleContainer(status, oneCollector.(*perf.PerfCollector), podUid)
+			perfCollector, ok := oneCollector.(*perf.PerfCollector)
+			if !ok {
+				klog.Errorf("PerfCollector type convert failed")
+				return
+			}
+			c.profilePerfOnSingleContainer(status, perfCollector, podUid)
 
-			err1 := unix.Close(oneCollector.(*perf.PerfCollector).CgroupFd)
-			if err1 != nil {
-				klog.Errorf("close CgroupFd %v, err : %v", oneCollector.(*perf.PerfCollector).CgroupFd, err1)
+			cleanErr := perfCollector.CleanUp()
+			if cleanErr != nil {
+				klog.Errorf("PerfCollector cleanup err : %v", cleanErr)
 			}
 		}(containerStatus, string(podUid))
 	}
--- a/pkg/util/perf/perf_linux.go
+++ b/pkg/util/perf/perf_linux.go
@@ -22,6 +22,7 @@ package perf
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/hodgesds/perf-utils"
 	"go.uber.org/multierr"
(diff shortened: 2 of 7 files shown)


pkg/util/perf/perf_linux.go
```diff
@@ -30,14 +31,17 @@ import (
 )
 
 type PerfCollector struct {
+	CgroupFile        *os.File
```
saintube, line 34: How about encapsulating a method like `(*PerfCollector).Cleanup()`, not exposing the detailed behavior like closing a cgroup fd.

songtao98: Good idea! Already modified.

pkg/util/perf/perf_linux.go
```diff
@@ -30,15 +31,18 @@ import (
 )
 
 type PerfCollector struct {
-	CgroupFd          int
+	cgroupFile        *os.File
```
saintube, line 34: do we need to store both the fd and the file pointer

songtao98: At first I store both for minimum code modification, but I think it's okay to abort usage of cgroupfd. Will remove it.

saintube:
@songtao98 UT pipeline failed to upload to codecov, please retry.
> You can use `git --amend` to trigger a commit without any code modification.

pkg/koordlet/metricsadvisor/performance_collector_linux.go
```diff
@@ -94,9 +93,9 @@ func (c *performanceCollector) collectContainerCPI() {
 			}
 			c.profilePerfOnSingleContainer(status, oneCollector.(*perf.PerfCollector), podUid)
 
-			err1 := unix.Close(oneCollector.(*perf.PerfCollector).CgroupFd)
+			err1 := oneCollector.(*perf.PerfCollector).CleanUp()
```
zwzhang0107, line 96: check type convert

pkg/util/perf/perf_linux.go
```diff
@@ -138,14 +139,22 @@ func (c *PerfCollector) stopAndClose() (err error) {
...
 func (c *PerfCollector) closeOnSingleCPU(cpu int) error {
 	if err := (*c.cpuHwProfilersMap[cpu]).Close(); err != nil {
-		return fmt.Errorf("close container %v, cpu: %v fd err : %v", c.CgroupFd, cpu, err)
+		return fmt.Errorf("close container %v, cpu: %v fd err : %v", int(c.cgroupFile.Fd()), cpu, err)
+	}
+	return nil
+}
+
+func (c *PerfCollector) CleanUp() error {
+	err := c.cgroupFile.Close()
```
zwzhang0107, line 155: what if the container already finished and the file is deleted

pkg/koordlet/metricsadvisor/performance_collector_linux.go
```diff
@@ -92,11 +91,16 @@ func (c *performanceCollector) collectContainerCPI() {
...
-			c.profilePerfOnSingleContainer(status, oneCollector.(*perf.PerfCollector), podUid)
+			perfCollector, ok := oneCollector.(*perf.PerfCollector)
+			if !ok {
+				klog.Errorf("PerfCollector type convert failed")
+				return
+			}
+			c.profilePerfOnSingleContainer(status, perfCollector, podUid)
 
-			err1 := unix.Close(oneCollector.(*perf.PerfCollector).CgroupFd)
+			err1 := perfCollector.CleanUp()
```
zwzhang0107, line 101: err1 -> cleanErr

PR #783, opened 2022-11-03
apis: remove dependency on controller-runtime

eahydra:

When publishing Koordinator APIs(https://github.com/koordinator-sh/apis) separately, it is necessary to remove unnecessary dependencies to reduce the impact on those who rely on Koordinator APIs. Therefore, it is necessary to remove the controller-runtime that the APIs package depends on.

### V. Checklist

### DIFF
--- a/apis/config/v1alpha1/groupversion_info.go
+++ b/apis/config/v1alpha1/groupversion_info.go
@@ -21,7 +21,8 @@ package v1alpha1
 
 import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
+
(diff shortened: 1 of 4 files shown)


FillZpp (commented):
It's ok to do removal like this. But since we have changed the code generated by kubebuilder, anyone that may add new CRD types in future should keep in mind that he/she has to remove the `init()` and add the types into scheme manually.

hormes:
> It's ok to do removal like this. But since we have changed the code generated by kubebuilder, anyone that may add new CRD types in future should keep in mind that he/she has to remove the `init()` and add the types into scheme manually.

Take a look at where to add this prompt, otherwise developers who add CRDs will easily step on this problem.

PR #786, opened 2022-11-05
koord-scheduler: fix runtime quota calculation error if the total amount of resources is insufficient

fjding:
Signed-off-by: fjding <dingfangjie@bytedance.com>

fixed two problems:
1、when the total resource is not enough, the runtime is error
2、quota  runtime random calculate error，because of traverse the key of the go map is random。
case：
total resource 100
quotaA request :60  quotaB request:70
but after syncHandler function run
the quotaA runtime= 60, quotaB runtime=70.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
+++ b/pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
@@ -17,6 +17,8 @@ limitations under the License.
 package core
 
 import (
+	"math"
+	"sort"
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
@@ -48,17 +50,21 @@ func NewQuotaNode(quotaName string, sharedWeight, request, min int64, allowLentR
 // quotaTree abstract the struct to calculate each resource dimension's runtime Quota independently
 type quotaTree struct {
 	quotaNodes map[string]*quotaNode
+	sortQuota  *SortQ
 }
 
 func NewQuotaTree() *quotaTree {
 	return &quotaTree{
 		quotaNodes: make(map[string]*quotaNode),
+		sortQuota:  &SortQ{},
 	}
 }
 
 func (qt *quotaTree) insert(groupName string, sharedWeight, request, min int64, allowLentResource bool) {
 	if _, exist := qt.quotaNodes[groupName]; !exist {
-		qt.quotaNodes[groupName] = NewQuotaNode(groupName, sharedWeight, request, min, allowLentResource)
+		node := NewQuotaNode(groupName, sharedWeight, request, min, allowLentResource)
+		qt.quotaNodes[groupName] = node
+		qt.sortQuota.insert(node)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
```diff
@@ -150,6 +164,20 @@ func (qt *quotaTree) iterationForRedistribution(totalRes, totalSharedWeight int6
 	}
 }
 
+type SortQ struct {
```
eahydra, line 167: Maybe we need to add some UT to test this function.

eahydra:
Hi @fjding, the question you mentioned "quota runtime random calculate error，because of traverse the key of the go map is random", can I understand that every time the runtime value is calculated, the runtime value of some ElasticQuota objects changes unexpectedly? For example, the runtime originally allocated to ElasticQuota A is 123, and the runtime allocated to ElasticQuota B is 456, but if other parameters remain unchanged, if the runtime is recalculated, it is possible that ElasticQuota A has become 456, and ElasticQuota B has become 123?

If this is what I understand, this really needs to be improved.

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
fjding: ok

eahydra:
Got It, nice work ! @fjding

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
```diff
@@ -48,17 +50,21 @@ func NewQuotaNode(quotaName string, sharedWeight, request, min int64, allowLentR
 // quotaTree abstract the struct to calculate each resource dimension's runtime Quota independently
 type quotaTree struct {
 	quotaNodes map[string]*quotaNode
```
hormes, line 52: Can treemap help in this scenario?

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
fjding: done

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
fjding: there need a simple sort struct  is ok,  this only have insert method

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
```diff
@@ -100,20 +106,28 @@ func (qt *quotaTree) redistribution(totalResource int64) {
 	toPartitionResource := totalResource
 	totalSharedWeight := int64(0)
 	needAdjustQuotaNodes := make([]*quotaNode, 0)
-	for _, node := range qt.quotaNodes {
+	for _, node := range qt.sortQuota.items {
 		if node.request > node.min {
 			// if a node's request > autoScaleMin, the node needs adjustQuota
```
buptcozy, line 111: 这个pr是不需要的；从算法设计角度，我们没有对order有任何的依赖； 这个pr的背景是当时在没有开scaleMin的时候，在这里额外实现了类似的scaleMin，所以印出来了random的问题； 经过讨论，如果需要部分quota组支持scaleMin的功能，在quota组级别增加scaleMin的开关即可，代码是支持部分quota组打开scaleMin，部分关闭的； 

换另外一个角度来说，如果没有修改类似scaleMin的逻辑(已经revert)，无论是否打开scaleMin，都不会出现随机的问题，所以这个pr的方向应该调整为per quota组支持scaleMin

fjding: Get了，那先关闭掉吧

PR #787, opened 2022-11-07
koord-manager: make webhookConfigurationName configurable

ZiMengSheng:

this PR make webhookConfigurationName configurable with env

NONE

### V. Checklist

### DIFF
--- a/pkg/webhook/util/configuration/configuration.go
+++ b/pkg/webhook/util/configuration/configuration.go
@@ -32,9 +32,9 @@ import (
 	webhookutil "github.com/koordinator-sh/koordinator/pkg/webhook/util"
 )
 
-const (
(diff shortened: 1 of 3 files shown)


pkg/webhook/util/configuration/configuration.go
```diff
@@ -32,19 +32,14 @@ import (
 	webhookutil "github.com/koordinator-sh/koordinator/pkg/webhook/util"
 )
 
-const (
```
eahydra, line 35: This approach works, but it is recommended to consider another modification, eg: make these const values into variable values, and use the `init` function to initialize the values from environment variable values.

eahydra (commented):
/lgtm

eahydra (approved):

FillZpp (approved):

PR #797, opened 2022-11-10
webhook: pod mutating handler add extended resource spec

saintube:

Update pod mutating webhook to add a pod annotation of extended resource spec.

Partially for #717.

### V. Checklist

### DIFF
--- a/pkg/webhook/pod/mutating/extended_resource_spec.go
+++ b/pkg/webhook/pod/mutating/extended_resource_spec.go
@@ -1,0 +1,108 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mutating
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+)
+
+func (h *PodMutatingHandler) extendedResourceSpecMutatingPod(ctx context.Context, req admission.Request, pod *corev1.Pod) error {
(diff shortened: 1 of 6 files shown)


pkg/webhook/pod/mutating/extended_resource_spec.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (h *PodMutatingHandler) mutateByExtendedResources(pod *corev1.Pod) error {
+	// dump batch-resource of pod.spec.containers[*].resources.requests/limits into ExtendedResourceSpec{}
+	extendedResourceSpec := &extension.ExtendedResourceSpec{
+		Containers: map[string]extension.ExtendedResourceContainerSpec{},
+	}
+
+	// TODO: count pod overhead
+	for i := range pod.Spec.InitContainers {
+		container := pod.Spec.InitContainers[i]
```
eahydra, line 48: here we can get container pointer.

pkg/webhook/pod/mutating/extended_resource_spec.go
```diff
@@ -0,0 +1,124 @@
...
+			extension.BatchCPU,
+			extension.BatchMemory,
+		})
+		if r == nil {
+			continue
+		}
+		extendedResourceSpec.Containers[container.Name] = *r
+	}
+	for i := range pod.Spec.Containers {
+		container := pod.Spec.Containers[i]
```
eahydra, line 59: and here

pkg/webhook/pod/mutating/extended_resource_spec.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil
+	}
+
+	// compare annotation values
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+
+	spec, err := extension.GetExtendedResourceSpec(pod.Annotations)
+	if err != nil {
```
eahydra, line 81: the function `GetExtendedResourceSpec`  will return nil, nil even if there is no extendResourceSpec annotation. Maybe we should return an empty extendResourceSpec object.

pkg/webhook/pod/mutating/extended_resource_spec.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// no requirement of specified extended resources
+	if len(extendedResourceSpec.Containers) <= 0 {
+		return nil
+	}
+
+	// compare annotation values
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
```
eahydra, line 77: we can move the code  to line 90

pkg/webhook/pod/mutating/extended_resource_spec.go
saintube: fixed

pkg/webhook/pod/mutating/extended_resource_spec.go
saintube: done

eahydra (approved):
/lgtm

PR #810, opened 2022-11-15
koord-scheduler: fix errors caused by incorrect lock in ElasticQuota

xulinfei1996:

When deleting QuotaGroup, the pod in the QuotaGroup will be moved to defaultQuota, while the background loop will move the defaultQuota to another quotaGroup. The two processes only have readLock constraints, it will result in some errors. So this PR will change the readLock to writeLock to fix the problem. 
As the migrate operation is low-frequency, the writeLock will not influence the performance.

### V. Checklist


jasonliu747:
[11_Run Go test.txt](https://github.com/koordinator-sh/koordinator/files/10008682/11_Run.Go.test.txt)

eahydra (commented):
/lgtm

jasonliu747 (commented):
/lgtm

PR #830, opened 2022-11-22
koord-scheduler: support force sync data from informer

eahydra:

Currently, the existing scheduling plugins of koord-scheduler will construct the internal state data of the plugin based on the data in the Informer, such as which Pods are maintained in NodeNUMAResource and which CPUs are allocated, which Pods are maintained in DeviceShare and which GPU instances are allocated, and so on. These data must be correct, otherwise it may cause problems with the resources allocated during Pod scheduling, or the orchestration results may not meet expectations, which will seriously affect the runtime problems or availability of workloads.

In the current existing code, the EventHandler is registered with the Informer, and the internal state can be constructed after WaitForCacheSync returns, but WaitForCacheSync only means that the data queried from the APIServer is synchronized to the Store inside the Informer, and does not perceive the Event in the EventHandler whether processing is complete. This also means that when WaitForCacheSync returns, EventHandler may not have finished processing the existing Add Event, and this problem will be more obvious in a larger cluster.

Therefore, we need a mechanism that can help the plugin to correctly convert the data in the Informer to the internal state of the plugin, and ensure that subsequent Add/Update/Delete Events can be processed correctly. Hence the introduction of `ForceSyncedEventHandler`.

When the plugin is initialized, you can register the EventHandler through `frameworkexthelp.ForceSyncFromInformer`. This function will automatically start the Informer, and wait for the data to be synchronized successfully before pouring the current data into the EventHandler.

### DIFF
--- a/pkg/scheduler/frameworkext/helper/synced_eventhandler.go
+++ b/pkg/scheduler/frameworkext/helper/synced_eventhandler.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helper
+
+import (
+	"reflect"
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachinerytypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+type forceSyncEventHandler struct {
+	handler cache.ResourceEventHandler
+	syncCh  chan struct{}
+	objects map[apimachinerytypes.UID]int64
+}
+
+func newForceSyncEventHandler(handler cache.ResourceEventHandler) *forceSyncEventHandler {
+	return &forceSyncEventHandler{
+		handler: handler,
+		syncCh:  make(chan struct{}, 1),
+		objects: map[apimachinerytypes.UID]int64{},
+	}
+}
+
+func (h *forceSyncEventHandler) syncDone() {
+	close(h.syncCh)
+}
+
+func (h *forceSyncEventHandler) waitForSyncDone() {
+	<-h.syncCh
+}
+
+func (h *forceSyncEventHandler) OnAdd(obj interface{}) {
+	h.waitForSyncDone()
+	if metaAccessor, ok := obj.(metav1.ObjectMetaAccessor); ok {
+		objectMeta := metaAccessor.GetObjectMeta()
+		objectUID := objectMeta.GetUID()
+		if oldResourceVersion, ok := h.objects[objectUID]; ok && oldResourceVersion != 0 {
+			resourceVersion, err := strconv.ParseInt(objectMeta.GetResourceVersion(), 10, 64)
+			if err == nil && resourceVersion <= oldResourceVersion {
+				return
+			}
+			delete(h.objects, objectUID)
+		}
+	}
+	h.handler.OnAdd(obj)
+}
+
+func (h *forceSyncEventHandler) OnUpdate(oldObj, newObj interface{}) {
+	h.waitForSyncDone()
+	if h.objects != nil {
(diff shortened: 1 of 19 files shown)


pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -21,6 +21,7 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jasonliu747, line 24: dupicate import

eahydra: removed

eahydra:
Discussion from kubernetes https://github.com/kubernetes/kubernetes/issues/113763

pkg/scheduler/frameworkext/helper/synced_eventhandler.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+type CacheSyncer interface {
+	Start(stopCh <-chan struct{})
+	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
+}
+
+// ForceSyncFromInformer ensures that the EventHandler will synchronize data immediately after registration,
+// helping those plugins that need to build memory status through EventHandler to correctly synchronize data
+func ForceSyncFromInformer(stopCh <-chan struct{}, cacheSyncer CacheSyncer, informer cache.SharedInformer, handler cache.ResourceEventHandler) {
```
ZiMengSheng, line 71: we may not need cacheSyncer arguments, we can use cache.WaitForCacheSync(stopCh, informer.HasSynced)

ZiMengSheng: if this function is waitForEventHandlerSync, we man can use it as follows? this may be better?

syncEventHandler := newForceSyncEventHandler(handler)
syncEventHandler.AddToInformer(informer)
syncEventHandler.waitForEventHandlerSync

pkg/scheduler/frameworkext/helper/synced_eventhandler.go
```diff
@@ -0,0 +1,82 @@
...
+func (h *forceSyncEventHandler) syncDone() {
+	close(h.syncCh)
+}
+
+func (h *forceSyncEventHandler) waitForSyncDone() {
+	<-h.syncCh
+}
+
+func (h *forceSyncEventHandler) OnAdd(obj interface{}) {
+	h.waitForSyncDone()
```
ZiMengSheng, line 46: OnAdd may be called repeatedly? this require eventHandler idempotent. We need add comment and check every eventhandler or find alternative approach.

pkg/scheduler/frameworkext/helper/synced_eventhandler.go
eahydra: We need cacheSyncer  because we should use cacheSyncer start infromer to keep the syncer internal state. e.g.  informers.SharedInformerFactory.Start will keep the started state of informer internal.

pkg/scheduler/frameworkext/helper/synced_eventhandler.go
eahydra: Yes, OnAdd will be called repeatedly with same object.  There is currently no particularly good way to avoid onAdd being called repeatedly. A feasible method is to record the resourceVersion of each object. Whenever onAdd comes in, check whether the version is greater than the recorded resourceVersion. If it is greater, the processing is allowed to continue.

pkg/scheduler/plugins/nodenumaresource/pod_eventhandler.go
```diff
@@ -30,9 +33,11 @@ type podEventHandler struct {
 }
 
 func registerPodEventHandler(handle framework.Handle, cpuManager CPUManager) {
-	handle.SharedInformerFactory().Core().V1().Pods().Informer().AddEventHandler(&podEventHandler{
+	podInformer := handle.SharedInformerFactory().Core().V1().Pods().Informer()
+	eventHandler := &podEventHandler{
 		cpuManager: cpuManager,
-	})
+	}
+	frameworkexthelper.ForceSyncFromInformer(context.TODO().Done(), handle.SharedInformerFactory(), podInformer, eventHandler)
```
ZiMengSheng, line 40: While the function is being called, there may be other pod Update and Delete events，how to guarantee their order?

eahydra: Event Order will not change. The reason is that forceSyncEventHandler.waitForSyncDone blocks the entire EventHandler. When waitForSyncDone returns, the original accumulated Events will be consumed.

buptcozy (approved):
/ltgm

buptcozy (approved):
/approve

PR #833, opened 2022-11-23
koord-manager: make initialization scalable

eahydra:

Simplify the initialization logic of the koord-manager main function to make it more scalable.

### DIFF
--- a/cmd/koord-manager/main.go
+++ b/cmd/koord-manager/main.go
@@ -27,7 +27,6 @@ import (
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/clock"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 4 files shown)


cmd/koord-manager/main.go
```diff
@@ -224,6 +207,20 @@ func main() {
 	}
 }
 
+func setupControllersWithManager(m manager.Manager) error {
+	for controllerName, addFn := range controllerAddFuncs {
+		if err := addFn(m); err != nil {
+			if kindMatchErr, ok := err.(*meta.NoKindMatchError); ok {
+				klog.Infof("CRD %v is not installed, its controller %s will perform noops!", kindMatchErr.GroupKind, controllerName)
```
saintube, line 214: what is the case where we can ignore to start a controller that some CRD not installed

eahydra: Some controllers expect to decide whether to enable or disable it based on whether there is a CRD in the cluster. But it can be removed first.

ZiMengSheng (approved):

cmd/koord-manager/main.go
saintube: OK

PR #720, opened 2022-10-19
docs: proposal koordlet support cgroups v2

saintube:
Signed-off-by: saintube <saintube@foxmail.com>

### Ⅰ. Describe what this PR does

Add a proposal that clarifies how to support cgroups v2 in the koordlet.

### V. Checklist

- [ ] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [ ] All checks passed in `make test`

PR #834, opened 2022-11-24
koordlet: refactor cgroup resource

saintube:

Refactor the CgroupResource module. Implement `system.Resource` interface (proposed in #720 ) to support file operations on cgroups-v2.

part of #407 

### V. Checklist

### DIFF
--- a/pkg/util/system/cgroup.go
+++ b/pkg/util/system/cgroup.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"math"
 	"os"
-	"path"
 	"strconv"
 	"strings"
 
@@ -46,26 +45,25 @@ type CPUStatRaw struct {
 	ThrottledNanoSeconds int64
 }
 
-func CgroupFileWriteIfDifferent(cgroupTaskDir string, file CgroupFile, value string) error {
+func CgroupFileWriteIfDifferent(cgroupTaskDir string, file Resource, value string) error {
 	currentValue, currentErr := CgroupFileRead(cgroupTaskDir, file)
 	if currentErr != nil {
 		return currentErr
 	}
 	if value == currentValue || value == CgroupMaxValueStr && currentValue == CgroupMaxSymbolStr {
 		// compatible with cgroup valued "max"
-		klog.V(6).Infof("read before write %s %s and got str value, considered as MaxInt64", cgroupTaskDir,
-			file.ResourceFileName)
+		klog.V(6).Infof("read before write %s and got str value, considered as MaxInt64", file.Path(cgroupTaskDir))
 		return nil
 	}
 	return CgroupFileWrite(cgroupTaskDir, file, value)
 }
 
-func CgroupFileReadInt(cgroupTaskDir string, file CgroupFile) (*int64, error) {
-	if file.IsAnolisOS && !HostSystemInfo.IsAnolisOS {
-		return nil, fmt.Errorf("read cgroup config : %s fail, need anolis kernel", file.ResourceFileName)
+func CgroupFileReadInt(cgroupTaskDir string, r Resource) (*int64, error) {
+	if supported, msg := r.IsSupported(cgroupTaskDir); !supported {
+		return nil, fmt.Errorf("read cgroup %s failed, msg: %s", r.ResourceType(), msg)
 	}
 
-	dataStr, err := CgroupFileRead(cgroupTaskDir, file)
+	dataStr, err := CgroupFileRead(cgroupTaskDir, r)
 	if err != nil {
 		return nil, err
 	}
@@ -75,8 +73,7 @@ func CgroupFileReadInt(cgroupTaskDir string, file CgroupFile) (*int64, error) {
 	if dataStr == CgroupMaxSymbolStr {
 		// compatible with cgroup valued "max"
 		data := int64(math.MaxInt64)
-		klog.V(6).Infof("read %s %s and got str value, considered as MaxInt64", cgroupTaskDir,
-			file.ResourceFileName)
+		klog.V(6).Infof("read %s and got str value, considered as MaxInt64", r.Path(cgroupTaskDir))
 		return &data, nil
 	}
 	data, err := strconv.ParseInt(strings.TrimSpace(dataStr), 10, 64)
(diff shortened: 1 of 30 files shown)


pkg/util/system/cgroup.go
```diff
@@ -46,26 +45,25 @@ type CPUStatRaw struct {
...
-	if file.IsAnolisOS && !HostSystemInfo.IsAnolisOS {
-		return nil, fmt.Errorf("read cgroup config : %s fail, need anolis kernel", file.ResourceFileName)
+func CgroupFileReadInt(cgroupTaskDir string, r Resource) (*int64, error) {
+	if supported, msg := r.IsSupported(); !supported {
+		return nil, fmt.Errorf("read cgroup %s failed, msg: %s", r.ResourceType(), msg)
 	}
 
-	dataStr, err := CgroupFileRead(cgroupTaskDir, file)
+	dataStr, err := CgroupFileRead(cgroupTaskDir, r)
 	if err != nil {
```
zwzhang0107, line 67: // TODO ignore not exist error(not supported)
so do the followings

pkg/util/system/resource.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	if valid, msg := r.IsValid(strconv.FormatInt(*value, 10)); !valid {
+		klog.V(4).Infof("failed to validate cgroup value, path:%s, msg:%s", r.Path(parentDir), msg)
+		return false
+	}
+	return true
+}
+
+func SupportedIfAnolisOS() (bool, string) {
+	if HostSystemInfo.IsAnolisOS {
```
zwzhang0107, line 59: // TODO IsAnolisOS can be abandoned since each resource has `IsSupported()` method

pkg/util/system/cgroup_resource.go
```diff
@@ -18,124 +18,290 @@ package system
...
+		MemoryPriority,
+		MemoryUsePriorityOom,
+		MemoryOomGroup,
+		BlkioReadIops,
+		BlkioReadBps,
+		BlkioWriteIops,
+		BlkioWriteBps,
+	}
+
+	CPUQuotaV2    = DefaultFactory.NewV2(CPUCFSQuotaName, CPUMaxName).WithCheckSupported(SupportedIfCgroupsV2)
```
zwzhang0107, line 214: WithCheckSupported(SupportedIfCgroupsV2) is not necessary since it's enabled by default

pkg/util/system/cgroup_resource.go
```diff
@@ -18,124 +18,290 @@ package system
...
+	NewV2(t ResourceType, filename string) Resource
+}
+
+type cgroupResourceFactoryImpl struct{}
+
+func NewCgroupResourceFactory() CgroupResourceFactory {
+	return &cgroupResourceFactoryImpl{}
+}
+
+func (f *cgroupResourceFactoryImpl) New(filename string, subfs string) Resource {
```
zwzhang0107, line 301: will there be lots of if-else in `executor` since the same ResourceType(e.g. cfs quota) are divided to two files

saintube: ResourceType defines the fine-grained cgroup resource.  `Executor` will pick the correct CgroupResource according to the ResourceType.
e.g. In cgroups-v1, CFSQuota, CFSPeriod refer to file `cpu.cfs_quota_us` and `cpu.cfs_period_us` respectively, while both of them refer to `cpu.max` in v2.

saintube: It is convenient for updating cgroups with almost the same writing pattern on v1 and v2 (e.g. `memory.min`). It does require some if-else for cgroups with different patterns (e.g. `cpu.cfs_period_us`).

pkg/util/system/cgroup.go
saintube: the existence should be checked in `IsSupported()`. If a supported cgroup file is missing, we may report the error.

pkg/util/system/cgroup_resource.go
```diff
@@ -18,124 +18,290 @@ package system
...
+	CPUProcs     = DefaultFactory.New(CPUProcsName, CgroupCPUDir)
+
+	CPUSet = DefaultFactory.New(CPUSetCPUSName, CgroupCPUSetDir)
+
+	CPUAcctUsage = DefaultFactory.New(CPUAcctUsageName, CgroupCPUAcctDir)
+
+	MemoryLimit            = DefaultFactory.New(MemoryLimitName, CgroupMemDir)
+	MemoryUsage            = DefaultFactory.New(MemoryUsageName, CgroupMemDir)
+	MemoryStat             = DefaultFactory.New(MemoryStatName, CgroupMemDir)
+	MemoryWmarkRatio       = DefaultFactory.New(MemoryWmarkRatioName, CgroupMemDir).WithValidator(MemoryWmarkRatioValidator).WithCheckSupported(SupportedIfAnolisOS)
```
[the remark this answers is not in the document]
saintube: OK, we can check inside the `IfAnolisOS()`

pkg/util/system/cgroup_resource.go
saintube: done

pkg/util/system/resource.go
saintube: done

pkg/util/system/cgroup_resource.go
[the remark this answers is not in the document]
zwzhang0107: I mean we should not use `IfAnolisOS` to check wether a cgroup resource is supported. Because some versions of AnolisOS may only support few cgroup resource(e.g. MemoryWmarkRatio supported but bvt is not supported)
for resources like MemoryWmarkRatio, we should use `if-the-file-exist` for `WithCheckSupported`.
for resources like `bvt`, we should use more complex function.

FillZpp:
/approve
/hold
Please fix the conflicts.

pkg/util/system/cgroup_resource.go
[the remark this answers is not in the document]
saintube: fixed. check if file exist in `IsSupported()`

PR #835, opened 2022-11-24
static cpu allocation

lucming:

assigned cpus that bepods can use.

usage:
1. add ` "cpusAssigned": "18-23,6-11"` to `resource-threshold-config` under `slo-controller-config` in this configmap
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: slo-controller-config
  namespace: koordinator-system
  labels:
    app.kubernetes.io/managed-by: Helm
  annotations:
    meta.helm.sh/release-name: koordinator
    meta.helm.sh/release-namespace: default
data:
  colocation-config: |
    {
      "enable": true
    }
  resource-threshold-config: |
    {
      "clusterStrategy": {
        "enable": true,
        "cpusAssigned": "18-23,6-11"
      }
    }

```
2. submit a deployment to stress cpu,eg:
 ```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: stress
  namespace: lm-test
  labels:
    app: stress
spec:
  replicas: 1
  selector:
    matchLabels:
      app: stress
  template:
    metadata:
      labels:
        app: stress
        koordinator.sh/enable-colocation: 'true'
    spec:
      containers:
        - name: stress
          image: tj1633/ubuntu-stress
          command:
            - /bin/bash
            - -c
          args:
            - '/usr/bin/stress --cpu 12'
          resources:
            limits:
              cpu: '12'
              memory: 2Gi
            requests:
              cpu: '1'
              memory: 1Gi
      nodeName: stark12.add.bjyt.qihoo.net.   # it is best to specify node
```
 3. checking bepod cpu usage on node,and we can see
截屏2022-11-24 02 07 20

### V. Checklist

### DIFF
--- a/apis/slo/v1alpha1/nodeslo_types.go
+++ b/apis/slo/v1alpha1/nodeslo_types.go
@@ -181,6 +181,9 @@ type ResourceThresholdStrategy struct {
 	CPUEvictBESatisfactionLowerPercent *int64 `json:"cpuEvictBESatisfactionLowerPercent,omitempty"`
 	// cpu evict start after continue avg(cpuusage) > CPUEvictThresholdPercent in seconds
 	CPUEvictTimeWindowSeconds *int64 `json:"cpuEvictTimeWindowSeconds,omitempty"`
+
+	// specify the cpus that can be used
+	CPUsAssigned string `json:"cpusAssigned,omitempty"`
 }
 
 // ResctrlQOSCfg stores node-level config of resctrl qos
--- a/config/crd/bases/slo.koordinator.sh_nodeslos.yaml
+++ b/config/crd/bases/slo.koordinator.sh_nodeslos.yaml
@@ -820,6 +820,9 @@ spec:
                     maximum: 100
                     minimum: 0
                     type: integer
+                  cpusAssigned:
+                    description: specify the cpus that can be used
+                    type: string
                   enable:
                     description: whether the strategy is enabled, default = false
                     type: boolean
--- a/pkg/koordlet/resmanager/cgroup_reconcile.go
+++ b/pkg/koordlet/resmanager/cgroup_reconcile.go
@@ -533,36 +533,34 @@ func getKubeQoSResourceQoSByQoSClass(qosClass corev1.PodQOSClass, strategy *slov
 	if strategy == nil {
 		return nil
 	}
-	var resourceQoS *slov1alpha1.ResourceQOS
-	switch qosClass {
-	case corev1.PodQOSGuaranteed:
-		resourceQoS = strategy.LSRClass
-	case corev1.PodQOSBurstable:
-		resourceQoS = strategy.LSClass
-	case corev1.PodQOSBestEffort:
-		resourceQoS = strategy.BEClass
+
+	qosClassMap := map[corev1.PodQOSClass]*slov1alpha1.ResourceQOS{
+		corev1.PodQOSGuaranteed: strategy.LSRClass,
+		corev1.PodQOSBurstable:  strategy.LSClass,
+		corev1.PodQOSBestEffort: strategy.BEClass,
 	}
-	return resourceQoS
+
+	return qosClassMap[qosClass]
 }
 
 func getPodResourceQoSByQoSClass(pod *corev1.Pod, strategy *slov1alpha1.ResourceQOSStrategy, config *Config) *slov1alpha1.ResourceQOS {
 	if strategy == nil {
 		return nil
 	}
-	var resourceQoS *slov1alpha1.ResourceQOS
-	podQoS := apiext.GetPodQoSClass(pod)
-	switch podQoS {
-	case apiext.QoSLSR:
-		resourceQoS = strategy.LSRClass
-	case apiext.QoSLS:
-		resourceQoS = strategy.LSClass
-	case apiext.QoSBE:
-		resourceQoS = strategy.BEClass
-	default:
-		// qos=None pods uses config mapped from kubeQoS
-		resourceQoS = getKubeQoSResourceQoSByQoSClass(util.GetKubeQosClass(pod), strategy, config)
-		klog.V(6).Infof("get pod ResourceQOS according to kubeQoS for QoS=None pods, pod %s, "+
-			"resourceQoS %v", util.GetPodKey(pod), util.DumpJSON(resourceQoS))
+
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by hormes** [comment]
Because this configuration is at the global level, and nodes in cluster may be different CPU cores, set CPU core seems to have some scalability problems. Therefore, maybe we need to discuss what scenario needs to introduce this limitation, and then look at how to do better abstraction?

**Comment by lucming** [comment]
> Because this configuration is at the global level, and nodes in cluster may be different CPU cores, set CPU core seems to have some scalability problems. Therefore, maybe we need to discuss what scenario needs to introduce this limitation, and then look at how to do better abstraction?

we can also assign cpus on node level,just add `nodeStrategies`  will be fine.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: slo-controller-config
  namespace: koordinator-system
  labels:
    app.kubernetes.io/managed-by: Helm
  annotations:
    meta.helm.sh/release-name: koordinator
    meta.helm.sh/release-namespace: default
data:
  colocation-config: |
    {
      "enable": true
    }
  resource-threshold-config: |
    {
       "clusterStrategy": {
         "enable": true,
         "cpuSuppressThresholdPercent": 40,
         "cpuSuppressPolicy":"cpuset",
         "memoryEvictThresholdPercent": 70, 
         "memoryEvictLowerPercent": 2, 
         "cpuEvictBESatisfactionUpperPercent":30, 
         "cpuEvictBESatisfactionLowerPercent":30,  
         "cpuEvictTimeWindowSeconds":30,
         "cpusAssigned": "18-23,6-11"
         },
      "nodeStrategies":[
          {"nodeSelector": {
            "matchLabels":{"kubernetes.io/hostname": "stark12.add.bjyt.qihoo.net"}
          },
          "cpusAssigned": "18-23",
          "enable": true,
         "cpuSuppressThresholdPercent": 70,
         "cpuSuppressPolicy":"cpuset",
         "memoryEvictThresholdPercent": 70, 
         "memoryEvictLowerPercent": 2, 
         "cpuEvictBESatisfactionUpperPercent":30, 
         "cpuEvictBESatisfactionLowerPercent":30,  
         "cpuEvictTimeWindowSeconds":30 }
      ]
     }

```

eahydra:
From the example given, is it expected to be bound to specific cores according to Pod Limits? But who is responsible for allocating these cores? Both koordinator and koordlet have the ability to allocate cores. They are responsible for different dimensions, but they can work together very well. This PR only modifies koordlet. I think it may not be complete enough, because the scheduler is not aware of slo-controller-config, and the cores set in slo-controller-config will still be allocated by the scheduler. Maybe you can consider adding a specific proposal, describing the specific background and goals.

lucming:
the background is that we have a group of machines running middleware applications, but the resource utilization on these machines is low，and they can specify the cpu they use, our aim is to use up the remaining free resources,and then we will register those machines as k8s node to run the mixed deploy.

eahydra:
> the background is that we have a group of machines running middleware applications, but the resource utilization on these machines is low，and they can specify the cpu they use, our aim is to use up the remaining free resources,and then we will register those machines as k8s node to run the mixed deploy.

So, expect a specific node to reserve CPU?

lucming:
en

hormes:
> the background is that we have a group of machines running middleware applications, but the resource utilization on these machines is low，and they can specify the cpu they use, our aim is to use up the remaining free resources,and then we will register those machines as k8s node to run the mixed deploy.

Whether the middleware workloads are running on k8s?

lucming:
> > the background is that we have a group of machines running middleware applications, but the resource utilization on these machines is low，and they can specify the cpu they use, our aim is to use up the remaining free resources,and then we will register those machines as k8s node to run the mixed deploy.
> 
> Whether the middleware workloads are running on k8s?

No, the middleware workload is running directly on the machine

eahydra:
hi @lucming,  now the problem seems to be clear, I hope to reserve some CPU for some nodes. But at the same time, it should also include Memory and other resource information. Not sure how you're reserving resources other than CPU right now. In addition, these reserved resources should be perceived by the scheduler and excluded during the scheduling process. Therefore, how to pass the information of these reserved resources to the scheduler is a problem that needs to be solved. An optional solution is to update the reserved CPU and other resources to Node in the form of annotation based on your current changes, or update to the existing NodeResourceTopology CRD instance, and then the scheduler processes the annotation and corrects it when scheduling processing reserved resources. This needs to be discussed in detail. We have a community meeting next Tuesday, do you think we can chat at this meeting?

cc @hormes @zwzhang0107 @FillZpp

lucming:
> hi @lucming, now the problem seems to be clear, I hope to reserve some CPU for some nodes. But at the same time, it should also include Memory and other resource information. Not sure how you're reserving resources other than CPU right now. In addition, these reserved resources should be perceived by the scheduler and excluded during the scheduling process. Therefore, how to pass the information of these reserved resources to the scheduler is a problem that needs to be solved. An optional solution is to update the reserved CPU and other resources to Node in the form of annotation based on your current changes, or update to the existing NodeResourceTopology CRD instance, and then the scheduler processes the annotation and corrects it when scheduling processing reserved resources. This needs to be discussed in detail. We have a community meeting next Tuesday, do you think we can chat at this meeting? Meeting information can be found here: https://github.com/koordinator-sh/koordinator#community
> 
> cc @hormes @zwzhang0107 @FillZpp

Okay, let's discuss in detail at the meeting.

hormes:
> hi @lucming, now the problem seems to be clear, I hope to reserve some CPU for some nodes. But at the same time, it should also include Memory and other resource information. Not sure how you're reserving resources other than CPU right now. In addition, these reserved resources should be perceived by the scheduler and excluded during the scheduling process. Therefore, how to pass the information of these reserved resources to the scheduler is a problem that needs to be solved. An optional solution is to update the reserved CPU and other resources to Node in the form of annotation based on your current changes, or update to the existing NodeResourceTopology CRD instance, and then the scheduler processes the annotation and corrects it when scheduling processing reserved resources. This needs to be discussed in detail. We have a community meeting next Tuesday, do you think we can chat at this meeting? Meeting information can be found here: https://github.com/koordinator-sh/koordinator#community
> 
> cc @hormes @zwzhang0107 @FillZpp

Yeah, the reserved resource should be excluded by scheduler, koordlet and the slo-controller. We need a proposal to let these changes be clearly presented to ensure that no omissions occur.

FillZpp:
/close
latest: #998

PR #843, opened 2022-11-28
koord-scheduler: LoadAware supports load scheduling by Prod Pods

### LINKED ISSUE #314: [proposal] enhance Load-aware scheduling adopts different strategies according to the difference of workload type
**What is your proposal**:

Load-aware scheduling adopts different strategies according to the difference of workload type.

**Why is this needed**:

Currently, load-aware scheduling does not distinguish the type of workloads, and only evaluates it according to the current load status of the node and the estimated value of the Pod. This approach also works efficiently, but it would be better if we supported some scheduling strategy for the workload type.

The situation caused by the current method is that the utilization among the nodes of the entire cluster is relatively balanced, but the utilization of latency-sensitive applications and best-effort applications is unbalanced.

If there are many BestEffort Pods scheduled in one Node, the latency-sensitive Pods may fail to schedule cause the load of node has reached the limit of usage.   

So we can do some optimizations and enhancements:

For latency-sensitive(LSE/LSR/LS) Pods, priority is given to scheduling to the nodes with low total utilization of the Prod Pods. BestEffort(BE) Pods are scheduled according to the utilization level of the whole node.  

eahydra:

fix #314.

Features:
1. Supports filter nodes by `ProdUsageThresholds` that defined in the config of scheduler or the annotation of nodes.
The `ProdUsageThresholds`  indicates the resource utilization threshold of Prod Pods compared to the whole machine. 
The default for CPU is 55%, and the default for memory is 75%.
2.  When scoring, estimate the usage of the recently scheduled Prod Pod or take the Pod usage reported by NodeMetric, and calculate the load water level of the Prod Pod after accumulating the total amount. The lower the water level, the higher the score.

### V. Checklist


pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -120,39 +125,83 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+	}
+	return nil
+}
+
+func (p *Plugin) filterProdUsage(node *corev1.Node, nodeMetric *slov1alpha1.NodeMetric, prodUsageThresholds map[corev1.ResourceName]int64) *framework.Status {
+	if len(nodeMetric.Status.PodsMetric) == 0 {
+		return nil
+	}
+
+	// TODO(joseph): maybe we should estimate the Pod that just be scheduled that have not reported
```
saintube, line 185: +1

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -120,39 +125,83 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+		used := nodeMetric.Status.NodeMetric.NodeUsage.ResourceList[resourceName]
+		usage := int64(math.Round(float64(used.MilliValue()) / float64(total.MilliValue()) * 100))
+		if usage >= threshold {
+			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf(ErrReasonUsageExceedThreshold, resourceName))
+		}
+	}
+	return nil
+}
+
+func (p *Plugin) filterProdUsage(node *corev1.Node, nodeMetric *slov1alpha1.NodeMetric, prodUsageThresholds map[corev1.ResourceName]int64) *framework.Status {
```
saintube, line 180: do we need count the system usage either, as some system processes or non-containerized apps can affect Prod pods?

eahydra: Actually it is needed. But this needs to extend NodeMetric CRD and be supported by koordlet. I think this extension can be used as a subsequent optimization item and implemented when it is really needed.

saintube: OK. Currently, there is an approximation where we can subtract all pods' usage from the node usage to estimate the system usage.

eahydra: Yes. Change the NodeMetric CRD will be planned in the future

pkg/scheduler/plugins/loadaware/load_aware.go
eahydra: Support in the next PR

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -188,64 +237,62 @@ func (p *Plugin) Score(ctx context.Context, state *framework.CycleState, pod *co
...
-		if resourceName == corev1.ResourceCPU {
-			allocatable[resourceName] = quantity.MilliValue()
-		} else {
-			allocatable[resourceName] = quantity.Value()
+	if prodPod {
+		for resourceName, quantity := range sumPodUsages(podMetrics, estimatedPods) {
+			estimatedUsed[resourceName] += getResourceValue(resourceName, quantity)
 		}
+	} else {
 		if nodeMetric.Status.NodeMetric != nil {
```
ZiMengSheng, line 253: some pod usage may be counted repeatedly.

eahydra: resolved.

ZiMengSheng (approved):

FillZpp (approved):
/lgtm

pkg/scheduler/plugins/loadaware/helper.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func getNodeMetricReportInterval(nodeMetric *slov1alpha1.NodeMetric) time.Duration {
+	if nodeMetric.Spec.CollectPolicy == nil || nodeMetric.Spec.CollectPolicy.ReportIntervalSeconds == nil {
+		return DefaultNodeMetricReportInterval
+	}
+	return time.Duration(*nodeMetric.Spec.CollectPolicy.ReportIntervalSeconds) * time.Second
+}
+
+func getPodNamespacedName(namespace, name string) string {
```
jasonliu747, line 46: 这个函数貌似在许多插件里都实现了一次，我们有空的时候或许可以把他扔进util里。

jasonliu747 (commented):
great work as always.

pkg/scheduler/plugins/loadaware/helper.go
eahydra: 是的。确实可以把一些函数挪到util的。

Issue #846, opened 2022-11-29
[proposal] scheduler support filter logs


eahydra:
**What is your proposal**:

When Pod scheduling fails, there is currently no good way to quickly diagnose why Pods cannot be scheduled to certain nodes. It is best to support printing out Filter error results when diagnostic analysis is required.

**Is there a suggested solution, if so, please add it**:

1. We can implement `RunFilterPlugins` in `frameworkExtenderImpl`, logs the filter statuses.  
2. Add Flags and setter like `debugTopNScores` named `logFilterStatuses`.

PR #847, opened 2022-11-29
koord-scheduler: support log filtering failure reasons as needed

ZiMengSheng:

this PR solves issue #846

image

### V. Checklist

- [] I have written necessary docs and comments

### DIFF
--- a/pkg/scheduler/frameworkext/framework_extender.go
+++ b/pkg/scheduler/frameworkext/framework_extender.go
@@ -227,7 +227,11 @@ func (ext *frameworkExtenderImpl) RunFilterPluginsWithNominatedPods(ctx context.
 			nodeInfo = newNodeInfo
 		}
 	}
-	return ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	status := ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	if !status.IsSuccess() && debugFilterFailure {
+		klog.Infof("Failed to filter for Pod %q on Node %q, failedPlugin: %s, reason: %s", klog.KObj(pod), klog.KObj(nodeInfo.Node()), status.FailedPlugin(), status.Message())
+	}
+	return status
 }
 
 func (ext *frameworkExtenderImpl) RunScorePlugins(ctx context.Context, state *framework.CycleState, pod *corev1.Pod, nodes []*corev1.Node) (framework.PluginToNodeScores, *framework.Status) {
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -273,6 +273,7 @@ func installProfilingHandler(pathRecorderMux *mux.PathRecorderMux, enableContent
 	debugFlags := utilroutes.NewDebugFlags(pathRecorderMux)
 	debugFlags.Install("v", utilroutes.StringFlagPutHandler(logs.GlogSetter))
 	debugFlags.Install("s", utilroutes.StringFlagPutHandler(frameworkext.DebugScoresSetter))
+	debugFlags.Install("f", utilroutes.StringFlagPutHandler(frameworkext.DebugFiltersSetter))
 }
 
 // newMetricsHandler builds a metrics server from the config.
--- a/pkg/scheduler/frameworkext/debug.go
+++ b/pkg/scheduler/frameworkext/debug.go
@@ -29,11 +29,13 @@ import (
 )
 
 var (
-	debugTopNScores = 0
+	debugTopNScores    = 0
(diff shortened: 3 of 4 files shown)


pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -227,7 +227,11 @@ func (ext *frameworkExtenderImpl) RunFilterPluginsWithNominatedPods(ctx context.
 			nodeInfo = newNodeInfo
 		}
 	}
-	return ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	status := ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	if !status.IsSuccess() && debugFilterFailure {
+		klog.Infof("Failed Filter Status For Pod %v on node %v: %v", klog.KObj(pod), klog.KObj(nodeInfo.Node()), status)
```
eahydra, line 232: recommend to use klog.Error

eahydra, line 232: we also need to print failed plugins and failed reason.

pkg/scheduler/frameworkext/debug_scores_filters.go
```diff
@@ -29,11 +29,13 @@ import (
...
 
 var (
-	debugTopNScores = 0
+	debugTopNScores    = 0
+	debugFilterFailure = false
 )
 
 func AddFlags(fs *pflag.FlagSet) {
 	fs.IntVarP(&debugTopNScores, "debug-scores", "s", debugTopNScores, "logging topN nodes score and scores for each plugin after running the score extension, disable if set to 0")
+	fs.BoolVarP(&debugFilterFailure, "debug-filters", "f", debugFilterFailure, "logging filter failures")
```
FillZpp, line 38: Why not reuse `--v` to set the log level?

ZiMengSheng: this already exists in cmd/koord-scheduler/app/server.go

func installProfilingHandler(pathRecorderMux *mux.PathRecorderMux, enableContentionProfiling bool) {
	routes.Profiling{}.Install(pathRecorderMux)
	if enableContentionProfiling {
		goruntime.SetBlockProfileRate(1)
	}
	// NOTE: Use utilroutes.DebugFlags instead of k8s.io/apiserver/pkg/server/routes.DebugFlags
	//  as using the latter will print a useless stack when installing multiple flags
	debugFlags := utilroutes.NewDebugFlags(pathRecorderMux)
	debugFlags.Install("v", utilroutes.StringFlagPutHandler(logs.GlogSetter))
	debugFlags.Install("s", utilroutes.StringFlagPutHandler(frameworkext.DebugScoresSetter))
	debugFlags.Install("f", utilroutes.StringFlagPutHandler(frameworkext.DebugFiltersSetter))
}

FillZpp: I mean you can reuse the flag instead of adding a new one.

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -227,7 +227,11 @@ func (ext *frameworkExtenderImpl) RunFilterPluginsWithNominatedPods(ctx context.
 			nodeInfo = newNodeInfo
 		}
 	}
-	return ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	status := ext.Framework.RunFilterPluginsWithNominatedPods(ctx, cycleState, pod, nodeInfo)
+	if !status.IsSuccess() && debugFilterFailure {
+		klog.Errorf("Failed to filter for Pod %q on Node %q, failedPlugin: %s, reason: %s", klog.KObj(pod), klog.KObj(nodeInfo.Node()), status.FailedPlugin(), status.Message())
```
jasonliu747, line 232: log messages should be lowercase, therefore `Failed` -> `failed`.

pkg/scheduler/frameworkext/debug_scores_filters.go
ZiMengSheng: If we only want to diagnose the filter failed reasons, we dont't need to much unrelated verbose logs.

jasonliu747 (commented):
/lgtm

pkg/scheduler/frameworkext/debug_scores_filters.go
eahydra: > Why not reuse `--v` to set the log level?

Most users set the log level to V(4) or even lower, and then decide whether to enable this diagnostic method according to their needs. Most of the time, it is OK to control whether to output this log through the log level, but it is a bit troublesome to only prohibit the output of this log.

FillZpp (approved):
/approve

Issue #856, opened 2022-12-05
[proposal] more detailed data is need in NodeMetric for scheduler/descheduler 


zwzhang0107:
**What is your proposal**:
Add more aggregate data in nodeMetricUsage (e.g. p90, p99).

**Why is this needed**:
Scheduler/De-scheduler need more data about resource usage for load-aware scheduling.
The short/mid-term (seconds, minutes) usage info should comes from koordlet.
The long-term (hours, even days) usage info should comes from other sources(e.g. prometheus). (FUTURE WORK)

**Is there a suggested solution, if so, please add it**:

FillZpp:
fixed by #861

/close

PR #861, opened 2022-12-06
api: add percentile node usage to NodeMetric

zwzhang0107:

add percentile usage info of node for scheduler/de-scheduler to solve the load-aware issue.

#856

### V. Checklist


apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -23,8 +23,25 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type AggregationType string
+
+const (
+	// max is not welcomed since it may import outliers
```
jasonliu747, line 29: do we have to mention `max` here?

jasonliu747 (commented):
/lgtm

apis/slo/v1alpha1/nodemetric_types.go
zwzhang0107: in order to explain the reason that why we don't add the `MAX` aggregate type temporarily.

eahydra (commented):
/lgtm

apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -23,8 +23,25 @@ import (
...
+	AVG    AggregationType = "avg"
+	P99    AggregationType = "p99"
+	P95    AggregationType = "p95"
+	P90    AggregationType = "p90"
+	MEDIAN AggregationType = "median"
+)
+
 type NodeMetricInfo struct {
-	NodeUsage ResourceMap `json:"nodeUsage,omitempty"`
+	NodeUsage            ResourceMap       `json:"nodeUsage,omitempty"`
```
FillZpp, line 38: If the field is optional, it had better be a pointer type like `*ResourceMap`, or we can remove the `omitempty` to make it a required field.

apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -23,8 +23,25 @@ import (
...
 
+type AggregationType string
+
+const (
+	// max is not welcomed since it may import outliers
+	AVG    AggregationType = "avg"
+	P99    AggregationType = "p99"
+	P95    AggregationType = "p95"
+	P90    AggregationType = "p90"
+	MEDIAN AggregationType = "median"
```
hormes, line 34: Why not use p50 to keep consistently

eahydra: Maybe in the percentile statistical method, the median is generally used to represent p50

apis/slo/v1alpha1/nodemetric_types.go
zwzhang0107: it is an internal protocol, and ResourceMap contains an inline map. so it does not need to be defined as pointer type.

FillZpp (commented):
/lgtm

PR #863, opened 2022-12-06
docs: add star history into README-zh_CN

jasonliu747:

### V. Checklist


FillZpp:
LG, but why not also add it into README?

jasonliu747:
> LG, but why not also add it into README?

Added in separate PR, #862.

PR #873, opened 2022-12-09
koord-scheduler: fix Coscheduling resource deadlock with groups

### LINKED ISSUE #872: [BUG]deadlock when schedule multi-gang-group
**What happened**:
if there are pod group A,B,C,D in the cluster, A and B are a group while C and D are a group.
if the pod groups are scheduled in the following order: A C B D, a deadlock may happen when pod group A and C occupy all resource is the cluster.
when pod group B find that the resource is not enough, it can reject pod group A so that pod group C and D can bind.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

KunWuLuan:
fix: deadlock when schedule multi-gang-group

fixes #872

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
 	listerv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -62,12 +63,14 @@ type Manager interface {
 	PostBind(context.Context, *corev1.Pod, string)
 	PostFilter(context.Context, *corev1.Pod, framework.Handle, string) (*framework.PostFilterResult, *framework.Status)
 	GetCreatTime(*framework.QueuedPodInfo) time.Time
+	GetGroupId(*corev1.Pod) (string, error)
 	GetAllPodsFromGang(string) []*corev1.Pod
 	ActivateSiblings(*corev1.Pod, *framework.CycleState)
 	AllowGangGroup(*corev1.Pod, framework.Handle, string)
 	Unreserve(context.Context, *framework.CycleState, *corev1.Pod, string, framework.Handle, string)
 	GetGangSummary(gangId string) (*GangSummary, bool)
 	GetGangSummaries() map[string]*GangSummary
+	IsGangMinSatisfied(*corev1.Pod) bool
 }
 
 // PodGroupManager defines the scheduling operation called
@@ -135,6 +138,28 @@ func (pgMgr *PodGroupManager) OnPodGroupDelete(obj interface{}) {
 	pgMgr.cache.onPodGroupDelete(obj)
 }
 
+func (pgMgr *PodGroupManager) GetGroupId(pod *corev1.Pod) (string, error) {
+	gang := pgMgr.GetGangByPod(pod)
+	if gang == nil {
+		return "", fmt.Errorf("gang doesn't exist in cache")
+	}
+
+	return gang.GangGroupId, nil
+}
+
+func (pgMgr *PodGroupManager) IsGangMinSatisfied(pod *corev1.Pod) bool {
+	gang := pgMgr.GetGangByPod(pod)
+	if gang == nil {
+		return false
+	}
+
+	if gang.isGangOnceResourceSatisfied() {
+		return true
+	}
+
+	return gang.MinRequiredNumber <= gang.getGangAssumedPods()
+}
+
 // ActivateSiblings stashes the pods belonging to the same PodGroup of the given pod
 // in the given state, with a reserved key "kubernetes.io/pods-to-activate".
 func (pgMgr *PodGroupManager) ActivateSiblings(pod *corev1.Pod, state *framework.CycleState) {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -62,12 +63,14 @@ type Manager interface {
 	PostBind(context.Context, *corev1.Pod, string)
 	PostFilter(context.Context, *corev1.Pod, framework.Handle, string) (*framework.PostFilterResult, *framework.Status)
 	GetCreatTime(*framework.QueuedPodInfo) time.Time
+	GetGroupId(*corev1.Pod) ([]byte, error)
```
eahydra, line 66: it's better to return string instead of byte slice.

pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -50,6 +53,7 @@ type Gang struct {
 	Mode              string
 	MinRequiredNumber int
 	TotalChildrenNum  int
+	GangGroupId       []byte
```
eahydra, line 56: use `string` instead of `[]byte`

pkg/scheduler/plugins/coscheduling/core/gang_cache_test.go
```diff
@@ -484,6 +487,15 @@ func TestGangCache_OnPodAdd(t *testing.T) {
...
 				gangCache.onPodAdd(pod)
 			}
+			for k, v := range tt.wantCache {
+				if !v.HasGangInit {
+					continue
+				}
+				sort.Strings(v.GangGroup)
+				s := strings.Join(v.GangGroup, ",")
+				encoded := md5.Sum([]byte(s))
+				tt.wantCache[k].GangGroupId = encoded[:]
```
eahydra, line 497: these code are duplicated in gang.go.

pkg/scheduler/plugins/coscheduling/core/gang_cache_test.go
```diff
@@ -669,6 +681,15 @@ func TestGangCache_OnPodDelete(t *testing.T) {
...
 				gangCache.onPodDelete(pod)
 			}
+			for k, v := range tt.wantCache {
+				if !v.HasGangInit {
+					continue
+				}
+				sort.Strings(v.GangGroup)
+				s := strings.Join(v.GangGroup, ",")
+				encoded := md5.Sum([]byte(s))
+				tt.wantCache[k].GangGroupId = encoded[:]
```
eahydra, line 691: same problem here

pkg/scheduler/plugins/coscheduling/core/gang_cache_test.go
```diff
@@ -815,6 +836,15 @@ func TestGangCache_OnPodGroupAdd(t *testing.T) {
...
 				gangCache.onPodGroupAdd(pg)
 			}
+			for k, v := range tt.wantCache {
+				if !v.HasGangInit {
+					continue
+				}
+				sort.Strings(v.GangGroup)
+				s := strings.Join(v.GangGroup, ",")
+				encoded := md5.Sum([]byte(s))
+				tt.wantCache[k].GangGroupId = encoded[:]
```
eahydra, line 846: same problem here

pkg/scheduler/plugins/coscheduling/core/gang_cache_test.go
```diff
@@ -931,6 +961,10 @@ func TestGangCache_OnGangDelete(t *testing.T) {
 		ChildrenScheduleRoundMap: map[string]int{},
 	}
 	cacheGang := cache.getGangFromCacheByGangId("default/gangb", false)
+	sort.Strings(wantedGang.GangGroup)
+	s := strings.Join(wantedGang.GangGroup, ",")
+	encoded := md5.Sum([]byte(s))
+	wantedGang.GangGroupId = encoded[:]
```
eahydra, line 967: same problem here

eahydra (changes requested):

eahydra (commented):
/lgtm

jasonliu747 (commented):
@KunWuLuan Hi, thanks for your hard work. Please fix the CI, and we're all set.

eahydra (commented):
/lgtm

pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -311,16 +328,40 @@ func (pgMgr *PodGroupManager) Unreserve(ctx context.Context, state *framework.Cy
 	gang.delAssumedPod(pod)
 
 	if !gang.isGangOnceResourceSatisfied() && gang.getGangMode() == extension.GangModeStrict {
-		// release resource of all assumed children of the gang
+		pgMgr.rejectGangGroupById(gang.Name, handle)
+	}
+}
+
+func (pgMgr *PodGroupManager) rejectGangGroupById(gangId string, handle framework.Handle) {
```
xulinfei1996, line 335: Maybe can add pluginName string as input, print waitingPod.Reject(pluginName,"gang rejection by another ganggroup").
Or modify the interface Manager's functions Unreserve、PostFilter.

eahydra: The pgMgr.PostFilter method already has a pluginName parameter. Can use this directly.

pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -151,7 +152,7 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
 			gang.Name, pod.Annotations[extension.AnnotationGangGroups])
 	}
 	gang.GangGroup = groupSlice
-
+	gang.GangGroupId = util.GetGangGroupId(groupSlice)
```
eahydra, line 155: An optimization can be considered here to ensure that the groupSlice contains at least one member and prevent the group id from being empty.

pkg/scheduler/plugins/coscheduling/core/core.go
KunWuLuan: good idea, thanks

pkg/scheduler/plugins/coscheduling/core/gang.go
KunWuLuan: You mean the groupId is a pointer to string?

eahydra (approved):
/lgtm
/approve

PR #882, opened 2022-12-14
koord-scheduler: LoadAware scheduling support percentile usage

### LINKED ISSUE #856: [proposal] more detailed data is need in NodeMetric for scheduler/descheduler

eahydra:

- The LoadAware scheduling plugin support filter/score with aggregated usages such as P99/P95.
- Not enabled by default.  Users can set `AggregatedUsages` via the annotation of Node or args of plugin.
- If the koordlet have not report the aggregated usages, koord-scheduler will use the estimated result as current usage to balance.

fix #856 

### V. Checklist


pkg/scheduler/plugins/loadaware/helper.go
```diff
@@ -43,6 +47,96 @@ func getNodeMetricReportInterval(nodeMetric *slov1alpha1.NodeMetric) time.Durati
...
+func stillInTheReportInterval(assignedTime, updateTime time.Time, reportInterval time.Duration) bool {
+	return assignedTime.Before(updateTime) && updateTime.Sub(assignedTime) < reportInterval
+}
+
+func getTargetAggregatedUsage(nodeMetric *slov1alpha1.NodeMetric, aggregatedDuration *metav1.Duration, aggregationType slov1alpha1.AggregationType) *slov1alpha1.ResourceMap {
+	if nodeMetric.Status.NodeMetric == nil || len(nodeMetric.Status.NodeMetric.AggregatedNodeUsages) == 0 {
+		return nil
+	}
+
+	if aggregatedDuration == nil || aggregatedDuration.Duration == 0 {
```
saintube, line 63: what is the scenario where `Duration == nil` and the aggregated result of the max duration is picked?

eahydra: It's just a default strategy, and it's also an empirical strategy. Some users expect to schedule according to the data of the maximum aggregation period.

saintube: The default strategy may be not obvious to understand. How about adding some comments?

eahydra:
optimize filter implement. cc @saintube

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -125,40 +125,42 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
 		if !status.IsSuccess() {
 			return status
 		}
-	}
-
-	if len(prodUsageThresholds) > 0 {
-		status := p.filterProdUsage(node, nodeMetric, prodUsageThresholds)
-		if !status.IsSuccess() {
-			return status
+	} else {
```
saintube, line 134: Here the `else` condition indicates `len(filterProfile.ProdUsageThresholds) == 0 || extension.GetPriorityClass(pod) != extension.PriorityProd`. Is it expected?

eahydra: yes. It's expected.

ZiMengSheng (approved):

buptcozy (commented):
/lgtm

buptcozy (approved):
/approve

FillZpp (approved):
/lgtm

PR #887, opened 2022-12-15
koordlet: release feature gates config

zwzhang0107:

### V. Checklist

### DIFF
--- a/config/manager/koordlet.yaml
+++ b/config/manager/koordlet.yaml
@@ -23,8 +23,7 @@ spec:
       containers:
         - args:
             - -cgroup-root-dir=/host-cgroup/
-            - -feature-gates=AllAlpha=true
-            - -runtime-hooks=AllAlpha=true
+            - -feature-gates=AllAlpha=true,AllBeta=true
             - -runtime-hooks-network=unix
             - -runtime-hooks-addr=/host-var-run-koordlet/koordlet.sock
             - -runtime-hooks-host-endpoint=/var/run/koordlet//koordlet.sock
--- a/pkg/features/koordlet_features.go
+++ b/pkg/features/koordlet_features.go
@@ -26,49 +26,96 @@ import (
 )
 
 const (
-	// AuditEvents is used to audit recent events
+	// owner: @zwzhang0107
+	// alpha: v0.1
+	//
+	// AuditEvents is used to audit recent events.
 	AuditEvents featuregate.Feature = "AuditEvents"
 
-	// AuditEventsHTTPHandler is used to get recent events from koordlet port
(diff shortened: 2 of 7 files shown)


jasonliu747 (changes requested):
CI failed, PTAL.

eahydra (approved):
/lgtm

config/manager/koordlet.yaml
```diff
@@ -23,8 +23,6 @@ spec:
       containers:
         - args:
             - -cgroup-root-dir=/host-cgroup/
-            - -feature-gates=AllAlpha=true
```
FillZpp, line 26: Why should we remove this? It is to make sure all feature-gates open in e2e or test cluster so that we can test them all.

zwzhang0107: some GA features maybe closed by default, `AllAlpha=true/AllBeta=true` cannot cover this.
how about set the e2e config to open All feature-gates, and use default value here.

jasonliu747 (approved):

config/manager/koordlet.yaml
FillZpp: Here is the YAMLs for test and e2e, so how about set the config to open all features here?

pkg/features/koordlet_features.go
```diff
@@ -84,15 +131,15 @@ var (
...
-		CPUBurst:               {Default: false, PreRelease: featuregate.Alpha},
-		RdtResctrl:             {Default: false, PreRelease: featuregate.Alpha},
-		CgroupReconcile:        {Default: false, PreRelease: featuregate.Alpha},
-		NodeTopologyReport:     {Default: false, PreRelease: featuregate.Alpha},
-		Accelerators:           {Default: false, PreRelease: featuregate.Alpha},
+		CPUBurst:               {Default: true, PreRelease: featuregate.Beta},
+		RdtResctrl:             {Default: true, PreRelease: featuregate.Beta},
+		CgroupReconcile:        {Default: true, PreRelease: featuregate.Alpha},
+		NodeTopologyReport:     {Default: true, PreRelease: featuregate.Beta},
+		Accelerators:           {Default: true, PreRelease: featuregate.Alpha},
```
hormes, line 142: why alpha feature default to true ?

config/manager/koordlet.yaml
zwzhang0107: sure, -runtime-hooks have been merged into feature-gates

zwzhang0107:
ping @fillzpp @eahydra @jasonliu747

FillZpp (approved):
/lgtm

PR #888, opened 2022-12-16
optimize logic of get node strategy

lucming:
Just rearranging the logic for getting the node strategy. It maybe easier to read when written like this.

### V. Checklist


pkg/slo-controller/config/config.go
```diff
@@ -189,20 +189,23 @@ func GetNodeColocationStrategy(cfg *ColocationCfg, node *corev1.Node) *Colocatio
 
 	nodeLabels := labels.Set(node.Labels)
 	for _, nodeCfg := range cfg.NodeConfigs {
+		if nodeCfg.NodeSelector == nil {
```
saintube, line 192: A nil NodeSelector is translated to match nothing via `LabelSelectorAsSelector()`. Maybe we do not need to check if it is nil.

lucming: en,indeed.

jasonliu747 (commented):
/lgtm

PR #889, opened 2022-12-16
koord-descheduler: support loadaware/lowNodeLoad descheduling

### LINKED ISSUE #201: [proposal] load-aware descheduling
**What is your proposal**:

Support load-aware descheduling

**Why is this needed**:

Although Koordinator now has load-aware scheduling capabilities, but it is a best-effort mechanism. Whether it is a latency-sensitive application or a best-effort workloads, it is always possible to generate local hot spots in actual operation. At this time, it is necessary to optimize local hot spots by rescheduling

eahydra:

Support LoadAware descheduling. Currently support `LowNodeLoad` strategy.  
`LowNodeLoad` evicts pods from overutilized nodes to underutilized nodes.  Note that the plugin refers to the actual usage of the node. `LowNodeLoad` implements the extension point `Balance`.  It will be called periodically.

Basic process:
1. Filter nodes according to the configured `nodeSelector`, and combine the configured `lowThresholds` and `highThresholds` to filter out which nodes are `sourceNodes` and which nodes are `lowNodes`(or named `destinationNodes`).
2. Calculate the resource capacity that can be used for migration based on the actual resource usage of the target node and the target threshold(`highThresholds`).
3. Filter the Pods in the `sourceNodes` as `removablePods` according to the configured `evictableNamespaces` and `podSelectors`.
5. Sort `removablePods` by `Koordinator PriorityClass`, `Koordinator QoSClass`, `K8s QoSClass`, `Priority`, `Usage`, and `creation time`.
6. One by one evicts each Pod and reduces the corresponding resource capacity that can be used for migration until all Pods are evicted or the available resource capacity is 0.

fix #201 

### V. Checklist


eahydra:
/hold cancel
ready for review

pkg/descheduler/framework/plugins/loadaware/sorter.go
```diff
@@ -0,0 +1,150 @@
...
+func sortPods(pods []*corev1.Pod, podMetrics map[types.NamespacedName]*slov1alpha1.ResourceMap, resourceNames []corev1.ResourceName) {
+	sort.Slice(pods, func(i, j int) bool {
+		iPod, jPod := pods[i], pods[j]
+		iKoordPriorityClassOrder := koordPriorityClassOrder[extension.GetPriorityClass(iPod)]
+		jKoordPriorityClassOrder := koordPriorityClassOrder[extension.GetPriorityClass(jPod)]
+		if iKoordPriorityClassOrder != jKoordPriorityClassOrder {
+			return iKoordPriorityClassOrder > jKoordPriorityClassOrder
+		}
+
+		iKoordQoSOrder := koordQoSOrder[extension.GetPodQoSClass(iPod)]
```
saintube, line 94: except the priority, do we need to check koord qosClass and kube qosClass?

pkg/descheduler/apis/config/validation/validation_pluginargs_test.go
```diff
@@ -1,3 +1,19 @@
+/*
```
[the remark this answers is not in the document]
eahydra: Part of the code in descheduler reuses k8s descheduler. So the license check of the entire descheduler code is skipped. Adjust the check configuration later.

pkg/descheduler/framework/plugins/loadaware/sorter.go
eahydra: I think Kube QoS should be checked, although the runtime quality assurance semantics corresponding to Kube QoS are not as strong as the semantics of koordinator QoS. In addition, we also need to consider that existing Pods do not have Koordinator QoS labels.

saintube: Maybe we can check both the priority and the kube qosClass, ignoring the koordinator QoS to simply the comparison

pkg/descheduler/apis/config/validation/validation_pluginargs_test.go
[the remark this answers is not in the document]
saintube: OK

pkg/descheduler/framework/plugins/loadaware/sorter.go
eahydra: I still want to use Koordinator QoS as sorting criteria. It also does not conflict with other parameters.

saintube: OK.

eahydra:
fix LowNodeLoad args error. cc @hormes @saintube

FillZpp (approved):
/lgtm

PR #894, opened 2022-12-20
koord-schedler: fix queuesortplugin of coscheduling

lucming:
`coscheduling.Less`, the id of the pg should be compared, and the old logic is compare with `{pod.namespace}/{pod.name}` maybe not right.
Our ultimate goal should be to put pods belonging to the same pg together in schedulerqueue.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -31,10 +31,9 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
-	pgformers "sigs.k8s.io/scheduler-plugins/pkg/generated/informers/externalversions"
-
 	"sigs.k8s.io/scheduler-plugins/pkg/apis/scheduling/v1alpha1"
 	pgclientset "sigs.k8s.io/scheduler-plugins/pkg/generated/clientset/versioned"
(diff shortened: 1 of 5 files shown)


eahydra:
Hi @lucming , It is also strange that this PR uses PodGroup UID as one of the sorting conditions, especially when the sorting condition of the time dimension is also removed.

I think your problem should be solved by #873 . You can take a look at the code of this PR.

lucming:
> Hi @lucming , It is also strange that this PR uses PodGroup UID as one of the sorting conditions, especially when the sorting condition of the time dimension is also removed.
> 
> I think your problem should be solved by #873 . You can take a look at the code of this PR.

okay,i will close this pr. IMO,we just want something to make sure the pod in the same pg，`pg.uid` maybe can use directly,is it not necessary to regenerate a hash code?

eahydra:
> 

yeah...The hash code is not necessary, I understand that it is just for comparison.

eahydra:
> > Hi @lucming , It is also strange that this PR uses PodGroup UID as one of the sorting conditions, especially when the sorting condition of the time dimension is also removed.
> > I think your problem should be solved by #873 . You can take a look at the code of this PR.
> 
> okay, i will close this pr. IMO,we just want something to make sure the pod in the same pg, `pg.uid` maybe can use directly, is it not necessary to regenerate a hash code? Of course that's fine, I'm just giving my opinion.

yeah...The hash code is not necessary, I understand that it is just for comparison.

eahydra:
/close 
resolved by #873

PR #900, opened 2022-12-22
koordlet: fix pleg to increase pod sync interval

zwzhang0107:

memory limits of BE maybe too late to set if the pod sync interval is too long

### V. Checklist


pkg/koordlet/statesinformer/states_pods.go
```diff
@@ -66,6 +66,7 @@ func NewPodsInformer() *podsInformer {
 		podMap:       map[string]*PodMeta{},
 		podHasSynced: atomic.NewBool(false),
 		pleg:         p,
+		podCreated:   make(chan string, 1),
```
saintube, line 69: /lgtm

PR #908, opened 2022-12-28
koord-scheduler: check if reservation port match the pod

KunWuLuan:
Signed-off-by: KunWuLuan <kunwuluan@gmail.com>

Increase the judgment on port when judging whether the pod matches the reservation.

Now koordinator only judge on resource and owners when judging whether the pod matches the reservation. And remove the ports according to the pods' requirement. (https://github.com/koordinator-sh/koordinator/blob/main/pkg/scheduler/plugins/reservation/hook.go#L240)

By the implementation, reserve ports are created by all reservations on node, so the pod may remove the ports which are reserved by more than one reservation.

If a pod requires ports which are reserved by more than one reservation and its resource requirement can be satisfied by one of the reservations, the pod can be bind to the node. Then this pod will use the ports that are reserved by other reservations, which will cause the port conflicts with the other pods.

For example, reservation A reserve 100m cpu, 100Mi mem and port 1000, reservation B reserve 1000m cpu and 1000Mi mem. Now a pod which require 1000m cpu and 1000Mi mem with port 1000 can pass the filter. In this case the pod will be bound to reservation B. Then the pod who requires 100m cpu, 100Mi mem and port 1000 can still pass the filter. And it will have port conflict with the previous pod.

### DIFF
--- a/pkg/scheduler/plugins/reservation/utils.go
+++ b/pkg/scheduler/plugins/reservation/utils.go
@@ -250,7 +250,18 @@ func getReservationRequests(r *schedulingv1alpha1.Reservation) corev1.ResourceLi
 }
 
 func matchReservation(pod *corev1.Pod, rMeta *reservationInfo) bool {
-	return matchReservationOwners(pod, rMeta.Reservation) && matchReservationResources(pod, rMeta.Reservation, rMeta.Resources)
+	return matchReservationOwners(pod, rMeta.Reservation) && matchReservationResources(pod, rMeta.Reservation, rMeta.Resources) && matchReservationPort(pod, rMeta)
+}
+
+func matchReservationPort(pod *corev1.Pod, rMeta *reservationInfo) bool {
+	for _, container := range pod.Spec.Containers {
+		for _, podPort := range container.Ports {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/reservation/utils.go
```diff
@@ -250,7 +250,18 @@ func getReservationRequests(r *schedulingv1alpha1.Reservation) corev1.ResourceLi
 }
 
 func matchReservation(pod *corev1.Pod, rMeta *reservationInfo) bool {
-	return matchReservationOwners(pod, rMeta.Reservation) && matchReservationResources(pod, rMeta.Reservation, rMeta.Resources)
+	return matchReservationOwners(pod, rMeta.Reservation) && matchReservationResources(pod, rMeta.Reservation, rMeta.Resources) && matchReservationPort(pod, rMeta)
```
saintube, line 253: please add unit tests

KunWuLuan: done. 💯

eahydra (approved):
/lgtm
/approve

PR #912, opened 2022-12-30
koordlet: add more cli flags for audit log and ensure the audit log directory exist

LeoLiuYan:

1.  add more cli flags for audit log
2. ensure the audit log directory exist

NONE

### V. Checklist

### DIFF
--- a/pkg/koordlet/audit/event_logger.go
+++ b/pkg/koordlet/audit/event_logger.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sync"
@@ -29,7 +30,8 @@ import (
 )
 
 const (
-	MaxFileSize = 1 << 21 // 2MB
+	MaxFileSize                   = 1 << 21 // 2MB
+	DefaultLogDirMode fs.FileMode = 0755
 )
 
 // NewEventLogger create an EventWriter, it won't open the underly file until do a log call.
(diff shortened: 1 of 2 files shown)


pkg/koordlet/audit/event_logger.go
```diff
@@ -212,6 +227,10 @@ func listAuditfiles(dir string) ([]string, error) {
 	pattern := fmt.Sprintf("%s/audit-*.log", dir)
 	files, err := filepath.Glob(pattern)
 	if err == nil {
+		// reverse files
```
zwzhang0107, line 230: hi, what is the purpose to reverse these files

LeoLiuYan: While reading audit files, we expect the audit files order by `date` desc, and the `Glob function` will do the asc by default.   @zwzhang0107

LeoLiuYan: Another question, should we add a filter flag in `CLI` to discard any events?

LeoLiuYan:
@zwzhang0107 @jasonliu747 @saintube  Please review the code when you having any idle time.

pkg/koordlet/audit/event_logger.go
zwzhang0107: now i see the result is already desc, or do you want to get the event from old to new(asc)? maybe add a query param to control this like (desc=true by deafult/false)
image

LeoLiuYan: In one file the result is correct, and returned content by `desc` order. But, if we have N files? 

For example, we have those files in the audit log directory: `audit.log`、`audit-2023-01-04-14-28-01.log`、`audit-2023-01-04-13-28-01.log`、`audit-2023-01-04-12-28-01.log`.

 the `Glob function` will return by this order: [`audit-2023-01-04-12-28-01.log`, `audit-2023-01-04-13-28-01.log`, `audit-2023-01-04-14-28-01.log`], when read logger finished `audit.log`, we expect the next file is `audit-2023-01-04-14-28-01.log` which is the latest audit files expect `audit.log`.   @zwzhang0107

zwzhang0107: i get your idea, this looks great, good job!!

PR #918, opened 2023-01-03
koord-scheduler: remove allocated in filter hook

KunWuLuan:
Currently, the resource reservation plugin doesn't subtract allocated resources from the reservepod after allocating resources to the pod, which will lead to fewer available resources on the nodes.

For example, a reservation reserved 2c on node and the reservation allocated 1c to pod. In this case available resource on node is `allocated-2c` instead of `allocated-3c`.

### DIFF
--- a/pkg/scheduler/plugins/reservation/hook.go
+++ b/pkg/scheduler/plugins/reservation/hook.go
@@ -19,6 +19,7 @@ package reservation
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
@@ -108,20 +109,24 @@ func (h *Hook) FilterHook(handle frameworkext.ExtendedHandle, cycleState *framew
 
 	// only continue if matchedCache is prepared in PreFilterHook (i.e. find any reservation matched)
 	state := getPreFilterState(cycleState)
-	if state == nil || state.skip {
+	if state == nil {
 		return nil, nil, false
 	}
 
+	allocatedResource, ok := state.allocatedResources[node.Name]
 	// skip hook when no reservation matched on this node
 	rOnNode := state.matchedCache.GetOnNode(node.Name)
-	if len(rOnNode) <= 0 {
+	if len(rOnNode) <= 0 && !ok {
 		return nil, nil, false
+	}
+	if !ok {
+		allocatedResource = util.NewZeroResourceList()
 	}
 
 	klog.V(5).InfoS("FilterHook get reservation matched on node",
 		"pod", klog.KObj(pod), "node", node.Name, "count", len(rOnNode))
 	// fix-up reserved resources and ports
-	return pod, prepareFilterNodeInfo(pod, nodeInfo, rOnNode), true
+	return pod, prepareFilterNodeInfo(pod, nodeInfo, rOnNode, allocatedResource), true
 }
 
 func (h *Hook) prepareMatchReservationState(handle frameworkext.ExtendedHandle, pod *corev1.Pod) (*stateData, error) {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -108,20 +109,27 @@ func (h *Hook) FilterHook(handle frameworkext.ExtendedHandle, cycleState *framew
 
 	// only continue if matchedCache is prepared in PreFilterHook (i.e. find any reservation matched)
 	state := getPreFilterState(cycleState)
-	if state == nil || state.skip {
+	if state == nil {
 		return nil, nil, false
 	}
+	newNodeInfo := nodeInfo.Clone()
+	if allocatedResource, ok := state.allocatedResources[node.Name]; ok {
+		newNodeInfo.Requested.Add(quotav1.Subtract(util.NewZeroResourceList(), allocatedResource))
```
saintube, line 117: Why is it necessary to add the allocated reserved resources? The former implementation already subtracts the node request with the pod request in `prepareFilterNodeInfo()`.

KunWuLuan: We only subtracts the resource when there are some reservations matched on node.
When the pod matches no reservation on node, and there are reservations which have allocated resources for pods, available resources will be wrong.

saintube: OK, got it. In summary, we should subtract the allocated reserved resources (or the requests of scheduled pods allocating reserved resources) when there is no reservation matching the pod, and subtract the total reserved when there is any reservation matching the pod. How about doing the subtraction in `prepareFilterNodeInfo()` if we want to change the NodeInfo?

KunWuLuan: Unfortunately, if there is no reservation matching the pod, Skip attribute in prefilterState will be true, and then FilterHook will skip the prepareFilterNodeInfo().
That''s why I put this code here.

eahydra:
@KunWuLuan thanks for your contribution. please add more unit tests for the cases.

pkg/scheduler/plugins/reservation/hook.go
saintube: We can simply change the order between `prepareFilterNodeInfo` and the check of `preFilterState.skip`.

pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -235,12 +251,15 @@ func preparePreFilterPod(pod *corev1.Pod) *corev1.Pod {
 	return rPod
 }
 
-func prepareFilterNodeInfo(pod *corev1.Pod, nodeInfo *framework.NodeInfo, rOnNode []*reservationInfo) *framework.NodeInfo {
+func prepareFilterNodeInfo(pod *corev1.Pod, nodeInfo *framework.NodeInfo, rOnNode []*reservationInfo, allocatedResources *corev1.ResourceList) *framework.NodeInfo {
```
eahydra, line 254: Can we directly use the type `corev1.ResourceList` for allocatedResources?
```go
if len(allocatedResources) > 0 {
  		newNodeInfo.Requested.Add(quotav1.Subtract(util.NewZeroResourceList(), *allocatedResources))
}
```

pkg/scheduler/plugins/reservation/hook.go
```diff
@@ -235,12 +251,15 @@ func preparePreFilterPod(pod *corev1.Pod) *corev1.Pod {
...
 }
 
-func prepareFilterNodeInfo(pod *corev1.Pod, nodeInfo *framework.NodeInfo, rOnNode []*reservationInfo) *framework.NodeInfo {
+func prepareFilterNodeInfo(pod *corev1.Pod, nodeInfo *framework.NodeInfo, rOnNode []*reservationInfo, allocatedResources corev1.ResourceList) *framework.NodeInfo {
 	newNodeInfo := nodeInfo.Clone()
 	// 1. ignore current pod requests by reducing node requests
 	//    newNode.requests = node.requests - pod.requests
 	podRequests, _ := resourceapi.PodRequestsAndLimits(pod)
 	newNodeInfo.Requested.Add(quotav1.Subtract(util.NewZeroResourceList(), podRequests))
+	if len(allocatedResources) > 0 {
```
saintube, line 260: can we omit the `if`

PR #920, opened 2023-01-03
koord-scheduler: protect reserve pod from being preempted

KunWuLuan:
Now reserve pod can be preempt by real pods, but these reserve pods won't be deleted by preemptor because they don't exist on apiserver. It will cause problem when a real pod try to preempt a reserve pod, the real pod will never find the reserve pod being terminated.

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	listercorev1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	pluginhelper "k8s.io/kubernetes/pkg/scheduler/framework/plugins/helper"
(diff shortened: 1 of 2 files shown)


eahydra:
@saintube PTAL

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -236,6 +236,24 @@ func (p *Plugin) PostFilter(ctx context.Context, state *framework.CycleState, po
...
+		return nil, framework.NewStatus(framework.Error)
+	}
+
+	p.parallelizeUntil(ctx, len(nodeinfos), func(piece int) {
+		nodeInfo := nodeinfos[piece]
+		for _, pi := range nodeInfo.Pods {
+			if util.IsReservePod(pi.Pod) {
+				maxPri := int32(math.MaxInt32)
+				newReservePod := pi.Pod.DeepCopy()
+				newReservePod.Spec.Priority = &maxPri
```
eahydra, line 251: Can only be processed when the reservePod's priority is lower than the Pod's priority.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -236,6 +237,24 @@ func (p *Plugin) PostFilter(ctx context.Context, state *framework.CycleState, po
...
 	}
+
+	nodeinfos, err := p.handle.SnapshotSharedLister().NodeInfos().List()
+	if err != nil {
+		return nil, framework.NewStatus(framework.Error)
+	}
+
+	p.parallelizeUntil(ctx, len(nodeinfos), func(piece int) {
+		nodeInfo := nodeinfos[piece]
+		for _, pi := range nodeInfo.Pods {
```
saintube, line 248: how about finding the reservations via `(*AvailableCache).GetOnNode()` rather than visit all pods on the node

KunWuLuan: good idea, thanks!

saintube (commented):
/lgtm

eahydra (approved):
/lgtm
/approve

PR #924, opened 2023-01-04
koordlet: add more metrics for cpu throttled and node used memory

LeoLiuYan:

1. Add Prometheus metrics for node's and pod's cpu throttled ratio, including metrics below:

```
koordlet_pod_cpu_throttled_ratio{}
koordlet_container_cpu_throttled_ratio{}
```

2. Add Node memory used metrics, including metrics below:

```
koordlet_node_used_memory_bytes{}
```

NONE

### V. Checklist


jasonliu747:
@LeoLiuYan you could split this PR into two. One for adding metrics and the other for ConfigMap example.

LeoLiuYan:
> @LeoLiuYan you could split this PR into two. One for adding metrics and the other for ConfigMap example.

OK.

pkg/koordlet/metrics/common.go
```diff
@@ -44,11 +44,18 @@ var (
 		Help:      "Number of cpu cores used by node in realtime",
 	}, []string{NodeKey})
 
+	NodeUsedMemory = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "node_used_memory_bytes",
```
zwzhang0107, line 49: I think these metrics have already exported by cadvisor, how about reuse this?
https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

LeoLiuYan: You are right. But should we  export the `node_used_memory_bytes`  metric?  @zwzhang0107

LeoLiuYan: The way of koordlet calculates `node_used_memory` is: `memInfo.MemTotal - memInfo.MemAvailable`, `memInfo` is read from `/proc/meminfo`, it's different with `cadvisor`.  And the `node_used_memory` metric is used by `memory evict` feature.

That's a good idea to record `node_used_memory ` metric with `QoS` label. I will write code for this idea.  @zwzhang0107

PR #925, opened 2023-01-04
koord-manager: Fix log print fmt and add an example for slo-controller-config

LeoLiuYan:

As the title says.

NONE.

### V. Checklist

### DIFF
--- a/pkg/slo-controller/config/constants.go
+++ b/pkg/slo-controller/config/constants.go
@@ -33,5 +33,187 @@ Koordinator uses configmap to manage the configuration of SLO, the configmap is 
    - <CPUBurstConfigKey>
 
 et.
-  TODO add a sample here
+
+For example, the configmap is as follows:
+
+```
+apiVersion: v1
+data:
+  colocation-config: |
+    {
+      "enable": false,
+      "metricAggregateDurationSeconds": 300,
+      "metricReportIntervalSeconds": 60,
+      "metricAggregatePolicy": {
+        "durations": [
+          {
+            "duration": "5m"
+          },
+          {
+            "duration": "10m"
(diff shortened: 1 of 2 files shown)


jasonliu747 (commented):
/lgtm

pkg/slo-controller/config/constants.go
```diff
@@ -33,5 +33,94 @@ Koordinator uses configmap to manage the configuration of SLO, the configmap is
...
+          }
+        }
+      ]
+    }
+  resource-threshold-config: |
+    {
+      "clusterStrategy": {
+        "enable": true,
+        "cpuSuppressThresholdPercent": 75,
+        "cpuSuppressPolicy": "cpuBurst",
```
saintube, line 99: the example is recommended to be valid. here the value of `cpuSuppressPolicy` may be `cpuset`

pkg/slo-controller/config/constants.go
```diff
@@ -33,5 +33,94 @@ Koordinator uses configmap to manage the configuration of SLO, the configmap is
...
+        {
+          "nodeSelector": {
+            "matchLabels": {
+              "kubernetes.io/kernel": "alios"
+            }
+          },
+          "lsClass": {
+            "memoryQOS": {
+              "enable": true,
+              "wmarkRatio": 50,
```
saintube, line 80: Actually, this wmarkRatio is too small for LS pods. How about using some of the recommended values defined in `DefaultResourceQOSStrategy()`

LeoLiuYan: Review again.  @saintube

jasonliu747 (approved):

PR #936, opened 2023-01-06
koord-scheduler:  add new method GetPodIsAssigned to QuotaInfo

kingeasternsun:
Signed-off-by: kingeasternsun <kingeasternsun@gmail.com>

when we get pods that is assigned from quota, we could just return them directly with lock once , instead of lock twice that with first lock we get all pods, with second lock we judge each pod and collect assigned pods.

no 
### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #941, opened 2023-01-09
koordlet: add event cache for audit log

LeoLiuYan:

As the title says.

NONE

### V. Checklist

### DIFF
--- a/pkg/koordlet/audit/auditor.go
+++ b/pkg/koordlet/audit/auditor.go
@@ -36,7 +36,7 @@ var (
 )
 
 func NewAuditor(c *Config) Auditor {
-	logWriter := NewFluentEventLogger(c.LogDir, c.MaxDiskSpaceMB, c.Verbose)
+	logWriter := NewFluentEventLogger(c.LogDir, c.MaxDiskSpaceMB, c.Verbose, c.EventCacheExpired)
 	logReader := NewEventReader(c.LogDir)
 	return &auditor{
 		config:        c,
@@ -228,6 +228,9 @@ func (a *auditor) HttpHandler() func(http.ResponseWriter, *http.Request) {
 }
 
 func (a *auditor) Run(stopCh <-chan struct{}) error {
+	// start the logWriter cache
+	a.logWriter.RunInit(stopCh)
+
 	timer := time.NewTicker(a.config.TickerDuration)
 	defer timer.Stop()
 	for {
@@ -275,6 +278,8 @@ func (e *emptyEventFluentWriter) Close() error {
 	return nil
 }
 
+func (e *emptyEventFluentWriter) RunInit(<-chan struct{}) {}
+
 type emptyEventWriter struct {
 }
 
@@ -289,6 +294,8 @@ func (e *emptyEventWriter) Flush() error {
 func (e *emptyEventWriter) Close() error {
 	return nil
 }
+
+func (e *emptyEventWriter) RunInit(<-chan struct{}) {}
 
 // SetupDefaultAuditor initialize the `Default` auditor.
 func SetupDefaultAuditor(c *Config, stopCh <-chan struct{}) {
--- a/pkg/koordlet/audit/config.go
+++ b/pkg/koordlet/audit/config.go
@@ -30,6 +30,7 @@ type Config struct {
 	DefaultEventsLimit   int
 	MaxEventsLimit       int
 	TickerDuration       time.Duration
+	EventCacheExpired    time.Duration
 }
 
 func NewDefaultConfig() *Config {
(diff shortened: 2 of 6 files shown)


jasonliu747:
What's the background for this feature?

LeoLiuYan:
> What's the background for this feature?

There are a lot of redundant events in `audit.log`, it's confusion to users which want to select the valuable events.  
So it's better to add event cache to reduce the repetitive events.

@jasonliu747

hormes:
The write operations of cgroups have been cached, maybe we can adjust the point of writing the log to solve the problem of repeated events?

LeoLiuYan:
> The write operations of cgroups have been cached, maybe we can adjust the point of writing the log to solve the problem of repeated events?

The `resourceexecutor` package which responsibility for write cgroups has cached part of audit log, that's true. But it may be not enough to reduce repeated events because of logging the audit operation have scattered elsewhere, for example:  https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/runtimehooks/protocol/container_context.go#L177.

Another viewpoint,  the cache of audit log may have longer gc time than `resourceexecutor` package.

@hormes

LeoLiuYan:
So, should we add cache capability in audit package?   @jasonliu747 @hormes

hormes:
> So, should we add cache capability in audit package? @jasonliu747 @hormes

The cache may be a little heavy, is there a lighter way?

zwzhang0107:
new module `koordlet/resourceexecutor` has been added, which has supported for cacheable read/write for cgroup file,  how about using these cacheable method and add audit log. This may solve the redundant logs. @LeoLiuYan

LeoLiuYan:
> > So, should we add cache capability in audit package? @jasonliu747 @hormes
> 
> The cache may be a little heavy, is there a lighter way?



> > So, should we add cache capability in audit package? @jasonliu747 @hormes
> 
> The cache may be a little heavy, is there a lighter way?

May be the cache is already lightweight?  Do we have an easy way?

LeoLiuYan:
> > The write operations of cgroups have been cached, maybe we can adjust the point of writing the log to solve the problem of repeated events?
> 
> The `resourceexecutor` package which responsibility for write cgroups has cached part of audit log, that's true. But it may be not enough to reduce repeated events because of logging the audit operator have scattered elsewhere, for example: https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/runtimehooks/protocol/container_context.go#L177.
> 
> Another viewpoint, the cache of audit log may have longer gc time than in `resourceexecutor` package.
> 
> @hormes

/cc @zwzhang0107

hormes:
Let's sort it out first, the current auditor is used to audit key resource changes, and it is quite dependent on the order of changes. Therefore, let's first look at which changes are frequent and ineffective, and then discuss feasible solutions together. Using event cache directly to deduplicate may violate the original intention. Of course, this whole thing can be discussed and changed.

LeoLiuYan:
I found out the main places where logging the audit events by searching `audit.V` keyword.

- package `koordlet/qosmanager`

    -  `cpu_suppress.go:359`    
        The command is: `audit.V(1).Node().Reason(resourceexecutor.AdjustBEByNodeCPUUsage).Message("update BE group to cpuset: %v", beCPUSet).Do()`.
        
        It will log events in fixed interval.

   
- package `koordlet.runtimehooks`

  -  `hooks/groupidentity/rule.go:138`
       The command is: `audit.V(2).Group(string(kubeQOS)).Reason(name).Message("set bvt to %v", bvtValue)`

      It will log events many times in one reconcile.

zwzhang0107:
i recommend that let `resource executor` do the audit log, since it is already cachable.
so maybe there are two jobs:
1. use `resource executor` for all cgroups file writes (most are already done)
2. mv audit log to `resource executor`, e.g. ResourceUpdateExecutorImpl.updateByCache

zwzhang0107:
> i recommend that let `resource executor` do the audit log, since it is already cachable. so maybe there are two jobs:
> 
> 1. use `resource executor` for all cgroups file writes (most are already done)
> 2. mv audit log to `resource executor`, e.g. ResourceUpdateExecutorImpl.updateByCache

what do you think about this, we can start some new issues to follow. @LeoLiuYan

LeoLiuYan:
I agree with you.    @zwzhang0107 
What do you think ? @hormes

Re-Grh:
new issues here

zwzhang0107:
since the new issue has been created, we will close this issue in following days

zwzhang0107:
now you can follow this PR #1055 @LeoLiuYan

PR #943, opened 2023-01-09
koord-scheduler: move set parent into addChildGroupQuotaInfo

kingeasternsun:
Signed-off-by: kingeasternsun <kingeasternsun@gmail.com>

When adding new child node to parent node in tree data structure, It would be better to put the code that set child's parent and the code set parent's child together.

none

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/core/quota_info.go
+++ b/pkg/scheduler/plugins/elasticquota/core/quota_info.go
@@ -383,6 +383,7 @@ func NewQuotaTopoNode(quotaInfo *QuotaInfo) *QuotaTopoNode {
 }
 
 func (qtn *QuotaTopoNode) addChildGroupQuotaInfo(childNode *QuotaTopoNode) {
+	childNode.parQuotaTopoNode = qtn
 	qtn.childGroupQuotaInfos[childNode.name] = childNode
 }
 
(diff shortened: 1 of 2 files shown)


kingeasternsun:
Thanks for your reviews @eahydra

eahydra:
> Thanks for your reviews @eahydra

Thanks for your contribution @kingeasternsun.
If you are interested in this work, you can take a look at the related implementation as a whole, and you can make some optimizations for the implementation.

pkg/scheduler/plugins/elasticquota/core/quota_info.go
```diff
@@ -383,6 +383,7 @@ func NewQuotaTopoNode(quotaInfo *QuotaInfo) *QuotaTopoNode {
 }
 
 func (qtn *QuotaTopoNode) addChildGroupQuotaInfo(childNode *QuotaTopoNode) {
+	childNode.parQuotaTopoNode = qtn
```
eahydra, line 386: From the name of this function, the original implementation is OK. On the whole, this function is not very necessary.

kingeasternsun:
> > Thanks for your reviews @eahydra
> 
> Thanks for your contribution @kingeasternsun. If you are interested in this work, you can take a look at the related implementation as a whole, and you can make some optimizations for the implementation.

Thanks for your suggestion, I will reconsider it.

PR #950, opened 2023-01-11
koord-descheduler: limits frequently migrated workloads

eahydra:

The migrationJob controller in koord-descheduler supports that limits frequently migrated workloads
- Add configuration field `objectLimiters` to support control the limit duration and thresholds.  
- Now just limits workload. By default if one workload's migrating Pods exceed the `maxMigrating` in the duration , the workload will be limited. By default the duration is `5 minutes`, `maxMigrating` is nil that means will use `maxMigratingPerWorkload` instead.

### V. Checklist


pkg/descheduler/apis/config/types_pluginargs.go
```diff
@@ -145,3 +149,19 @@ type MigrationControllerArgs struct {
...
+
+const (
+	MigrationFrozenWorkload MigrationFrozenObjectType = "workload"
+)
+
+type FrozenGateMap map[MigrationFrozenObjectType]MigrationFrozenGate
+
+// MigrationFrozenGate means that if the specified dimension has multiple migrations within the configured time period
+// and exceeds the configured threshold, it will be frozen for a certain period of time.
+type MigrationFrozenGate struct {
```
zwzhang0107, line 163: how about define as
``` go
type MigrationFrozenGate struct {
  Duration       metav1.Duration
  MaxMigrating   int
}
```
which means max migrate M pods in D duration (just like QPS=M/D, Burst=0)
what do you think @FillZpp @hormes @jasonliu747

eahydra: This configuration is indeed OK. The existing implementation is relatively simple, and a workload can be automatically processed through the cache mechanism, without an additional GC mechanism. If the ratelimit method is used, it means maintaining a ratelimit instance for each workload and specifying an additional cycle time to execute GC logic.

pkg/descheduler/apis/config/types_pluginargs.go
```diff
@@ -145,3 +149,19 @@ type MigrationControllerArgs struct {
...
+	Duration       metav1.Duration
+	MaxMigrating   int
```
hormes, line 165: How is the duration refreshed? After the previous duration is over, start a new round of counting?

eahydra: Yes. If there is no new eviction operation within the period specified by the duration, a new round will be started after the next eviction is initiated, and the number of evictions will be recorded; otherwise, the time will be extended after the last eviction operation. Similar to a window mechanism, the time window will be adjusted passively.

pkg/descheduler/apis/config/types_pluginargs.go
eahydra: updated with rate limiter.

pkg/descheduler/apis/config/types_pluginargs.go
eahydra: updated.

PR #952, opened 2023-01-12
koordlet: fix batch resource plugin for empty label pods

saintube:

Fix the batch resource plugin when the BE pod has no label or annotation.

Also, fix some log format errors.

Currently, the batch resource plugin aborts set cgroup resources when a BE pod has no label or annotation since it checks if both of them are non-nil.

### V. Checklist


jasonliu747 (commented):
/lgtm

PR #974, opened 2023-02-01
runtimeproxy: fix trap of nil interface{}

yuexian1234:
Signed-off-by: yuexian <694547990@qq.com>

### Ⅰ. fix bug,nil interface

 ref  https://cloud.tencent.com/developer/article/1911289

No

1. stop koordlet,before fix the  bug,it will be panic. because the hookResp which return from Dispatch is not nil,although the real value is nil
2. the return type is interface{}   https://github.com/koordinator-sh/koordinator/blob/ced52523f941978835950bf47b00062434b417d8/pkg/runtimeproxy/dispatcher/dispatcher.go#L68-L69
3. while rsp is nil, but if it return as interface{}, it will not be equal with nil  https://github.com/koordinator-sh/koordinator/blob/ced52523f941978835950bf47b00062434b417d8/pkg/runtimeproxy/dispatcher/dispatcher.go#L88-L89
4. after fix,it will be work fine

#978

### DIFF
--- a/pkg/runtimeproxy/dispatcher/dispatcher.go
+++ b/pkg/runtimeproxy/dispatcher/dispatcher.go
@@ -86,6 +86,9 @@ func (rd *RuntimeHookDispatcher) Dispatch(ctx context.Context, runtimeRequestPat
 			// currently, only one hook be called during one runtime
 			// TODO: multi hook server to merge response
 			rsp, err := rd.dispatchInternal(ctx, hookType, client, request)
+			if err != nil {
+				return nil, err, hookServer.FailurePolicy
+			}
 			return rsp, err, hookServer.FailurePolicy
 		}
 	}
--- a/pkg/runtimeproxy/dispatcher/dispatcher_test.go
+++ b/pkg/runtimeproxy/dispatcher/dispatcher_test.go
@@ -121,8 +121,12 @@ func TestRuntimeHookDispatcher_Dispatch(t *testing.T) {
 			hookManager: configManager,
 			cm:          clientManager,
 		}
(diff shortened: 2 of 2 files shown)


jasonliu747 (commented):
Hi, thanks for your contribution. The code is not properly formatted, you may follow the instruction below to fix it.
image

saintube:
@yuexian1234 hi, could you please add the unit test to verify the fix?

yuexian1234:
> @yuexian1234 hi, could you please add a unit test to verify the fix?

of course,but i think it is a obvious go error trap about return nil object as interface {}.
this is some blog about this "https://cloud.tencent.com/developer/article/1911289"

saintube:
> > @yuexian1234 hi, could you please add a unit test to verify the fix?
> 
> of course,but i think it is a obvious go error trap about return nil object as interface {}. this is some blog about this "https://cloud.tencent.com/developer/article/1911289"

Yes, there is clearly a trap of nil interface. It will be appreciated if you also provide a test case that can verify the bug is fixed and does not reproduce in the next versions.

yuexian1234:
> > > @yuexian1234 hi, could you please add a unit test to verify the fix?
> > 
> > 
> > of course,but i think it is a obvious go error trap about return nil object as interface {}. this is some blog about this "https://cloud.tencent.com/developer/article/1911289"
> 
> Yes, there is clearly a trap of nil interface{}. It will be helpful if you also provide a test case that can verify the bug is fixed and does not reproduce in the next versions :).

ok

jasonliu747 (commented):
/lgtm

PR #976, opened 2023-02-01
koord-descheduler: set default values for evictQPS/evictBurst

eahydra:

1. Set evictQPS of MigrationControllerArgs to 10
2. Set evictBurst of MigrationControllerArgs to 1
3. evictQPS support string or float64 type and users expect to write numbers directly.
4. Set MaxMigratingPerNode to 2

### V. Checklist


pkg/descheduler/apis/config/v1alpha2/types_pluginargs.go
```diff
@@ -143,9 +145,9 @@ type MigrationControllerArgs struct {
 	DefaultJobTTL *metav1.Duration `json:"defaultJobTTL,omitempty"`
 
 	// EvictQPS controls the number of evict per second
-	EvictQPS string `json:"evictQPS,omitempty"`
+	EvictQPS *config.Float64OrString `json:"evictQPS,omitempty"`
```
jasonliu747, line 148: tha's very thoughtful, but is it really necessary?

eahydra: Yes..Users naturally think that the evictQPS field needs to be filled with numbers and needs to be of floating point type. In order to be compatible with stock scenarios, it is now necessary to support both floating point and string types

jasonliu747: I see, that makes sense to me.

FillZpp (approved):
/lgtm

PR #980, opened 2023-02-02
koordlet:fix a bug in topologyinformer when start-up

Re-Grh:
Signed-off-by: Mew2 <1271013391@qq.com>

fix a small bug

### V. Checklist


pkg/koordlet/statesinformer/states_noderesourcetopology.go
```diff
@@ -135,9 +134,7 @@ func (s *nodeTopoInformer) Start(stopCh <-chan struct{}) {
 }
 
 func (s *nodeTopoInformer) HasSynced() bool {
-	if !s.initialTopology {
-		return false
-	}
+	klog.V(4).Infof("nodeTopoInformer need collector to start")
```
saintube, line 137: how about adding a comment instead of a log

zwzhang0107: V5 xxx ready to start
// TODO only node cpu info collector relies on node topo informer

PR #981, opened 2023-02-02
koord-descheduler: use DefaultInsecureDeschedulerPort(10251) as healthz/metrics default port

eahydra:

koord-descheduler needs to use DefaultInsecureDeschedulerPort(10251)  as the default port of healthz/metrics, but the current implementation uses DefaultDeschedulerPort(10258) , which causes koord-descheduler to fail to start when the bind address of healthz/metrics is not configured.

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #987, opened 2023-02-05
koord-descheduler: LowNodeLoad supports anomaly detector

eahydra:

One of the commonly encountered scenarios: the load of the node suddenly rises, but then drops back quickly. If LowNodeLoad happens to be executing at this time, an unexpected eviction action is performed. 

Therefore, in order to avoid such sudden scenarios, LowNodeLoad introduces an anomaly detector to help determine whether a node is really overutilized. The default detection is that if the load of a node is higher than HighThresholds for five consecutive times, then the node is considered really overutilized.

Users can configure the detection thresholds via `LowNodeArgs.AnomalyThresholds`. 

1.  create stress pods
2.  run the latest descheduler with LowNodeLoad plugin.
3.  watch the logs, you can see the logs: 
```log
I0206 00:52:33.366587   14886 low_node_load.go:184] "None of the nodes were detected as anomalous, nothing to do here"
```
4.  After run LowNodeLoad.Balance five times, the stress pods on the nodes that are overutilized will be evicted.

### V. Checklist


pkg/descheduler/apis/config/types_loadaware.go
```diff
@@ -64,6 +64,11 @@ type LowNodeLoadArgs struct {
 
 	// LowThresholds defines the low usage threshold of resources
 	LowThresholds ResourceThresholds
+
+	// AnomalyThresholds indicates the node load anomaly thresholds,
+	// the default is 5 consecutive times exceeding HighThresholds,
+	// it is determined that the node is abnormal, and the Pod needs to be migrated to reduce the load.
+	AnomalyThresholds *LoadAnomalyThresholds
```
saintube, line 71: It seems to be a smooth threshold for node overutilization. Perhaps LoadHighThresholds is more concise?

eahydra: Already have a field named `HighThresholds`. Maybe named `LoadSmoothHighThresolds`? 😂

saintube: Can we combine the fields of the high threshold? Or name it like `LoadHighConditions`...

jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

PR #988, opened 2023-02-06
koordlet: filter out Pod not run by runc to support multiple runtime

jasonliu747:

### V. Checklist

### DIFF
--- a/pkg/koordlet/metricsadvisor/collector.go
+++ b/pkg/koordlet/metricsadvisor/collector.go
@@ -227,6 +227,11 @@ func (c *collector) collectPodResUsed() {
 	podMetas := c.statesInformer.GetAllPods()
 	for _, meta := range podMetas {
 		pod := meta.Pod
+		// TODO: currently koordlet only supports collecting metrics from Pod launched by runc.
+		// Other runtimes collector plugins are currently under development for better compatibility.
+		if ignoreNonRuncPod(pod) {
+			continue
+		}
 		uid := string(pod.UID) // types.UID
 		collectTime := time.Now()
 		podCgroupDir := koordletutil.GetPodCgroupDirWithKube(meta.CgroupDir)
--- a/pkg/koordlet/metricsadvisor/utils.go
+++ b/pkg/koordlet/metricsadvisor/utils.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


pkg/koordlet/util/runtimeclass.go
```diff
@@ -0,0 +1,35 @@
...
+// NOTE: functions in this file can be overwritten for extension
+
+// GetRuntimeClassName returns RuntimeClassName which refers to a RuntimeClass object in
+// the node.k8s.io group, which should be used to run this pod.
+func GetRuntimeClassName(pod *corev1.Pod) *string {
+	return pod.Spec.RuntimeClassName
+}
+
+// ShouldIgnorePod will return true if Pod uses RuntimeClass other than runc.
+func ShouldIgnorePod(pod *corev1.Pod) bool {
```
zwzhang0107, line 32: since this only works for `metric advisor`, how about mv inside as private(e.g. ignorePodCollection)

jasonliu747: done

pkg/koordlet/metricsadvisor/utils.go
```diff
@@ -0,0 +1,19 @@
...
+
+// getRuntimeClassName returns RuntimeClassName which refers to a RuntimeClass object in
+// the node.k8s.io group, which should be used to run this pod.
+func getRuntimeClassName(pod *corev1.Pod) *string {
+	return pod.Spec.RuntimeClassName
+}
+
+// shouldIgnorePod will return true if Pod uses RuntimeClass other than runc.
+func shouldIgnorePod(pod *corev1.Pod) bool {
+	runtimeClass := getRuntimeClassName(pod)
```
zwzhang0107, line 17: return true directly?

pkg/koordlet/metricsadvisor/collector.go
```diff
@@ -227,6 +227,10 @@ func (c *collector) collectPodResUsed() {
 	podMetas := c.statesInformer.GetAllPods()
 	for _, meta := range podMetas {
 		pod := meta.Pod
+		// koordlet only supports standard OCI-runtime, such as runc.
+		if shouldIgnorePod(pod) {
```
saintube, line 231: We may check the pod meta (e.g. pod runtimeclass, pod annotations) for the corresponding collector plugin, but not ignore all non-runc pods.

e.g. By default, we keep the collection for runc pods. For a kata-runtime pod, we collect its pod metric via a collector plugin of kata-runtime.

jasonliu747: I got your point. 
Although we ignore all non-runc pods here, users are still free to create their own plugins to collect Pods run by other runtimes.

zwzhang0107:
we also need a TODO for "making each collector as an independent plugin in metricsAdvisor for better extension", just like those in statesInformers.

pkg/koordlet/metricsadvisor/utils.go
jasonliu747: Even though this function might be overwritten, I think we should still keep this simple check. WDYT?

jasonliu747:
> we also need a TODO for "making each collector as an independent plugin in metricsAdvisor for better extension", just like those in statesInformers.

Totally agree with you!

PR #996, opened 2023-02-08
koord-scheduler: set Reservation indexer before informer start to avoid error

tzzcfrank:

If there is a new plugin that also starts the reservation informer during initialization, it will cause the reservation plugin to fail to register the indexer internally. The reason is that the k8s scheduling framework does not guarantee that the plugins are initialized in the order of configuration, even if the reservation plugin is configured first.

### V. Checklist

### DIFF
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -338,11 +338,14 @@ func Setup(ctx context.Context, opts *options.Options, schedulingHooks []framewo
 
 	// NOTE(joseph): K8s scheduling framework does not provide extension point for initialization.
 	// Currently, only by copying the initialization code and implementing custom initialization.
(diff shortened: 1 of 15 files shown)


cmd/koord-scheduler/app/server.go
```diff
@@ -351,6 +353,16 @@ func Setup(ctx context.Context, opts *options.Options, schedulingHooks []framewo
 		}
 	}
 
+	reservationInterface := cc.KoordinatorSharedInformerFactory.Scheduling().V1alpha1().Reservations()
```
eahydra, line 356: recommend move the indexer register codes to frameworkext.NewExtendedHandle.

eahydra (approved):
/lgtm
/approve

FillZpp (approved):
Thanks!

PR #1007, opened 2023-02-09
koord-scheduler: compatible with lightweight-coscheduling

eahydra:

Some users are still using [lightweight-coscheduling](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/2-lightweight-coscheduling/README.md), so require the coscheduling plugin to be compatible with lightweight-coscheduling. PodGroups are not created when the Pod uses the lightweight-coscheduling API.

### V. Checklist


FillZpp (approved):
/lgtm

Issue #1011, opened 2023-02-13
[proposal] add name to slo-controller-config cfg->NodeStrategies as selector alias


chzhj:
**What is your proposal**:

![image](https://user-images.githubusercontent.com/19923484/218403325-680c9042-4211-4c56-814a-1740edc51493.png)
add name to slo-controller-config cfg->NodeStrategies as selector alias

![image](https://user-images.githubusercontent.com/19923484/218403167-bb79a0d6-6dd5-481f-939a-0cfc2db87902.png)

![image](https://user-images.githubusercontent.com/19923484/218403105-57c88e62-a9ac-4b8e-aed7-8174b1408dfb.png)

**Why is this needed**:
It's difficult to build a console to manager this config，if have not name(or id) for this nodeStrategies

**Is there a suggested solution, if so, please add it**:

PR #1020, opened 2023-02-16
add name to slo-controller-config cfg->NodeStrategies as selector alias

chzhj:
add name to slo-controller-config cfg->NodeStrategies as selector alias
see details https://github.com/koordinator-sh/koordinator/issues/1011

### V. Checklist


apis/extension/slo_controller_config.go
```diff
@@ -34,6 +34,13 @@ const (
 	SystemConfigKey            = "system-config"
 )
 
+// +k8s:deepcopy-gen=true
+type NodeCfgProfile struct {
```
zwzhang0107, line 38: how about add some ut as yaml format for compatibility checking, to make sure that old format configmap yaml can be parsed to new struct.

apis/extension/slo_controller_config.go
```diff
@@ -34,6 +34,13 @@ const (
...
+	Name string `json:"name,omitempty"`
```
saintube, line 39: Perhaps we need to print the name of the node config for the config merged failed. Otherwise, the field makes sense only when shown to end users.

apis/extension/slo_controller_config.go
chzhj: ut of colocation_cm_event_handler_test have case for old config json to new struct

apis/extension/slo_controller_config.go
saintube: we'd better add some comments about preserving this field

FillZpp (approved):
/lgtm

PR #1025, opened 2023-02-20
apis: deprecated device resource names with prefix kuberenetes.io

eahydra:

The resource names starting with the prefix `kubernetes.io` in k8s are native resource types, and device plugins cannot use these resource names, otherwise the kubelet will reject them.

### V. Checklist


eahydra:
/hold cancel
TBR

jasonliu747 (approved):
/lgtm

Issue #1022, opened 2023-02-17
[proposal] Pod Ready is considered to be migration complete


eahydra:
**What is your proposal**:

The migrationJob in koord-descheduler uses Reservation to help the Pod migrate, and the migration is considered to be complete, but it is possible that the newly created Pod after the eviction is not yet fully Ready. Although there is already a protection mechanism for the workload dimension, it would be better if it can support checking the Pod Ready status to determine whether it is completed.

**Why is this needed**:

Improve the migration process to better protect the stability of workload.

**Is there a suggested solution, if so, please add it**:

PR #1026, opened 2023-02-20
koord-descheduler: check pod ready when migrating

lucming:
fix: https://github.com/koordinator-sh/koordinator/issues/1022

### V. Checklist

### DIFF
--- a/pkg/descheduler/controllers/migration/controller.go
+++ b/pkg/descheduler/controllers/migration/controller.go
@@ -489,6 +489,19 @@ func (r *Reconciler) doMigrate(ctx context.Context, job *sev1alpha1.PodMigration
 
 	boundPod := reservationObj.GetBoundPod()
 	podNamespacedName := types.NamespacedName{Namespace: boundPod.Namespace, Name: boundPod.Name}
+	podReady, result, err := r.waitForPodReady(ctx, job, podNamespacedName)
+	if err != nil {
+		return result, err
+	} else if !podReady {
+		return result, nil
+	}
+
+	msg := fmt.Sprintf("Bind pod %q is ready", podNamespacedName)
+	r.eventRecorder.Eventf(job, nil, corev1.EventTypeNormal, "BoundPodReady", "Migrating", msg)
+	if err := r.handleBoundPodReadySuccess(ctx, job); err != nil {
+		return reconcile.Result{}, err
+	}
+
 	job.Status.PodRef = boundPod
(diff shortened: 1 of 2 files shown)


pkg/descheduler/controllers/migration/evictor/interpreter.go
```diff
@@ -100,13 +100,10 @@ func (p *interpreterImpl) Evict(ctx context.Context, job *sev1alpha1.PodMigratio
 		evictionPolicy = getCustomEvictionPolicy(job.Labels)
 	}
 
-	var evictor Interface
+	evictor := p.defaultEvictor
 	if evictionPolicy != "" {
```
eahydra, line 104: At present, the `evictionPolicy` parameter is not checked. It is possible that the policy configured by the user is not supported yet, and it may be nil

pkg/descheduler/controllers/migration/controller.go
```diff
@@ -787,6 +791,22 @@ func (r *Reconciler) waitForPodBindReservation(ctx context.Context, job *sev1alp
...
+	klog.V(4).Infof("checks if the boundPod %q is ready or not", podNamespacedName)
+	podObj := &corev1.Pod{}
+	if err := r.Client.Get(ctx, podNamespacedName, podObj); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{RequeueAfter: defaultRequeueAfter}, fmt.Errorf("failed to get pod %q", podNamespacedName)
+	}
+	if isReady := k8spodutil.IsPodReady(podObj); !isReady {
+		return reconcile.Result{RequeueAfter: defaultRequeueAfter}, fmt.Errorf("boundPod: %q is not ready", podNamespacedName)
```
eahydra, line 804: Maybe we can add event for the MigrationJob indicates that it is waiting for Pod ready, otherwise the user may think that MigrationJob hangs because the Pod has not been ready for a long time. In addition, a type can also be added to the Job Status Condition to indicate whether the Pod is ready

pkg/descheduler/controllers/migration/evictor/interpreter.go
lucming: en, already rolled back.

pkg/descheduler/controllers/migration/controller.go
lucming: done

eahydra (approved):
/lgtm
/approve

jasonliu747 (commented):
/lgtm

PR #1027, opened 2023-02-20
e2e: add an e2e case BatchResource

saintube:

1. Add an e2e case BatchResource for checking the extended resources is correctly updated when colocation is enabled in the slo-controller-config.
2. Fix the typos of package names.

Just run E2E tests.

e.g.

```bash
ginkgo ./test/e2e -- --allowed-not-ready-nodes=5 --system-pods-startup-timeout=10s --e2e-verify-service-account=false
```

### V. Checklist

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -43,6 +43,7 @@ import (
 
 	// test sources
 	_ "github.com/koordinator-sh/koordinator/test/e2e/scheduling"
+	_ "github.com/koordinator-sh/koordinator/test/e2e/slocontroller"
 )
 
 // handleFlags sets up all flags and parses the command line.
--- a/test/e2e/framework/manifest/manifest.go
+++ b/test/e2e/framework/manifest/manifest.go
@@ -139,3 +139,21 @@ func DaemonSetFromData(data []byte) (*appsv1.DaemonSet, error) {
 	}
 	return &ds, nil
 }
+
+// ConfigMapFromManifest reads a .json/yaml file and returns the pod in it.
+func ConfigMapFromManifest(filename string) (*v1.ConfigMap, error) {
+	var configMap v1.ConfigMap
+	data, err := e2etestfiles.Read(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	json, err := utilyaml.ToJSON(data)
+	if err != nil {
+		return nil, err
+	}
+	if err := runtime.DecodeInto(scheme.Codecs.UniversalDecoder(), json, &configMap); err != nil {
+		return nil, err
+	}
+	return &configMap, nil
+}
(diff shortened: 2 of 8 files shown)


test/e2e/node/batchresource.go
```diff
@@ -0,0 +1,184 @@
...
+				for i := range nodeList.Items {
+					node := &nodeList.Items[i]
+					isAllocatable, msg := isNodeHasBatchResourcesAllocatable(node)
+					if isAllocatable {
+						allocatableCount++
+					} else {
+						framework.Logf("node %s has no batch resources allocatable, msg: %s", node.Name, msg)
+					}
+				}
+				if float64(allocatableCount) > float64(totalCount)*minNodesBatchResourceAllocatableRatio {
```
zwzhang0107, line 133: can we also check the exactly batch total resource by mocking NodeMetric CR of node?

test/e2e/node/batchresource.go
```diff
@@ -0,0 +1,184 @@
+/*
```
zwzhang0107, line 1: e2e/node -> e2e/slocontroller ?

eahydra:
NICE! E2E work has finally started!

test/e2e/node/OWNERS
```diff
@@ -0,0 +1,3 @@
+reviewers:
+  - koordlet-approvers
+  - runtime-proxy-approvers
```
eahydra, line 3: also need add `approvers` section.
```yaml
approvers:
   - koordlet-approvers
   - runtime-proxy-approvers
```

test/e2e/node/batchresource.go
saintube: Yes, but a little complicated.
By default, the test case runs when the koordinator is deployed in a general configuration. Before mocking a NodeMetric CR, we should disable the node metric reporting by mutating the spec of the koordlet DaemonSet.

saintube: We may not mock a NodeMetric CR. Instead, we get the latest NodeMetric CR and roughly verify the node allocatable batch resources.

test/e2e/node/batchresource.go
saintube: done

test/e2e/node/OWNERS
saintube: done. change the approvers as the koordlet's and the slo-controller's

test/e2e/node/batchresource.go
zwzhang0107: the later one sound good, since NodeMetric won't update frequently.

jasonliu747 (commented):
Nice! Should we add a workflow specifically for E2E test?

test/e2e/node/batchresource.go
saintube: Done. Now we estimate the lower bound of the batch resources using `batch.total >= node.total - node.total * cpuReclaimRatio - nodeMetric.usage - node.total * maxDiffRatio`.

saintube:
> Nice! Should we add a workflow specifically for E2E test?

Yes. Let's add the pipeline later. For now we can use the described command to run locally.

eahydra (approved):
/approve

Issue #911, opened 2022-12-30
[BUG] reportNodeTopology logs error messages when the initial metrics have not been collected


saintube:
**What happened**:

The newly-created koordlet pods sometimes log error messages about reportNodeTopology which cannot calculate a NodeResourceTopology since the nodeCPUInfo metric has not been collected.

```bash
I1230 10:32:22.155071 3724360 states_noderesourcetopologyinformer.go:305] failed to calculate node topology, err: failed to calculate cpu topology, err: failed to get node cpu info
```

It does not affect the koordlet's follow-up running and should be considered a false alarm.
It is mainly because the `(*nodeTopoInformer).reportNodeTopology` is not synced with the `(*collector).collectNodeCPUInfo` and the metricCache. The nodeTopoInformer may fetch the NodeCPUInfo before the first metric collection has been done.

**What you expected to happen**:

The reportNodeTopology's running should wait for the initial metric collection of the NodeCPUInfo (maybe with a timeout). Then there should be no error message when the koordlet pod starts normally.

**How to reproduce it (as minimally and precisely as possible)**:

1. Start the koordlet Daemonset.
2. Check some koordlet Pods' logs about `states_noderesourcetopologyinformer`.

**Anything else we need to know?**:

**Environment**:
- App version: 1.1.0
- Kubernetes version (use `kubectl version`): 1.24.6
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: systemd
- Others:

PR #1047, opened 2023-02-23
koordlet: ensure report nodetopo after nodecpuinfo have already collected

lucming:
fix:  https://github.com/koordinator-sh/koordinator/issues/911

### V. Checklist

### DIFF
--- a/pkg/koordlet/metriccache/metric_cache.go
+++ b/pkg/koordlet/metriccache/metric_cache.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	"gorm.io/gorm"
+
 	"k8s.io/apimachinery/pkg/api/resource"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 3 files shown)


pkg/koordlet/statesinformer/states_noderesourcetopology.go
```diff
@@ -353,6 +353,10 @@ func (s *nodeTopoInformer) reportNodeTopology() {
 		klog.V(5).Infof("feature %v not enabled, node topo will not be reported", features.NodeTopologyReport)
 	}
 
+	if !koordletutil.NodeCPUInfoInited {
+		return
+	}
+
 	nodeTopoAnnotations, err := s.calcNodeTopo()
```
zwzhang0107, line 360: how about we just ignore the not exist error from metric cache, so that we don't need to add an atomic flag in util(actually that's not the duty of util pkg).

lucming: well

pkg/koordlet/metriccache/metric_cache.go
```diff
@@ -147,7 +149,7 @@ func (m *metricCache) GetNodeResourceMetric(param *QueryParam) NodeResourceQuery
 		return result
 	}
 	if len(metrics) == 0 {
-		result.Error = fmt.Errorf("get node resource metric not exist, query params %v", param)
+		result.Metric = &NodeResourceMetric{}
```
zwzhang0107, line 152: will GetNodeResourceMetric also import error logs?

lucming: `GetNodeResourceMetric `  also should ignore error log, maybe it's the same kind of problem as report-nodetopo

PR #1050, opened 2023-02-23
koordlet: add some logs for cpi and psi metric cache

zwzhang0107:
### V. Checklist


pkg/koordlet/metriccache/storage_test.go
```diff
@@ -848,6 +856,14 @@ func Test_storage_ContainerPSIMetric_CRUD(t *testing.T) {
 				}
 			}
 
+			gotNum, err := s.CountContainerPSIMetric()
```
saintube, line 859: To count the metrics individually, we may create the metric cache via `newStorage()` while the `NewStorage` would open the same DB.

zwzhang0107: both are the same db, and needs defer close now.

jasonliu747 (approved):
/lgtm

PR #1059, opened 2023-02-28
draft----koord-scheduler: enhanced reservation

### LINKED ISSUE #566: [proposal] Reservation supports fine-grained CPU orchestration
**What is your proposal**:

Reservation should support fine-grained CPU orchestration. i.e. A reservation can allocate fine-grained CPU resources via our NodeNUMAResource plugin, and an owner pod can allocate reserved fine-grained CPU resources and get the same CPU binding result.

Furthermore, we expect a generic approach for other scheduling plugins to get compatible with the resource reservation. We've added some utility functions in #550 to facilitate this work.

**Why is this needed**:

Reservation is designed to reserve node resources for specific pods or workloads.  When an LSR/LSE pod is rescheduled, we hope to use resource reservation to ensure node resources. Nevertheless, it cannot reserve fine-grained CPU resources with our NodeNUMAResource plugin, because the reservation plugin does not identify the CPU allocation protocol in the pod annotations.

eahydra:

Enhanced reservation scheduling to support fine-grained resource scheduling plugins such as `NodeNUMAResource` and `DeviceShare` etc..

fix #566 

### V. Checklist


eahydra:
It has been split into three PRs according to the relevant modules, #1139, #1140, #1141

Issue #1034, opened 2023-02-21
[proposal] descheduler support PDB or Kruise PUB


eahydra:
**What is your proposal**:

Currently descheduler is aware of the workload of the Pod, and provides a default stability policy to ensure that the services provided by the workload are as unaffected as possible. This default policy is actually a built-in PDB. Therefore, the existing PDB and PUB of the workload should also be supported. This is also a point mentioned in the previous design. https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220701-pod-migration-job.md#filter-podmigrationjob

**Why is this needed**:

Support the existing PDB and PUB of the workload, which can better cooperate with the PaaS layer.

**Is there a suggested solution, if so, please add it**:

PR #1064, opened 2023-03-01
koord-descheduler: dry-run evict pod to avoid unnecessary reservation creation

ZiMengSheng:

This PR provides a more practicable approach to avoid unnecessary reservation creation. PodMigrationJob will dry run evict pod before create reservation. If dry run failed, abort the job.

Fix https://github.com/koordinator-sh/koordinator/issues/1034

### V. Checklist

### DIFF
--- a/pkg/descheduler/controllers/migration/controller.go
+++ b/pkg/descheduler/controllers/migration/controller.go
@@ -413,6 +413,10 @@ func (r *Reconciler) doMigrate(ctx context.Context, job *sev1alpha1.PodMigration
 	}
 
 	if job.Spec.ReservationOptions == nil || job.Spec.ReservationOptions.ReservationRef == nil {
+		dryRunEvictPodFailed, err := r.requeueJobIfDryRunEvictFailed(ctx, job)
+		if dryRunEvictPodFailed || err != nil {
+			return reconcile.Result{RequeueAfter: defaultRequeueAfter}, err
+		}
 		err = r.createReservation(ctx, job)
 		return reconcile.Result{}, err
 	} else {
@@ -912,6 +916,37 @@ func (r *Reconciler) evictPod(ctx context.Context, job *sev1alpha1.PodMigrationJ
 		r.eventRecorder.Eventf(job, nil, corev1.EventTypeNormal, sev1alpha1.PodMigrationJobReasonEvicting, "Migrating", "%s", cond.Message)
 	}
 	return false, reconcile.Result{RequeueAfter: defaultRequeueAfter}, err
+}
+
+func (r *Reconciler) requeueJobIfDryRunEvictFailed(ctx context.Context, job *sev1alpha1.PodMigrationJob) (failed bool, err error) {
+	_, cond := util.GetCondition(&job.Status, sev1alpha1.PodMigrationJobConditionEviction)
+	if cond != nil {
+		return false, nil
+	}
+
+	klog.V(4).Infof("MigrationJob %s dry run evict pod before create reservation", job.Name)
+	pod := &corev1.Pod{}
+	podNamespacedName := types.NamespacedName{Namespace: job.Spec.PodRef.Namespace, Name: job.Spec.PodRef.Name}
+	err = r.Client.Get(ctx, podNamespacedName, pod)
+	if errors.IsNotFound(err) {
+		err = r.abortJobByMissingPod(ctx, job, podNamespacedName)
+		return true, err
+	}
+	if err != nil {
+		klog.Errorf("Failed to get target Pod %q, MigrationJob: %s, err: %v", podNamespacedName, job.Name, err)
+		return true, err
+	}
+	job = job.DeepCopy()
+	if job.Spec.DeleteOptions == nil {
+		job.Spec.DeleteOptions = &metav1.DeleteOptions{}
+	}
+	job.Spec.DeleteOptions.DryRun = []string{metav1.DryRunAll}
+	err = r.evictorInterpreter.Evict(ctx, job, pod)
+	if err != nil {
+		r.eventRecorder.Eventf(job, nil, corev1.EventTypeWarning, "Requeue", "Migrating", "Failed to dry run evict pod")
+		return true, nil
+	}
+	return false, nil
 }
 
 func (r *Reconciler) prepareJobWithReservationScheduleSuccess(ctx context.Context, job *sev1alpha1.PodMigrationJob, reservationObj reservation.Object) error {
--- a/pkg/descheduler/controllers/migration/controller_test.go
+++ b/pkg/descheduler/controllers/migration/controller_test.go
@@ -576,6 +576,49 @@ func TestEvictPod(t *testing.T) {
 	assert.Equal(t, expectCond, cond)
 }
 
+func TestRequeueJobIfDryRunEvictFailed(t *testing.T) {
+	reconciler := newTestReconciler()
+	reconciler.evictorInterpreter = fakeEvictionInterpreter{err: fmt.Errorf("dry run failed")}
+
+	job := &sev1alpha1.PodMigrationJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:              "test",
+			CreationTimestamp: metav1.Time{Time: time.Now()},
+		},
+		Spec: sev1alpha1.PodMigrationJobSpec{
+			Paused: true,
+			PodRef: &corev1.ObjectReference{
+				Namespace: "default",
+				Name:      "test-pod",
+			},
+		},
+	}
+	assert.Nil(t, reconciler.Client.Create(context.TODO(), job))
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "default",
+			Name:      "test-pod",
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "apps/v1",
+					Controller: pointer.Bool(true),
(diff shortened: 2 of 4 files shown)


apis/extension/pub.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+const (
+	PodRelatedPubAnnotation = "kruise.io/related-pub"
```
eahydra, line 20: Why is the annotation of kruise defined here? It is not the protocol of koordinator.

cmd/koord-descheduler/app/options/scheme.go
```diff
@@ -19,6 +19,7 @@ package options
 import (
 	appsv1alpha1 "github.com/openkruise/kruise-api/apps/v1alpha1"
 	appsv1beta1 "github.com/openkruise/kruise-api/apps/v1beta1"
+	policyv1alpha1 "github.com/openkruise/kruise-api/policy/v1alpha1"
```
eahydra, line 22: Add the kruise prefix before the package name, k8s itself also has a policy package

pkg/descheduler/controllers/migration/controller_test.go
```diff
@@ -106,12 +110,35 @@ func (f *fakeControllerFinder) GetExpectedScaleForPods(pods []*corev1.Pod) (int3
 	return f.replicas, f.err
 }
 
+//
+//var _ podbudgetfinder.Interface = &fakePodBudgetFinder{}
```
eahydra, line 114: remove these comments.

pkg/descheduler/controllers/migration/controller_test.go
```diff
@@ -161,6 +188,7 @@ func newTestReconciler() *Reconciler {
 		reservationInterpreter: nil,
 		evictorInterpreter:     nil,
 		controllerFinder:       controllerFinder,
+		podBudgetFinder:        &podbudgetfinder.PodBudgetFinder{Client: runtimeClient, ControllerFinder: controllerFinder},
```
eahydra, line 191: keep same style with other fields.

pkg/descheduler/controllers/migration/controller_test.go
```diff
@@ -1874,6 +1902,325 @@ func TestFilterMaxMigratingPerNamespace(t *testing.T) {
...
+			maxUnavailable:     2,
+			sameWorkload:       false,
+			want:               true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			reconciler := newTestReconciler()
+			//intOrString := intstr.FromInt(int(tt.totalReplicas - 1))
```
eahydra, line 2037: remove these comments.

pkg/descheduler/controllers/migration/controller.go
```diff
@@ -79,6 +80,7 @@ type Reconciler struct {
 	reservationInterpreter reservation.Interpreter
 	evictorInterpreter     evictor.Interpreter
 	controllerFinder       controllerfinder.Interface
+	podBudgetFinder        podbudgetfinder.Interface
```
eahydra, line 83: podbudget is not short for `podunavailablebudget`. Or renaming it to pubfinder is more appropriate.

pkg/descheduler/controllers/migration/podbudgetfinder/pod_budget_finder.go
```diff
@@ -0,0 +1,119 @@
...
+type PodBudgetFinder struct {
+	client.Client
+	ControllerFinder controllerfinder.Interface
+}
+
+// GetPodsForPub returns Pods protected by the pub object.
+// return two parameters
+// 1. podList
+// 2. expectedCount, the default is workload.Replicas
+func (r *PodBudgetFinder) GetPodsForPub(pub *policyv1alpha1.PodUnavailableBudget, active bool) ([]*corev1.Pod, int32, error) {
```
eahydra, line 36: add more tests for the package.

pkg/descheduler/controllers/migration/podbudgetfinder/pod_budget_finder.go
```diff
@@ -0,0 +1,119 @@
+package podbudgetfinder
```
eahydra, line 1: does the codes be copied from kruise? keep the kruise's copyright.

pkg/descheduler/controllers/migration/filter.go
```diff
@@ -140,6 +140,49 @@ func (r *Reconciler) filterMaxMigratingPerNamespace(pod *corev1.Pod) bool {
...
+	migratingPods := map[types.NamespacedName]struct{}{}
+	for _, v := range pods {
+		if v.UID != pod.UID {
+			podNamespacedName := types.NamespacedName{
+				Namespace: v.Namespace,
+				Name:      v.Name,
+			}
+			if r.existingPodMigrationJob(v) {
+				migratingPods[podNamespacedName] = struct{}{}
+			}
```
eahydra, line 172:
```suggestion
			if r.existingPodMigrationJob(v) {
			      podNamespacedName := types.NamespacedName{
				  Namespace: v.Namespace,
				  Name:      v.Name,
			  }
				migratingPods[podNamespacedName] = struct{}{}
			}
```

pkg/descheduler/controllers/migration/filter.go
```diff
@@ -247,7 +293,7 @@ func (r *Reconciler) trackEvictedPod(pod *corev1.Pod) {
 	}
 
 	var maxMigratingReplicas int
-	if expectedReplicas, err := r.controllerFinder.GetExpectedScaleForPods([]*corev1.Pod{pod}); err == nil {
+	if expectedReplicas, err := r.podBudgetFinder.GetExpectedScaleForPods([]*corev1.Pod{pod}); err == nil {
```
eahydra, line 296: why use podBudgetFinder instead of controllerFinder here?

pkg/descheduler/controllers/migration/podbudgetfinder/pod_budget_finder.go
```diff
@@ -0,0 +1,119 @@
...
+	utilclient "github.com/koordinator-sh/koordinator/pkg/util/client"
+)
+
+type Interface interface {
+	GetPodsForPub(pub *policyv1alpha1.PodUnavailableBudget, active bool) ([]*corev1.Pod, int32, error)
+	GetPubForPod(pod *corev1.Pod) (*policyv1alpha1.PodUnavailableBudget, error)
+	GetExpectedScaleForPods(pods []*corev1.Pod) (int32, error)
+}
+
+type PodBudgetFinder struct {
```
eahydra, line 27: How does the code work if there is no PUB CRD in the cluster?

eahydra (changes requested):

apis/extension/pub.go
ZiMengSheng: OK, I will move it to pkg/descheduler/controllers/migration/podbudgetfinder

pkg/descheduler/controllers/migration/filter.go
ZiMengSheng: because pod budget will getExpectedScale from pub if controllerFinder find nothing

pkg/descheduler/controllers/migration/podbudgetfinder/pod_budget_finder.go
ZiMengSheng: filterMaxUnavailableByPUB will return false. ok I will make it true if no pub crd.

pkg/descheduler/controllers/migration/podbudgetfinder/pod_budget_finder.go
```diff
@@ -0,0 +1,119 @@
...
+	return pub, nil
+}
+
+func (r *PodBudgetFinder) GetExpectedScaleForPods(pods []*corev1.Pod) (int32, error) {
+	expectedReplicas, err := r.ControllerFinder.GetExpectedScaleForPods(pods)
+	if err != nil {
+		return 0, err
+	}
+	if expectedReplicas == 0 {
+		pub, err := r.GetPubForPod(pods[0])
```
eahydra, line 101: here maybe got panic if pods are empty.

pkg/descheduler/controllers/migration/filter.go
```diff
@@ -140,6 +140,49 @@ func (r *Reconciler) filterMaxMigratingPerNamespace(pod *corev1.Pod) bool {
...
+		return false
+	}
+	if pub == nil {
+		return true
+	}
+	pods, expectedReplicas, err := r.pubFinder.GetPodsForPub(pub, false)
+	if err != nil {
+		return false
+	}
+	maxUnavailable := pub.Status.UnavailableAllowed
```
veophi, line 155: This implementation approach has a significant issue in the context of concurrent eviction, as the presence of synchronous delay in PUB Informer can result in inaccurate `UnavailableAllowed` data being obtained here.

eahydra:
If we just want to prevent unnecessary Reservation from being created through DrynRun mode, it seems not so reliable. The entire migration process is still relatively long, so we have to think about where to execute DryRun. Assuming that DryRun has passed and Reservation is created, do we need to check DryRun after the Reservation scheduling is successful?

PR #1066, opened 2023-03-01
koord-scheduler: clarify framework extender

eahydra:

The internal concepts of the `frameworkext` package of koord-scheduler are somewhat confusing and need some clarification. 
- `ExtenderFactory` is responsible for creating `Extender`.
- `Extender` implements `ExtendedHandle`, which means the extension of `framework.Handle`, and implements some functions of `framework.Framework`, responsible for intercepting some function calls. The `Extender` corresponds to each profile, and some transformers inside each `Extender` are initialized through the type assert mechanism when the plugin is built. 
- Rename the Hook class interface to Transformer, which can better express the actual meaning.
- In most scenarios, the global transformer is basically not needed. If necessary, the necessary global transformer (also called default transformer) can be registered with frameworkext. 
- In the previous implementation, we did not give the default implementation of the sharedLister adapter, but this time we added it.

### V. Checklist


hormes:
The [design document](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220513-enhanced-scheduler-extension.md) needs to be updated. Another picture saves a thousand words, draw a diagram of the internal architecture of the scheduler after expansion?

eahydra:
> 

Yeah.. Should update the design proposal..I want to update the proposal later when the enhanced implementation of Reservation is almost complete.  If need, I also can update the proposal in the PR. WDYT? @hormes

eahydra:
> The [design document](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220513-enhanced-scheduler-extension.md) needs to be updated. Another picture saves a thousand words, draw a diagram of the internal architecture of the scheduler after expansion?

Yeah.. Should update the design proposal..I want to update the proposal later when the enhanced implementation of Reservation is almost complete. If need, I also can update the proposal in the PR. WDYT? @hormes

hormes:
Or create an issue to track all the changes you mentioned, so that everyone will not miss it

eahydra:
> Or create an issue to track all the changes you mentioned, so that everyone will not miss it

OKay.

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -18,216 +18,89 @@ package frameworkext
...
 }
 
-func (i *frameworkExtenderFactoryImpl) New(f framework.Framework) FrameworkExtender {
-	return &frameworkExtenderImpl{
-		Framework:      f,
-		handle:         i.handle,
-		preFilterHooks: i.preFilterHooks,
-		filterHooks:    i.filterHooks,
-		scoreHooks:     i.scoreHooks,
+func (ext *frameworkExtenderImpl) updatePlugins(pl framework.Plugin) {
```
ZiMengSheng, line 64: what is the use case for this function

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -0,0 +1,168 @@
...
+func RegisterDefaultTransformers(transformers ...SchedulingTransformer) {
+	DefaultTransformers = append(DefaultTransformers, transformers...)
+}
+
+type extendedHandleOptions struct {
+	servicesEngine                   *services.Engine
+	koordinatorClientSet             koordinatorclientset.Interface
+	koordinatorSharedInformerFactory koordinatorinformers.SharedInformerFactory
+	sharedListerAdapter              SharedListerAdapter
+	defaultTransformers              []SchedulingTransformer
```
ZiMengSheng, line 41: why not just call it transformers

eahydra: `transformers` is so general.

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+func WithKoordinatorSharedInformerFactory(informerFactory koordinatorinformers.SharedInformerFactory) Option {
+	return func(options *extendedHandleOptions) {
+		options.koordinatorSharedInformerFactory = informerFactory
+	}
+}
+
+func WithSharedListerFactory(adapter SharedListerAdapter) Option {
```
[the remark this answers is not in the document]
eahydra: No,  Abstract `SharedListerAdaptor`  to support custom different shareLister adapte strategy. And It can work with WithDefaultTransformers  to replace the default strategy. And `Option function` is  a pattern, we should be consistent.

PR #1071, opened 2023-03-02
Fix stale comment

FillZpp:

Fix stale comment.

### V. Checklist

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -13,9 +13,8 @@ exemptLabels:
 staleLabel: lifecycle/stale
 
 # Comment to post when marking an issue as stale. Set to `false` to disable
-markComment: >
-  This issue has been automatically marked as stale because it has not had
-  recent activity.
+markComment: |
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -13,9 +13,8 @@ exemptLabels:
 staleLabel: lifecycle/stale
 
 # Comment to post when marking an issue as stale. Set to `false` to disable
-markComment: >
```
jasonliu747, line 16: what's the difference? it looks the same to me.
image

FillZpp: Look like:
Value of
```yaml
key: >
  a
  b
```
is
```
ab
```
***
Value of
```yaml
key: |
  a
  b
```
is
```
a
b
```

jasonliu747 (approved):
/lgtm

.github/stale.yml
jasonliu747: makes sense to me then.

PR #1075, opened 2023-03-03
koord-manager: allow plugins configured by configmap then updated in NodeSLO

AlbeeSo:
 - Allow plugins configured by configmap then synchronously updated in ExtensionMap of NodeSLO.
-  Here's an example:
Modified the data of CM:
> an-extension-key:
{
 "clusterStrategy": {
  "enable": true,
  "a-key": "a-value"
    }
}

Then related will be synchronized to NodeSLO like:

> Spec:
   ...
   Cpu Burst Strategy:
   ...
   Extensions:
       Object:
            An Extension:
                Enable:             false
                An Key:             An Value
   Resource QOS Strategy:
   ...
- extendor_config: Set the default config of the extension (if exists)
- extendor_plugin: Allow plugins implement func getExtensionsConfigSpec in nodeslo_controller and func calculateExtensionsCfgMerged in nodeslo_cm_event_handler

### V. Checklist


pkg/slo-controller/nodeslo/extender_plugin.go
```diff
@@ -0,0 +1,102 @@
...
+func RegisterNodeSLOMergedExtender(name string, extender NodeSLOMergedPlugin) error {
+	if _, exist := globalNodeSLOMergedExtender[name]; exist {
+		return fmt.Errorf("node merged plugin %s already exist", name)
+	}
+	globalNodeSLOMergedExtender[name] = extender
+	return nil
+}
+
+func calculateExtensionsCfgMerged(oldCfg extension.ExtensionCfg, configMap *corev1.ConfigMap, recorder record.EventRecorder) extension.ExtensionCfg {
+	oldCfgCopy := *oldCfg.DeepCopy()
```
zwzhang0107, line 44: need checking oldCfg == nil

pkg/slo-controller/nodeslo/extender_plugin.go
```diff
@@ -0,0 +1,102 @@
...
+	return &extMap
+}
+
+func UnregisterNodeSLOMergedExtender(name string) {
+	delete(globalNodeSLOMergedExtender, name)
+}
+
+type NodeSLOMergedPlugin interface {
+	// calculate each extension cfg merged
+	MergeNodeSLO(oldCfg extension.ExtensionCfg, configMap *corev1.ConfigMap, recorder record.EventRecorder) (extension.ExtensionCfg, error)
```
zwzhang0107, line 99: consider naming as:
MergeNodeSLOExtension
GetNodeSLOExtension

pkg/slo-controller/nodeslo/extender_plugin.go
```diff
@@ -0,0 +1,102 @@
...
+
+}
+
+func getExtensionsConfigSpec(node *corev1.Node, cfg *extension.ExtensionCfg) *slov1alpha1.ExtensionsMap {
+	extMap := slov1alpha1.ExtensionsMap{}
+	if cfg == nil || cfg.ClusterStrategy == nil || cfg.ClusterStrategy.Object == nil {
+		return &extMap
+	}
+	for name, extender := range globalNodeSLOMergedExtender {
+		extKey, extStrategy, err := extender.GetNodeSLO(node, cfg)
```
zwzhang0107, line 76: here we can check whether to use clusterStrategy or nodeStrategy.
so that we don't need to implement in each extender plugin.

pkg/slo-controller/nodeslo/nodeslo_cm_event_handler.go
```diff
@@ -49,6 +49,7 @@ type SLOCfg struct {
 	ResourceQOSCfgMerged extension.ResourceQOSCfg       `json:"resourceQOSCfgMerged,omitempty"`
 	CPUBurstCfgMerged    extension.CPUBurstCfg          `json:"cpuBurstCfgMerged,omitempty"`
 	SystemCfgMerged      extension.SystemCfg            `json:"systemCfgMerged,omitempty"`
+	ExtensionCfgMerged   extension.ExtensionCfg         `json:"extensionCfgMerged,omitempty"` // for third-party extension
```
zwzhang0107, line 52: what if there are multi extensions, it seems that the format of ExtensionCfgMerged should be map

zwzhang0107:
PTAL @jasonliu747 @saintube

pkg/slo-controller/nodeslo/extender_plugin_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+type ManageNodeSLO struct{}
+
+func (m *ManageNodeSLO) MergeNodeSLO(oldCfg extension.ExtensionCfg, configMap *corev1.ConfigMap, recorder record.EventRecorder) (extension.ExtensionCfg, error) {
+	newCfg := oldCfg.DeepCopy()
+	if cfgIf, ok := configMap.Data[testExtKey]; ok {
+		newCfg.ClusterStrategy.Object[testExtKey] = cfgIf
+	}
+	return *newCfg, nil
+}
```
saintube, line 43: please leave a newline here

pkg/slo-controller/nodeslo/extender_plugin.go
AlbeeSo: In order to be consistent with the serires functions "calculatexxxxCfgMerged", oldCfg here is a struct instead of a pointer. It's no need to check if oldCfg is nil.

pkg/slo-controller/nodeslo/extender_plugin.go
AlbeeSo: I guess it's better implemented in each GetNodeSLOExtension, as only after parsing the cfg can we know whether an extension has clusterStrategy  or nodeStrategy. And parsing the cfg need extra info and cannot done in precheck.

pkg/slo-controller/nodeslo/nodeslo_cm_event_handler.go
AlbeeSo: Strategy in Cfg is a map. 
`// +k8s:deepcopy-gen=true
type ExtensionCfg struct {
	ClusterStrategy *slov1alpha1.ExtensionsMap `json:"clusterStrategy,omitempty"`
	NodeStrategies  []NodeExtensionStrategy    `json:"nodeStrategies,omitempty"`
}

// +k8s:deepcopy-gen=true
type NodeExtensionStrategy struct {
	NodeCfgProfile             `json:",inline"`
	*slov1alpha1.ExtensionsMap // for third-party extension
}`

AlbeeSo: It can reuse the  *slov1alpha1.ExtensionsMap struct.

pkg/slo-controller/nodeslo/extender_config.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/koordinator-sh/koordinator/apis/extension"
+	slov1alpha1 "github.com/koordinator-sh/koordinator/apis/slo/v1alpha1"
+)
+
+var (
+	defaultExtensionStrategy = slov1alpha1.ExtensionsMap{}
+	defaultExtensionCfg      = extension.ExtensionCfgMap{}
+)
+
+func getDefaultExtensionStrategy() *slov1alpha1.ExtensionsMap {
```
zwzhang0107, line 31: seems that `getDefaultExtensionStrategy` and `defaultExtensionStrategy` only used in UT, which can be deleted

FillZpp (approved):

PR #1077, opened 2023-03-06
koord-scheduler: improve the readability of the PreFilter error message of the DeviceShare plugin

### LINKED ISSUE #840: [proposal] Improve the readability of the PreFilter error message of the DeviceShare plugin
**What is your proposal**:

When I created a Pod with resource requests `kubernetes.io/gpu-core=50`, `kubernetes.io/gpu-memory=2Gi`  and  `kubernetes.io/gpu-memory-ratio=50`, I got the PreFilter error like this: `request is not valid, current combination: 010001000` .

This error message cannot be understood intuitively.

**Why is this needed**:

make sense of error messages.

jasonliu747:
Currently, the error message of PreFiltre in DeviceShare plugin cannot be understood intuitively.
This PR will improve the readability.

fix #840

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/utils.go
+++ b/pkg/scheduler/plugins/deviceshare/utils.go
@@ -21,6 +21,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	"k8s.io/klog/v2"
 
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/deviceshare/utils.go
```diff
@@ -117,7 +117,7 @@ var ValidateGPURequest = func(podRequest corev1.ResourceList) (uint, error) {
 		return gpuCombination, nil
 	}
 
-	return gpuCombination, fmt.Errorf("request is not valid, current combination: %b", gpuCombination)
+	return gpuCombination, fmt.Errorf("request is not valid, current combination: %v", convertGPURequestCombination(gpuCombination))
```
ZiMengSheng, line 120: maybe we can give what is right?

jasonliu747: image

eahydra: hmm.. we should update the comments..

PR #1078, opened 2023-03-06
koordlet: require gc started for cacheable update

saintube:

To avoid memory leaks, require the cache GC to be started when the executor invokes cache updating.

### V. Checklist

### DIFF
--- a/pkg/koordlet/resourceexecutor/executor.go
+++ b/pkg/koordlet/resourceexecutor/executor.go
@@ -17,6 +17,7 @@ limitations under the License.
 package resourceexecutor
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -110,6 +121,11 @@ func (e *ResourceUpdateExecutorImpl) UpdateBatch(cacheable bool, updaters ...Res
 func (e *ResourceUpdateExecutorImpl) LeveledUpdateBatch(cacheable bool, updaters [][]ResourceUpdater) {
```
zwzhang0107, line 121: remove cacheable since it is never supported?

saintube: done

PR #1086, opened 2023-03-07
koordlet: merge extensions in nodeslo and add general util methods

AlbeeSo:
- merge NodeSLOSpec.Extensions when updating nodeSLO
- add general method in koordlet/util, which allows to read NumaStat info in cgroup

### V. Checklist

### DIFF
--- a/pkg/koordlet/util/system/cgroup.go
+++ b/pkg/koordlet/util/system/cgroup.go
@@ -17,8 +17,10 @@ limitations under the License.
 package system
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 const (
@@ -54,6 +56,11 @@ type MemoryStatRaw struct {
 	ActiveAnon   int64
 	Unevictable  int64
 	// add more fields
+}
+
+type NumaMemoryPages struct {
+	NumaId   int
+	PagesNum uint64
 }
 
 func (m *MemoryStatRaw) Usage() int64 {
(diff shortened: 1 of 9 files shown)


pkg/koordlet/util/system/cgroup.go
```diff
@@ -412,3 +417,39 @@ func CalcCPUThrottledRatio(curPoint, prePoint *CPUStatRaw) float64 {
 	}
 	return throttledRatio
 }
+
+func GetCgroupCurNumaStat(cgroupPath string) ([]NumaMemoryPages, error) {
```
hormes, line 421: Is it possible to add an unittest?

pkg/koordlet/util/container.go
```diff
@@ -203,6 +211,14 @@ func GetContainerCurTasks(podParentDir string, c *corev1.ContainerStatus) ([]int
 	return system.GetCgroupCurTasks(cgroupPath)
 }
 
+func GetContainerCurNumaStat(podParentDir string, c *corev1.ContainerStatus) ([]system.NumaMemoryPages, error) {
```
saintube, line 214: To support different cgroup versions, it is recommended to add a `(*resourceexecutor.CgroupReader).ReadNumaStat()` instead of the `GetContainerCurNumaStat()`.

pkg/util/config.go
```diff
@@ -271,3 +272,7 @@ func DefaultSystemStrategy() *slov1alpha1.SystemStrategy {
 		WatermarkScaleFactor: pointer.Int64Ptr(150), // 1 means 1/10000
 	}
 }
+
+func DefaultExtensions() *slov1alpha1.ExtensionsMap {
+	return &slov1alpha1.ExtensionsMap{}
```
zwzhang0107, line 277: the default value of slo.ExtMap should be extensible(e.g. new plugin can define its own default value in independent files)
consider define as var and `DefaultExtensions()` return the var value

zwzhang0107: // config.go
var defautExtensions = &slov1alpha1.ExtensionsMap{}

// ack_dsa_config.go
var dsaCfg = {}
defautExtensions[dsaDefault] = dsaCfg

// contorller
Regist(dsaCfg)

PR #1098, opened 2023-03-09
ci: add commit id as tag when releasing docker images

jasonliu747:
### V. Checklist


jasonliu747 (commented):
/approve

PR #1065, opened 2023-03-01
koord-manager: refactor noderesource and add resource calculation extenders

saintube:
### Ⅰ. Describe what this PR does

Add the following extensions for the node-resource controller:

- NodeSync: allow the plugin to implement the node update check.
- ResourceReset: allow the plugin to reset the corresponding resources for degradation or colocation-disabled.
- ResourceCalculate: allow the plugin to implement different resource calculation algorithms other than the batch resource overcommitment.

### Ⅱ. Does this pull request fix one issue?

Fixes #1060.

### Ⅳ. Special notes for reviews

Just keep the GPU resources reconciliation since it will be refactored in the future.

### V. Checklist

- [x] I have written necessary docs and comments
- [x] I have added necessary unit tests and integration tests
- [X] All checks passed in `make test`

PR #1101, opened 2023-03-09
koord-manager: revise node resource framework

### LINKED ISSUE #1060: [proposal] noderesource controller supports resource calculator extenders
**What is your proposal**:

The node-resource controller of the koord-manager should allow extensions for resource calculation.

**Why is this needed**:

The resource calculation algorithm in the node-resource controller should be extensible. For instance, a non-overcommit extender can be used for some nodes' batch resource calculation, while the other nodes use the default.

Add an extender framework in the node-resource controller.

saintube:

Revise node resource framework:

1. Add a sequence of the plugins' execution by the registered order.
2. Make the degraded resource calculation private in each ResourceCalculatePlugin.
3. Use structural logs in the node-resource controller.

fixes #1060

### V. Checklist


pkg/slo-controller/noderesource/noderesource_controller.go
```diff
@@ -86,21 +85,13 @@ func (r *NodeResourceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-
-	if r.isDegradeNeeded(nodeMetric, node) {
-		klog.Warningf("node %v need degradation, reset BE resource", req.Name)
-		if err := r.resetNodeResource(node, "degrade node resource because of abnormal nodeMetric, reason: "+degradeByKoordController); err != nil {
+		if !errors.IsNotFound(err) {
+			klog.ErrorS(err, "failed to get nodeMetric", "node", req.Name)
 			return ctrl.Result{Requeue: true}, err
 		}
-		return ctrl.Result{}, nil
+		nodeMetric = &slov1alpha1.NodeMetric{}
```
jasonliu747, line 92: Do we need to reset NodeMetric here? 
Please correct me if I'm wrong. I thought `nodeMetric` would remain empty if there is an error when calling `Get()`.

pkg/slo-controller/noderesource/noderesource_controller_test.go
```diff
@@ -18,8 +18,8 @@ package noderesource
 
 import (
 	"context"
+
```
jasonliu747, line 21: no need to add a new line between `context` and `testing`

pkg/slo-controller/noderesource/noderesource_controller_test.go
```diff
@@ -197,20 +198,15 @@ func Test_NodeResourceController_ColocationEnabled(t *testing.T) {
 	assert.Equal(t, int64(65000), batchCPU)
 
 	// reset node resources
-	r.Clock = clock.NewFakeClock(r.Clock.Now().Add(time.Hour))
```
jasonliu747, line 200: will `isDegradeNeeded` be covered after this fake clock got deleted?

jasonliu747: never mind. I saw you add a test for `isDegradeNeeded` exclusively in `pkg/slo-controller/noderesource/plugins/batchresource/plugin_test.go`

pkg/slo-controller/noderesource/resource_calculator.go
```diff
@@ -120,7 +104,8 @@ func (r *NodeResourceReconciler) updateNodeResource(node *corev1.Node, nr *frame
 func (r *NodeResourceReconciler) updateNodeExtensions(node *corev1.Node, nodeMetric *slov1alpha1.NodeMetric, podList *corev1.PodList) error {
 	// update device resources
 	if err := r.updateDeviceResources(node); err != nil {
```
jasonliu747, line 106: should we add a config or feature-gate for device calculation?

pkg/slo-controller/noderesource/resource_calculator.go
```diff
@@ -120,7 +104,8 @@ func (r *NodeResourceReconciler) updateNodeResource(node *corev1.Node, nr *frame
...
-		klog.Errorf("failed to update device resources for node %s, err: %v", node.Name, err)
+		klog.V(4).InfoS("failed to update device resources for node", "node", node.Name,
```
jasonliu747, line 107: why use InfoS here?

saintube: The `Reconcile` will log the error of `updateNodeExtensions`, so here it just log for `updateDeviceResources` with a higher verbosity.

pkg/slo-controller/noderesource/resource_calculator.go
saintube: It has been discussed in https://github.com/koordinator-sh/koordinator/pull/1065#issuecomment-1457451506, so we do not change the device calculation in the issue #1060.

pkg/slo-controller/noderesource/noderesource_controller.go
saintube: OK. The nodeMetric object here is got by `CacheReader`, so it should remain empty when an error is returned.

pkg/slo-controller/noderesource/noderesource_controller_test.go
saintube: nice catch. fixed

pkg/slo-controller/noderesource/framework/extender_plugin.go
```diff
@@ -17,122 +17,138 @@ limitations under the License.
...
+	globalNodePrepareExtender       = NewRegistry("NodePrepare")
+	globalNodeSyncExtender          = NewRegistry("NodeSync")
+	globalResourceCalculateExtender = NewRegistry("ResourceCalculate")
 )
 
+// Plugin has its name and priority. Registered plugins are executed in order of the priority. The high-priority plugin
+// runs earlier than the low-priority plugin. If the priorities are equal, compare using the lexicographic order.
+type Plugin interface {
+	Name() string
+	Priority() int32
```
zwzhang0107, line 38: consider naming like sequence or something

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -40,12 +42,18 @@ const PluginName = "BatchResource"
...
+var Clock clock.Clock = clock.RealClock{} // for testing
+
 type Plugin struct{}
 
 func (p *Plugin) Name() string {
 	return PluginName
 }
 
+func (p *Plugin) Priority() int32 {
+	return framework.DefaultPluginPriority
```
zwzhang0107, line 54: priority/seq can be overwrite in local repo, maybe defined all seq in an independent file.

pkg/slo-controller/noderesource/framework/extender_plugin.go
saintube: removed. just use the order of registration

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
saintube: done. the ordering and registration are placed in `/noderesource/plugins_profile.go`

hormes:
Setting the priority on the plugin will make understanding very complicated, because the logic is scattered in different files, it is difficult to add a plugin to know where it should be placed. If there is a strong order dependency between plugins, then we should have a place to register each plugin uniformly and arrange their order here.

hormes:
I see that priority has been removed, and some comments and functions have not been removed

saintube:
> I see that priority has been removed, and some comments and functions have not been removed

Fixed.

pkg/slo-controller/noderesource/framework/registry_test.go
```diff
@@ -0,0 +1,63 @@
...
+		r.Unregister(p0.Name())
+		assert.Equal(t, wantPlugins1, r.GetAll())
+	})
+}
+
+var _ Plugin = (*testPlugin)(nil)
+
+type testPlugin struct {
+	name     string
+	priority int32
```
hormes, line 52: the priority field is no longer valid

saintube: removed

jasonliu747 (commented):
/lgtm
others PTAL @hormes @zwzhang0107

PR #1106, opened 2023-03-10
koord-scheduler: PodGroups are no longer created if pods use Gang annotations

### LINKED ISSUE #1099: [proposal] Remove the verification of the Gang name defined in the pod
**What is your proposal**:

Remove the verification of the Gang name defined in the pod.

**Why is this needed**:

Currently koord-scheduler will verify the format of the Gang name defined in the Pod, such as `gang.scheduling.koordinator.sh` or and `pod-group.scheduling.sigs.k8s.io/name` defined in k8s scheduler-plugins  to used to represent Gang Name, the purpose of verification is to create a PodGroup next. But there are some users who are using `pod-group.scheduling.sigs.k8s.io/name` to specify the Gang Name. So we may need to consider remove this format verification, or even not creating a PodGroup if the Pod uses annotations.

eahydra:

1.  Remove the verification of the Gang name defined in the pod.
2. PodGroups are no longer created if pods use Gang annotations.

fix #1099 

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/gang.go
+++ b/pkg/scheduler/plugins/coscheduling/core/gang.go
@@ -35,9 +35,8 @@ var (
 )
 
 const (
-	GangFromPodGroupCrd       string = "GangFromPodGroupCrd"
-	GangFromPodAnnotation     string = "GangFromPodAnnotation"
-	PodGroupFromPodAnnotation string = "PodGroupFromPodAnnotation"
+	GangFromPodGroupCrd   string = "GangFromPodGroupCrd"
+	GangFromPodAnnotation string = "GangFromPodAnnotation"
 )
 
 // Gang  basic podGroup info recorded in gangCache:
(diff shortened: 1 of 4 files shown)


hormes:
After removing the gang cr, will the gang declared through the annotation have a memory object in the gangcache? Will it recover after process restart?

eahydra:
> After removing the gang cr, will the gang declared through the annotation have a memory object in the gangcache? Will it recover after process restart?

Yes, the Coscheduling plugin will create a Gang object in the GangCache if the Pod declares a Gang in the annotation. And recover from the Pod's annotations when the process starts.

eahydra:
> After removing the gang cr, will the gang declared through the annotation have a memory object in the gangcache? Will it recover after process restart?

When the Pod declares the Gang in the annotation, the reason why PodGroup is created is to give the user a chance to perceive the status of the Gang, but in fact, the user may not care about it at all.

PR #1121, opened 2023-03-15
apis: Reservation enables AllocateOnce by default

### LINKED ISSUE #934: [proposal] change default value allocateOnce in reservation to true
Can we change default value of allocateOnce in reservation to true?
Do you have many scenarios where allocateOnce needs to be set to false？

**What is your proposal**:

**Why is this needed**:

eahydra:

Scenarios where the Reservation is released immediately after being used are more common. And using AllocateOnce can also avoid many resource allocation problems. So the default value of AllocateOnce is changed to true

fix #934 

### V. Checklist

### DIFF
--- a/apis/scheduling/v1alpha1/reservation_types.go
+++ b/apis/scheduling/v1alpha1/reservation_types.go
@@ -57,9 +57,9 @@ type ReservationSpec struct {
 	// reservation would be waiting to be available until free resources are sufficient.
 	// +optional
 	PreAllocation bool `json:"preAllocation,omitempty"`
-	// By default, reserved resources are always allocatable as long as the reservation phase is Available. When
-	// `AllocateOnce` is set, the reserved resources are only available for the first owner who allocates successfully
-	// and are not allocatable to other owners anymore.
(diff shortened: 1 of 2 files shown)


jasonliu747 (approved):
/lgtm

apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -57,9 +57,9 @@ type ReservationSpec struct {
 	// reservation would be waiting to be available until free resources are sufficient.
 	// +optional
 	PreAllocation bool `json:"preAllocation,omitempty"`
-	// By default, reserved resources are always allocatable as long as the reservation phase is Available. When
```
ZiMengSheng, line 60: if AllocateOnce, the unused reservation resource (i.e. reservation.allocatable-reservation.allocated) should be ignored when schedule unmatched pod. this logic is already exists?

eahydra: yes. already exists

FillZpp (approved):

PR #1125, opened 2023-03-16
koord-scheduler: ElasticQuota skips reserve pod in PostFilter phase

eahydra:

ElasticQuota should skip reserve pod in PostFilter phase.  If we don't do this, some resource return logic of the subsequent Reservation will be affected.

### V. Checklist


jasonliu747 (approved):
/lgtm

ZiMengSheng:
The reason behing this logic is quota temporarily ignore reservation?

hormes (approved):

PR #1126, opened 2023-03-16
koord-scheduler: enhanced frameworkext to support Reservation

### LINKED ISSUE #1076: [proposal] clarify scheduler framework extender
**What is your proposal**:

The internal concepts of the frameworkext package of koord-scheduler are somewhat confusing and need some clarification.

- ExtenderFactory is responsible for creating Extender.
- Extender implements ExtendedHandle, which means the extension of framework.Handle, and implements some functions of framework.Framework, responsible for intercepting some function calls. The Extender corresponds to each profile, and some transformers inside each Extender are initialized through the type assert mechanism when the plugin is built.
- [design document](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220513-enhanced-scheduler-extension.md) describes the new extension points named `Transformer` but now named `Hook` in the current implementation.
- And the `Hook` should be registered in the main function, but we can use type assert with plugin instance to determine whether the plugin implements the `Hook`
- And we should add `PreBindReservation` extension point to clearly indicate whether a plugin needs to update the Reservation in the PreBind phase

**Why is this needed**:

- [x] refactor the scheduler framework extender.  Already in review.. #1066 
- [ ] implements the `PreBindReservation`
- [ ] update the doc

eahydra:

In order to support the subsequent enhanced Reservation (#1059), the mechanism in frameworkext needs to be further enhanced and optimized.

1. Add new plugin interfaces: 
    - `ReservationPreFilterExtension` is used to support the return of fine-grained resources held by Reservation, such as CPU Cores, GPU Devices, etc. 
    -  `ReservationFilterPlugin` is an interface for Filter Reservation plugins.These plugins will be called during the Reserve phase to determine whether the Reservation can participate in the Reserve`
    - `ReservationNominator` nominates a most suitable Reservation in the Reserve stage and Pod will bind this Reservation.
    - `ReservationScorePlugin` is an interface that must be implemented by "ScoreReservation" plugins to rank reservations that passed the reserve phase.
    - `ReservationPreBindPlugin` performs special binding logic specifically for Reservation in the PreBind phase.
2. Add method `AfterPreFilter` in PreFilterTransformer to support triggering the correction of the State data of each plugin after the PreFilter.
3. Add method `RunReservationPreFilterExtensionRemoveReservation`, `RunReservationPreFilterExtensionAddPodInReservation`, `RunReservationFilterPlugins` and `RunReservationScorePlugins` for the convenience of calling these new plugin interfaces.
4. Add built-in scheduling cycle state named `nominatedReservationKey` to  save the result of `ReservationNominator`.
 

fix #1076 

### V. Checklist


eahydra:
/hold cancel
TBR, PTAL @hormes @FillZpp @saintube @ZiMengSheng

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -127,3 +199,131 @@ func (ext *frameworkExtenderImpl) RunScorePlugins(ctx context.Context, state *fr
...
+
+	for _, pl := range ext.reservationRecommender {
+		reservation, status := pl.RecommendReservation(ctx, cycleState, pod, nodeName)
+		if !status.IsSuccess() {
+			return status
+		}
+		if reservation != nil {
+			SetRecommendReservation(cycleState, reservation)
+			break
+		}
```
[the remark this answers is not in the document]
eahydra: There will only be one Recommender.

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -76,22 +112,58 @@ func (ext *frameworkExtenderImpl) KoordinatorSharedInformerFactory() koordinator
...
+	if ext.temporarySnapshot == nil {
+		// Only test scenarios are encountered here.
+		sharedLister := ext.Framework.SnapshotSharedLister()
+		if ext.sharedListerAdapter != nil {
+			sharedLister = ext.sharedListerAdapter(sharedLister)
+		}
+		return sharedLister
+	}
+	return ext.temporarySnapshot
+}
```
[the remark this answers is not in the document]
eahydra: see the framework_extender.go line 139 comments.

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -18,27 +18,48 @@ package frameworkext
...
 	koordinatorClientSet             koordinatorclientset.Interface
 	koordinatorSharedInformerFactory koordinatorinformers.SharedInformerFactory
 
 	preFilterTransformers []PreFilterTransformer
 	filterTransformers    []FilterTransformer
 	scoreTransformers     []ScoreTransformer
+
+	reservationFilterPlugins       []ReservationFilterPlugin
+	reservationScorePlugins        []ReservationScorePlugin
+	reservationRecommender         []ReservationRecommender
```
saintube, line 49: Perhaps `reservationNominator` is better. And we may need to clarify why a ReservationRecommender is needed though we have some ReservationScorePlugins (I think it is like the DefaultBinder for the BindPlugin).

eahydra: Nominator is indeed better. There is a description of why recommender is needed. For details, see the document https://github.com/koordinator-sh/koordinator/pull/1126/files#diff-88ed0b7eb4820fe7359f4af52e61b5fe0e1050c97de1c81790e34e2faf7c7ab3R172

saintube: OK

hormes:
/lgtm
much clearer than before

jasonliu747 (approved):
/lgtm

saintube (commented):
/lgtm

Issue #1102, opened 2023-03-09
[proposal] descheduler should users take over eviction behavior themselves


eahydra:
**What is your proposal**:

Currently koord-descheduler support pod eviction based on K8s Eviction API, but in some scenarios, users expect to be able to take over the eviction behavior themselves.  

**Why is this needed**:

1. Users want to scale new Pod before eviction, and when the pod ready, users delete  the old Pod themselves.  Such as in search services, they want to transfer the local data to the new Pod before eviction, and when transfer completed, they delete the old Pod.
2.  Some scenarios have only one replica per workload, such as internal test clusters or some small clusters, but they also need to keep the service continuously available to users.

**Is there a suggested solution, if so, please add it**:

Koordinator descheduler should defined some API and patch the API in Pod to trigger the users take over the eviction behavior.
Such as patch some labels that configured in descheduler profile to the target Pod, and users can implement controller themselves to scale new Pod, and when the pod ready, delete the target pod.

leason00:
For the case2, we had an internal implementation. For this single pod workload, we modified a system environment variable in the workload to achieve the scrolling effect.

LeoLiuYan:
So,  is it having same effect to `dryrun` mode?   In `dryrun` mode, we just selected out the target pods and logging the pod's information? 

Can we filter the target pods which should be evicted in `stdout log`?

PR #1128, opened 2023-03-20
koord-descheduler: support soft eviction takeover eviction behavior

leason00:
By tagging, the user takes over the eviction.

fixes  https://github.com/koordinator-sh/koordinator/issues/1102

1. Enable this feature in koord-descheduler-config
```yaml
pluginConfig:
- name: MigrationController
  args:
    apiVersion: descheduler/v1alpha2
    kind: MigrationControllerArgs
    evictionPolicy: SoftEviction
    evictLocalStoragePods: true
    ignorePvcPods: true
    namespaces:
      include:
          - qa
    evictQPS: "10"
    evictBurst: 1
```

2. A annotation will be written on Pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
   scheduling.koordinator.sh/soft-eviction: |-
          {
            "timestamp": 1679970640,
            "deleteOptions": {},
            "initiator": "xxx",
            "reason": "xxx"
          }
```

### V. Checklist

### DIFF
--- a/apis/extension/descheduling.go
+++ b/apis/extension/descheduling.go
@@ -17,8 +17,11 @@ limitations under the License.
 package extension
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


eahydra:
Hi @leason00, thanks for your contribution.

Do we need to implement a new Evict Plugin? For issue #1102, my original idea was to add a new implementation to the evictor in PodMigrationController. Users can enable this new evictor implementation in MigrationControllerArgs.

In addition, we need to define the annotation added when the evict is executed, and define it as an API in apis/extension/descheduling.go , for example, scheduling.koordinator.sh/eviction=true means triggering eviction. In addition, you may have to consider adding some additional information, such as the time to trigger the eviction, the reason, who triggered it, and whether to set gracePeriodSeconds and forceDeletion, etc.

I think maybe a simple proposal can be given to clearly define this new change and API.

hormes:
Do you want to reuse other logic in MigrationController?

leason00:
It is better to reuse MigrationController. I'm going to redesign the MigrationController-based approach.

leason00:
@eahydra Resubmitted. Review it again when you have time.

apis/extension/descheduling.go
```diff
@@ -34,6 +36,20 @@ const (
 	AnnotationEvictionCost = SchedulingDomainPrefix + "/eviction-cost"
 )
 
+const (
+	// AnnotationEvictionMark indicates custom eviction. It can be used to set to an "true".
+	AnnotationEvictionMark = SchedulingDomainPrefix + "/custom-eviction"
```
eahydra, line 41: Maybe we can directly set the annotation key to `koordinator.sh/eviction`, and the value of this annotation is the specific field of the eviction request.



 For example:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
      koordinator.sh/eviction: |-
          {
               "timestamp": "the timestamp of eviction requests, and the format is metav1.Time",
               "deleteOptions":  "metav1.DeleteOptions",
               "initiator": "We need this field to help quickly locate the initiator to judge whether the eviction is reasonable", 
               "reason": "This information is also useful",
               "controller": "There are some complex scenarios where different workloads are evicted or deleted by different controllers. So we need this field. It is not set by default, and when it is not set, a default controller can be responsible for eviction."
          }
```

WDYT @leason00 cc @hormes @FillZpp

leason00: Except for `controller`, everything else seems to be fine. From the above description, the controller needs to be configured by the user, or declared on the Pod, to determine whether theirs custom controller will take over and evict the annotation Pod

eahydra: Yes. Your description is more accurate. The controller should indeed be specified by the user on the Pod, not by our descheduler.

hormes:
nice work!

apis/extension/descheduling.go
```diff
@@ -34,6 +37,37 @@ const (
...
 )
 
+const (
+	// AnnotationSoftEviction indicates custom eviction. It can be used to set to an "true".
+	AnnotationSoftEviction = SchedulingDomainPrefix + "/soft-eviction"
+)
+
+type SoftEvictionSpec struct {
+	// Timestamp indicates time when custom eviction occurs . It can be used to set a second timestamp.
+	Timestamp int64 `json:"timestamp,omitempty"`
```
eahydra, line 47: It is recommended to change the type of the timestamp field to *metav1.Time so that it can be formatted as a time string. More readable than Unix Seconds.

PR #1129, opened 2023-03-21
koordlet: use cgroup parent path straightforward

saintube:

Directly store the cgroup parent path with the kubepods directory in the pod meta. No longer do concatenation while generating the full cgroup file path. It is also helpful when the pod cgroup path is moved from the upstream pattern.

### V. Checklist

### DIFF
--- a/pkg/koordlet/common/utils/pod.go
+++ b/pkg/koordlet/common/utils/pod.go
@@ -27,7 +27,7 @@ func GetPodMetas(pods []*corev1.Pod) []*statesinformer.PodMeta {
 	podMetas := make([]*statesinformer.PodMeta, len(pods))
 
 	for index, pod := range pods {
-		cgroupDir := util.GetPodKubeRelativePath(pod)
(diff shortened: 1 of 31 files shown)


docs/proposals/koordlet/20221018-koordlet-support-cgroups-v2.md
```diff
@@ -522,7 +522,7 @@ func (p *Plugin) applyContainerCFSQuota(podMeta *statesinformer.PodMeta, contain
 		}
 
 		// for container-level
-		containerDir, err := util.GetContainerCgroupPathWithKube(podMeta.CgroupDir, containerStat)
+		containerDir, err := util.GetContainerCgroupParentDir(podMeta.CgroupDir, containerStat)
```
zwzhang0107, line 525: how about save the dir in podMeta, so that we don't need to change lots of interfaces

saintube: done.

PR #1132, opened 2023-03-23
koord-scheudler: fix ElasticQuota PreFilterExtension error with missing Pod

### LINKED ISSUE #1131: [BUG] ElasticQuota reports error "pod is not in cache"
**What happened**:

image

Schedule new pod successfully.

Can not reproduce it.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

eahydra:

The PreFilterExtension.AddPod/Remove implemented by ElasticQuota will conflict with the scheduler framework snapshot mechanism. When the framework calls these two methods, it is possible that the Pod corresponding to these two parameters has been deleted. ElasticQuota senses this change through cache.EventHandler and clears the relevant data from its internal state. At this time, AddPod/RemovePod will Check whether the corresponding Pod is still there, and report an error if it is not. This stage actually expects a best effort, so try not to report errors here. 

fix #1131 

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1139, opened 2023-03-25
koord-scheduler: enhanced reservation scheduling

eahydra:

Enhanced reservation scheduling.
1. Weaken the restriction of the matching strategy, and select the Reservation only based on whether the Owner matches.
2. After the PreFilter, the resources reserved by the Reservation are returned uniformly.
3. Supports filtering invalid Reservations in Reserve phase , such as Reservations that have no remaining resources, or Reservations that have AllocateOnce enabled and Pods have been allocated, or Reservations deemed invalid by other plugins. 
4. To further simplify the implementation of multiple stages such as reserve/unreserve/prebind, we have abstracted a separate Controller that is responsible for updating the state of the Reservation. This makes the entire Pod scheduling process simpler and more reliable.
5. When AllocateOnce=false, a new constraint is added: only one Reservation with the same Owner can be scheduled on the same node, and when returning the resources reserved by the Reservation (with AllocateOnce=false), only the one with the earliest creation time is selected.

### V. Checklist


pkg/scheduler/plugins/reservation/controller/controller.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	if reservationutil.IsReservationFailed(reservation) ||
+		reservationutil.IsReservationSucceeded(reservation) {
+		return result{}, nil
+	}
+
+	reservation = reservation.DeepCopy()
+
+	if reservationutil.IsReservationActive(reservation) {
```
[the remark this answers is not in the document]
eahydra: Yeah, fixed.

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,275 @@ import (
...
+		// For example, on a 32C machine, ReservationA reserves 8C, and then PodA uses ReservationA to allocate 4C,
+		// then the record on NodeInfo is that 12C is allocated. But in fact it should be calculated according to 8C,
+		// so we need to return some resources.
+		for _, rInfo := range reservationInfos {
+			if len(rInfo.allocated) == 0 {
+				continue
+			}
+
+			reservePod := reservationutil.NewReservePod(rInfo.reservation)
+			if err := nodeInfo.RemovePod(reservePod); err != nil {
```
tzzcfrank, line 78: In unmatched case, the real pod should be deleted, not the reserve pod, as there may be available resources in the reservation.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -219,412 +190,165 @@ func (p *Plugin) Filter(ctx context.Context, cycleState *framework.CycleState, p
...
-		// avoid concurrency conflict inside the scheduler (i.e. scheduling cycle vs. binding cycle)
-		if !matchReservation(pod, rInfo) {
-			klog.V(5).InfoS("failed to reserve reservation since the reservation does not match the pod",
-				"pod", klog.KObj(pod), "reservation", klog.KObj(target), "reason", dumpMatchReservationReason(pod, rInfo))
-			continue
-		}
+	// NOTE: Having entered the Reserve stage means that the Pod scheduling is successful,
+	// even though the associated Reservation may have expired, but in fact the real impact
+	// will not be encountered until the next round of scheduling.
+	assumed := nominatedReservation.DeepCopy()
```
tzzcfrank, line 273: When multiple reservations are consumed in a pod, only one reservation is recorded, which may cause inconsistencies between the scheduler accounting and reservation status in some cases.

pkg/scheduler/plugins/reservation/transformer.go
eahydra: line 85-94 calc the remaining resources and re-add into NodeInfo

pkg/scheduler/plugins/reservation/plugin.go
eahydra: Most of the scenes use AllocateOnce when setting up a Reservation, which is basically consistent with the Resource Requests of Pods and generally there will be no fragmentation. When AllocateOnce=false, a new constraint is added: only one Reservation with the same Owner can be scheduled on the same node, and when returning the resources reserved by the Reservation (with AllocateOnce=false), only the one with the earliest creation time is selected.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -95,90 +90,72 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
...
 		rName := reservationutil.GetReservationNameFromReservePod(pod)
-		r, err := p.rLister.Get(rName)
-		if errors.IsNotFound(err) {
-			klog.V(3).InfoS("skip the pre-filter for reservation since the object is not found",
-				"pod", klog.KObj(pod), "reservation", rName)
-			return nil
-		} else if err != nil {
+		r, err := pl.rLister.Get(rName)
+		if err != nil {
+			if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
eahydra: You have already solved this logic before. In a separate Reservation EventHandler, remove the corresponding Pod from the queue on delete event handler.

apis/extension/reservation.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	reservationAllocated := &ReservationAllocated{}
+	err := json.Unmarshal([]byte(data), reservationAllocated)
+	if err != nil {
+		return nil, err
+	}
+	return reservationAllocated, nil
+}
+
+func SetReservationAllocated(pod *corev1.Pod, r *schedulingv1alpha1.Reservation) {
```
saintube, line 59: Why no longer set the allocateOnce reservation succeeded here

eahydra: ReservationStatus is no longer set in the extension point implemented by the Reservation plugin, and the Controller is responsible for this matter. If a Pod is bound to a Reservation using AllocateOnce, the internal reservationInfo will have corresponding records.

saintube: Why not keep setting the phase of reservation as succeeded in the plugin? The separation above makes criteria that require the scheduler to check if the allocateOnce reservation is assumed but still in the Available phase.

saintube: ok

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,361 @@ import (
...
-		return nil, nil, false
-	}
-	// abort if node is not found
+func (pl *Plugin) BeforeFilter(handle frameworkext.ExtendedHandle, cycleState *framework.CycleState, pod *corev1.Pod, nodeInfo *framework.NodeInfo) (*corev1.Pod, *framework.NodeInfo, bool) {
 	node := nodeInfo.Node()
 	if node == nil {
 		return nil, nil, false
 	}
+	// Generation equals -1 means that the Reservation return logic has been executed once, but it should not happen.
+	if nodeInfo.Generation == -1 {
```
ZiMengSheng, line 93: how does nodeInfo.Generation=-1 works

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,361 @@ import (
...
+	if err := pl.restoreUnmatchedReservations(cycleState, pod, nodeInfo, unmatchedReservationInfos, false); err != nil {
 		return nil, nil, false
 	}
 
-	klog.V(5).InfoS("BeforeFilter get reservation matched on node",
-		"pod", klog.KObj(pod), "node", node.Name, "count", len(rOnNode))
-	// fix-up reserved resources and ports
-	return pod, prepareFilterNodeInfo(pod, nodeInfo, rOnNode, allocatedResource), true
+	if !reservationutil.IsReservePod(pod) {
+		if err := pl.restoreMatchedReservations(cycleState, pod, nodeInfo, matchedReservationInfos, false); err != nil {
```
ZiMengSheng, line 110: 1. Why does resource returning logic executed twice in beforeFilter and afterPreFilter?
2. Why afterPreFilter not beforePreFilter

pkg/scheduler/plugins/reservation/nominator.go
```diff
@@ -0,0 +1,123 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+
+	schedulingv1alpha1 "github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1"
+	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext"
+	reservationutil "github.com/koordinator-sh/koordinator/pkg/util/reservation"
+)
+
+func (pl *Plugin) NominateReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeName string) (*schedulingv1alpha1.Reservation, *framework.Status) {
```
ZiMengSheng, line 33: NonimateReservation isn't called in score Phase. Reservation Filter and Score doesn't matter while scoring node?

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,361 @@ import (
...
+	}
+
+	// Regardless of whether the Reservation enables AllocateOnce
+	// and whether the Reservation uses high-availability constraints such as InterPodAffinity/PodTopologySpread,
+	// we should trigger the plugins to update the state and erase the Reservation-related state in this round
+	// of scheduling to ensure that the Pod can pass these filters.
+	// Although Reservation may reserve a large number of resources for a batch of Pods to use,
+	// and it is possible to use these high-availability constraints to cause resources to be unallocated and wasteful,
+	// users need to bear this waste.
+	status := handle.RunPreFilterExtensionRemovePod(context.Background(), cycleState, pod, reservePodInfo, nodeInfo)
```
ZiMengSheng, line 223: the core points of the plugin cache returning logic of matchedReservation are RemoveReservePod 、RemoveReservation、AddPodInReservation. What's the difference？Why not just call RemoveReservePod.

ZiMengSheng, line 223: the core points of the plugin cache returning logic of matchedReservation are RemoveReservePod 、RemoveReservation、AddPodInReservation. What's the difference？Why not just call RemoveReservePod.

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,361 @@ import (
...
+		if len(matched) == 0 {
 			return
 		}
 
-		// NOTE: when the pod can allocate any reservation on the node, we should alter the nodeInfo snapshot to skip
-		//  the affinity/anti-affinity/topo constrains filtering in InterPodAffinity and PodTopologySpread plugins.
-		preparePreFilterNodeInfo(handle, nodeInfo, pod, matchedCache)
-		klog.V(4).InfoS("BeforePreFilter get matched reservations", "pod", klog.KObj(pod),
-			"node", node.Name, "count", count)
+		restorePVCRefCounts(pl.handle.SharedInformerFactory(), nodeInfo, pod, matched)
```
ZiMengSheng, line 359: why not restore PVCRefcounts afterPrefilter?

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -22,210 +22,361 @@ import (
...
+		reservePod := reservationutil.NewReservePod(rInfo.reservation)
+		if err := nodeInfo.RemovePod(reservePod); err != nil {
+			klog.Errorf("Failed to remove reserve pod %v from node %v, err: %v", klog.KObj(rInfo.reservation), nodeInfo.Node().Name, err)
+			return err
+		}
+
+		for i := range reservePod.Spec.Containers {
+			reservePod.Spec.Containers[i].Resources.Requests = corev1.ResourceList{}
+		}
+		remainedResource := quotav1.SubtractWithNonNegativeResult(rInfo.allocatable, rInfo.allocated)
```
ZiMengSheng, line 137: afterPreFilter restore unmatched（将unmatched且Allocated>0的Reservation重复计算资源归还掉）为啥不直接减去Allocated
why not just subtract allocated(repeatedly counted)

pkg/scheduler/plugins/reservation/transformer.go
eahydra: The purpose of updating Generation is to reduce the calculation amount of nodeinfo snapshot. Most nodes may not change during the scheduling phase, that is, the generation is the original, so this node does not need to be updated. The Generation is -1, which means that nNodeInfo has been modified and must be updated.

pkg/scheduler/plugins/reservation/transformer.go
eahydra: 1. In Filter phase, the parameter `NodeInfo` comes from the internal cache of `GenericScheduler` that is original data, this data also retains Reserve Pod data, so it needs to be processed again before executing Filter.
2. Processing in AfterPreFilter, relatively speaking, the implementation will be unified, and both upstream plug-ins and internal plug-ins can be considered, and the implementation is simpler and easier to maintain

pkg/scheduler/plugins/reservation/nominator.go
eahydra: NominateReservation is a method called ScoreReservation that triggers more plug-ins to be called, and the most suitable one among multiple Reservations on a node is selected based on the scores. The method Score is to score the nodes. These are two completely different intentions.

pkg/scheduler/plugins/reservation/transformer.go
eahydra: There is no such function called RemoveReservePod, but the k8s scheduling framework PreFilterExtension.RemovePod is reused to correct the state of the native plug-in, and its input parameter podToRemove is Reserve Pod. RemoveReservation and AddPodInReservation are to support fine-grained resource scheduling scenarios. The reason why these two functions are added is that RemovePod represents a preemption scenario, and the state returned by podToRemove can be preempted, but RemoveReservation represents that the current Pod should use the resources reserved by Reservation first.

pkg/scheduler/plugins/reservation/transformer.go
eahydra: There are related comments in this function, mainly to solve the problem of VolumeRestriction.

pkg/scheduler/plugins/reservation/transformer.go
eahydra: Allocated is likely to be greater than Allocatable. Because the Pod may use the remaining resources of the Reservation and the remaining resources on the node.

docs/proposals/scheduling/20220609-resource-reservation.md
```diff
@@ -105,135 +107,199 @@ In this section, a Custom Resource Definition (CRD) named `Reservation` is propo
...
-	// Resource reserved and allocatable for owners.
-	Allocatable corev1.ResourceList `json:"allocatable,omitempty"`
-	// Resource allocated by current owners.
-	Allocated corev1.ResourceList `json:"allocated,omitempty"`
+    // The `phase` indicates whether is reservation is waiting for process, available to allocate or failed/expired to
+    // get cleanup.
+    // +optional
+    Phase ReservationPhase `json:"phase,omitempty"`
+    // The `conditions` indicate the messages of reason why the reservation is still pending.
+    // +optional
```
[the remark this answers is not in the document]
eahydra: This field has been removed a long time ago. It's just that the documentation has not been updated.

PR #1141, opened 2023-03-26
koord-scheduler: support Reservation reserve devices

eahydra:

Enhanced DeviceShare scheduling plugin:
- Support reserving device via Reservation.

There are some differences between Device Resource Reservation and other resource processing methods. Other resources, such as CPU/Memory or CPU Core, are continuous, single-dimensional resources. During Pod scheduling, resources reserved by Reservation and the remaining resources on the node can be allocated together. However, Devices such as GPUs are device instances. They are continuous within a single instance, and it is impossible to combine the Device resources reserved by Reservation with the remaining Device resources of the same type on the node. This may lead to fragmentation. However, the current device resource abstraction and resource protocol can only handle this for the time being.

For example, Reservation A reserves `koordiantor.sh/gpu-core:100`, and the scheduler allocates a GPU device instance for it. If an owner Pod A allocates `koordinator.sh/gpu-core:80` first, and then an owner Pod B applies for `koordinator.sh/gpu-core:50`, Pod B cannot reuse the resources reserved by this Reservation to be allocated from the node.

Additionally, the device resources specified when Reservation reserves resources need to correspond to the Requests of the owner Pod. For example, if Reservation A reserves `koordinator.sh/gpu-core:100`, the Pod must also write `koordinator.sh/gpu-core:100` instead of `nvidia.com/gpu:1`; otherwise, the Reservation cannot be reused.

I am exploring ways to optimize this issue in future upgrades, including upgrading to k8s version1.24 and introducing the informer transformer.

### V. Checklist

### DIFF
--- a/apis/extension/scheduling.go
+++ b/apis/extension/scheduling.go
@@ -150,9 +150,10 @@ var GetDeviceAllocations = func(podAnnotations map[string]string) (DeviceAllocat
 	return deviceAllocations, nil
 }
 
-func SetDeviceAllocations(pod *corev1.Pod, allocations DeviceAllocations) error {
-	if pod.Annotations == nil {
-		pod.Annotations = map[string]string{}
+func SetDeviceAllocations(obj metav1.Object, allocations DeviceAllocations) error {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
 	}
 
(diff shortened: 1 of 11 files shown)


jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

PR #1166, opened 2023-04-02
koordlet: add event info when cgroups resources first set

### LINKED ISSUE #897: [proposal] add more events info to pod when cgroup resources (cpuset.cpus) first set
**What is your proposal**:
add events info for pods when cgroup resources first set.

**Why is this needed**:
for better observation purpose.

add an event recorder in resource-executor of koordlet.

halegreen:
add event info when cgourps resources first set, fix issue #897

fix #897

### V. Checklist

### DIFF
--- a/pkg/koordlet/resourceexecutor/executor.go
+++ b/pkg/koordlet/resourceexecutor/executor.go
@@ -21,8 +21,14 @@ import (
 	"sync"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	apiruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/audit"
 	sysutil "github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
 	"github.com/koordinator-sh/koordinator/pkg/util/cache"
 )
@@ -30,7 +36,7 @@ import (
 var _ ResourceUpdateExecutor = &ResourceUpdateExecutorImpl{}
 
 type ResourceUpdateExecutor interface {
-	Update(cacheable bool, updater ResourceUpdater) (updated bool, err error)
+	Update(cacheable bool, updater ResourceUpdater, eventHelper *audit.EventHelper) (updated bool, err error)
 	UpdateBatch(cacheable bool, updaters ...ResourceUpdater)
 	// LeveledUpdateBatch is to cacheable update resources by the order of resources' level.
 	// For cgroup interfaces like `cpuset.cpus` and `memory.min`, reconciliation from top to bottom should keep the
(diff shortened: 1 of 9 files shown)


pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -47,16 +53,31 @@ type ResourceUpdateExecutorImpl struct {
...
+func NewResourceExecutorEventRecorder(schema *apiruntime.Scheme, kubeClient kubernetes.Interface, nodeName string, node *corev1.Node) {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	recorder := eventBroadcaster.NewRecorder(schema, corev1.EventSource{Component: "koordlet-resource-executor", Host: nodeName})
+	recorderSingleton = recorder
+	nodeSingleton = node
+}
+
 func NewResourceUpdateExecutor() ResourceUpdateExecutor {
+	singleton.eventRecorder = recorderSingleton
```
saintube, line 79: Why separate the initialization of the executor and the recorder but not new the event recorder in the singleton's definition?

pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -69,7 +90,14 @@ func (e *ResourceUpdateExecutorImpl) Update(cacheable bool, resource ResourceUpd
 		}
 		return e.updateByCache(resource)
 	}
-	return true, e.update(resource)
+
+	if err := e.update(resource); err == nil {
+		executorMessage := fmt.Sprintf("successfully set resource: %s to value: %s", resource.Key(), resource.Value())
+		e.eventRecorder.Eventf(e.node, corev1.EventTypeNormal, reason.SetCPUSetCPUS, executorMessage)
```
saintube, line 96: It might broadcast too many events at the node, though the events are not complete since the cgroup could be updated in the `updateByCache` and the `UpdateBatch`.

halegreen: `It might broadcast too many events at the node.`  
Is here only need to record event when cgroups resource first set? 
while I cannot find where the  cgroups resource first set, would you give some tips ?  it's would be helpful.

pkg/koordlet/resourceexecutor/executor.go
halegreen: OK , fixed it.

pkg/koordlet/resourceexecutor/executor.go
halegreen: OK, changed it to only record event when inject resource change from runtimehooks,  where I guess do the init of the pod cgourp resources.

pkg/koordlet/resourceexecutor/executor.go
halegreen: OK, fixed it, PTAL

pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -220,19 +243,25 @@ func (e *ResourceUpdateExecutorImpl) needUpdate(updater ResourceUpdater) bool {
...
 	err := updater.update()
 	if err != nil {
 		klog.V(4).Infof("failed to update resource %s to %v, err: %v", updater.Key(), updater.Value(), err)
 		return err
 	}
 	klog.V(6).Infof("successfully update resource %s to %v", updater.Key(), updater.Value())
+
+	if eventHelper != nil && e.eventRecorder != nil {
+		executorMessage := fmt.Sprintf("successfully update resource: %s to value: %s", updater.Key(), updater.Value())
+		e.eventRecorder.Eventf(e.node, corev1.EventTypeNormal, eventHelper.Event.Reason, executorMessage)
```
zwzhang0107, line 256: event is better to add on pod, besides, it seems that all callers of update function set the eventHelper as nil?

saintube: +1

pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -47,29 +53,46 @@ type ResourceUpdateExecutorImpl struct {
...
+func NewResourceExecutorEventRecorder(schema *apiruntime.Scheme, kubeClient kubernetes.Interface, nodeName string, node *corev1.Node) {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	recorder := eventBroadcaster.NewRecorder(schema, corev1.EventSource{Component: "koordlet-resource-executor", Host: nodeName})
+	recorderSingleton = recorder
+	nodeSingleton = node
 }
 
 func NewResourceUpdateExecutor() ResourceUpdateExecutor {
+	if singleton == nil {
```
saintube, line 75: need to resolve the concurrency problem
Why not just initialize the event broadcaster in the definition of the executor?

e.g.

```go
var singleton = &ResourceUpdateExecutorImpl{
  eventBroadCaster: record.NewBroadcaster(),
  // ...
}
```

pkg/koordlet/resourceexecutor/executor.go
```diff
@@ -83,7 +106,7 @@ func (e *ResourceUpdateExecutorImpl) UpdateBatch(cacheable bool, updaters ...Res
 		}
 
 		for _, updater := range updaters {
-			isUpdated, err := e.updateByCache(updater)
+			isUpdated, err := e.updateByCache(updater, nil)
```
saintube, line 109: Strangely, there are some updates that have events while some other updates do not. Could you please adopt the event recorder uniformly or provide some information about why should not?

halegreen: the `eventHelper` is not nil ,  when only passed from the `update` function at `/pkg/koordlet/runtimehooks/protocal/protocal.go` ,  cause I think it's where we have the very first create pod/container event ?  
please let me know  if  I got something wrong, thanks!

```
func update(e resourceexecutor.ResourceUpdateExecutor, updater resourceexecutor.ResourceUpdater, eventHelper *audit.EventHelper) error {
	_, err := e.Update(true, updater, eventHelper)
	if err != nil {
		return err
	}
	return nil
}
```

pkg/koordlet/resourceexecutor/executor.go
halegreen: @zwzhang0107 
the eventHelper is not nil , when only passed from the update function at /pkg/koordlet/runtimehooks/protocal/protocal.go , cause I think it's where we have the very first  set cgourps event when creating pod/container ?
please let me know if I got something wrong, thanks!

pkg/koordlet/resourceexecutor/executor.go
saintube: I don't get your point. Could you please elaborate on what cgroup events you think need to be set?
Please note that cgroups set by CPUBurst, CPUSuppress, and other features can be different from the `very first set cgroups` set by the runtime hooks on cgroup files and values. For instance, the CPUBurst sets the pod cgroup's `cpu.cpu_burst_us` while no runtime hook does.

zwzhang0107:
hi @halegreen, is there any new idea about this?

PR #1168, opened 2023-04-02
koord-scheduler:  DeviceShare ensure NodeName is not empty when handling new Pod

jasonliu747:
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/pod_handler.go
+++ b/pkg/scheduler/plugins/deviceshare/pod_handler.go
@@ -46,6 +46,10 @@ func (n *nodeDeviceCache) onPodAdd(obj interface{}) {
 		return
 	}
 
+	if pod.Spec.NodeName == "" {
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/deviceshare/pod_handler.go
```diff
@@ -46,6 +46,11 @@ func (n *nodeDeviceCache) onPodAdd(obj interface{}) {
 		return
 	}
 
+	if pod.Spec.NodeName == "" {
+		klog.Errorf("failed to get node name from pod %v", klog.KObj(pod))
```
eahydra, line 50: There is no need to add logs here, pod.spec.nodeName is normal.  method `onPodDelete` also need to check the spec.nodeName

jasonliu747: fixed

eahydra (approved):
/lgtm
/approve

PR #1174, opened 2023-04-03
koord-descheduler: the reservation created during migration skips the original node

### LINKED ISSUE #1114: [proposal]How to avoid A evicted pod is scheduled to the node, which is  just evicted a pod 。
How to avoid A evicted pod is scheduled to the node, which is  just evicted a pod 。

eahydra:

The Reservation created by the migrationController of koord-descheduler may be scheduled to the node where the migrated Pod is located. Although this is the most suitable node from the perspective of the scheduler, it is not suitable from the perspective of descheduling. Therefore, when migrating The created Reservation should skip the original node.

When creating a Reservation, skipped terms will be added to each required term in the original Pod.Spec.Affinity.NodeAffinity to ensure that any condition will not be dispatched to the original node.

fix #1114 

### V. Checklist

### DIFF
--- a/pkg/descheduler/controllers/migration/reservation/util_test.go
+++ b/pkg/descheduler/controllers/migration/reservation/util_test.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/descheduler/controllers/migration/reservation/util_test.go
```diff
@@ -0,0 +1,192 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
ZiMengSheng, line 184: Maybe we should check whether UpdatedReservationOptions works by getRequiredNodeAffnity and matching node.

PR #1197, opened 2023-04-08
koord-scheduler: fix pods outside the Reservation preempt pods in the Reservation

### LINKED ISSUE #1187: [BUG] When a Pod preempts a Pod in a Reservation, it frequently preempts and fails to obtain resources
**What happened**:

Create a Reservation to reserve all the remaining amount of a resource on a node, then create a batch of low-priority Pods to use up the reserved resources, and then create a high-priority Pod, which will preempt the low-priority use of the Reservation Class Pods. However, high-priority Pods cannot use the preempted resources because they are occupied by the Reservation. 

When the preempted Pod subsequently creates a new instance, although it can continue to use the Reservation, it will be preempted again.

If the Pod uses Reservation, then the external Pod should not preempt the Pod in the Reservation, only the Pod in the Reservation can preempt according to rules such as priority.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

eahydra:

The pods outside the Reservation cannot preempt pods in Reservation.

fix #1187 

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_cache.go
+++ b/pkg/scheduler/plugins/deviceshare/device_cache.go
@@ -324,13 +324,18 @@ func (n *nodeDevice) tryAllocateDeviceByType(podRequest corev1.ResourceList, dev
 
 	// freeDevices is the rest of the whole machine, or is the rest of the reservation
 	freeDevices := n.deviceFree[deviceType]
+	deviceUsed := n.deviceUsed[deviceType]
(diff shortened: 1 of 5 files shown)


eahydra:
E2E is being supplemented, submitted later

eahydra:
TBR

hormes (approved):

jasonliu747 (approved):
/lgtm

PR #1211, opened 2023-04-12
koordlet: reduce the number of drop callback events

xigang:

The current size of `callbackChans` channel is set to 1. When there are too many nodeSLO and NodeTopo events, `SendCallback` will lose events. Therefore, adjust the size of `callbackChans` channel to 1000, reduce the number of drop callback events.

```
NONE
```

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/callback_runner.go
+++ b/pkg/koordlet/statesinformer/callback_runner.go
@@ -20,6 +20,8 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const maxQueueUpdateCallBack = 1000
+
 type RegisterType int64
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/koordlet/statesinformer/callback_runner.go
```diff
@@ -104,7 +106,7 @@ func (s *callbackRunner) SendCallback(objType RegisterType) {
 			klog.Infof("last callback runner %v has not finished, ignore this time", objType.String())
```
zwzhang0107, line 106: actually, it is not necessary to maintain multi callback events, since each callback will always get the latest obj after event processed.

PR #1217, opened 2023-04-13
apis: Node Resource Reservation supports apply policy

### LINKED ISSUE #1216: [proposal] NodeResourceReservation supports different apply policies
**What is your proposal**:

[Node Resource Reservation](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20221227-node-resource-reservation.md) can help reserve some resources on the node for use by components other than K8s. However, there will still be conflicts with the reservation mechanism of kubelet in some scenarios. Some new strategies need to be added to solve this conflict problem.

**Why is this needed**:

Assume that the X86 CPU of a node A has a total of 64 physical cores and 128 CPUs. A part is reserved for components other than k8s, assuming 10 CPUs, and a part of CPUs is reserved in K8s, such as 6 CPUs for some other system components, and finally `node.status.allocatable.cpu` is expected to be equal to 112 CPUs. Then kubelet needs to set the reserved parameter to 16000m. According to the existing implementation of Node Resource Reservation, in koord-scheduler, the total amount of NodeInfo.Allocatable will be modified = node.status.allocatable - 10CPUs, and finally become 102CPUs. 

This is where the conflicts arise.

Node Resource Reservation supports different strategies, which can support only excluding specified CPUs, but does not update NodeInfo.Allocatable.

eahydra:

Add `applyPolicy` to enhance compatibility of Node Resource Reservation with kubelet reservation mechanism.

fix #1216 

### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20221227-node-resource-reservation.md
+++ b/docs/proposals/scheduling/20221227-node-resource-reservation.md
@@ -10,35 +10,37 @@ reviewers:
   - "@zwzhang0107"
   - "@jasonliu747"
 creation-date: 2022-12-27
-last-updated: 2022-12-27
+last-updated: 2023-04-13
 ---
 
 # Resource Reservation
 
 ## Table of Contents
 
-<!--ts-->
+<!-- TOC -->
 
 - [Resource Reservation](#resource-reservation)
-  - [Table of Contents](#table-of-contents)
-  - [Summary](#summary)
-  - [Motivation](#motivation)
-    - [Goals](#goals)
-    - [Non-Goals/Future Work](#non-goalsfuture-work)
-  - [Proposal](#proposal)
-    - [User Stories](#user-stories)
-      - [Story 1](#story-1)
-      - [Story 2](#story-2)
-      - [Story 3](#story-3)
-    - [Implementation Details](#implementation-details)
-      - [koordlet](#koordlet)
-      - [koord-manager](#koord-manager)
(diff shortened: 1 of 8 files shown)


eahydra:
@lucming PTAL

docs/proposals/scheduling/20221227-node-resource-reservation.md
```diff
@@ -80,19 +82,44 @@ so for K8S, the 0-11 core is reserved and it can not be used.
 #### Story 2
 As a K8S cluster administrator, set the total amount of resources to be reserved directly via `node.annotation`, just like: `1C2Gi`.
 
+#### Story 3
+
+The cluster administrator just wants to reserve some cpu cores on the nods, and also wants to use reservation mechanism of kubelet to trim the allocatable resources.  For example, suppose the CPU capacity of node A is 32000m, the administrator reserve `0-3` CPU cores to the processes outside the Kubernetes, and kubelet reserve 8000m CPU to keep the `Node.Status.Allocatable` is 24000m.
```
lucming, line 87: If kubelet.reserved < node.anno.reserved, don't need to change `nodeinfo.alloc` as well?

eahydra: Yes.

lucming: ok, awesome

hormes (approved):

PR #1220, opened 2023-04-14
runtime-proxy: extract annotations from labels for docker

Solomonwisdom:

runtime-proxy: extract annotations from labels for docker

In dockershim，annotations and labels are merged because docker supports only labels.
We should extract annotations from labels for docker in failOver as we do in HandleCreateContainer of RuntimeManagerDockerServer.

None

### V. Checklist

### DIFF
--- a/pkg/runtimeproxy/server/docker/server.go
+++ b/pkg/runtimeproxy/server/docker/server.go
@@ -125,10 +125,11 @@ func (d *RuntimeManagerDockerServer) failOver(dockerClient proxyDockerClient) er
 
 	// need to backup pod meta first
 	for _, s := range sandboxes {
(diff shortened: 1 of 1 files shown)


hormes:
Why do you need to do this, can you add some description to help everyone understand its purpose?

Solomonwisdom:
In dockershim，annotations and labels are merged because docker supports only labels.
We should extract annotations from labels for docker in failOver as we do in HandleCreateContainer of RuntimeManagerDockerServer.
You can look [this](https://github.com/koordinator-sh/koordinator/blob/main/pkg/runtimeproxy/server/docker/handler.go#L57)

hormes:
I got it, thanks for contributing this pr.

hormes (approved):

PR #1223, opened 2023-04-14
koordlet: adjust node comparison method to improve program performance

xigang:

adjust node comparison method to improve program performance for nodeInformer event handler.

```
NONE
```

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/states_node.go
+++ b/pkg/koordlet/statesinformer/states_node.go
@@ -77,7 +77,7 @@ func (s *nodeInformer) Setup(ctx *pluginOption, state *pluginState) {
 				klog.Errorf("unable to convert object to *corev1.Node, old %T, new %T", oldObj, newObj)
 				return
 			}
-			if reflect.DeepEqual(oldNode, newNode) {
+			if newNode.ResourceVersion == oldNode.ResourceVersion {
 				klog.V(5).Infof("find node %s has not changed", newNode.Name)
 				return
 			}


zwzhang0107:
how about also checking the content after version, since other field maybe updated.

xigang:
> how about also checking the content after version, since other field maybe updated.

@zwzhang0107 If the resource object is persisted to etcd, the `resourceVersion` of the resource object is increased. Do you mean that the `Node` resource object is also determined by replacing the `Generation` with the `resourceVersion`?

zwzhang0107:
> > how about also checking the content after version, since other field maybe updated.
> 
> @zwzhang0107 If the resource object is persisted to etcd, the `resourceVersion` of the resource object is increased. Do you mean that the `Node` resource object is also determined by replacing the `Generation` with the `resourceVersion`?
> 
> `Generation` only when the `Spec` content of the resource object changes, will be increases.

I mean how about this:
`if newNodeMetric.Generation == oldNodeMetric.Generation || deepEqualCheck()`
in case got spec with the same value

xigang:
> > > how about also checking the content after version, since other field maybe updated.
> > 
> > 
> > @zwzhang0107 If the resource object is persisted to etcd, the `resourceVersion` of the resource object is increased. Do you mean that the `Node` resource object is also determined by replacing the `Generation` with the `resourceVersion`?
> > `Generation` only when the `Spec` content of the resource object changes, will be increases.
> 
> I mean how about this: `if newNodeMetric.Generation == oldNodeMetric.Generation || deepEqualCheck()` in case got spec with the same value

@zwzhang0107 I don't think it's necessary to add a `deepEqualCheck()`. If `newNodeMetrics` and `oldNodeMetrics` correspond to the same `Spec`, then their `metadata.Generaton` is the same.
https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation

novahe:
@xigang According to [what-is-the-difference-between-a-resourceversion-and-a-generation](https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation)
> in some objects, generation is incremented by the server as part of persisting writes affecting the spec of an object.

`in some objects`: This should not include the `Node`. After looking at the source code implementation, the generation is modified by the `Deployment`(used for `Deployment` version rollback)  or other controller, but  the controller of the `Node` ignores this field.

Back to the problem, when code enter from `Informer.UpdateFunc`, the resource has been modified. Therefore, `deepEqualCheck()` is always false. (refer to: deploy/ds/job‘s controller)

https://github.com/kubernetes/kubernetes/blob/8bba479f03fd27f1a54fcd1e02347c1a5b98b30c/pkg/controller/daemon/daemon_controller.go#L231-L250

zwzhang0107:
> > @xigang According to [what-is-the-difference-between-a-resourceversion-and-a-generation](https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation)
> > > in some objects, generation is incremented by the server as part of persisting writes affecting the spec of an object.
> > 
> > 
> > `in some objects`: This should not include the `Node`. After looking at the source code implementation, the generation is modified by the `Deployment`(used for `Deployment` version rollback) or other controller, but the controller of the `Node` ignores this field.
> > Back to the problem, when code enter from `Informer.UpdateFunc`, the resource has been modified. Therefore, `deepEqualCheck()` is always false. (refer to: deploy/ds/job‘s controller)
> > https://github.com/kubernetes/kubernetes/blob/8bba479f03fd27f1a54fcd1e02347c1a5b98b30c/pkg/controller/daemon/daemon_controller.go#L231-L250
> 
> @novahe Yes, you're right. Some object updates ignore `metadata.Generation` field (for example:`Node`, `ConfigMaps`). But the `NodeMetric` object can determine whether it changes by generation. So we don't need `deepEqualCheck()`. /cc @zwzhang0107 Please review the code and see if there is any problem?
> 
> Thanks.

spec data maybe updated twice with same value( A-> B -> A), is it possible that `the second update(B)` maybe lost because of network problem?

xigang:
> > > @xigang According to [what-is-the-difference-between-a-resourceversion-and-a-generation](https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation)
> > > > in some objects, generation is incremented by the server as part of persisting writes affecting the spec of an object.
> > > 
> > > 
> > > `in some objects`: This should not include the `Node`. After looking at the source code implementation, the generation is modified by the `Deployment`(used for `Deployment` version rollback) or other controller, but the controller of the `Node` ignores this field.
> > > Back to the problem, when code enter from `Informer.UpdateFunc`, the resource has been modified. Therefore, `deepEqualCheck()` is always false. (refer to: deploy/ds/job‘s controller)
> > > https://github.com/kubernetes/kubernetes/blob/8bba479f03fd27f1a54fcd1e02347c1a5b98b30c/pkg/controller/daemon/daemon_controller.go#L231-L250
> > 
> > 
> > @novahe Yes, you're right. Some object updates ignore `metadata.Generation` field (for example:`Node`, `ConfigMaps`). But the `NodeMetric` object can determine whether it changes by generation. So we don't need `deepEqualCheck()`. /cc @zwzhang0107 Please review the code and see if there is any problem?
> > Thanks.
> 
> spec data maybe updated twice with same value( A-> B -> A), is it possible that `the second update(B)` maybe lost because of network problem?

@zwzhang0107 There is no problem with network jitter. When the `NodeMetric` spec value changes from `A->B->A`, the `NodeMetric` `metadata.Generation` value increases continuously. For example:

- A->B: `metadata.Generation` value from 1 to 2.
- B->A: `metadata.Generation` value from 2 to 3.

zwzhang0107:
> ( A-> B -> A)

yep, so when koordlet receives metadata.Generation bumps from 1 to 3, there is no need to do the update since the spec info not changed.

xigang:
> 

@zwzhang0107  Yes, you are right, there is such a case. Let me remove part of the checking logic.

xigang:
@zwzhang0107 Please take a look this code needs to merge?

zwzhang0107:
/lgtm
/approve

Thanks! Acturally, we can still keep the generation comparing as an `OR` condition, which can save the cost of deep-equal comparing. @xigang 
`if newNodeMetric.Generation == oldNodeMetric.Generation || deepEqualCheck()`

xigang:
> /lgtm /approve
> 
> Thanks! Acturally, we can still keep the generation comparing as an `OR` condition, which can save the cost of deep-equal comparing. @xigang `if newNodeMetric.Generation == oldNodeMetric.Generation || deepEqualCheck()`

@zwzhang0107 I will submit a pr for this issue later.

xigang:
> > /lgtm /approve
> > Thanks! Acturally, we can still keep the generation comparing as an `OR` condition, which can save the cost of deep-equal comparing. @xigang `if newNodeMetric.Generation == oldNodeMetric.Generation || deepEqualCheck()`
> 
> @zwzhang0107 I will submit a pr for this issue later.

@zwzhang0107 pr has been submitted.

PR #1224, opened 2023-04-16
proposal: NUMA Topology Scheduling

### LINKED ISSUE #228: koord-scheduler supports NUMA Topology Alignment policy
**What is your proposal**:

The Topology Manager is a kubelet component that aims to coordinate the set of components that are responsible for these optimizations. After Topology Manager was introduced the problem of launching pod in the cluster where worker nodes have different NUMA topology and different amount of resources in that topology became actual. The Pod could be scheduled in the node where the total amount of resources is enough, but resource distribution could not satisfy the appropriate Topology policy. 

When allocating CPUs for LSE/LSR Pods to implement static binding, NUMA Topology also needs to be considered.

More details please refer to #96 

**Why is this needed**:

eahydra:

The scheduler and koordlet of Koordinator have the ability to allocate resources according to NUMA Topology and align them with NUMA Node as much as possible, but currently only support CPU resources. In scenarios such as model training in the field of machine learning, it is required to allocate GPU, RDMA, and CPU at the same time, and also requires them to be on the same NUMA Node or even the same PCI-E, so as to obtain optimal performance.

Although Kubernetes’ kubelet provides Topology Manager, it works on the node side and is not aware of the scheduler. This may cause the scheduled Pod to encounter Topology-related errors when it is scheduled to a certain node, resulting in the Pod being unable to start. And this kind of error is sometimes unacceptable. For example, some job-class workloads may cause the entire job to fail due to this reason. Another problem is that it is impossible to obtain cluster-level optimal NUMA Node resources.

Therefore, a mechanism needs to be implemented to solve this requirement, which can coordinate different resource allocation plugins to uniformly align according to NUMA Topology and obtain optimal performance.

fix #228 

### V. Checklist


docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,236 @@
...
+# Proposal
+
+## Design Detail
+
+### NUMA Topology Policy
+
+The label `node.koordinator.sh/numa-topology-policy` represents that how to aligning resource allocation according to the NUMA topology. 
+- `None` is the default policy and does not perform any topology alignment.
+- `BestEffort` indicates that preferred select NUMA Node that is topology alignment, and if not, continue to allocate resources to Pods.
+- `Restricted` indicates that each resource requested by a Pod on the NUMA Node that is topology alignment, and if not, koord-scheduler will skip the node when scheduling. This strategy requires that all resources be on the same NUMA Node. There can be multiple NUMA Nodes. For example, if a Pod requests 33C and each NUMA Node has 32C, then it can be allocated to use two NUMA Nodes. If this Pod also needs to request GPU/RDMA, then it needs to be on the same NUMA Node as the CPU. This strategy is more flexible.
```
hormes, line 82: what's difference between BestEffort and Restricted?

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,236 @@
...
+	// Health indicates whether the device is normal
+	Health bool `json:"health,omitempty"`
+	// Resources is a set of (resource name, quantity) pairs
+	Resources corev1.ResourceList `json:"resources,omitempty"`
++	// Topology represents the topology information about the device
++	Topology *DeviceTopology `json:"topology,omitempty"`
++}
++
++type DeviceTopology struct {
++	NUMASocketID int32  `json:"numaSocketID"`
```
hormes, line 213: socketId is enough

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,236 @@
...
++	NUMANodeID   int32  `json:"numaNodeID"`
```
hormes, line 214: nodeID is enough

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: Updated description.

- `BestEffort` indicates that the node *does not strictly* allocate resources according to NUMA Topology alignment. As long as the remaining total amount of nodes meets the needs of Pods, the scheduler can always allocate such nodes to Pods.
- `Restricted` indicates that the node allocates resources *strictly* according to NUMA Topology alignment, that is, the scheduler is required to only select the same one or more NUMA Nodes when allocating multiple resources, otherwise the node should not be used.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: update

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: update

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Compatibility
+
+- Because Koordinator Scheduler adopts the same Preferred NUMA Node Affinity calculation method as kubelet Topology Manager, and Koordinator Scheduler can perceive the Topology Policy configured by kubelet, theoretically, the two can maintain an identical logic, so there will no longer be TopologyError-related errors even when Topology Manager is enabled on the node. However, in a production environment, it is still advisable to consider disabling kubelet's Topology Manager to avoid unknown risks as much as possible.
+- Koordinator Scheduler only supports resource allocation at the Pod level, so if kubelet is configured at the Container level, conflicts may occur.
+- Koordinator's NUMA Topology Scheduling can only be applied to newly imported nodes. When there are existing nodes and Pods running on them, it is best to clear the Pods on the node before configuring the NUMA Topology Policy.
```
zwzhang0107, line 234: Is it a strict requirement that we must drain out all pods on node before we apply koord-topo-policy on Node?

Otherwise, if there are already some pods allocated by kubelet, koordlet should report the allocated info on NodeTopo CR to let koord-scheduler knows it.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,238 @@
...
+
+If the current NUMA Topology Policy of the node is BestEffort, we can also consider handling it according to the original elastic mechanism, and only enhance the restrictions when it is Restricted/SingleNUMANode.
+
+In the reporting logic of the Device CRD, we need to adapt to support new fields. When reporting the NodeResourceTopology, we need to report the NUMA Topology policy of kubelet and write it to the field NodeResourceTopology.TopologyPolicies according to the NodeResourceTopology specification.
+
+### Compatibility
+
+- Because Koordinator Scheduler adopts the same Preferred NUMA Node Affinity calculation method as kubelet Topology Manager, and Koordinator Scheduler can perceive the Topology Policy configured by kubelet, theoretically, the two can maintain an identical logic, so there will no longer be TopologyError-related errors even when Topology Manager is enabled on the node. However, in a production environment, it is still advisable to consider disabling kubelet's Topology Manager to avoid unknown risks as much as possible.
```
zwzhang0107, line 232: For example, If koord-topo-policy=bestEffort and kubelet-topo-policy=single-numa-node, TopologyError may occur.
Maybe we should limit that kubelet-topo-policy must be set as BestEffort for simplification.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,238 @@
...
+- `Restricted` indicates that the node allocates resources *strictly* according to NUMA Topology alignment, that is, the scheduler is required to only select the same one or more NUMA Nodes when allocating multiple resources, otherwise the node should not be used. There can be multiple NUMA Nodes. For example, if a Pod requests 33C and each NUMA Node has 32C, then it can be allocated to use two NUMA Nodes. If this Pod also needs to request GPU/RDMA, then it needs to be on the same NUMA Node as the CPU. This strategy is more flexible.
+- `SingleNUMANode` is similar to Restricted, that is, it is also strictly aligned according to NUMA Topology, but unlike Restricted, Restricted allows multiple NUMA Nodes to be used, while SingleNUMANode only allows the use of one NUMA Node.
+
+If there is no `node.koordinator.sh/numa-topology-policy` in the node's label and TopologyPolicies=None in NodeResourceTopology, it will be executed according to the policy configured by the koord-scheduler.
```
zwzhang0107, line 85: What is the behavior if we change the `node.koordinator.sh/numa-topology-policy` for old and new allocated pods, for example from BestEffort to SingleNUMANode.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,238 @@
...
+
+The user applies for CPU, GPU and RDMA at the same time, and expects the allocated resources to be aligned according to the NUMA node. At this time, it may be necessary to use multiple NUMA nodes.
+
+# Proposal
+
+## Design Detail
+
+### NUMA Topology Policy
+
+The label `node.koordinator.sh/numa-topology-policy` represents that how to aligning resource allocation according to the NUMA topology.
```
zwzhang0107, line 79: How about considering support policy protocol on pod labels, for example the user story:

There is a workload have performance issue, and I want to make a quick fix to improve it. So I change the label on pod template and redeploy it.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: Yes, before applying the NUMA policy, the user needs to ensure that there are no other Pods on the node (except DaemonSet).

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: Yes. It is better to turn off the topology manager of kubelet or set none or bestEffort when users actually use it

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: We don’t do anything in this scenario. Of course, you can also consider that once a node has set the NUMA Policy, if you want to change it, you need to ensure that there are no other Pods on the node, or allow setting from a strict policy to a loose policy.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
eahydra: It's a good idea. But it also brings additional complexity. The reason is that it is necessary to ensure that Pods with the same policy can be effective together.

docs/proposals/scheduling/20230415-numa-topology-scheduling.md
```diff
@@ -0,0 +1,238 @@
...
+
+By default, MostAllocated algorithm is used for scoring, which prioritizes nodes with fewer remaining NUMA Nodes. It also supports scoring based on strategies specified by the node.koordinator.sh/numa-allocate-strategy label on nodes, but only supports two strategies corresponding to this label: MostAllocated and LeastAllocated.
+
+MostAllocated and LeastAllocated strategies are both calculated based on the relationship between the number of used NUMA Nodes and the total number of NUMA Nodes used. Taking MostAllocated as an example, if a node has 4 NUMA Nodes but one has been allocated completely and one more is needed during this scheduling process, then the calculation formula is as follows: *Score = (2 * 100)/4 = 50*
```
saintube, line 123: Do we need to prioritize the absolute number of available NUMA Nodes? There may be machines with different NUMA nodes, and in some cases, the machine with the most numa nodes is preferred.

eahydra: Good idea. It can be used as a follow-up optimization item.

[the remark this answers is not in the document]
eahydra: updated.

saintube: OK

hormes (approved):

PR #1235, opened 2023-04-20
koord-manager: add args for common controllers

saintube:

Add command options for the common controllers. We can enable/disable the known controllers via the arg `--controllers`.

### V. Checklist

### DIFF
--- a/cmd/koord-manager/main.go
+++ b/cmd/koord-manager/main.go
@@ -18,15 +18,18 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"math/rand"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


jasonliu747 (approved):
/lgtm

FillZpp:
Should this be in feature-gate or a command-line flag?

For example, kube-controller-manager has `--feature-gates` to enable those features in controllers and `--controllers` to indicate which controllers should be started.

saintube:
> Should this be in feature-gate or a command-line flag?
> 
> For example, kube-controller-manager has `--feature-gates` to enable those features in controllers and `--controllers` to indicate which controllers should be started.

I think both options are acceptable since these controllers have different maturities. The main reason I chose the feature gate is that the webhook plugins are maintained in this way.

hormes:
> --controllers to indicate which controllers should be started

I think this is more accurate. The feature gate is a function switch, and the controller is more like a program. Some behaviors inside the controller should be controlled by the feature gate.

saintube:
> > --controllers to indicate which controllers should be started
> 
> I think this is more accurate. The feature gate is a function switch, and the controller is more like a program. Some behaviors inside the controller should be controlled by the feature gate.

@FillZpp @hormes Fixed. PTAL.  
Now we use `--controllers` to control which controllers should be added to the manager.

cmd/koord-manager/main.go
```diff
@@ -101,6 +111,9 @@ func main() {
 	flag.BoolVar(&enablePprof, "enable-pprof", true, "Enable pprof for controller manager.")
 	flag.StringVar(&pprofAddr, "pprof-addr", ":8090", "The address the pprof binds to.")
 	flag.StringVar(&syncPeriodStr, "sync-period", "", "Determines the minimum frequency at which watched resources are reconciled.")
+	pflag.StringSliceVar(&controllers, "controllers", allControllers, fmt.Sprintf("A list of controllers to enable. '*' enables all on-by-default controllers,"+
```
eahydra, line 114: We still have a small number of controllers inside, which are controlled by feature-gate. Then follow this to modify?

cmd/koord-manager/main.go
```diff
@@ -78,10 +80,18 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+var controllers []string
+
+var allControllers = []string{
```
eahydra, line 85: Do we need to define allControllers? Adding a controller also requires changing two variables at the same time. Consider simplifying.

saintube: fixed. `controllers = []string{...}` is enough

cmd/koord-manager/main.go
```diff
@@ -78,10 +80,16 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+var controllers = []string{
```
eahydra, line 83: Here we can do a little optimization
```suggestion
var controllerAddFuncs = map[string]func(manager.Manager) error{
	nodemetric.Name:   nodemetric.Add,
	noderesource.Name: noderesource.Add,
	nodeslo.Name:      nodeslo.Add,
}

var controllers = set.StringKeySet(controllerAddFuncs)
```

eahydra (commented):
/lgtm

hormes (approved):

PR #1242, opened 2023-04-24
koord-descheduler: add leaderelection metric

ZiMengSheng:

Provide Leaderelection metrics for koord-descheduler.

### V. Checklist

### DIFF
--- a/cmd/koord-descheduler/main.go
+++ b/cmd/koord-descheduler/main.go
@@ -27,6 +27,9 @@ import (
 	"k8s.io/component-base/logs"
 
 	"github.com/koordinator-sh/koordinator/cmd/koord-descheduler/app"
+
+	// Ensure metric package is initialized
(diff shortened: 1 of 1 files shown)


cmd/koord-descheduler/app/server.go
```diff
@@ -236,7 +240,18 @@ func buildHandlerChain(handler http.Handler) http.Handler {
...
+		ErrorHandling: promhttp.HTTPErrorOnError,
+	}))
+}
+
+type MergedRegistry struct {
+	prometheus.Gatherers
+}
+
+func (mr MergedRegistry) Gather() ([]*dto.MetricFamily, error) {
+	mfs, _ := mr.Gatherers.Gather()
```
eahydra, line 253: we should not ignore the error.

FillZpp (approved):

PR #1259, opened 2023-04-28
bugfix: when cpuset is an empty string, error log

leason00:
error:
image

### V. Checklist


zwzhang0107:
for proxy mode, empty string is necessary to erase BE cpuset if koordlet is setup with static cpu-policy. @saintube @leason00

PR #1268, opened 2023-05-05
koord-scheduler: take over snapshot generation

### LINKED ISSUE #1188: [BUG] The scoring results of k8s built-in scheduling plugins for nodes with Reservation are not as expected
**What happened**:

Create a Reservation to node A, and then create a Pod, and you will find that the Score result of the NodeResources plugin is incorrect. The reason is that these built-in plug-ins are obtained from the NodeInfo in the Snapshot in the framework.Handle, and these NodeInfo include the Reserve Pod of the Reservation.

The Reserve Pod of the Reservation should also be excluded from the NodeInfo seen during the built-in plug-in scoring to ensure that the final result is correct.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

eahydra:

Take over the field `generation` in snapshot defined in framework.Framework. If the nodeInfo transformers updated the NodeInfo or reservation scheduling updated the NodeInfo, force update the `generation` to zero to force update snapshot from cache.  So the built-in plugins in kube-scheduler and koordinator's plugins have same data view.

The disadvantage is that once a change occurs, the NodeInfo needs to be refreshed in full.

fix #1188 

### V. Checklist


jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

hormes (approved):

PR #1273, opened 2023-05-08
apis: defined Reservation.Spec.AllocateOnce as pointer 

eahydra:

Defined Reservation.Spec.AllocateOnce as pointer. 
Previously, the default value of AllocateOnce was set to true, but because of the empty tag, AllocateOnce could not be set to false, which caused problems in Reservation reuse scenarios. 

In the previous test, this problem was not exposed because the CRD Schema was not updated into test cluster.

### V. Checklist


jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

PR #1021, opened 2023-02-17
koordlet: use libcontainer check cgroupfsv2

huiwq1990:
### Ⅰ. Describe what this PR does

Use libcontainer check if the os support cgroupfsv2, and the default value for `system.UseCgroupsV2` maybe change to 
true.

### V. Checklist

- [ ] I have written necessary docs and comments
- [] I have added necessary unit tests and integration tests
- [x] All checks passed in `make test`

PR #1275, opened 2023-05-08
koordlet: fix host cgroup check failed when the config is not the default

saintube:
…ault

Fix the koordlet panic when the host path configs are changed.

After #1021, the koordlet panics when the host cgroup path is not correctly configured. However, the check is invoked in `init()` in which the value of the `CgroupRoot` has not been parsed from the command line arguments. So the koordlet always panics when the host cgroup is not mounted at the default path `/host-cgroup`. 

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1284, opened 2023-05-10
koord-scheduler: optimize the flow of restore Reservation

eahydra:

Optimize the flow of restore Reservation from NodeInfo in BeforePreFilter instead of AfterPreFilter.

In the previous implementation, it was necessary to cooperate to modify the scheduling state before and after the PreFilter to ensure the consistency of the data view. The overall performance is relatively poor, and some of the original interfaces cannot help to process data correctly. For example, unmatched reservations are not passed to each plugin, resulting in some data not being corrected, resulting in scheduling results in some scenarios that do not meet expectations.

Now it is optimized to process data concurrently in BeforePreFilter, and removes a lock, which improves the overall performance, and can better coordinate each plugin to process their own internal state.

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/framework_extender.go
+++ b/pkg/scheduler/frameworkext/framework_extender.go
@@ -46,11 +46,11 @@ type frameworkExtenderImpl struct {
 	filterTransformers    []FilterTransformer
 	scoreTransformers     []ScoreTransformer
 
-	reservationFilterPlugins       []ReservationFilterPlugin
-	reservationScorePlugins        []ReservationScorePlugin
-	reservationNominatorPlugins    []ReservationNominator
-	reservationPreBindPlugins      []ReservationPreBindPlugin
-	reservationPreFilterExtensions []ReservationPreFilterExtension
+	reservationFilterPlugins    []ReservationFilterPlugin
+	reservationScorePlugins     []ReservationScorePlugin
+	reservationNominatorPlugins []ReservationNominator
+	reservationPreBindPlugins   []ReservationPreBindPlugin
+	reservationRestorePlugins   []ReservationRestorePlugin
 }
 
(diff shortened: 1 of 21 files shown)


pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -236,12 +237,14 @@ func (pl *Plugin) PostFilter(ctx context.Context, cycleState *framework.CycleSta
 		}
 		reservationInfos := pl.reservationCache.listReservationInfosOnNode(node.Name)
 		for _, rInfo := range reservationInfos {
-			newReservePod := reservationutil.NewReservePod(rInfo.Reservation)
```
eahydra, line 239: In the original implementation, the data that has been corrected by NodeInfo may be added back again

jasonliu747 (approved):
/lgtm

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -267,27 +267,41 @@ func (ext *frameworkExtenderImpl) RunPreBindPlugins(ctx context.Context, state *
...
+			return nil, status
+		}
+		pluginToRestoreState[pl.Name()] = state
+	}
+	return pluginToRestoreState, nil
+}
+
+func (ext *frameworkExtenderImpl) RunReservationExtensionFinalRestoreReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, states PluginToNodeReservationRestoreStates) *framework.Status {
+	for _, pl := range ext.reservationRestorePlugins {
+		s, ok := states[pl.Name()]
```
ZiMengSheng, line 297: Is s already exists in plugin state when pl.RestoreReservation is called? Why make it as auguments for pl.FinalRestoreReservation here. Is it useless?

PR #1289, opened 2023-05-12
koord-scheduler: fix duplicate allocate GPU after leader selection changed

eahydra:

The master and slave instances of koord-scheduler  both sync pods from api-server via informer and both build nodeDeviceCache with add events. But the slave instance ignore the update event that missed the bound pod by master instance that leaky the pod's deviceAllocations. So if the leader election changed, the slave instance will serve with old states that allocates GPUs that already be allocated for the new Pod. 

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1291, opened 2023-05-13
koord-scheduler: support Reservation allocate policy

### LINKED ISSUE #1226: [proposal]Reservations support two modes: preferred or required
**What is your proposal**:
Reservations support two modes: preferred or required.
prefferred: means the pod selected by the reservation owner uses the Reservation resource first.
required: means the pod selected by the owner must use the Reservation resource.

**Why is this needed**:

eahydra:

Support Reservation allocate policy
1. The policies `Aligned` and `Restricted` support Pod only allocates from one Reservation instance and node. 
2. If a node has multiple Reservations with `Aligned` or `Restricted` policies, and these Reservations cannot satisfy the pod’s request at all during filtering, the node will be filtered out.
3. The default policy (currently is none) cannot coexist with the policies `Aligned` and `Restricted`. 
4. Deprecated the constraint about Reservation.Spec.Owners during filter, so the Pods can allocate from resources via more than one Reservations.  If the result is not expected, users should set the `Reservation.Spec.AllocatePolicy` with `Aligned` or `Restricted`.
5. **Currently, the implementation of AllocatePolicy does not support the port reservation scenario**

fix #1226 

### V. Checklist


eahydra:
TBD,PTAL @hormes @FillZpp @tzzcfrank @ZiMengSheng @saintube @jasonliu747

apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -62,8 +62,29 @@ type ReservationSpec struct {
...
+	// ReservationAllocatePolicyDefault means that there is no restriction on the policy of reserved resources,
+	// and allocated from the Reservation first, and if it is insufficient, it is allocated from the node.
+	ReservationAllocatePolicyDefault ReservationAllocatePolicy = ""
+	// ReservationAllocatePolicyAligned indicates that the Pod allocates resources from the Reservation first.
+	// If the remaining resources of the Reservation are insufficient, it can be allocated from the node,
+	// but it is required to strictly follow the resource specifications of the Pod.
+	// This can be used to avoid the problem that a Pod uses multiple Reservations at the same time.
+	ReservationAllocatePolicyAligned ReservationAllocatePolicy = "Aligned"
+	// ReservationAllocatePolicyRestricted means that the resources requested by the Pod overlap with the resources reserved by the Reservation,
+	// then these intersection resources can only be allocated from the Reservation, but other resources can be allocated from the node.
```
hormes, line 83: > but other resources can be allocated from the node

How to understand the `other resources`, is the resource claimed by the Pod but not by the Reservation?

apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -62,8 +62,29 @@ type ReservationSpec struct {
...
+type ReservationAllocatePolicy string
+
+const (
+	// ReservationAllocatePolicyDefault means that there is no restriction on the policy of reserved resources,
+	// and allocated from the Reservation first, and if it is insufficient, it is allocated from the node.
+	ReservationAllocatePolicyDefault ReservationAllocatePolicy = ""
+	// ReservationAllocatePolicyAligned indicates that the Pod allocates resources from the Reservation first.
+	// If the remaining resources of the Reservation are insufficient, it can be allocated from the node,
+	// but it is required to strictly follow the resource specifications of the Pod.
+	// This can be used to avoid the problem that a Pod uses multiple Reservations at the same time.
```
hormes, line 80: > This can be used to avoid the problem that a Pod uses multiple Reservations at the same time

This means that the default policy can use multiple Reservations？

apis/scheduling/v1alpha1/reservation_types.go
eahydra: Other resources refer to resources other than the intersection of Pod and Reservation. For example, if a Pod applies for 4C8Gi and 1 ENI, and the Reservation reserves 4C8Gi, then 1 ENI is other resource.

apis/scheduling/v1alpha1/reservation_types.go
eahydra: Indeed. At present, in the existing implementation of the main branch, there will be situations where the Pod uses multiple Reservations at the same time. For example, the user creates ReservationA and reserves 4C, and ReservatonB reserves 8Gi of memory, and Pod A matches the two Reservations, so that Pod A uses the resources of the two Reservations. Of course, the scheduler ensures that the entire node dimension will not be oversold.

eahydra: We can consider changing the default behavior here, such as changing to Aligned or Restricted. But here is a scenario that needs to be confirmed: If all Aligned or all Restricted on a node are not satisfied, should this node be filtered out? The current implementation in this PR is filtering out, but if the default behavior is Aligned/Restricted, this filtering behavior may not be correct. Or whether to decide to filter out this node depends on whether the user's Pod has set Reservation Affinity?

WDYT @hormes

apis/scheduling/v1alpha1/reservation_types.go
hormes: Add a comment `(resources declared in the Pod but not reserved in the Reservation)` to clarify the semantics?

eahydra: Good suggestion, I will update it.

eahydra: updated.

FillZpp (approved):
/lgtm

hormes (approved):

PR #1315, opened 2023-05-20
koord-scheduler: support reserve by pod

eahydra:

Supports using real Pods to reserve resources.
- If the Pod has the special label `scheduling.koordinator.sh/operating-mode` and the value is `Reservation`, it means that the purpose of the Pod is to reserve resources.
- Such Pods are collectively called Operating Pods. It can also be allocated from a Reservation CRD, or it can be preempted like a normal Pod.
- Pod whose Operating Mode is `Reservation`, its Reservation allocation policy defaults to `Aligned`, and it is assigned only once.
- Users can set annotation `scheduling.koordinator.sh/reservation-owners` such as `Reservation.Spec.Owners` to indicate which Pods can use resources reserved by real Pods.
- Once the Operating Pod is bound by other Pods, the Operating Pod will be annotated with `scheduling.koordinator.sh/reservat-current-owner` to indicate who uses the Pod.

### V. Checklist

### DIFF
--- a/apis/extension/operating_pod.go
+++ b/apis/extension/operating_pod.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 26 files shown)


jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

hormes (approved):

hormes (approved):

PR #1317, opened 2023-05-21
koord-scheduler: implement PreBindExtensions to support one-time apply patches

eahydra:

Optimize the Patch behavior of the koord-scheduler custom plugins:
1. Added the `DefaultPrebind` plugin to initiate a Patch request. The plugin implements the new interface `PreBindExtensions`. 
2. `DefaultPreBind` will be added to `KubeSchedulerConfig` by default. The plugin can be disabled by setting the `Disabled` parameter, but after disabling, a new plugin implementation of `PreBindExtensions` needs to be provided.
3. Custom plugins such as NodeNUMAResources and DeviceShare only modify Pods during PreBind and do not initiate Patch requests.
4. In this way, multiple Patch requests can be converted into one. And it can support custom implementation of different Patch strategies.

### V. Checklist


jasonliu747 (approved):
/lgtm

saintube (commented):
/lgtm

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -270,6 +283,8 @@ func (ext *frameworkExtenderImpl) RunPreBindPlugins(ctx context.Context, state *
 	reservation = reservation.DeepCopy()
 	reservation.Status.NodeName = nodeName
 
+	original := reservation
+	reservation = reservation.DeepCopy()
```
FillZpp, line 287: Is there any way to avoid deepcopy the reservation twice?

eahydra: updated

FillZpp (approved):
/lgtm

buptcozy (approved):
/arrove

PR #1321, opened 2023-05-23
koord-descheduler: allow annotated pod only pass non-retriable filter

ZiMengSheng:

The user of podMigrationJob may know the risk of migrationPod. So we allow annotated job to pass filter. Then user can use annotation to indicate he/she knows the risk. 
However, user just want the annotated job to pass non-retryable filter and the retryable-filter(flow-control ability) should remain unchanged.

None

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1322, opened 2023-05-23
koord-manager: ClusterColocationProfile support grayscale

### LINKED ISSUE #1309: [proposal] ColocationProfile supports enhanced grayscale capabilities for matching Pods
**What is your proposal**:

It is recommended that ColocationProfile supports enhanced grayscale capabilities for matching Pods, for example, only 10% of Pods are expected to be affected.

**Why is this needed**:

Although ColocationProfile can select a batch of Pods for grayscale according to Namespace or Selector, the total number of affected Pods may still be large. Therefore, if the proportional grayscale mechanism is supported, stability risks can be better controlled. , to better guarantee the stability of the public application.

Add field `Probability *intstr.IntOrString` in ColocationProfile.Spec. The value range is 0-100. If it's 0, disable the profile; or if it's 100, all Pods can be modified; finnaly, generate a random value like rand.Intn(100), if the value is in 0-targetValue, modify the Pod.

ZiMengSheng:

Add field Probability *intstr.IntOrString in ColocationProfile.Spec. The value range is 0-100. If it's 0, disable the profile; or if it's 100, all Pods can be modified; finnaly, generate a random value like rand.Intn(100), if the value is in 0-targetValue, modify the Pod.

fixes #1309 

### V. Checklist


eahydra (changes requested):

eahydra (approved):
/lgtm
/approve

hormes (approved):

PR #1339, opened 2023-05-31
koordlet: revise cgroup update err when unsupported or path not exist

saintube:

- Revise the error handling when the cgroup resource update fails because the cgroup is unsupported or the file does not exist.

### V. Checklist

### DIFF
--- a/pkg/koordlet/resourceexecutor/cgroup.go
+++ b/pkg/koordlet/resourceexecutor/cgroup.go
@@ -56,6 +56,8 @@ func cgroupFileWriteIfDifferent(cgroupTaskDir string, r sysutil.Resource, value 
 	if currentErr != nil {
 		return false, currentErr
 	}
+	// FIXME(saintube): Instead of handling cpuset resource in writing function, we should use a updater and do
+	//  MergeUpdate in resourceexecutor's LeveledUpdateBatch.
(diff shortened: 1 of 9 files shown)


pkg/koordlet/resourceexecutor/config.go
```diff
@@ -32,15 +32,21 @@ const (
...
-		ResourceForceUpdateSeconds: 60,
+		ResourceForceUpdateSeconds:    60,
+		IgnoreResourceUnsupportedErr:  true,
+		IgnoreResourcePathNotExistErr: true,
 	}
 }
 
 func (c *Config) InitFlags(fs *flag.FlagSet) {
 	fs.IntVar(&c.ResourceForceUpdateSeconds, "resource-force-update-seconds", c.ResourceForceUpdateSeconds, "executor force update resources interval by seconds")
+	fs.BoolVar(&c.IgnoreResourceUnsupportedErr, "ignore-resource-unsupported-err", c.IgnoreResourceUnsupportedErr, "executor ignore resource unsupported error while updating resources")
```
zwzhang0107, line 50: maybe not necessary for these args

saintube: removed

PR #1343, opened 2023-06-01
api: add gpu-model-detail

ZiMengSheng:
### V. Checklist


apis/extension/resource.go
```diff
@@ -38,6 +38,7 @@ const (
 
 const (
 	LabelGPUModel         string = NodeDomainPrefix + "/gpu-model"
+	LabelGPUModelDetail   string = NodeDomainPrefix + "gpu-model-detail"
```
eahydra, line 41: missing '/'

eahydra (commented):
The label `node.koordinator.sh/gpu-model` already contains the detail model information about GPU.   I thought this label was the short name of the gpu model.

Maybe we should add a new label to represent the short name of the gpu model. For example `node.koordinator.sh/gpu-model-series`.

Or, we may have to consider making a break change. gpu-model means short name, and gpu-model-detail means complete GPU Model information.

Issue #1361, opened 2023-06-07
[proposal] Mid tier 


hormes:
![image](https://github.com/koordinator-sh/koordinator/assets/4395943/1c48340a-3645-463c-ac90-a9cccbede2ce)

PR #1365, opened 2023-06-07
add a histogram util

hormes:
Ref to #1361 

This tool comes from [kubernetes/autoscaler](https://github.com/kubernetes/autoscaler), thanks to the open source predecessors.


FillZpp (approved):
/lgtm

PR #1371, opened 2023-06-12
koord-scheduler: enhance the preemption when common resources are reserved

### LINKED ISSUE #1187: [BUG] When a Pod preempts a Pod in a Reservation, it frequently preempts and fails to obtain resources

eahydra:

After the introduction of Reservation, Pods outside the Reservation cannot preempt Pods in the Reservation. In the previous PR, this problem was solved for the Device Share scenario, but for common resources, such as CPU/Memory, this PR can be used to solve it.

fix #1187

1. Create a Reservation reserving all remained CPUs of one Node . 
2. Create a Pod using the Reservation to allocate 4000m CPU.
3. Create a high-priority Pod but does not use the Reservation and try to allocate the node that Reservation reserved.
4. If not using the PR, the high-priority Pod will preempt the Pod that using Reservation, but both of the Pods cannot be scheduled.
5. If using the PR, the high-priority Pod will not preempt the Pod that using Reservation.

More cases see the E2E tests.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


eahydra:
All E2E passed

ZiMengSheng (commented):
/lgtm

hormes (approved):

Issue #940, opened 2023-01-09
[proposal] support fastfail for gang and multiple types of permit policies, correct scheduled semantics in coscheduling


KunWuLuan:
**FastFail**
In strict mode, we mark the gang failed when any pod of the gang can''t be scheduled. Pod in this gang can''t be scheduled until all pods in this gang has come.

The process of marking failure can be optimized as follows:
- When the total num of the pods in a pod group is larger than min-available, the number of scheduled pod can still meet the min-available requirement even if some pods fail to schedule. We can count the number of remaining pods that have not been scheduled and the number of pods that have failed to be scheduled to determine whether the podgroup can meet the min available requirements. If not, we mark the pod group failed.
- When any pod group in a gang group is in the status of schedule cycle invalid, any successful schedule in this gang group is useless. Only the pods in the gang group where all gangs are available should pass the prefilter.

**Matching Policy**

Koordinator support only `OnceResourceSatisfied` mode. Sometimes some pods in the pod group failed for some reasons, and we hope that the subsequently reconstructed Pod will still be constrained by the gang semantics. So we think that the logic to determine whether gang is satisfied can be expanded to allow users to freely choose which logic to use.

- For GPU topology, all my tasks need to be scheduled at the same time to ensure the best performance, so I hope that gang can only be satisfied when the number of pods waiting on permit is larger than the min-available.
- For task of elastic training, the new workers'' arrival is random. So I hope the gang can be satisfied when the total number of the running pods and waiting pods is large than the min-available. 
- Sometimes in elastic training, some workers complete earlier than the others, and the new workers should start to training at its arrival. So I hope the gang can be satisfied when the total number of the running pods, the succeed pods and waiting pods is large than the min-available. 
- In some specific scenarios, the completed task Pod may not exist due to node recycling. At this time, it is impossible to count the number of completed tasks. In this case, OnceResourceSatisfied can be used.

**Scheduled Semantics in Status**

I think the Scheduled attribute in the status means how many pods have been scheduled, not how many existing pods have been scheduled. And the implementation in koordinator can make patch the wrong status on the pod groups because the patch are requests from multiple goroutines. I think the status of pod groups should be updated in stead of patched.

buptcozy:
“We can count the number of remaining pods that have not been scheduled and the number of pods that have failed to be scheduled to determine whether the podgroup can meet the min available requirements”

[...]

PR #1380, opened 2023-06-13
koord-scheduler: coscheduling support multiple gang match policy

xulinfei1996:
Expand the logic of determining whether gang is satisfied by "match-policy".

In some specific scenarios, the completed task Pod may not exist. At this time, it is impossible to count the number of completed tasks. In this case, once Gang has been satisfied, all subsequent resource allocations are no longer constrained by Gang rules, which is called once-satisfied, the default gang 'match-policy'.

For GPU topology, all the tasks need to be scheduled at the same time to ensure the best performance, so the gang only consider the number of pods waiting on permit is larger than the min-available, which is called only-waiting. What's more, if the workload support failover, but workload controller can't persist the new gangName or support reconstruct all pods at one-time, only-waiting can help.

For tasks of elastic training, the new workers' arrival is random, so the gang only consider the sum of running pods and waiting pods is larger than the min-available, which is called waiting-and-running. For example, if the gang's total number is 5, min available is 3 and there are 3 pods running at first. After two of the three pods completed,  the fourth pod comes, if the match-policy is once-satisfied, the fourth pod will be scheduled and bound. If the match-policy is waiting-and-running, the fourth pod will keep pending, since there are only 2 pods' status is waiting-and-running which is smaller than min-available.

fixes https://github.com/koordinator-sh/koordinator/issues/940 Matching Policy part.

### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20220901-gang-scheduling.md
+++ b/docs/proposals/scheduling/20220901-gang-scheduling.md
@@ -112,12 +112,30 @@ which allow user to bundle different Gangs together.
 It should be noted that, if the resource accumulation conditions of Gang are met, then some pods failed in the process of binding,
 or some bound pods are preempted\rescheduled, should the constraints of Gang still be effective in the process of resource reallocation? 
 Because the initial purpose of Gang is to require pods to be pulled up at the same time, if some pods have been pulled up, 
-then the subsequent Gang behavior is meaningless. Therefore, when once Gang has been satisfied, all subsequent resource allocations 
-are no longer constrained by Gang rules, and their performance is similar to ordinary pod.
+then the subsequent Gang behavior is meaningless. Therefore, once Gang has been satisfied and its match policy is once-satisfied, all 
+subsequent resource allocations are no longer constrained by Gang rules, and their performance is similar to ordinary pod.
 
 As mentioned above, `WaitTime` is the max wait time since first pod comes to permit stage. If `WaitTime` is timeout, 
 scheduler will roll back all assumed pods, update each pod's annotation with `gang.scheduling.koordinator.sh/timeout=true`, and
 won't schedule these pods anymore. User should pay attention to this status and delete pods timely.
+
+#### Gang Match Policy
+When using Gang scheduling, you can expand the logic of determining whether gang is satisfied by `match-policy`. 
+
+In some specific scenarios, the completed task Pod may not exist. At this time, it is impossible to count the number of completed tasks. In this
+case, once Gang has been satisfied,  all subsequent resource allocations are no longer constrained by Gang rules, which is called `once-satisfied`,
+the default gang 'match-policy'.
+
+For GPU topology, all the tasks need to be scheduled at the same time to ensure the best performance, so the gang only consider the 
+number of pods waiting on permit is larger than the min-available, which is called `only-waiting`. What's more, if the workload support
+failover, but workload controller can't persist the new gangName or support reconstruct all pods at one-time, `only-waiting` can
+help.
+
+For tasks of elastic training, the new workers' arrival is random, so the gang only consider the sum of running pods and waiting pods 
+is larger than the min-available, which is called `waiting-and-running`. For example, if the gang's total number is 5, min available is 3 and
+there are 3 pods running at first. When two of the three pods completed, here comes the fourth pod, the fourth pod will be scheduled and bound 
+if the `match-policy` is `once-satisfied`. If the `match-policy` is `waiting-and-running`, the fourth pod will keep pending, since there are
+only 2 pods' status is `waiting-and-running` which is smaller than min-available.
 
 ### API
 #### Definition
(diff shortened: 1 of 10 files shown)


eahydra:
Please describe more user scenarios to explain why these policies are needed? Are these policies really necessary to implement in the scheduler? 
WDYT @hormes @KunWuLuan

KunWuLuan:
I also wonder in what situations we need these match policies. Could you please describe more, thanks?

xulinfei1996:
> Please describe more user scenarios to explain why these policies are needed? Are these policies really necessary to implement in the scheduler? WDYT @hormes @KunWuLuan

Edited the gang scheduling proposal and this PR's introduction to describe the scenarios why these three policies are needed, please take a look.

docs/proposals/scheduling/20220901-gang-scheduling.md
```diff
@@ -112,13 +112,25 @@ which allow user to bundle different Gangs together.
...
-are no longer constrained by Gang rules, and their performance is similar to ordinary pod.
+then the subsequent Gang behavior is meaningless. Therefore, once Gang has been satisfied and its match policy is once-satisfied, all 
+subsequent resource allocations are no longer constrained by Gang rules, and their performance is similar to ordinary pod.
 
 As mentioned above, `WaitTime` is the max wait time since first pod comes to permit stage. If `WaitTime` is timeout, 
 scheduler will roll back all assumed pods, update each pod's annotation with `gang.scheduling.koordinator.sh/timeout=true`, and
 won't schedule these pods anymore. User should pay attention to this status and delete pods timely.
 
+#### Gang Match Policy
+When using Gang scheduling, you can expand the logic of determining whether gang is satisfied by `match-policy`.
```
eahydra, line 123: I still don't quite understand the scenes behind this. There are several issues to discuss:
1. From these descriptions, it seems that after minAvailable is satisfied, you want to continue to do Gang Scheduling for subsequent Pods again? 
2. Do these special policies have to be implemented in the scheduler? When the Job Controller (no matter which Job Controller it is) creates  a new batch of Pods, can it not set new Gang parameters for a new batch of Pods?
3. Which of these three policies is the default strategy, and is it compatible with existing implementations?

xulinfei1996: To issue 1, the answer is yes.
To issue 2, if the Job Controller need reconstruct all pods to enhance fault toleration and the pods' reconstruction behavior is to reconstruct pod one by one, Koordinator now will schedule reconstructed pod immediately, which may not meet the requirement of scheduling all pods at the same time to ensure the best performance.
To issue3,as the Koordinator's current implementation is  "once-satisfied" policy, so the default strategy is "once-satisfied" policy to compat with existing implementation.

KunWuLuan: In what situation we have to do Gang for subsequent Pods?
If job's replica is larger than min available, subsequent Pods should pass.
If job's replica is same with min available, why pods come again? If pods failed, the job status will be set as failed for tfjobs.

xulinfei1996: If job's replica is same with min available and pods failed, but the job supports failover, pods will come again.

docs/proposals/scheduling/20220901-gang-scheduling.md
```diff
@@ -112,13 +112,31 @@ which allow user to bundle different Gangs together.
...
+number of pods waiting on permit is larger than the min-available, which is called `only-waiting`. What's more, if the workload support
+failover, but workload controller can't persist the new gangName or support reconstruct all pods at one-time, `only-waiting` can
+help.
+
+For tasks of elastic training, the new workers' arrival is random, so the gang only consider the sum of running pods and waiting pods 
+is larger than the min-available, which is called `waiting-and-running`. For example, if the gang's total number is 5, min available is 3 and
+there are 3 pods running at first. When two of the three pods completed, here comes the fourth pod, the fourth pod will be scheduled and bound 
+if the `match-policy` is `once-satisfied`. If the `match-policy` is `waiting-and-running`, the fourth pod will keep pending, since there are
+only 2 pods' status is `waiting-and-running` which is smaller than min-available.
+
```
hormes, line 139: The description here is much clearer, thanks for the more detailed expansion.

docs/proposals/scheduling/20220901-gang-scheduling.md
KunWuLuan: > If job's replica is same with min available and pods failed, but the job supports failover, pods will come again.

If job supports failover, why we should block the subsequent arrival pods? I mean, if all pods should be deleted and created, we can create a new gang, otherwise once-satisfied is enough.

xulinfei1996: Yes, create a new gang is a solution. However, not all workload controller can persist the new gangName, and only-waiting may help.

KunWuLuan: Is there any open source job controller supports failover? How do they support failover and distinguish pods from different rounds？ 
If a job supports failover and all pods should be deleted and created, I think the better way is to include rounds in pod name, this is not difficult. 
@eahydra WDYT

hormes (approved):

PR #1389, opened 2023-06-14
koordlet: add a resource usage predictserver to estimate long-term reclaimed resource

hormes:
Ref #1361

This PR contains:
- [x]  collecte metrics and store in the histogram
- [x]  checkpoint histogram and restore

Capture Pod/Node utilization data and update it into the model, and provide the ability to query results.

### DIFF
--- a/pkg/koordlet/prediction/checkpoint_test.go
+++ b/pkg/koordlet/prediction/checkpoint_test.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prediction
+
+import (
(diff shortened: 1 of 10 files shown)


pkg/koordlet/prediction/checkpoint_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/koordinator-sh/koordinator/pkg/util/histogram"
+)
+
+func TestSaveAndRestore(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := ioutil.TempDir("/tmp", "checkpoints")
```
saintube, line 31: `io/ioutil` is deprecated since golang 1.16. please use `t.TempDir` or `os.TempDir`

pkg/koordlet/prediction/prediction.go
```diff
@@ -0,0 +1,181 @@
...
+	queryPodMetric := func(m metriccache.MetricResource) (float64, error) {
+		meta, err := m.BuildQueryMeta(podProperties)
+		if err != nil {
+			return 0, err
+		}
+		result := metriccache.DefaultAggregateResultFactory.New(meta)
+		if err = querier.Query(meta, nil, result); err != nil {
+			return 0, err
+		}
+		return result.Value(metriccache.AggregationTypeLast)
```
saintube, line 140: This adds the last metric point in the last 2 mins. It may be inaccurate when the pod usage varies in seconds.
How about using the avg/max/percentile metric and zooming in the window?

saintube (commented):
/lgtm

PR #1396, opened 2023-06-18
koord-manager: use GenerationChangedPredicate to prevent unnessary repeated reconcile calls

jasonliu747:

This predicate will skip update events that have no change in the object's metadata.generation field. The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object. This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.

### V. Checklist

### DIFF
--- a/pkg/slo-controller/nodemetric/nodemetric_controller.go
+++ b/pkg/slo-controller/nodemetric/nodemetric_controller.go
@@ -27,8 +27,10 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


pkg/slo-controller/noderesource/noderesource_controller.go
```diff
@@ -137,7 +139,7 @@ func (r *NodeResourceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	handler := config.NewColocationHandlerForConfigMapEvent(r.Client, *sloconfig.NewDefaultColocationCfg(), r.Recorder)
 	r.cfgCache = handler
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&corev1.Node{}).
+		For(&corev1.Node{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
saintube, line 142: The node-resource controller should watch the change in the `Node.Status`. e.g. The batch resource allocatable would be updated when the node allocatable changed.

jasonliu747: You are indeed correct!

hormes (approved):

PR #1397, opened 2023-06-19
Add nodemetrics api

hormes:
Part of #1361 

Add the NodeMetric API. 

What needs to be specially discussed here is the way the metrics are reported, whether to use the Pod level or aggregate to the Node level. The current definition adopts aggregation to the Node level, but reports Reclaimable resources. From the perspective of the scheduling view, when the newly added Pod is not counted into the model, Reclaimable is 0, which will not cause ledger problems. For the case where the Pod is deleted, we need to update the reclaimable resource in the Terminating phase to update the scheduling view as soon as possible.


apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -89,6 +95,9 @@ type NodeMetricStatus struct {
 
 	// PodsMetric contains the metrics for pods belong to this node.
 	PodsMetric []*PodMetricInfo `json:"podsMetric,omitempty"`
+
+	// ProdPeclaimableMetric is the indicator statistics of Prod type resources reclaimable
+	ProdPeclaimableMetric *ReclaimableMetric `json:"prodPeakMetric,omitempty"`
```
saintube, line 100: the field's naming does not match the json tag

apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -79,6 +79,12 @@ type AggregatePolicy struct {
 	Durations []metav1.Duration `json:"durations,omitempty"`
 }
 
+// ReclaimableMetric defines the reclaimable metric of resource priority
+type ReclaimableMetric struct {
+	// Resource is the resource usage of the prediction
+	Resource ResourceMap `json:"nodeUsage,omitempty"`
```
saintube, line 85: the field name does not match the json tag

apis/slo/v1alpha1/nodemetric_types.go
```diff
@@ -89,6 +95,9 @@ type NodeMetricStatus struct {
 
 	// PodsMetric contains the metrics for pods belong to this node.
 	PodsMetric []*PodMetricInfo `json:"podsMetric,omitempty"`
+
+	// ProdPeclaimableMetric is the indicator statistics of Prod type resources reclaimable
+	ProdPeclaimableMetric *ReclaimableMetric `json:"prodReclaimableMetric,omitempty"`
```
zwzhang0107, line 100: typo
ProdReclaimableMetric

eahydra, line 100: typo. ProdPeclaimableMetric -> ProdReclaimableMetric

PR #1398, opened 2023-06-19
all: upgrade to k8s v1.24.15

eahydra:

- Upgrade to k8s 0.24.15 (v1.24+).
- All tests passed.
- When building components, you need to upgrade to golang 1.18 and above.

known compatibility issues:
- The scheduler no longer supports insecure serving, and can only query metrics through the secure port. However, the insecure configuration is retained in this PR, and the IP and port to be bound need to be set through the command line parameters port and address.
- The built-in `volumebinding` plugin of k8s scheduler framework depends on v1.CSIStorageCapacity , which will cause koord-scheduler to fail to run in v1.22 and earlier versions.
- The k8s scheduler framework also Watch v1.CSIStorageCapacity, causing the same problem as above.  The specific code can be found in [pkg/scheduler/eventhandlers.go#L353](https://github.com/kubernetes/kubernetes/blob/de2c95dfadc8f63feb361d394768eb72b351cf7a/pkg/scheduler/eventhandlers.go#L353)
- The k8s scheduler framework changed the PreFilter interface to support returning a list of schedulable nodes.
- CRI v1alpha2.RuntimeServiceServer adds two new methods PodSandboxStats/ListPodSandboxStats
 
**NOTE:**

Compatibility issues brought by v1.CSIStorageCapacity will be more troublesome, which will involve the EventHandlers code of the scheduler framework. This place is not easy to modify, unless we maintain a scheduler framework code separately and do some special processing.

According to the discussion, the solution is as follows:
- Intercept the Informer construction method of v1.CSIStorageCapacity and forge a custom Informer. This Informer will be responsible for converting v1beta1's CSIStorageCapacity to v1.CSIStorageCapacity.
- Versions below k8s v1.22 that does not support v1beta1.CSIStorageCapacity or v1.CSIStorageCapacity, the users need to enable the feature-gate `DisableCSIStorageCapacityInformer` to disable sync the objects via informer.
- The k8s v1.22 version that support v1beta1.CSIStorageCapacity, enable the feature-gate `CompatibleCSIStorageCapacity` to sync v1beta1.CSIStorageCapacity and tranform the objects to v1.CSIStorageCapacity.  

### V. Checklist


eahydra:
Several methods to solve v1.CSIStorageCapacity's incompatibility with k8s v1.22 version: 
1. Intercept the Informer construction method of v1.CSIStorageCapacity and forge a custom Informer. This Informer will be responsible for converting v1beta1's CSIStorageCapacity to v1.CSIStorageCapacity.
- Pros. is that it can be seamlessly compatible with versions 1.22 and earlier; 
- Cons. is that the implementation is tricky
2. Fork k8s v1.24 code to koordinator-sh and modify Scheduler Framework related code. And use the fork version by replacing in the go.mod of koordinator.
- Pros. is that the changes are clearer; 
- Cons. is that a separate Fork branch is maintained, and the maintenance cost is relatively high.

WDYT @hormes @FillZpp

FillZpp:
> Several methods to solve v1.CSIStorageCapacity's incompatibility with k8s v1.22 version:
> 
> 1. Intercept the Informer construction method of v1.CSIStorageCapacity and forge a custom Informer. This Informer will be responsible for converting v1beta1's CSIStorageCapacity to v1.CSIStorageCapacity.
> 
> * Pros. is that it can be seamlessly compatible with versions 1.22 and earlier;
> * Cons. is that the implementation is tricky
> 
> 2. Fork k8s v1.24 code to koordinator-sh and modify Scheduler Framework related code. And use the fork version by replacing in the go.mod of koordinator.
> 
> * Pros. is that the changes are clearer;
> * Cons. is that a separate Fork branch is maintained, and the maintenance cost is relatively high.
> 
> WDYT @hormes @FillZpp

@eahydra Prefer #1. We could have a flag/featuregate or automatically discovery to decide whether v1 or v1 wrapper should be used.

eahydra:
> > Several methods to solve v1.CSIStorageCapacity's incompatibility with k8s v1.22 version:
> > 
> > 1. Intercept the Informer construction method of v1.CSIStorageCapacity and forge a custom Informer. This Informer will be responsible for converting v1beta1's CSIStorageCapacity to v1.CSIStorageCapacity.
> > 
> > 
> > * Pros. is that it can be seamlessly compatible with versions 1.22 and earlier;
> > * Cons. is that the implementation is tricky
> > 
> > 
> > 2. Fork k8s v1.24 code to koordinator-sh and modify Scheduler Framework related code. And use the fork version by replacing in the go.mod of koordinator.
> > 
> > 
> > * Pros. is that the changes are clearer;
> > * Cons. is that a separate Fork branch is maintained, and the maintenance cost is relatively high.
> > 
> > WDYT @hormes @FillZpp
> 
> @eahydra Prefer #1. We could have a flag/featuregate or automatically discovery to decide whether v1 or v1 wrapper should be used.

OK.  According to the first method, it is controlled by featureGate.

eahydra:
TBR.  @hormes @FillZpp @zwzhang0107 @saintube @ZiMengSheng

cmd/koordlet/main.go
```diff
@@ -38,6 +38,7 @@ import (
 func main() {
 	cfg := config.NewConfiguration()
 	cfg.InitFlags(flag.CommandLine)
+	klog.InitFlags(flag.CommandLine)
```
saintube, line 41: duplicate with the imported initialization in `k8s.io/component-base/logs`

eahydra: fixed. But `k8s.io/component-base/logs` no longer registers flags of klog to flag.CommandLine in init, manual registration is required.

saintube: ok

saintube (commented):
/lgtm

hormes:
> * d earlier;

Prefer 1, the tricky part is for compatibility and is worth the cost.

hormes (approved):

PR #1400, opened 2023-06-20
koordlet: refactor statesinformer to solve the circular dependency problem

hormes:
Part of #1361 

Separate the interface and implementation of stateinformer to solve the problem of circular dependency. The current abstraction of stateinformer still has some problems, which will not be resolved for the time being.


hormes:
> there is a race problem in ut

I will fix it later.

PR #1402, opened 2023-06-21
koordlet: fix can not collect psi metrics

lucming:
koordlet can not collect psi metrics.

1. kernel with psi feature on
```bash
sudo grubby --update-kernel="/boot/vmlinuz-4.19.91-24.8.an8.x86_64" --args="psi=1 psi_v1=1"
```
`update-kernel` need to be replaced with the actual kernel version
截屏2023-06-21 10 31 51

2. open koordlet's feature for collecting psi metrics.
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: koordlet
spec:
  template:
    spec:
      containers:
        - name: koordlet
          image: registry.cn-beijing.aliyuncs.com/koordinator-sh/koordlet:v1.2.0
          command:
            - /koordlet
          args:
            - '-cgroup-root-dir=/host-cgroup/'
            - >-
              -feature-gates=BECPUEvict=true,BEMemoryEvict=true,CgroupReconcile=true,Accelerators=true,CPICollector=true,PSICollector=true
            - '-runtime-hooks-host-endpoint=/var/run/koordlet/koordlet.sock'
            - '--logtostderr=true'
            - '--v=4'
```
3. then we can see
  3.1. Access to koordlet's metrics api, but didn't get any info about psi
  3.2. check koordlet's log:
截屏2023-06-20 17 04 19
It seems to be saying that there are no pressure related files under cpuacct cgroup
  3.3. looking at the cgroup, we can see that psi-related files exist
截屏2023-06-21 10 35 30

then there may be a problem with koordlet handling

The essential cause of the problem： 
In golang, global variables are initialized before normal functions, so even though there is logic in the koordlet that modifies `CgroupPathFormatter`, the initialization of global variables only happens once and before the modification, so the logic in `SupportedIfFileExistsInKubepods` will use `cgroupPathFormatterInSystemd`, then the file related to psi can not be find, then making koordlet decide not support psi.
 
截屏2023-06-21 11 07 10

截屏2023-06-21 11 14 00

### V. Checklist

### DIFF
--- a/pkg/koordlet/util/system/cgroup_driver.go
+++ b/pkg/koordlet/util/system/cgroup_driver.go
@@ -193,7 +193,19 @@ var cgroupPathFormatterInCgroupfs = formatter{
 }
 
 // CgroupPathFormatter uses the Systemd cgroup driver by default.
-var CgroupPathFormatter = cgroupPathFormatterInSystemd
+var CgroupPathFormatter = GetCgroupFormatter()
+
+func GetCgroupPathFormatter(driver CgroupDriverType) formatter {
+	switch driver {
+	case Systemd:
+		return cgroupPathFormatterInSystemd
+	case Cgroupfs:
+		return cgroupPathFormatterInCgroupfs
+	default:
+		klog.Warningf("cgroup driver formatter not supported: '%s'", string(driver))
+		return cgroupPathFormatterInSystemd
+	}
+}
 
 func SetupCgroupPathFormatter(driver CgroupDriverType) {
 	switch driver {
--- a/pkg/koordlet/koordlet.go
+++ b/pkg/koordlet/koordlet.go
@@ -17,16 +17,13 @@ limitations under the License.
 package agent
 
 import (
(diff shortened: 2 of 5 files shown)


pkg/koordlet/util/system/cgroup_driver.go
```diff
@@ -193,7 +200,61 @@ var cgroupPathFormatterInCgroupfs = formatter{
 }
 
 // CgroupPathFormatter uses the Systemd cgroup driver by default.
-var CgroupPathFormatter = cgroupPathFormatterInSystemd
+var CgroupPathFormatter = GetCgroupFormatter()
```
saintube, line 203: could you please add a unit test to verify the fix

lucming: done

saintube:
/lgtm
@lucming Please remember to mark the conversation as *resolved*.

Issue #1403, opened 2023-06-21
[BUG] sharePoolCPUCore not exclude LSE


j4ckstraw:
**What happened**:
I found sharePoolCPUCore logic not exclude LSE pod in function `getNodeStateForBurst` https://github.com/koordinator-sh/koordinator/blob/382f7a03b7dd42eac5866c9fffbc25760807080d/pkg/koordlet/resmanager/cpu_burst.go#L234

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

j4ckstraw:
how about this? `sharePoolCPUCoresTotal = nodeCPUCoresTotal - LSE - LSR`

j4ckstraw:
how about this? 
`sharePoolCPUCoresTotal = nodeCPUCoresTotal - LSE - LSR`
`sharePoolCPUCoresUsage = nodeCPUCoresUsage - LSE - LSR - BE`

j4ckstraw:
> how about this? `sharePoolCPUCoresTotal = nodeCPUCoresTotal - LSE - LSR` `sharePoolCPUCoresUsage = nodeCPUCoresUsage - LSE - LSR - BE`

or this? 
`sharePoolCPUCoresTotal = nodeCPUCoresTotal - LSE`
`sharePoolCPUCoresUsage = nodeCPUCoresUsage - LSE - LSR - BE`

saintube:
> how about this? `sharePoolCPUCoresTotal = nodeCPUCoresTotal - LSE - LSR` `sharePoolCPUCoresUsage = nodeCPUCoresUsage - LSE - LSR - BE`

This is correct.

eahydra:
fixed by #1404

PR #1404, opened 2023-06-21
koordlet: exclude LSE from share cpu pool

j4ckstraw:
we have definition of share cpu pool which should exclude LSE; while calculate share pool cpu usage, we should exclude LSR. see https://koordinator.sh/zh-Hans/docs/architecture/qos

#1403 

### V. Checklist

### DIFF
--- a/pkg/koordlet/resmanager/cpu_burst.go
+++ b/pkg/koordlet/resmanager/cpu_burst.go
@@ -274,18 +274,18 @@ func (b *CPUBurst) getNodeStateForBurst(sharePoolThresholdPercent int64,
 	sharePoolCPUCoresUsage := nodeCPUCoresUsage
 	for _, podMeta := range podsMeta {
 		podQOS := apiext.GetPodQoSClass(podMeta.Pod)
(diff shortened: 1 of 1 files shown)


pkg/koordlet/resmanager/cpu_burst.go
```diff
@@ -274,18 +274,18 @@ func (b *CPUBurst) getNodeStateForBurst(sharePoolThresholdPercent int64,
 	sharePoolCPUCoresUsage := nodeCPUCoresUsage
 	for _, podMeta := range podsMeta {
 		podQOS := apiext.GetPodQoSClass(podMeta.Pod)
-		// exclude LSR pod cpu from cpu share pool
-		if podQOS == apiext.QoSLSR {
+		// exclude LSE/LSR pod cpu from cpu share pool
+		if podQOS == apiext.QoSLSE || podQOS == apiext.QoSLSR {
```
saintube, line 278: Thanks for your contribution! We'd better add a UT to verify the fix and avoid regression.

PR #1407, opened 2023-06-21
koordlet: report the amount of reclaimable resources of the node

hormes:
Part of #1361

### DIFF
--- a/pkg/koordlet/prediction/peak_predictor.go
+++ b/pkg/koordlet/prediction/peak_predictor.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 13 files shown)


pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -0,0 +1,168 @@
...
+
+	podRequests := util.GetPodRequest(pod, v1.ResourceCPU, v1.ResourceMemory)
+	podCPURequest := podRequests[v1.ResourceCPU]
+	podMemoryRequest := podRequests[v1.ResourceMemory]
+
+	reclaimableCPUMilli := int64(0)
+	reclaimableMemoryBytes := int64(0)
+
+	if p95CPU, ok := p95Resources[v1.ResourceCPU]; ok {
+		peakCPU := util.MultiplyMilliQuant(p95CPU, 1.1)
```
saintube, line 141: This ratio can be inaccurate and not universal. How about adding a parameter to adjust it when necessary?

hormes: later add a pr separately to add configurations.

saintube (approved):
/lgtm

PR #1408, opened 2023-06-21
koord-manager: enhance noderesource framework for updating node labels

saintube:

Enhance node-resource framework by adding NodeMetaSyncPlugin for updating node metadata.
It's helpful when we want to update node labels/annotations rather than the allocatable in node status.

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/resource_calculator.go
+++ b/pkg/slo-controller/noderesource/resource_calculator.go
@@ -24,7 +24,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/koordinator-sh/koordinator/apis/extension"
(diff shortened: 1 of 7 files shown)


pkg/slo-controller/noderesource/resource_calculator.go
```diff
@@ -71,33 +73,85 @@ func (r *NodeResourceReconciler) updateNodeResource(node *corev1.Node, nr *frame
...
+			return r.updateNodeStatus(node, strategy, nr)
+		}); err != nil {
+			klog.ErrorS(err, "failed to update node status for node", "node", node.Name)
+			errList = append(errList, err)
+		}
+	}
 
-	if needSync := r.isNodeResourceSyncNeeded(strategy, node, nodeCopy); !needSync {
+	if needSyncMeta {
+		if err := util.RetryOnConflictOrTooManyRequests(func() error {
```
hormes, line 93: If the status has been updated once before, there will be a conflict for the first time here. Do you need to update the data in advance?

saintube: Yes. In the `updateNodeStatus()` and the `updateNodeMeta`, we Get the node object via the controller-runtime client first and then update the status/metadata of the node.
It should avoid most of the conflicts, while for some others (since the controller-runtime client reads from the informer cache), we can simply retry.

hormes (approved):

PR #1418, opened 2023-06-26
koord-controller-manager: noderesource supports configuration plugins

hormes:
Part of #1361 

Support for user-configurable plugins that controllers need to enable


saintube (commented):
/lgtm

PR #1422, opened 2023-06-27
koord-scheduler: refresh specific NodeInfos on demand instead of all

eahydra:

When there is a Reservation on the node, the NodeInfo in the Snapshot will be modified, which requires that the modified NodeInfo in the snapshot be refreshed from the Cache during the next scheduling. The existing implementation is to take over the generation in the snapshot and force it to be set to 0. Although this solves the problem, it will refresh all node data, which leads to serious performance regression.
Because now we can operate the Scheduler.Cache, we can mark the NodeInfo in the Cache as changing by forging the Pod update and delete actions, thereby triggering the update of the Snapshot. In this way, when there is no Reservation, there will be no impact. When there is a Reservation, only update the corresponding node.

In addition, Reservation.PostFilter will modify the priority of the Reserve Pod in order to prevent the Reservation from being preempted. In theory, NodeInfo also needs to be updated, but in fact, we can optimize this implementation by directly modifying the priority to prohibit preemption before adding it to the Cache.

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/scheduler_adapter.go
+++ b/pkg/scheduler/frameworkext/scheduler_adapter.go
@@ -18,6 +18,8 @@ package frameworkext
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/kubernetes/pkg/scheduler"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 )
@@ -38,6 +40,7 @@ type SchedulerCache interface {
 	IsAssumedPod(pod *corev1.Pod) (bool, error)
 	GetPod(pod *corev1.Pod) (*corev1.Pod, error)
 	ForgetPod(pod *corev1.Pod) error
+	InvalidNodeInfo(nodeName string) error
 }
 
 type SchedulingQueue interface {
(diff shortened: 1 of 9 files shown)


eahydra:
/hold cancel.

All E2E passed.   TBR

eahydra:
/hold cancel

All E2E passed. TBR

pkg/scheduler/frameworkext/scheduler_adapter.go
```diff
@@ -69,6 +72,58 @@ func (s *SchedulerAdapter) MoveAllToActiveOrBackoffQueue(event framework.Cluster
...
+
+func (c *cacheAdapter) GetPod(pod *corev1.Pod) (*corev1.Pod, error) {
+	return c.scheduler.Cache.GetPod(pod)
+}
+
+func (c *cacheAdapter) ForgetPod(pod *corev1.Pod) error {
+	return c.scheduler.Cache.ForgetPod(pod)
+}
+
+func (c *cacheAdapter) InvalidNodeInfo(nodeName string) error {
```
KunWuLuan, line 108:
```
// SetNode sets the overall node information.
func (n *NodeInfo) SetNode(node *v1.Node) {
	n.node = node
	n.Allocatable = NewResource(node.Status.Allocatable)
	n.Generation = nextGeneration()
}
```

We can also use `SetNode` to update generation, maybe this cost is lower?

eahydra: Yes, using `SetNode` is also a good method, but this may conflict with the Node's Update Event, because the version of Node we get when we call it may be later than the latest version.

KunWuLuan: That make sense

hormes (approved):

PR #1424, opened 2023-06-28
koord-scheduler: load-aware supports estimate node allocatable

eahydra:

Some special oversold mechanisms will adjust Node.Status.Allocatable, and zoom in or out according to the actual situation. The enlarged situation will be special and needs to be evaluated according to the actual total amount of physical resources.
So extend the Estimator in the LoadAware plugin to support custom node.status.allocatable. 

### V. Checklist


jasonliu747 (approved):
/lgtm

ZiMengSheng (commented):
/lgtm

PR #1432, opened 2023-06-30
chore: remove the tag in internal config as usual

### LINKED ISSUE #1431: [proposal] remove the tag in internal config as usual
**What is your proposal**:

The component config of koord-scheduler does not need to define json tag, only set in special version. 

**Why is this needed**:

eahydra:

The component config of koord-scheduler does not need to define json tag, only set in special version.

fix #1431

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1434, opened 2023-06-30
koord-scheduler: improve throughput with reservation via return PreFilterResult

### LINKED ISSUE #1435: [proposal] use the framework.PreFilterResult to improve throughput with Reservation
**What is your proposal**:

Use the framework.PreFilterResult to improve throughput with Reservation.

**Why is this needed**:

In a cluster there are some Reservation objects but maybe just few nodes, and the cluster maybe has 5K nodes, so we can return PreFilterResult in Reservation to improve the performance.

eahydra:

Kubernetes Scheduling Framework v1.24  has supported return `framework.PreFilterResult` to improve the throughput, and Reservation plugin can return the nodeNames which has reservations that matched `ReservationAffinity`.

fix #1435

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1446, opened 2023-07-04
apis: add default PriorityClass and QoSClass, revise batch resource calculation

### LINKED ISSUE #1441: [proposal] Revise batch resource calculation
**What is your proposal**:

The batch resource calculation in the koord-manager should be revised.
Besides the QoS class of pods, the calculator should also check their priority classes.
Note that a pod without a priority class is regarded as Prod by default, and a pod without a QoS class is regarded as LS by default.

**Why is this needed**:

Since the Mid-tier is supported, there can be BE pods in different priority classes, where the requests and usages of Mid+BE pods should be counted as the high-priority and the batch resource should reclaim the unused part.

Revise the BatchResource plugin's implementation in the koord-manager.

saintube:
…yclass and qosclass

- koord-manager: revise batch resource calculation, identifying the pods' priorityclasses.
- koord-manager: fix a bug where the `noderesource-plugin` arg allows no plugin by default.
- util: add default PriorityClass and QoSClass when any of them is missing.

Fixes #1441.

### V. Checklist

### DIFF
--- a/apis/extension/priority.go
+++ b/apis/extension/priority.go
@@ -33,10 +33,6 @@ const (
 	PriorityNone  PriorityClass = ""
 )
 
-var (
-	DefaultPriorityClass = PriorityNone
-)
-
 // Define Koordinator priority as a variable value to support customizing different priority ranges
 var (
 	PriorityProdValueMax int32 = 9999
--- a/apis/extension/priority_utils.go
+++ b/apis/extension/priority_utils.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+import corev1 "k8s.io/api/core/v1"
+
+// NOTE: functions in this file can be overwritten for extension
+
+var DefaultPriorityClass = PriorityNone
+
+// GetPodPriorityClassWithDefault gets the pod's PriorityClass with the default config.
+func GetPodPriorityClassWithDefault(pod *corev1.Pod) PriorityClass {
+	priorityClass := GetPriorityClass(pod)
+	if priorityClass != PriorityNone {
+		return priorityClass
+	}
+
+	return GetPodPriorityClassWithQoS(GetPodQoSClassWithDefault(pod))
+}
+
+// GetPodPriorityClassWithQoS returns the default PriorityClass according to its QoSClass when the pod does not specify
+// a PriorityClass explicitly.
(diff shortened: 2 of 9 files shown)


pkg/util/slo_utils.go
```diff
@@ -0,0 +1,75 @@
...
+	DefaultQoSClass      = extension.QoSLS
+)
+
+var (
+	// DefaultPriorityClassWithQoS is the default PriorityClass according to its QoSClass when the pod does not specify
+	// a PriorityClass explicitly.
+	// NOTE: The pod with PriorityClassNone is considered as PriorityClassProd, and the pod with QoSNone is considered
+	// as qos=LS. A special case is that the pod with PriorityClassNone+QoSBE is considered as PriorityBatch to get
+	// compatible to with existing BE pods.
+	DefaultPriorityClassWithQoS = map[extension.QoSClass]extension.PriorityClass{
```
eahydra, line 38: Why not move the file to package "apis/extensions"? Should we replace all code that cares about PriorityClass and QoS ?

pkg/util/slo_utils.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+)
+
+// NOTE: functions in this file can be overwritten for extension
+
+const (
+	DefaultPriorityClass = extension.PriorityProd
```
eahydra, line 28: In package `apis` already has the same definition but DefaultPriorityClass is PriorityNone.
The new definition looks to change the default behavior of Koordinator models.

saintube: done

pkg/util/slo_utils.go
saintube: OK

eahydra (commented):
/lgtm

apis/extension/priority_utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+// NOTE: functions in this file can be overwritten for extension
+
+var (
+	DefaultPriorityClass = PriorityNone
+
+	// DefaultPriorityClassWithQoS is the default PriorityClass according to its QoSClass when the pod does not specify
+	// a PriorityClass explicitly.
+	DefaultPriorityClassWithQoS = map[QoSClass]PriorityClass{
+		QoSLSE:  PriorityProd,
```
zwzhang0107, line 27: QoSSystem

saintube: fixed

apis/extension/priority.go
```diff
@@ -52,6 +48,16 @@ var (
...
 )
 
+// GetPodPriorityClassWithDefault gets the pod's PriorityClass with the default config.
+func GetPodPriorityClassWithDefault(pod *corev1.Pod) PriorityClass {
+	priorityClass := GetPriorityClass(pod)
+	if priorityClass != PriorityNone {
+		return priorityClass
+	}
+
+	return DefaultPriorityClassWithQoS[GetPodQoSClass(pod)]
```
hormes, line 58: 现在 QoS 也缺少兼容逻辑，如果用啥啥都没填，应该兼容 k8s qos

eahydra (commented):
/lgtm

apis/extension/priority.go
saintube: Done

hormes (approved):

PR #1456, opened 2023-07-09
koord-scheduler: optimize nominate reservation

eahydra:

In the existing implementation, Nominate Reservation is executed in the Reserve stage. Although this can work, but there are still some problems to be solved. For example, when a plugin needs to implement Score, it needs to consider each Reservation on the current node, which means traversing each Reservation to calculate a Score separately, and compare and find the maximum value as the node score.
Then in the Reserve stage, Nominate Reservation will calculate the score of each Reservation on the candidate node again and select the one with the highest score.
This approach brings additional complexity to the plugin, which should be considered to simplify.

The current PR provides an optimization: execute Nominate Reservation in Reservation.PreScore, calculate the most suitable Reservation on each node in advance, and then directly use this result for calculation in the Score and Reserve stages. This will make the Score implementation of the plugin easier and clearer.

### V. Checklist


jasonliu747 (commented):
/lgtm

PR #1459, opened 2023-07-09
koord-descheduler: LowNodeLoad supports descheduling by node pool

### LINKED ISSUE #1430: [proposal] descheduler supports threshold config with node selector
**What is your proposal**:
support lowThresholds and highThresholds config with node selector.

also print idle, hotspot and normal nodes number of different pools(selector) for easy to debug.

**Why is this needed**:
clusters are always divided into multiple pools by node selectors for running different types of workloads.

eahydra:

At present LowNodeLoad  plugin in koord-descheduler just supports descheduling Pods by nodeSelector or all nodes in cluster, but users or cluster administrators divided into multiple pools by node selectors that the plugin cannot support.

To fix this, a new field "nodePools" is added to support multiple pools. Original fields such as `nodeSelector`, `useDeviationThresholds`, `highThresholds`, `lowThresholds`, `resourceWeights` and `anomalyCondition` will be automatically migrated to NodePool by default. Therefore, the user's original usage method does not need to be changed, but when necessary, different parameters can be configured for multiple batches of nodes.

In the future, support for existing NodePool CRDs can also be considered.

fix #1430 

### V. Checklist


pkg/descheduler/apis/config/validation/validation_loadaware.go
```diff
@@ -34,12 +34,6 @@ func ValidateLowLoadUtilizationArgs(path *field.Path, args *deschedulerconfig.Lo
 		allErrs = append(allErrs, field.Invalid(path.Child("evictableNamespaces"), args.EvictableNamespaces, "only one of Include/Exclude namespaces can be set"))
 	}
 
-	if args.NodeSelector != nil {
```
zwzhang0107, line 37: how about we just don't allow set NodeSelector and NodePools at the same time.
we can treat all nodes as one "pool" so that the NodeSelector config still works.

eahydra: At present, it supports automatic conversion of the original nodeSelector to nodePool

jasonliu747 (commented):
/lgtm

PR #1461, opened 2023-07-10
apis: add priority-class compatible label

hormes:

Provide a solution using Koordinator PriorityClass for already running a large number of Pods clusters by adding a label `koordinator.sh/priority-class`.

### DIFF
--- a/docs/proposals/api-machinery/20230707-add-compatibility-interface-for-priority-class.md
+++ b/docs/proposals/api-machinery/20230707-add-compatibility-interface-for-priority-class.md
@@ -1,0 +1,105 @@
+---
+title: Add compatibility interface for priority class
+authors:
+- "@hormes"
+reviews:
+- "@eahydra"
+- "@saintube"
+- "@zwzhang0107"
+- "@jasonliu747"
+---
+
+# Add compatibility interface for priority class
+
+## Table of Contents
+
+- [Add compatibility interface for priority class](#add-compatibility-interface-for-priority-class)
+  - [Table of Contents](#table-of-contents)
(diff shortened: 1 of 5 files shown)


docs/proposals/api-machinery/20230707-add-compatibility-interface-for-priority-class.md
```diff
@@ -0,0 +1,105 @@
...
+  priorityClassName: user-defined-priority-class
+```
+
+When users install Koordinator, they cannot adjust all these Pods at once (affecting business stability), and hope that
+there is a way to regulate these Pods into the best practice category of Koordinator.
+
+### Implementation
+
+1. Defines a new label `koordinator.sh/priority-class` to compatible with Pods that are already running.
+2. When a user install Koordinator, find a way to label the existing Pods with this label. There is a default value inference mechanism for unlabeled Koordinator, but please do not rely on it.
```
saintube, line 88: Does the new label affect the priority comparison between pods? If it does, we may need to define the priority value for each priority label.

e.g. When a pod `p0` with the new label (`koordinator.sh/priority-class=Prod`) is compared with a pod `p1` with koord priority class specified (`pod.spec.priorityClassName=koord-prod`), we need to determine which one is considered high-priority.

hormes: It does not affect the priority comparison. There is a section in it that explains that when the priority is reversed by introducing a compatible label, you need to be aware of this problem. It is recommended to turn off priority preemption, and consider turning it on after completing the transition of the plan.

PR #1462, opened 2023-07-10
koordlet: enhance node prediction

saintube:

koordlet: enhance the node prediction.

- Add the node priority predictor and the min predictor. s.t.

```
Reclaimable[Prod] := max(0, min(Requests[Prod] - Peak[Prod], sum(Request[ProdPod] - Peak[ProdPod])))
```

### V. Checklist

### DIFF
--- a/pkg/koordlet/prediction/peak_predictor.go
+++ b/pkg/koordlet/prediction/peak_predictor.go
@@ -22,8 +22,11 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	quotav1 "k8s.io/apiserver/pkg/quota/v1"
+	"k8s.io/klog/v2"
 
 	"github.com/koordinator-sh/koordinator/apis/extension"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/metrics"
 	"github.com/koordinator-sh/koordinator/pkg/util"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/util/resource.go
```diff
@@ -44,6 +44,26 @@ func MultiplyQuant(quant resource.Quantity, factor float64) resource.Quantity {
 	return *newQuant
 }
 
+// Min returns the result of Min(a, b) for each named resource, corresponded to the quotav1.Max().
+func Min(a corev1.ResourceList, b corev1.ResourceList) corev1.ResourceList {
```
hormes, line 48: The semantics of this function are a bit strange, such as A is {cpu = 1}, B is {memory = 1}, after Min is {cpu = 1, memory = 1}. For quotav1.Max, union keys is semantic, but min is a bit problematic.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -65,18 +68,33 @@ func NewPredictorFactory(predictServer PredictServer, coldStartDuration time.Dur
 func (f *predictorFactory) New(t PredictorType) Predictor {
 	switch t {
 	case ProdReclaimablePredictor:
-		return &prodReclaimablePredictor{
+		podPredictor := &podReclaimablePredictor{
```
hormes, line 71: It is elegant to implement the combination of two predictors here.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -65,18 +68,33 @@ func NewPredictorFactory(predictServer PredictServer, coldStartDuration time.Dur
...
 			reclaimable:         util.NewZeroResourceList(),
 			pods:                make(map[string]bool),
 		}
+		priorityPredictor := &priorityReclaimablePredictor{
+			predictServer:         f.predictServer,
+			safetyMarginPercent:   f.safetyMarginPercent,
+			priorityClassFilterFn: isPriorityClassReclaimableForProd,
+			reclaimRequest:        util.NewZeroResourceList(),
+		}
+		return &minPredictor{
```
hormes, line 85: It is elegant to implement the combination of two predictors here.

pkg/util/resource.go
saintube: Fixed. Now the unmatched resource names will be dropped in the result.
e.g. a = {"cpu": "1"}, b = {"memory", "2Gi"} => {}

PR #1478, opened 2023-07-14
all: fix GetPriorityClass with default

hormes:

1. Koordinator is using PriorityClass to make a lot of logical judgments, which need to consider compatible scenarios (default value)
2. Renamed GetPriorityClass and GetQoSClass at the same time to avoid misuse

### DIFF
--- a/pkg/descheduler/utils/sorter/pod_test.go
+++ b/pkg/descheduler/utils/sorter/pod_test.go
@@ -130,7 +130,7 @@ func TestSortPods(t *testing.T) {
 		corev1.ResourcePods:   1,
 	}
 	SortPodsByUsage(pods, podMetrics, nodeAllocatableMap, resourceToWeightMap)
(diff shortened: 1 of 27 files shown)


pkg/descheduler/utils/sorter/pod_test.go
```diff
@@ -130,7 +130,7 @@ func TestSortPods(t *testing.T) {
 		corev1.ResourcePods:   1,
 	}
 	SortPodsByUsage(pods, podMetrics, nodeAllocatableMap, resourceToWeightMap)
```
hormes, line 132: Here you need to focus on it, because it will affect the sorting results. @eahydra

jasonliu747 (commented):
/lgtm

saintube (commented):
/lgtm

pkg/descheduler/utils/sorter/pod_test.go
eahydra: This new ordering looks reasonable.

eahydra (approved):
/lgtm
/approve

PR #1495, opened 2023-07-25
proposal: add limit-aware scheduling

ZiMengSheng:

This PR proposes a resource limit aware plugin to mitigate the risk of overcommitment.

### V. Checklist

### DIFF
--- a/docs/proposals/20230712-resource-limit-aware-scheduling.md
+++ b/docs/proposals/20230712-resource-limit-aware-scheduling.md
@@ -1,0 +1,165 @@
+---
+title: Limit Aware Scheduling
+authors:
+- "@ZiMengSheng"
+reviewers:
+- "@hormes"
+- "@eahydra"
+creation-date: 2023-07-12
+last-updated: 2023-07-12
+---
+
+# Limit Aware Scheduling
+
+## Motivation
+
+Kubernetes divides pod QoS into guaranteed, burstable, and best effort. Among them, the limits of burstable pods is greater than requests. However, the current scheduler actually only considers pod requests when scheduling pods. In this way, if a large number of burstable pods are scheduled to the node, the total resource usage of the pods on the node will probably exceed the allocatable capacity of the node, that is, the resources are over-subscriped.  Resource over-subscription may lead to resource contention, performance degradation and even pod failure (e.g., OOM).
+
+This proposal uses the limit to allocatable ratio of the node to measure the degree of resource over-subscription, and proposes a scheduling plugin named LimitAware to control the limit/allocatable of the node.
+
+### Goals
+
+1. Introduce a filtering plugin, determine whether the Pod can be scheduled on the node according to limit to allocatable ratios of nodes.
+2. Introduce a scoring plugin to mitigate resource over-subscription issue caused by burstable pods through "spreading" or "balancing" pod's resource limits across nodes.
+
+### Non-Goals
+
+Replace existing resource allocated plugin.
+
+## Proposal
+
+### User Story
+
+#### Story 1: Use limit aware score plugin to mitigate resource over-subscription
+
+Consider two nodes with the same resource allocatable (8 CPU cores) and each has two pods running.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/20230712-resource-limit-aware-scheduling.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
+
+### filter out nodes with high risk of resource over-subscription
+
+From before, we can already determine the Allocatable Limit and Allocated Limit of Node. When scheduling a new Pod, all nodes which satisfy requested limit + allocated limit > allocatable limit will be filtered out
+
+### score nodes to favor less risk of resource over-subscription
+
+Since the main purpose of the LimitAware Score plugin is to reduce the risk of resource over-subscription through Spreading or Balancing, this proposal uses the Least Allocated Limit strategy to score nodes.  We calculate each node’s raw score by this formula: `Score = (allocatable limit - allocated limit) * MaxNodeScore / allocatable limit` . For multiple resources, the raw score is the weighted sum of all allocatable resources. The resources and their weights are configurable as the plugin arguments. Like the other `NodeResources` scoring plugins, the plugin supports standard resource types: CPU, Memory, Ephemeral-Storage plus any Scalar-Resources.
```
eahydra, line 144: We need to redesign a new scoring algorithm. The reason is that the sum of pod limits is indistinguishable by the leastAllocated algorithm in the scenario where the maximum limit of the node is not set by default.

eahydra:
There are several issues to consider:
1. Whether to handle DaemonSet specially during the Filter stage. It is possible for DaemonSet not to set Request, but to set Limit; and DaemonSet is quite special, generally it is a basic component, and it is possible to use DaemonSet to expand its functions.
2. In the scoring stage, Pods of the BestEffort type also need to be considered, and a default limit can be set for these Pods.

docs/proposals/20230712-resource-limit-aware-scheduling.md
```diff
@@ -0,0 +1,165 @@
...
+#### Story 1: Use limit aware score plugin to mitigate resource over-subscription
+
+Consider two nodes with the same resource allocatable (8 CPU cores) and each has two pods running.
+
+- node1: pod1(request: 2, limit: 6), pod2(request: 2, limit: 4); total request/allocatable: 4/8, total limit/allocatable: 10/8
+- node2: pod3(request: 3, limit: 3), pod4(request: 2, limit: 2); total request/allocatable: 5/8, total limit/allocatable: 5/8
+
+To schedule a new pod pod5 (request: 1, limit: 4), although node1's total request/allocatable ratio is smaller than node2's, our `LimitAware` plugin considers node1's resource limit is already oversubscribed and hence place pod5 on node 2 instead.
+
+#### Story 2: Use limit aware filter plugin to filter out nodes with high risk of resource over-subscription
```
Huang-Wei, line 42: I'd say we should be extra cautious on introducing a Filter plugin. Several reasons:

- users, to my experience, may prefer their pods get scheduled (and _possible_ OOMed / CPU throttled afterwards) than not being able to get scheduled at first place
- it'd introduce a new semantics on KSM, which has a large blasting radius - CA has to be aware of this, so does other integrators

eahydra: @Huang-Wei Thanks for reviewing the PR.

Thanks for reviewing this PR.
Indeed, as you said, introducing a new Filter will cause many problems.
Specific to the Limit-aware scheduling scenario, considering the runtime stability of the node, it is risky if the total limit of a node is too high.  IMO there are two ways to control this risk:
1. Lower the weight of such nodes based on the scoring algorithm;
2. Introduce Filter to strongly restrict the total amount of Limit to not exceed the threshold.

The first method is a must. As for the second method, it can be given to users as an option, so that they have the opportunity to use it when necessary.

WDYT? @Huang-Wei

Huang-Wei: The 2nd option is not necessarily a big no, it depends on your use-case. It's just when offering it, it'd break KRM (kubernetes resource model), you have to document the restrictions clearly like vanilla CA are unable to handle this kind of filter error (unless it gets re-compiled), etc.

PR #1512, opened 2023-07-31
apis: exclude slo pkg imports from extension to avoid cycle ref

zwzhang0107:

exclude the imports of `slo` pkg from `extension` pkg

extension pkg is a basic pkg, it should not import other pkg like slo. Otherwise it will cause cycle reference.

### V. Checklist


apis/slo/v1alpha1/pod.go
```diff
@@ -14,33 +14,32 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package extension
+package v1alpha1
```
eahydra, line 17: I have a little doubt: will some codes in the extension package be adjusted according to this new layout? For example, slo-related extension protocols are putting into slo/v1alpha1, and scheduling-related extension protocols are putting into scheduling/v1alpha1.

apis/slo/v1alpha1/pod.go
```diff
@@ -14,33 +14,32 @@ See the License for the specific language governing permissions and
...
 
 import (
 	"encoding/json"
+	apiext "github.com/koordinator-sh/koordinator/apis/extension"
```
saintube, line 21: need goimports

apis/slo/v1alpha1/pod.go
zwzhang0107: i think it depends, for example the `CPUBurstConfig` struct, it is slo-relate and specific for slo config, which already defined in slo/v1alpha1 and no one else will use it except slo-controller/koordlet. It's better to mv related functions to slo/v1alpha1. While for other common slo-related functions like `GetPodQoSClassByName`, which can be used for other modules, maybe should be kept inside extension.

eahydra: If it is determined to adjust the code layout as a whole, I think it is OK.

eahydra (approved):
/lgtm

saintube (approved):
/lgtm

jasonliu747 (commented):
/lgtm

hormes (approved):

PR #1516, opened 2023-07-31
koordlet: support numa topology reporting

### LINKED ISSUE #1503: [proposal] koordlet identifies and reports the NUMA stats
**What is your proposal**:

The koordlet should support identifying, collecting, and reporting the NUMA stats. e.g. The memory total of the numa node 0.

**Why is this needed**:

Currently, the koordlet reports the node memory usage without NUMA awareness while it does have part of NUMA information like CPU topology. For some QoS strategies like BECPUSuppress, it checks the cpu topology to pick a good binding result. However, for some platforms like ARM the NUMA locality can be essential for the app performance. The koordlet is expected to identify the NUMA stats and report them so that the NUMA locality problem can be better handled in the scheduling and QoS management. e.g. To balance the memory util between numa nodes when an LSR pod is scheduled.

Related Issues/PRs: #228, #1224.

See #1516.

- Add collection for the cpu/memory capacity collection per NUMA.
     - [x] util/system: Add collection logic for NUMA stats. Related system interfaces: `/sys/devices/system/node/nodeX/meminfo`.
     - [x] metrics advisor & metric cache: Store the NUMA info into the storage with the key `NodeMemoryInfo`.
- Add the numa stats reporting in the NRT states informer.
     - [x] states informer: In the nodetopo informer, fetch the NUMA info and kubelet topology policy, report them onto NRT.

NRT example:

```yaml
apiVersion: topology.node.k8s.io/v1alpha1
kind: NodeResourceTopology
metadata:
  name: test-node
  annotations:
    node.koordinator.sh/cpu-shared-pools: "{...}"
    node.koordinator.sh/cpu-topology: "{...}"
    # ...
# change fields: `topologyPolicies`, `zones`
topologyPolicies: ["RestrictedContainerLevel"]
zones:
  - name: numa-node-0
    type: Node
    resources:
      - name: cpu
        capacity: 26
        allocatable: 26
      - name: memory
        capacity: 184Gi
        allocatable: 180Gi
  - name: numa-node-1
    type: Node
    resources:
      - name: cpu
        capacity: 4
        allocatable: 3
      - name: memory
        capacity: 184Gi
        allocatable: 180Gi
```

saintube:

- Add the node NUMA info collection in the NodeInfoCollector.
- Add the NUMA topology reporting in the NRT state informer.
- Fix a bug where a prom metric is not registered.
- Refactor the internal LocalCPUInfo to save the cpu maps instead of the cpu counts.

Fixes #1503.

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
+++ b/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
@@ -28,11 +28,11 @@ import (
 	"time"
 
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
-	topov1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
 	topologylister "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -40,6 +40,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
+	kubeletconfiginternal "k8s.io/kubernetes/pkg/kubelet/apis/config"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpumanager"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpumanager/state"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpumanager/topology"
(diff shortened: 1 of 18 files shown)


pkg/koordlet/util/meminfo.go
```diff
@@ -114,13 +142,95 @@ func readMemInfo(path string) (*MemInfo, error) {
...
+	memInfo, err := readMemInfo(meminfoPath, false)
 	if err != nil {
 		return 0, err
 	}
 	usage := int64(memInfo.MemTotal - memInfo.MemAvailable)
 	return usage, nil
 }
+
+type NUMAInfo struct {
+	NodeID  int32    `json:"nodeID,omitempty"`
```
zwzhang0107, line 166: NumaNode vs Node, maybe confused

pkg/koordlet/util/meminfo.go
```diff
@@ -114,13 +142,95 @@ func readMemInfo(path string) (*MemInfo, error) {
...
-	memInfo, err := readMemInfo(meminfoPath)
+	memInfo, err := readMemInfo(meminfoPath, false)
 	if err != nil {
 		return 0, err
 	}
 	usage := int64(memInfo.MemTotal - memInfo.MemAvailable)
 	return usage, nil
 }
+
+type NUMAInfo struct {
```
zwzhang0107, line 165: is socket id can be identified?

pkg/koordlet/util/meminfo.go
```diff
@@ -114,13 +142,95 @@ func readMemInfo(path string) (*MemInfo, error) {
...
+	}
+
+	result := &NodeNUMAInfo{
+		MemInfoMap: map[int32]*MemInfo{},
+	}
+	maxNodeID := int32(-1)
+	for _, n := range nodeDirs {
+		dirName := n.Name() // assert string pattern `nodeX`
+		if len(dirName) < 4 || dirName[:4] != "node" {
+			klog.V(5).Infof("failed to get node NUMA info, err: invalid dir name %s", dirName)
```
zwzhang0107, line 192: these error should be v4 if the prefix of dir is node

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
```diff
@@ -56,13 +57,35 @@ import (
...
+}
+
+func (n *nodeTopologyStatus) updateNRT(nrt *v1alpha1.NodeResourceTopology) {
+	if nrt.Annotations == nil {
+		nrt.Annotations = map[string]string{}
+	}
+	for k, v := range n.Annotations {
+		nrt.Annotations[k] = v
+	}
+
```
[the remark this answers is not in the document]
saintube: Added deep-copy for the origin NRT object. The annotation protocol still contains the sharepool info.

pkg/koordlet/util/meminfo.go
saintube: I'm afraid not. There is no info about the socket and core relations under the numa info directory. If we want to look up the socket ID of the NUMA node, we need to check the cpu info instead.

pkg/koordlet/util/meminfo.go
saintube: done. renamed as `NUMANodeID`

pkg/koordlet/util/meminfo.go
saintube: done

eahydra (commented):
Nice work!
/lgtm

PR #1521, opened 2023-08-02
koordlet: fix the cpu.shares conversion on cgroups-v2

saintube:

koordlet: fix the value conversion between `cpu.shares` and `cpu.weights` (cgroups-v2), use the same conversion as the kubelet.

link: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2254-cgroup-v2#phase-1-convert-from-cgroups-v1-settings-to-v2

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1534, opened 2023-08-07
koord-scheduler: DeviceShare supports scoring

### LINKED ISSUE #1457: [proposal] DeviceShare should support scoring.
**What is your proposal**:

DeviceShare plugin now only supports allocating Devices but missing Scoring.  DeviceShare should support scoring.

**Why is this needed**:

1. The DeviceShare plugin in koord-scheduler supports multiple resource APIs such as `nvidia.com/gpu`, `koordinator.sh/gpu-memory`, `koordinator.sh/gpu-memory-ratio` and `koordinator.sh/gpu` - core, etc. Therefore, a user can create some Pods on the same node using the above resource API, and the DeviceShare plugin will normalize the resource request to "koordinator.sh/gpu-memory" etc., for example, convert "nvidia.com/gpu=1" to "koordinator .sh/gpu-core=100" and "koordinator.sh/gpu-memory-ratio=100", this normalization behavior breaks the "NodeResourceFit" scoring algorithm.

2. In addition, in the GPU Share scenario, multiple Pods may run on one GPU device instance. This usage method can improve resource utilization and reduce costs, but it also has the risk of reducing the high availability of applications. Therefore, it is necessary to provide the ability to spread at the device instance level.

eahydra:

The DeviceShare plugin supports scoring.
- Nodes are scored according to the normalized resource name.
- When allocating devices in the Reserve stage, device instances are sorted according to the scoring strategy.

fix #1457 

### V. Checklist


jasonliu747 (commented):
/lgtm

PR #1535, opened 2023-08-08
koord-descheduler: accumulates the sum of system and Pod usage as node usage

### LINKED ISSUE #990: [proposal] koord-descheduler use sys usage in NodeMetric
**What is your proposal**:
- [ ] [Report sysUsage in NodeMetric](https://github.com/koordinator-sh/koordinator/issues/991).
- [ ] Koord-descheduler use sysUsage for LowNodeLoad.

**Why is this needed**:

For example, the descheduler plugin `LowNodeLoad` uses `nodeMetric.Status.NodeMetric.NodeUsage` to determine whether the load of the node exceeds the high threshold. `nodeMetric.Status.NodeMetric.NodeUsage` does not change immediately after some pods are evicted by the plugin, because by default nodeUsage is an average over 5 minutes. NodeUsage in NodeMetric takes a while to update to the current latest actual utilization. This may cause LowNodeLoad to incorrectly evict some Pods.

There are several ways to alleviate the problem:
- One solution is to reduce the aggregation calculation period of NodeMetric, for example, to 1 minute. However, this method may cause the total amount of resources that can be oversold to fluctuate.
- Another solution is to adjust the run cycle of descheduler, which can alleviate some problems.
- The third way is to add a sysUsage field, and then calculate nodeUsage = ∑usage(existing Pod) + sysUsage as needed.

eahydra:

The `nodeMetric.Status.NodeMetric.NodeUsage` does not change immediately after some pods are evicted by the plugin, because by default nodeUsage is an average over 5 minutes.  So change to the sum of Pod utilization plus sysUsage as Node Usage.

fix #990 

### V. Checklist

### DIFF
--- a/pkg/descheduler/controllers/migration/filter.go
+++ b/pkg/descheduler/controllers/migration/filter.go
@@ -88,9 +88,10 @@ func (r *Reconciler) initFilters(args *deschedulerconfig.MigrationControllerArgs
 		excludedNamespaces = sets.NewString(args.Namespaces.Exclude...)
 	}
 
+	filterPlugin := defaultEvictor.(framework.FilterPlugin)
 	wrapFilterFuncs := podutil.WrapFilterFuncs(
 		util.FilterPodWithMaxEvictionCost,
-		defaultEvictor.(framework.FilterPlugin).Filter,
+		filterPlugin.Filter,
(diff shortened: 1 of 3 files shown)


jasonliu747 (approved):
/lgtm

PR #1537, opened 2023-08-09
koord-scheduler: invalidate nodeDevice when deleting Device CRD object

### LINKED ISSUE #1186: [BUG] When the Device CRD object is deleted, the nodeDevice is deleted unexpectedly
**What happened**:

When the user deletes the device crd object of a certain node, the nodeDevice in the DeviceShare plugin is deleted. Even if the node still exists and the Pod requesting the device is running. As a result, when a new device is created, the device resource on the node is considered ready to be allocated, so the problem of repeated allocation occurs.

Clean up internal state only when the node is gone

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

eahydra:

The user may accidentally delete the Device CRD object, and then the Device CRD object will be recreated by koordlet/DevicePlugin. During this period, the internal state can only be marked as invalid, otherwise the GPU may be repeatedly allocated to different Pods.

And the DeviceShare plugin will start a GC thread to GC the remaining NodeDevice objects that are no longer present.

fix #1186

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_cache.go
+++ b/pkg/scheduler/plugins/deviceshare/device_cache.go
@@ -17,18 +17,28 @@ limitations under the License.
 package deviceshare
 
 import (
+	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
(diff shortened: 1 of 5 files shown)


jasonliu747 (approved):
/lgtm

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -482,3 +512,40 @@ func (n *nodeDeviceCache) getAllNodeDeviceSummary() map[string]*NodeDeviceSummar
...
+			if !nodeNames.Has(name) {
+				delete(n.nodeDeviceInfos, name)
+				klog.InfoS("nodeDevice has been removed since missing Node object", "node", name)
+			}
+		}
+	}
+
+	nodeLister := informerFactory.Core().V1().Nodes().Lister()
+
+	for {
```
jasonliu747, line 533: 建议用 wait.Until

PR #1533, opened 2023-08-07
koord-descheduler: ignore create pmj when pod.schedulerName is not koord-scheduler

lucming:
### Ⅰ. Describe what this PR does
this pr fixed tow problems.
1. `koord-descheduler` should reject create pmj when pmj.mode is `ReservationFirst` and  pod.schedulerName is not koord-scheduler  detail in  https://github.com/koordinator-sh/koordinator/issues/1527 this issue.
2.  update `deschuler` in go.mod, because some bug have been fixed.  eg: https://github.com/kubernetes-sigs/descheduler/pull/1104/files 

### V. Checklist

- [ ] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [ ] All checks passed in `make test`

PR #1540, opened 2023-08-11
koord-descheduler: pmj controller support assign many schedulers that can handle reservation

lucming:
enhancement of https://github.com/koordinator-sh/koordinator/pull/1533
pmj controller can assign many schedulers that can handle reservation.

### V. Checklist


pkg/descheduler/controllers/migration/filter.go
```diff
@@ -121,6 +122,11 @@ func (r *Reconciler) Filter(pod *corev1.Pod) bool {
 	if !r.filterExistingPodMigrationJob(pod) {
 		return false
 	}
+
+	if !r.reservationFilter(pod) {
```
eahydra, line 126: pmj reconcile also needs to check if Reservation is used, but pod.spec.schedulerName is not in the list range, should it also be rejected?

lucming: maybe this should be handled in a webhook, what do you think?

eahydra (approved):
/lgtm
/approve

PR #1543, opened 2023-08-14
koord-scheduler: make sure NodeInfo and other objects have synced before scheduling

### LINKED ISSUE #1544: [BUG] Pod is scheduled on Node which have already remaining nothing
**What happened**:

A Pending Pod that cannot be scheduled on any Node because there are no resources in the cluster, but in fact, once the koord-scheduler starts, the Pod is scheduled on a Node that has no resources.

Pod is still in Pending state.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

eahydra:

When the K8s Scheduler Framework starts, the thread that constructs NodeInfo and the scheduling thread are not synchronized. In this way, when the Pod on a Node is not filled in the NodeInfo, the Node is scheduled for a new Pod. This behavior is not expected. The K8s community itself has also noticed this issue [#116717](https://github.com/kubernetes/kubernetes/issues/116717), but it was only fixed in the K8s v1.28 version [#116729](https://github.com/kubernetes/kubernetes/pull/116729).

The PR force synced objects into cache.ResourceEventHandler  in `AddEventHandler` method.

fix #1544 

### V. Checklist


jasonliu747 (approved):
/lgtm

FillZpp (approved):
/lgtm

hormes (approved):

PR #1551, opened 2023-08-15
koord-scheduler: fix missing AddEventHandler with forceSyncEventHandler

eahydra:

Run E2E and find that the Case cannot pass, and the location discovery is a problem introduced by #1543 .

### V. Checklist


jasonliu747 (commented):
/lgtm

Issue #1470, opened 2023-07-11
[question] Use cycles instead of ref-cycles


Rouzip:
**What happened:**
Use ref-cycles as CPI factor.
**What you expected to happen:**
Use cycles as CPI factor.
**Environment:**

- Koordinator version: - v0.6.2
- Kubernetes version (use kubectl version): v1.22.5
- docker/containerd version: containerd 1.5.0
- OS (e.g: cat /etc/os-release): Ubuntu 20.04.4 LTS
- Kernel (e.g. uname -a): Linux 5.10.112-11.al8.x86_64 #1 SMP Tue May 24 16:05:50 CST 2022 x86_64 x86_64 x86_64 GNU/Linux

**Anything else we need to know:**
CPI is typically measured in cycles rather than ref-cycles as a performance evaluation metric.
References:
1. https://www.brendangregg.com/blog/2014-10-31/cpi-flame-graphs.html
2. https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2023-0/cpu-metrics-reference.html

songtao98:
@Rouzip Sorry for being late to answer. 

When CPI collector is implemented in koordlet, the reference is [_CPI2 : CPU performance isolation for shared compute clusters_](https://static.googleusercontent.com/media/research.google.com/zh-CN//pubs/archive/40737.pdf) It was clarified in Chapter 3.1 as **CPI data is derived from hardware counters,
and is defined as the value of the CPU CLK UNHALTED.REF
counter divided by the INSTRUCTIONS RETIRED counter.**

Besides, personally speaking, the CPU CLK UNHALTED.REF is not affected by thread frequency changes(CPU's dynamic frequency scaling mechanism). So I'll prefer this one but also want to hear more ideas from you. 

/cc @saintube @zwzhang0107

hormes:
> But the retired instructions in constant time doesn't reflect the performance of the application. Under the influence of similar technologies such as Intel SST, the calculation of CPI by the existing algorithm will lead to different results under different frequencies of the same machine, but it does not reflect the corresponding change in the performance of the application. @songtao98 @saintube

Can you explain the difference between these two conters in detail, what is the calculation logic of these two indicators inside the CPU when the frequency changes?

Rouzip:
[...]

PR #1554, opened 2023-08-17
koordlet: add libpfm4&perf group

### LINKED ISSUE #1476: [proposal] Use libpfm4 in perf collector
**What is your proposal**:
Improve observability by adding more core events. Refactor the perf module to support perf group to calculate metrics more accurately.

**Why is this needed**:
`perf-utils` can only collect limited events and it cannot collect group events, we can use `libpfm4` to get more core events and refactor the perf module.
`libpfm4` is developed by Google, and it can parse all kinds of core/uncore events. We can use `CGO` to use it in golang without adding too much dependency.
Some events need to be measured at the same time(such as CPI and MPKI), and limited PMUs may lead to PMU multiplexing, resulting in large deviations in measurement data.

We can use [libpfm4](https://sourceforge.net/p/perfmon2/libpfm4/ci/master/tree/) to parse events in perf collector.
Steps:

1. Init `libpfm4` library, create perfValue pool and byte buffer pool
2. Create perf collector for each container
3. Use syscall `perf_event_open` to create fd for perf group and save the `id` for each event
4. Enable leader perf group to start events schedule in kernel space
5. Wait a period and stop all perf group, read data from leader group fd
6. Close all fds and cgroup fd

bowen-intel:

This pull request enhances the precision of CPI calculation through the utilization of the perf group. Additionally, it introduces the integration of libpfm4 as an event parser, thereby expanding the potential to incorporate additional TMA analyses in subsequent developments.

fixes #1476 #1470 

Use Prometheus.

This pr uses `unsafe.Pointer` to mock `syscall.Syscall6` in unit test, so I have to add unsafeptr tag for `go vet` tool and ignore `github.com/koordinator-sh/koordinator/pkg/koordlet/util/perf` package in test since it use -race tag.
Any comments is welcome.

### V. Checklist


pkg/koordlet/metricsadvisor/collectors/performance/performance_collector_linux.go
```diff
@@ -79,8 +84,22 @@ func (p *performanceCollector) Run(stopCh <-chan struct{}) {
 	if features.DefaultKoordletFeatureGate.Enabled(features.PSICollector) {
 		p.collectPSI(stopCh)
 	}
-	if features.DefaultKoordletFeatureGate.Enabled(features.CPICollector) {
-		go wait.Until(p.collectContainerCPI, p.cpiCollectInterval, stopCh)
+	if p.EnabledPerf() {
```
zwzhang0107, line 87: Is it possible that we provide two utils: one is using perf-util, the other is using libpfm4.
This could be an option, both for comparing and stability in production env.

bowen-intel: Are you suggesting adding the `libpfm4` approach while retaining the existing CPI calculation method?

zwzhang0107: yes, it could be an args, perf/libpfm.

bowen-intel:
Make a A/B test for two method:
## workload
100 nginx
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-nginx
  namespace: nginx
  labels:
    run: my-nginx
spec:
  replicas: 100
  selector:
    matchLabels:
      run: my-nginx
  template:
    metadata:
      labels:
        run: my-nginx
    spec:
      nodeName: inspur-icx-1
      containers:
        - name: my-nginx
          image: nginx:latest
          ports:
            - containerPort: 80
          resources:
            requests:
              cpu: "1"
              memory: 1Gi
            limits:
              cpu: "1"
              memory: 1Gi
---
apiVersion: v1
kind: Service
metadata:
  name: my-nginx
  namespace: nginx
  labels:
    run: my-nginx
spec:
  ports:
    - port: 80
      protocol: TCP
      targetPort: 80
  selector:
    run: my-nginx
```
## environment
Ubuntu 22.04
containerd github.com/containerd/containerd v1.6.8 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
runc version 1.1.4
koordinator(based on v1.3.0)
kubernetes v1.27.2
## libpfm4
image
## perf-util
image

bowen-intel:
Any comments and questions is welcome! 😀 @zwzhang0107 @songtao98

pkg/koordlet/metricsadvisor/collectors/performance/performance_collector_linux.go
```diff
@@ -79,8 +84,25 @@ func (p *performanceCollector) Run(stopCh <-chan struct{}) {
...
+		if features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) {
```
songtao98, line 88: I'm not sure about this logic here: If both CPICollector and Libpfm4 feature-gate are enabeld, use Libpfm4 to collect CPI, if only CPICollector is enabled, use perf-util way. So should it be 

`
if features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) {
...
} else {
...
}
`
Or I misunderstood you? @bowen-intel

bowen-intel: > I'm not sure about this logic here: If both CPICollector and Libpfm4 feature-gate are enabeld, use Libpfm4 to collect CPI, if only CPICollector is enabled, use perf-util way. So should it be
> 
> `if features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) { ... } else { ... }` Or I misunderstood you? @bowen-intel

I use interface and flag to change the `perfCollector`/`perfGroupCollector` in the `getAndStartCollectorOnSingleContainer`, so if {...} else {...} logic is not used in `Run`.

songtao98: > > I'm not sure about this logic here: If both CPICollector and Libpfm4 feature-gate are enabeld, use Libpfm4 to collect CPI, if only CPICollector is enabled, use perf-util way. So should it be
> > `if features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) { ... } else { ... }` Or I misunderstood you? @bowen-intel
> 
> I use interface and flag to change the `perfCollector`/`perfGroupCollector` in the `getAndStartCollectorOnSingleContainer`, so if {...} else {...} logic is not used in `Run`.

Yes, I think it's ok to decide which collector to use there. And please see my second comment to determine whether the logic controlling libpfm4 initialization needs to be decoupled from the logic collecting CPI : )
cc @zwzhang0107

bowen-intel: > Here's some of my opinions:
> 
> 1. `p.EnablePerf()` is related to feature-gate `CPICollector`, but in future we may use Perf to collect other metrics, so this function can be deleted.
> 2. Then we use feature-gate `Libpfm4` to controll if we need to init libpfm4 or init buffer pool, etc.
> 3. To init buffer pool of libpfm4, we need to generate a `eventsMap`, and now you use `if features.DefaultKoordletFeatureGate.Enabled(features.CPICollector)`(line 91). I suggest provide a function `getLibpfm4EventMap`. Then if we add new metrics, only this function's implementation need to be modified. Or, a global featuregate map can be added in file performance_collector_linux.go so `getLibpfm4EventMap` can traverse it.
> 
> The resulting code structure looks like below:
> 
> ```
> func (p *performanceCollector) Run(stopCh <-chan struct{}) {
>         ...
> 
>         if features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) {
>                 ...
>                 eventsMap := p.getLibpfm4EventMap()
>                 perf.InitBufferPool(eventsMap)
>                 ...
>         }
>         
>         if features.DefaultKoordletFeatureGate.Enabled(features.CPICollector) {
>                  ... // p.collectContainerCPI will check Libpfm4 feature-gate again to really decide which way to use to collect CPI(perf-util or libpfm4)
>         }
> }
> 
> func (p *performanceCollector) getLibpfm4EventMap() map[int]struct{} {
> 	...
> }
> ```
> 
> @bowen-intel

1. I think `p.EnablePerf()` can judge whether perf related feature flag used. if I remove it, I also need to check all perf related flags to init `libpfm4` and events map.
2. `Libpfm4` flag is a sub feature gate for perf related flag. If user try flag `Libpfm4` but forget to use `CPI... `, just ignore it maybe a good choice?
3. Good idea! If I understand your meaning, you want to extract these logic like this?
```go
func (p *performanceCollector) getLibpfm4EventMap() map[int]struct{} {
	eventsMap := make(map[int]struct{})
	if features.DefaultKoordletFeatureGate.Enabled(features.CPICollector) {
		eventsMap[len(perf.EventsMap[string(features.CPICollector)])] = struct{}{}
	}
	return eventsMap
}
```

songtao98: > 3. Good idea! If I understand your meaning, you want to extract these logic like this?

Yes it is. Then if more perf related feature gates are enabled, we can only modify here to add more entries in eventsMap. 

> 1. I think p.EnablePerf() can judge whether perf related feature flag used. if I remove it, I also need to check all perf related flags to init libpfm4 and events map. 
> 2. Libpfm4 flag is a sub feature gate for perf related flag. If user try flag Libpfm4 but forget to use CPI... , just ignore it maybe a good choice?

So you are considering that if user only enable `Libpfm4 `flag but not enable `CPICollector`, to initialize libpfm4 is kind of a waste. To use `p.EnablePerf()` is a double-check to ensure this won't happen? 

If so, just to confirm, if another perf related collector is added in the future, we need to and a OR condition in `p.EnablePerf() `return value. If I'm right, I think it's an acceptable implementation. And this PR will be LGTM after `getLibpfm4EventMap()` function is added. 
@bowen-intel 
cc @zwzhang0107

songtao98 (commented):
/lgtm

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
+	buf := bufPool.Get().(*[]byte)
+	defer bufPool.Put(buf)
+	_, err := p.leaderFd.Read(*buf)
+	if err != nil {
+		return err
+	}
+
+	header := &perfValueHeader{}
+	reader := bytes.NewReader(*buf)
+	if err := binary.Read(reader, binary.LittleEndian, header); err != nil {
```
zwzhang0107, line 351: is it necessary to check whether it is LittleEndian or BigEndian

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
+				errMutex.Lock()
+				err = multierr.Append(err, fmt.Errorf("failed to enable perf group, Error: %s, cpu: %d, events: %s", unix.ErrnoName(e1), cpu, events))
+				errMutex.Unlock()
+				return
+			}
+			collector.perfCollectors.Store(cpu, &pc)
+		}(cpu)
+	}
+	wg.Wait()
+	for _, attr := range attrMap {
```
zwzhang0107, line 235: better to do this with defer once the attrMap defined, so that the pointer can be freed, event there is come error happend.

pkg/koordlet/metricsadvisor/collectors/performance/performance_collector_linux.go
```diff
@@ -191,9 +222,11 @@ func (p *performanceCollector) profilePerfOnSingleContainer(status *corev1.Conta
 
 	cpiMetrics = append(cpiMetrics, cpiCycle, cpiInstruction)
 
-	err1 := collectorOnSingleContainer.CleanUp()
-	if err1 != nil {
-		klog.Errorf("collectorOnSingleContainer cleanup err : %v", err1)
+	if !features.DefaultKoordletFeatureGate.Enabled(features.Libpfm4) {
```
zwzhang0107, line 225: the CleanUp can be done with defer or just after GetContainerCyclesAndInstructions.

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
+		"CPICollector": {"cycles", "instructions"},
+	}
+)
+
+func InitBufferPool(eventsNums map[int]struct{}) {
+	BufPools = make(map[int]*sync.Pool)
+	for eventNum := range eventsNums {
+		BufPools[eventNum] = &sync.Pool{
+			New: func() interface{} {
+				pool := make([]byte, 24+eventNum*16)
```
zwzhang0107, line 65: we need more comments about these magic numbers

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
+	var errMutex sync.Mutex
+	for _, cpu := range cpus {
+		go func(cpu int) {
+			// create perf group
+			var pc perfCollector
+			pc.syscall6 = syscallFunc
+			pc.fds = make([]io.ReadCloser, 0, len(events)-1)
+			pc.cpu = cpu
+			defer wg.Done()
+			attr := attrMap[events[0]]
```
zwzhang0107, line 169: what is the difference between events 0 and the others, maybe use different functions for cycles and instructions?

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
+
+	return collector.resultMap, err
+}
+
+func GetContainerCyclesAndInstructionsGroup(collector *PerfGroupCollector) (float64, float64, error) {
+	resMap, err := GetContainerPerfResult(collector)
+	if err != nil {
+		return 0, 0, err
+	}
+	return resMap["cycles"], resMap["instructions"], nil
```
zwzhang0107, line 287: better to define as consts

zwzhang0107:
> Make a A/B test for two method:
> 
> ## workload
> 100 nginx
> 
> ```yaml
> apiVersion: apps/v1
> kind: Deployment
> metadata:
>   name: my-nginx
>   namespace: nginx
>   labels:
>     run: my-nginx
> spec:
>   replicas: 100
>   selector:
>     matchLabels:
>       run: my-nginx
>   template:
>     metadata:
>       labels:
>         run: my-nginx
>     spec:
>       nodeName: inspur-icx-1
>       containers:
>         - name: my-nginx
>           image: nginx:latest
>           ports:
>             - containerPort: 80
>           resources:
>             requests:
>               cpu: "1"
>               memory: 1Gi
>             limits:
>               cpu: "1"
>               memory: 1Gi
> ---
> apiVersion: v1
> kind: Service
> metadata:
>   name: my-nginx
>   namespace: nginx
>   labels:
>     run: my-nginx
> spec:
>   ports:
>     - port: 80
>       protocol: TCP
>       targetPort: 80
>   selector:
>     run: my-nginx
> ```
> 
> ## environment
> CPU Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz Memory 186Gi Ubuntu 22.04 containerd github.com/containerd/containerd v1.6.8 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6 runc version 1.1.4 koordinator(based on v1.3.0) kubernetes v1.27.2
> 
> ## libpfm4
> image
> 
> ## perf-util
> image

this looks great!

it will be better if we have the comparison on accuracy between perf and libpfm4

pkg/koordlet/util/perf/perf_linux.go
```diff
@@ -22,15 +22,377 @@ package perf
...
 )
 
+/*
+#cgo CFLAGS: -I/usr/include
+#cgo LDFLAGS: -lpfm
+#include <perfmon/pfmlib.h>
+#include <stdlib.h>
+#include <string.h>
+*/
+import "C"
```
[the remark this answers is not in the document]
bowen-intel: done

pkg/koordlet/util/perf/perf_linux.go
bowen-intel: Keep same with `perf-utils`, most of the system is LittleEndian.

pkg/koordlet/util/perf/perf_linux.go
bowen-intel: events[0] is leader event. Some perf attr bit is different.

bowen-intel:
> > Make a A/B test for two method:
> > ## workload
> > 100 nginx
> > ```yaml
> > apiVersion: apps/v1
> > kind: Deployment
> > metadata:
> >   name: my-nginx
> >   namespace: nginx
> >   labels:
> >     run: my-nginx
> > spec:
> >   replicas: 100
> >   selector:
> >     matchLabels:
> >       run: my-nginx
> >   template:
> >     metadata:
> >       labels:
> >         run: my-nginx
> >     spec:
> >       nodeName: inspur-icx-1
> >       containers:
> >         - name: my-nginx
> >           image: nginx:latest
> >           ports:
> >             - containerPort: 80
> >           resources:
> >             requests:
> >               cpu: "1"
> >               memory: 1Gi
> >             limits:
> >               cpu: "1"
> >               memory: 1Gi
> > ---
> > apiVersion: v1
> > kind: Service
> > metadata:
> >   name: my-nginx
> >   namespace: nginx
> >   labels:
> >     run: my-nginx
> > spec:
> >   ports:
> >     - port: 80
> >       protocol: TCP
> >       targetPort: 80
> >   selector:
> >     run: my-nginx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ## environment
> > CPU Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz Memory 186Gi Ubuntu 22.04 containerd github.com/containerd/containerd v1.6.8 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6 runc version 1.1.4 koordinator(based on v1.3.0) kubernetes v1.27.2
> > ## libpfm4
> > image
> > ## perf-util
> > image
> 
> this looks great!
> 
> it will be better if we have the comparison on accuracy between perf and libpfm4

## libpfm4
image
cycles: 9694773088
instructions: 2817277063

## perf
```bash
perf stat -e '{cycles,instructions}' -G xxx.scope -a sleep 10
```
image

pkg/koordlet/util/perf_group/perf_group_linux.go
```diff
@@ -0,0 +1,409 @@
...
+				err = multierr.Append(err, fmt.Errorf("failed to enable perf group, Error: %s, cpu: %d, events: %s", unix.ErrnoName(e1), cpu, events))
+				errMutex.Unlock()
+				return
+			}
+			collector.perfCollectors.Store(cpu, &pc)
+		}(cpu)
+	}
+	wg.Wait()
+	for _, attr := range attrMap {
+		C.free(unsafe.Pointer(attr))
```
zwzhang0107, line 251: is the memory leak possible if return at L162 with error

Rouzip: Fix with `defer C.free(attr)` after create attr.

Rouzip: > is the memory leak possible if return at L162 with error

Fix with defer C.free(attr) after create attr.

hormes (approved):

PR #1555, opened 2023-08-17
koord-scheduler: export koordinator-root-quota as CRD object

### LINKED ISSUE #1519: [proposal] Add the koordinator-root-quota CRD in koordinator and include the koordinator-root-quota information in quota debug API.
**What is your proposal**:

- Background
  - When I use [quota debug API](https://koordinator.sh/docs/user-manuals/multi-hierarchy-elastic-quota-management/#quota-debug-api) in the ElasticQuota plugin to view quotas information, I can query koordinator-default-quota and koordinator-system-quota, but there is no information about koordinator-root-quota.
image

- Proposal
  - We hope to add the koordinator-root-quota CRD in koordinator and include the koordinator-root-quota information in quota debug API.(apis/v1/plugins/ElasticQuota/quotas).

**Why is this needed**:

- Adding this information to the debug API of the ElasticQuota plugin can make it more convenient for users to observe the quota allocation information, mainly for runtime information. Moreover, from the perspective of logical completeness, the quotas information in the debug API should also include information related to koordinator-root-quota.

- Solution: Similar to koordinator-default-quota and koordinator-system-quota, once the groupQuotaManager is created, create the koord-root-quota CRD in the code. In this way, the corresponding quota information can be directly viewed in the debug API.

tan90github:

Add the koordinator-root-quota CRD in koordinator.

fixes #1519
### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -172,9 +172,6 @@ type ElasticQuotaArgs struct {
 	// SystemQuotaGroupMax limit the maxQuota of SystemQuotaGroup
 	SystemQuotaGroupMax corev1.ResourceList
 
-	// RootQuotaGroupMax limit the maxQuota of RootQuotaGroup
-	RootQuotaGroupMax corev1.ResourceList
-
 	// QuotaGroupNamespace is the namespace of the DefaultQuotaGroup/SystemQuotaGroup
 	QuotaGroupNamespace string
 
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -172,6 +172,9 @@ type ElasticQuotaArgs struct {
 	// SystemQuotaGroupMax limit the maxQuota of SystemQuotaGroup
 	SystemQuotaGroupMax corev1.ResourceList
 
+	// RootQuotaGroupMax limit the maxQuota of RootQuotaGroup
+	RootQuotaGroupMax corev1.ResourceList
+
 	// QuotaGroupNamespace is the namespace of the DefaultQuotaGroup/SystemQuotaGroup
 	QuotaGroupNamespace string
 
--- a/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
+++ b/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
@@ -809,7 +809,6 @@ func (gqm *GroupQuotaManager) resetRootQuotaUsedAndRequest() {
 	systemQuotaInfo := gqm.getQuotaInfoByNameNoLock(extension.SystemQuotaName)
 	defaultQuotaInfo := gqm.getQuotaInfoByNameNoLock(extension.DefaultQuotaName)
 
-	rootQuotaInfo.CalculateInfo.Used = quotav1.Add(systemQuotaInfo.GetUsed(), defaultQuotaInfo.GetUsed()).DeepCopy()
-	rootQuotaInfo.CalculateInfo.Request = quotav1.Add(systemQuotaInfo.GetRequest(), defaultQuotaInfo.GetRequest()).DeepCopy()
-
+	rootQuotaInfo.CalculateInfo.Used = quotav1.Add(systemQuotaInfo.GetUsed(), defaultQuotaInfo.GetUsed())
+	rootQuotaInfo.CalculateInfo.Request = quotav1.Add(systemQuotaInfo.GetRequest(), defaultQuotaInfo.GetRequest())
 }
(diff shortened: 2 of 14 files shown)


pkg/scheduler/apis/config/types.go
```diff
@@ -171,6 +171,9 @@ type ElasticQuotaArgs struct {
 	// SystemQuotaGroupMax limit the maxQuota of SystemQuotaGroup
 	SystemQuotaGroupMax corev1.ResourceList
 
+	// RootQuotaGroupMax limit the maxQuota of RootQuotaGroup
+	RootQuotaGroupMax corev1.ResourceList
```
eahydra, line 175: IMO  the root-quota represents the total resources of the cluster, there should not be a separate parameter to set the default size. And what this PR expects is to expose the internal root-quota data through ElasticQuota CRD, so I understand that it is enough to modify the syncHandler in elasticquota/controller.go? Of course, there are some supporting modifications, such as the verification of webhook.

tan90github: You're right, it's not appropriate to set the max of root-quota through separate parameters. I'll make that change. However, just modifying the syncHandler in elasticquota/controller.go may not be enough to update certain data within the root-quota (e.g. request, used, runtime, etc.), because the existing code for updating used and request often "skips" the root-quota. So, we prefer to update the relevant data of the root-quota through the code we've modified. If you have any better suggestions for updating the root-quota parameters,  I am willing to consider them.

eahydra: In synchandler, count all the used and requested, and use the results as the used and requested of root-quota. The total capacity is the max of root-quota. Is this possible?

tan90github: That may be feasible, but wouldn't it look strange to update the root-quota information within the syncHandler? Why not directly update the 'used' value of root-quota in the updateGroupDeltaUsedNoLock? Wouldn't that be more logical? Our development approach is to update the relevant information of root-quota directly using the existing 'used' update method, to integrate with the existing code as much as possible and eliminate unnecessary 'skips'. Is there any specific design in the existing code development?

tan90github: fixed

pkg/scheduler/plugins/elasticquota/plugin_helper.go
```diff
@@ -20,9 +20,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"sort"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/resource"
```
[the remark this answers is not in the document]
tan90github: fixed

pkg/scheduler/plugins/elasticquota/plugin_helper.go
```diff
@@ -144,6 +146,42 @@ func (g *Plugin) createSystemQuotaIfNotPresent() {
...
 }
 
+// createRootQuotaIfNotPresent create RootQuotaGroup's CRD
+func (g *Plugin) createRootQuotaIfNotPresent() {
+	eq, _ := g.quotaLister.ElasticQuotas(g.pluginArgs.QuotaGroupNamespace).Get(extension.RootQuotaName)
+	if eq != nil {
+		klog.Infof("RootQuota already exists, skip create it.")
+		return
+	}
+	defaultRootQuotaGroupMax := v1.ResourceList{
```
eahydra, line 156: 这里只是为了填充一个默认值，那么我理解不设置也OK？

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
```diff
@@ -710,3 +719,15 @@ func getPodName(oldPod, newPod *v1.Pod) string {
...
+
+func (gqm *GroupQuotaManager) resetRootQuotaUsedAndRequest() {
+	rootQuotaInfo := gqm.getQuotaInfoByNameNoLock(extension.RootQuotaName)
+	rootQuotaInfo.lock.Lock()
+	defer rootQuotaInfo.lock.Unlock()
+
+	systemQuotaInfo := gqm.getQuotaInfoByNameNoLock(extension.SystemQuotaName)
+	defaultQuotaInfo := gqm.getQuotaInfoByNameNoLock(extension.DefaultQuotaName)
+
+	rootQuotaInfo.CalculateInfo.Used = quotav1.Add(systemQuotaInfo.GetUsed(), defaultQuotaInfo.GetUsed()).DeepCopy()
```
eahydra, line 731: quotav1包里的函数会返回的结果本身就是一个独立的resourceList了。所以这里不需要DeepCopy

pkg/scheduler/plugins/elasticquota/plugin_helper.go
tan90github: 是的，这里是将rootQuotaMax字段设置为和System等一样的最大值了，可以不设置。因为后续没有使用到rootQuota的MAX。

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
tan90github: done

eahydra (commented):
/lgtm

hormes (approved):

eahydra (approved):
/lgtm

PR #1556, opened 2023-08-18
koordlet: resource qos config supports SYSTEM qos

saintube:

koordlet: Add the ResourceQOS configurations for qos=SYSTEM pods.

When a pod reclaims its QoSClass as SYSTEM, the current ResourceQOS strategy can fail to parse the config for the pod since there is no SYSTEM class config defined. Especially, it can cause panic when CgroupReconcile is enabled and a SYSTEM pod is running on the node.

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1559, opened 2023-08-19
koord-manager: improve batch resource calculation with system metrics

### LINKED ISSUE #1427: [BUG] SLO-Controller calculation batch-cpu fluctuates frequently
**What happened**:

The BE Resource is calculated by the formula ```Node.Total - Node.Reserved - System.Used - Pod(non-BE).Used```, where ```System.Used = Node.Used - Pod(All).Used```. However, the System.Used calculation implementation has some problems, as follows: 
* At time t1, NodeMetric is updated; at time t2, the controller calculates NodeResource based on the update of NodeMetric. 
* Pod-B ends at time t2, and the controller regards the usage of Pod-B as System.Used

| component  | t1 | t2 | t3 | 
| ------------- | ------------- | ------------- | ------------- |
| koordlet  | node usage：2.5 core <br/> BE Pod-A: 1 core <br/> BE Pod-B: 1 core <br/> update NodeMetrics  | node usage：1.5 core <br/> BE Pod-A: 1 core <br/> BE Pod-B: terminated <br/> update NodeMetrics |
| slo-controller  | | node usage：2.5 core <br/> BE Pod-A: 1 core <br/> BE Pod-B: terminated <br/> calculate node resource | node usage：1.5 core <br/> BE Pod-A: 1 core <br/> BE Pod-B: terminated <br/> calculate node resource |

Calculated according to the code of https://github.com/koordinator-sh/koordinator/blob/v1.2.0/pkg/slo-controller/noderesource/plugins/batchresource/plugin.go#L82

| t1 | t2 | t3 | 
| ------------- | ------------- | ------------- |
| System.Used：0.5 core <br/> Batch-CPU: x core |  System.Used：1.5 core <br/> Batch-CPU: x-1 core  | System.Used：0.5 core <br/> Batch-CPU: x core |

Expecting time t2, the controller calculates System.Used = 0.5 core

Node.Used uses the average value of the last 60 seconds, and Pod(All).Used uses the current value, which also causes the System.Used value to be inaccurate.

saintube:

koord-manager: Improve the calculation of the batch resource allocatable with the newly-added system metrics.

Fixes #1427.

Please note that the calculated allocatable can be lower than the real result at the nodeMetric reported time. It becomes accurate when there are no more pods terminated or newly-scheduled between the nodeMetric reported time and the controller calculating time.

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
@@ -81,6 +81,11 @@ func (p *Plugin) Reset(node *corev1.Node, message string) []framework.ResourceIt
 
 // Calculate calculates Batch resources using the formula below:
 // Node.Total - Node.Reserved - System.Used - Pod(High-Priority).Used, System.Used = Node.Used - Pod(All).Used.
+// As node and podList are the nearly latest state at time T1, the resourceMetrics are the node metric and pod
+// metrics collected and snapshot at time T0 (T0 < T1). There can be gaps between the states of T0 and T1.
+// We firstly calculate an infimum of the batch allocatable at time T0.
+// `BatchAllocatable0 = NodeAllocatable * ratio - SystemUsed0 - Pod(HP and in Pods1).Used0` - Pod(not in Pods1).Used0.
+// Then we minus the sum requests of the pods newly scheduled but have not been reported metrics to give a safe result.
 func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *corev1.Node, podList *corev1.PodList,
 	resourceMetrics *framework.ResourceMetrics) ([]framework.ResourceItem, error) {
 	if strategy == nil || node == nil || podList == nil || resourceMetrics == nil || resourceMetrics.NodeMetric == nil {
@@ -96,15 +101,18 @@ func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *cor
 
 	// compute the requests and usages according to the pods' priority classes.
 	// HP means High-Priority (i.e. not Batch or Free) pods
-	// pod(HP).Used = pod(All).Used - pod(Batch/Free).Used
-	podAllUsed := util.NewZeroResourceList()
 	podHPRequest := util.NewZeroResourceList()
 	podHPUsed := util.NewZeroResourceList()
+	// podAllUsed is the sum usage of all pods reported in NodeMetric.
+	// podKnownUsed is the sum usage of pods which are both reported in NodeMetric and shown in current pod list.
+	podAllUsed := util.NewZeroResourceList()
+	podKnownUsed := util.NewZeroResourceList()
 
 	nodeMetric := resourceMetrics.NodeMetric
 	podMetricMap := make(map[string]*slov1alpha1.PodMetricInfo)
 	for _, podMetric := range nodeMetric.Status.PodsMetric {
 		podMetricMap[util.GetPodMetricKey(podMetric)] = podMetric
+		podAllUsed = quotav1.Add(podAllUsed, getPodMetricUsage(podMetric))
 	}
 
 	for i := range podList.Items {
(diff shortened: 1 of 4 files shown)


pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -113,36 +119,47 @@ func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *cor
 			continue
 		}
 
+		// mark listed pod metric
+		podKey := util.GetPodKey(pod)
+		podMetric, hasMetric := podMetricMap[podKey]
```
zwzhang0107, line 124: define another map to mark whether the podMetric has been added

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -96,10 +101,11 @@ func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *cor
 
 	// compute the requests and usages according to the pods' priority classes.
 	// HP means High-Priority (i.e. not Batch or Free) pods
-	// pod(HP).Used = pod(All).Used - pod(Batch/Free).Used
-	podAllUsed := util.NewZeroResourceList()
 	podHPRequest := util.NewZeroResourceList()
 	podHPUsed := util.NewZeroResourceList()
+	// podMissingUsed denotes the sum usage of pods which is reported in NodeMetric but not listed currently.
+	// It could be regarded as the HP part.
+	podMissingUsed := util.NewZeroResourceList()
```
zwzhang0107, line 108: danglingPod/unknownPriorityPod

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -113,36 +119,47 @@ func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *cor
...
+	for podKey, podMetric := range podMetricMap {
+		if podMetric == nil { // pod is in the current pod list
+			continue
 		}
-		podAllUsed = quotav1.Add(podAllUsed, getPodMetricUsage(podMetric))
+		podMissingUsed = quotav1.Add(podMissingUsed, getPodMetricUsage(podMetric))
+		hasMissingPod = true
+		klog.V(6).InfoS("batch resource got missing pods", "node", node.Name, "pod", podKey)
+	}
+	if hasMissingPod {
```
zwzhang0107, line 152: hasMissingPod seems not necessary since podMissingUsed will be zero if hasMissingPod is false

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
saintube: no longer change the map, use the subtraction between all pods' used and shown pods's used

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
saintube: fixed

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
saintube: this check was just for logging. now always log the dangling used with v=6

pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -113,36 +121,39 @@ func (p *Plugin) Calculate(strategy *configuration.ColocationStrategy, node *cor
...
+		if !isPodHighPriority {
 			continue
 		}
-
-		if isPodHighPriority {
+		podHPRequest = quotav1.Add(podHPRequest, podRequest)
+		if hasMetric {
 			podHPUsed = quotav1.Add(podHPUsed, getPodMetricUsage(podMetric))
+		} else {
+			podHPUsed = quotav1.Add(podHPUsed, podRequest)
```
j4ckstraw, line 142: why use Request if hos no metric, how about skip?

PR #1563, opened 2023-08-21
koord-scheduler: skip checking for nodes whose nodemetrics update time has expired

lucming:
In some cases, `nodemetric` is not updated for a long time, eg: koordlet be deleted. which will result in nodemetric not being able to report, and if the latest load in `nodemetric` is very high, it will result in the node not being able to run any pods, even if the actual load is very low

At this point, the `loadaware` scheduling plugin's calculations should not be counted in the final scheduling result, because the plugin strongly relies on `nodemetric`

in filter phase: do not handle these nodes that nodemetric have not been updated for a long time
in score phase: score 0 and return directly

### V. Checklist


pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -142,10 +141,9 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 		return framework.NewStatus(framework.Error, err.Error())
 	}
 
-	if p.args.FilterExpiredNodeMetrics != nil && *p.args.FilterExpiredNodeMetrics && p.args.NodeMetricExpirationSeconds != nil {
-		if isNodeMetricExpired(nodeMetric, *p.args.NodeMetricExpirationSeconds) {
-			return framework.NewStatus(framework.Unschedulable, ErrReasonNodeMetricExpired)
-		}
+	if p.args.FilterExpiredNodeMetrics != nil && *p.args.FilterExpiredNodeMetrics &&
```
eahydra, line 144: Maybe we should mark the args.FiterExpiredNodeMetrics always as true? WDYT? @lucming

lucming: > Maybe we should mark the args.FiterExpiredNodeMetrics always as true? WDYT? @lucming

agree with you, or how about just removing args.FiterExpiredNodeMetrics?

eahydra: If it is removed, there will be some compatibility issues. For example, users who are already using it need to modify the configuration. Of course, you can also consider implementing a custom conversion to solve this compatibility problem. You can refer to:

https://github.com/koordinator-sh/koordinator/blob/11f1bbb4aee81838b7d4b1c46fe481df7f379a67/pkg/descheduler/apis/config/v1alpha2/conversion_plugins.go#L25-L47

lucming: I've marked FiterExpiredNodeMetrics as deprecated, this field can be removed when the api is upgraded next time

eahydra (approved):
/lgtm
/approve

jasonliu747 (commented):
/lgtm

PR #1575, opened 2023-08-24
Fix webhook patch manifests for pod update mutation

FillZpp:

Fix webhook patch manifests for pod update mutation


eahydra (approved):
/lgtm
/approve

PR #1576, opened 2023-08-24
koord-scheduler: fix missing IsParent when rebuild tree

shaloulcy:

Fix IsParent in the quota tree rebuild

some parent quotas has no is-parent label. we should fix it in the rebuild phase

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1578, opened 2023-08-25
koordlet: revise nri server start

saintube:

koordlet: revise nri server starting, check if nri.sock exists before new the server.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/nri/server.go
+++ b/pkg/koordlet/runtimehooks/nri/server.go
@@ -34,7 +34,7 @@ import (
 	rmconfig "github.com/koordinator-sh/koordinator/pkg/runtimeproxy/config"
 )
 
-type nriconfig struct {
+type nriConfig struct {
 	Events []string `json:"events"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -47,8 +47,23 @@ type Options struct {
 	Executor      resourceexecutor.ResourceUpdateExecutor
 }
 
+func (o Options) Validate() error {
+	// a fast check for the NRI support status
+	var completeNriSocketPath string
+	if len(o.NriSocketPath) > 0 {
+		completeNriSocketPath = filepath.Join(system.Conf.VarRunRootDir, o.NriSocketPath)
+	} else {
+		completeNriSocketPath = filepath.Join(system.Conf.VarRunRootDir, "nri/nri.sock")
```
kangclzjc, line 56: it seems this line will never be executed. o.NriSocketPath will never be empty because when we NewDefaultConfig, it will always set RuntimeHooksNRISocketPath to nri/nri.sock. see this https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/runtimehooks/config.go#L102.

saintube: Yes, it was only for testing. Let me remove it.

Issue #1570, opened 2023-08-23
[proposal] CPU Normalization


zqzten:
**What is your proposal**:
Introduce mechanism to normalize CPU computing power of heterogeneous Kubernetes Nodes for CPUShare workloads.
In other words, it makes workloads with the same CPU request get the same CPU computing power on different Nodes with different CPUs.

**Why is this needed**:
It's quite common for users to have a Kubernetes cluster with heterogeneous nodes nowadays. As CPUs with different architecture and generations may have large performance gap, different pods of workload with the same CPU request may get very different CPU computing power and thus leads to resource waste (for those running on powerful CPUs) or unexpected application performance degradation (for those running on less powerful CPUs).

PR #1591, opened 2023-08-29
apis: add cpu normalization ratio and cpu basic info annotations

saintube:

apis: add the annotations of the cpu normalization ratio and cpu basic info, and add related funcs.

Related issue #1570.

### V. Checklist

### DIFF
--- a/apis/extension/cpu_normalization.go
+++ b/apis/extension/cpu_normalization.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+import (
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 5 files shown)


apis/extension/cpu_normalization.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"fmt"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// LabelCPUNormalizationRatio denotes the cpu-normalization ratio of the node CPU.
+	LabelCPUNormalizationRatio = NodeDomainPrefix + "/cpu-normalization-ratio"
```
zwzhang0107, line 28: better to be an annotation since it won't be used for select and can be extended as a struct with other overcommit fields.

zqzten (approved):

apis/extension/cpu_normalization.go
saintube: done

apis/extension/cpu_normalization.go
```diff
@@ -0,0 +1,118 @@
...
+	return &info, nil
+}
+
+// SetCPUBasicInfo sets the cpu basic info at the node-level annotations.
+// It returns true if the annotations changes.
+func SetCPUBasicInfo(annotations map[string]string, info *CPUBasicInfo) bool {
+	b, _ := json.Marshal(info)
+	s := string(b)
+
+	old, ok := annotations[AnnotationCPUBasicInfo]
```
zqzten, line 111: can be simplified to below code?
```
if old := annotations[AnnotationCPUBasicInfo]; old == s {
    return false
}
```

saintube: done

hormes (approved):

PR #1592, opened 2023-08-29
koord-scheduler: support NUMA-aware scheduling

eahydra:

- Plugin `NodeNUMAResource` supports scheduling Pods based on NUMA node topology and policies defined on Node or NRT.
- Currently the plugin supports all strategies including `BestEffort`, `Restricted` and `SingleNUMANode`.
- And it is also compatible with the original CPUSet scheduling logic
- But it does not support Pod allocation and preemption from Reservation and Node based on NUMA Node Topology, which I will implement later.
- And it does not support scoring based on the NUMA node topology. I haven't figured out how to implement a new scoring algorithm to be compatible with the original CPUSet scenario, or I can consider scoring only according to the resources of the NUMA Node.

All E2E cases have passed.

part of #228 

### V. Checklist


jasonliu747 (commented):
/lgtm

hormes (approved):

PR #1052, opened 2023-02-24
runtime-proxy: support cri v1 api

maaoBit:
### Ⅰ. Describe what this PR does

add cri v1 api to koord-runtime-proxy

### Ⅱ. Does this pull request fix one issue?

fix #1051

### Ⅳ. Special notes for reviews
* runtime-proxy registry cri server according to the cri verion supported by backendruntime.
* `RuntimeResourceExecutor` move up to cri v1 api
* v1alpha2 cri server uses `RuntimeResourceExecutor` by converting v1alpha2 object to v1 object

### V. Checklist

- [x] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [x] All checks passed in `make test`

Issue #1051, opened 2023-02-24
[proposal] add cri v1 api to koord-runtime-proxy


maaoBit:
**What is your proposal**:
Koord-runtime-proxy support cri v1 api, and support v1alpha in parallel

**Why is this needed**:
According to https://github.com/kubernetes/kubernetes/pull/110618, k8s v1.26 removed cri v1alpha2 support. 
In order to run on k8s 1.26+, koord-runtime-proxy support cri v1 api.

**Is there a suggested solution, if so, please add it**:

hormes:
Let me sort out the ideas, the support of cri-api api version:

version | cri-api supported | kubelet import
----- | ------- | --------
1.19 | v1alpha2 | v1alpha2
1.20 | v1alpha2, v1  | v1alpha2
1.21 |  v1alpha2, v1 |  v1alpha2
1.22 | v1alpha2, v1 |  v1alpha2
1.23 | v1alpha2, v1 |  v1
1.24 | v1alpha2, v1 |  v1
1.25 | v1alpha2, v1  |  v1
1.26  | v1             | v1

In order to simplify the compatibility cost of runtime-proxy, we only plan to support v1 frontend and backend. For kubernetes versions prior to 1.23, use RuntimeProxy 1.3 or prior. For kubernetes >= 1.23 clusters, use RuntimeProxy 1.4 to support.

PR #1593, opened 2023-08-29
Fix runtime proxy compatibility with CRI interface version

hormes:
Ref #1051 

This is an extension of the previous work on #1052 . The updated content here is:

1. For unsupported rpc calls, they are transparently forwarded to the backend by default to deal with the increase or decrease of interfaces on minor versions
2. Delete the processing of the image service part, and go to the process of 1
3. Removed some code generated by mockgen, I should avoid using this way to write one-sided


eahydra (approved):
/lgtm
/approve

PR #1594, opened 2023-08-29
koord-scheduler: deviceshare plugin skips handling nodes without device cr

lucming:
background:
Initially, gpu scheduling was implemented through the gpu device plugin, and after switching to koord-scheduler scheduling, the pod could not be scheduled, prompts that no device resources.

reason:
The gpu node doesn't have a koordlet, but it does report gpu info to node.status via the device plugin, so it can be scheduled by default-scheduler. But koord-scheduler's deviceshare plugin intercepts scheduling pod because there are no device resources.

resolve:
deviceshare skips handling nodes without device

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/plugin.go
+++ b/pkg/scheduler/plugins/deviceshare/plugin.go
@@ -293,7 +293,7 @@ func (p *Plugin) Filter(ctx context.Context, cycleState *framework.CycleState, p
 
 	nodeDeviceInfo := p.nodeDeviceCache.getNodeDevice(node.Name, false)
 	if nodeDeviceInfo == nil {
-		return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrMissingDevice)
+		return nil
 	}
 
 	reservationRestoreState := getReservationRestoreState(cycleState)
@@ -343,7 +343,7 @@ func (p *Plugin) FilterReservation(ctx context.Context, cycleState *framework.Cy
 
 	nodeDeviceInfo := p.nodeDeviceCache.getNodeDevice(nodeName, false)
 	if nodeDeviceInfo == nil {
-		return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrMissingDevice)
+		return nil
 	}
 
 	preemptible := appendAllocated(nil, restoreState.mergedUnmatchedUsed, state.preemptibleDevices[nodeName])
@@ -372,7 +372,7 @@ func (p *Plugin) Reserve(ctx context.Context, cycleState *framework.CycleState, 
 
 	nodeDeviceInfo := p.nodeDeviceCache.getNodeDevice(nodeName, false)
 	if nodeDeviceInfo == nil {
-		return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrMissingDevice)
+		return nil
 	}
 
 	reservationRestoreState := getReservationRestoreState(cycleState)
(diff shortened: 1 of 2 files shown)


eahydra:
We've also had this problem for the last few days. But such modifications may not be enough. 
For example, there may be both nvidia device-plugin and koordlet on a node, then the scheduler needs to check whether the Device CRD object exists at this time; but if there is only nvidia device-plugin, then there is no need to check.

So I have an idea to add a label on the node dimension, such as `node.koordinator.sh/gpu-resource-model`, the corresponding value is as follows:
- `Shared` indicates that only the Koordinator GPU Shared protocol is supported
- `Vendor` indicates that only vendor protocols are supported, such as nvidia.com/gpu
- `Mixed` means that both of the above are supported
- `Native` indicates that the node is in native mode and does not need to be processed by the device-share plugin of the scheduler.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -293,7 +293,9 @@ func (p *Plugin) Filter(ctx context.Context, cycleState *framework.CycleState, p
 
 	nodeDeviceInfo := p.nodeDeviceCache.getNodeDevice(node.Name, false)
 	if nodeDeviceInfo == nil {
-		return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrMissingDevice)
+		state.skip = true
```
eahydra, line 296: And the lifetime of the state is one scheduling cycle, so cannot fix as this.

lucming: Yes, it will work for the entire scheduling cycle, and I'm handling it this way in the hope that the plugin wouldn't handle nodes without devices.

so, if there is no device it will only be affected by the result of noderesourcefit, and if there is a device it will be handled by deviceshare

eahydra: But state.skip means that Pod does not need to request GPU or other devices.

eahydra:
I think it's complicated. In fact, it is only useful in the `Native` scenario, but in this scenario, judging whether it is `Native` is equivalent to not checking the Device CRD.

eahydra:
> We've also had this problem for the last few days. But such modifications may not be enough. For example, there may be both nvidia device-plugin and koordlet on a node, then the scheduler needs to check whether the Device CRD object exists at this time; but if there is only nvidia device-plugin, then there is no need to check.
> 
> So I have an idea to add a label on the node dimension, such as `node.koordinator.sh/gpu-resource-model`, the corresponding value is as follows:
> 
> * `Shared` indicates that only the Koordinator GPU Shared protocol is supported
> * `Vendor` indicates that only vendor protocols are supported, such as nvidia.com/gpu
> * `Mixed` means that both of the above are supported
> * `Native` indicates that the node is in native mode and does not need to be processed by the device-share plugin of the scheduler.
> 
> WDYT? @lucming


I think it's complicated. In fact, it is only useful in the Native scenario, but in this scenario, judging whether it is Native is equivalent to not checking the Device CRD.

lucming:
> > We've also had this problem for the last few days. But such modifications may not be enough. For example, there may be both nvidia device-plugin and koordlet on a node, then the scheduler needs to check whether the Device CRD object exists at this time; but if there is only nvidia device-plugin, then there is no need to check.
> > So I have an idea to add a label on the node dimension, such as `node.koordinator.sh/gpu-resource-model`, the corresponding value is as follows:
> > 
> > * `Shared` indicates that only the Koordinator GPU Shared protocol is supported
> > * `Vendor` indicates that only vendor protocols are supported, such as nvidia.com/gpu
> > * `Mixed` means that both of the above are supported
> > * `Native` indicates that the node is in native mode and does not need to be processed by the device-share plugin of the scheduler.
> > 
> > WDYT? @lucming
> 
> I think it's complicated. In fact, it is only useful in the Native scenario, but in this scenario, judging whether it is Native is equivalent to not checking the Device CRD.

agree with you (^^)

lucming:
We also need to clean up node.status.allocatable after the device resource has been deleted, this is part of the logic in koord-manger, I'll fix it by another pr.

eahydra (approved):
/lgtm
/approve

PR #1598, opened 2023-08-31
proposal: CPU Normalization

zqzten:
This is the proposal of #1570.


zqzten:
Update: Move the scheduler part to #1700 and update some impl details with the latest actual impls.

/unhold PTAL @hormes

zqzten:
Update: Move the scheduler part to https://github.com/koordinator-sh/koordinator/pull/1700 and update some impl details with the latest actual impls.

/unhold
PTAL @hormes

hormes (approved):

PR #1600, opened 2023-09-01
koord-scheduler: supports plugins to adjust reservation reusable resources

eahydra:

Plugins like DeviceShare will add additional normalized resources after the scheduling is successful, and there may be more resources than Pod expects to apply for. In the Reservation scenario, these need to be backfilled to the Reservation and be perceived by subsequent links, so as to complete similar scenarios. resource reuse. This PR is to solve this problem.

### V. Checklist


eahydra:
/hold cancel
PTAl @hormes @FillZpp  @jasonliu747 @saintube  @ZiMengSheng

hormes (approved):

Issue #1539, opened 2023-08-11
[proposal] Support node resource amplification


zqzten:
**What is your proposal**:
Introduce mechanism to support node resource amplification by configuring a factor.

**Why is this needed**:
There're cases where ppl would like to amplify the resources of a node, for CPUs e.g.
1. The node is equipped with a new-gen CPU, which is way faster than those on other nodes.
2. The node's CPU allocatable is nearly full but the actual CPU usage remains at an extremely low level.

PR #1602, opened 2023-09-01
koord-scheduler: support node cpu amplification

zqzten:

This is the scheduler part of Node CPU amplification. It implements below logic described in the proposal:

> On koord-scheduler side, a scheduling plugin filters and scores Nodes with similar logic to the `NodeResourcesFit` plugin, but with one difference: It amplifies the request of CPUSet Pods with the Node's CPU amplification ratio, as CPUSet Pods would get real physical CPUs requested so that their requests would occupy more CPU resources than what they specify on Nodes with allocatable CPU amplified. For example, a CPUSet Pod with CPU request 1 would occupy 1 physical CPU on a Node, if the Node's CPU amplification ratio is 2, the Pod would actually occupy 2 amplified CPUs (which are mapped to 1 physical CPU) on this Node.

Part of #1539

### V. Checklist


pkg/scheduler/plugins/cpunormalization/plugin.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Plugin checks if a node has sufficient CPU resources under CPU normalization.
+// It works like the NodeResourcesFit plugin but with CPU requests of CPUSet Pods amplified on Nodes with CPU normalization.
+type Plugin struct {
+	handle framework.Handle
+	resourceAllocationScorer
+}
+
+func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error) {
```
saintube, line 76: Do we need to replace the native ResourceFit plugin (and keep compatible with it in the next versions), or use the Transformer mechanism to mutate NodeInfo and SharedPodList?
@eahydra WDYT?

zqzten: I've discussed with @eahydra before and he suggested not to replace the native plugin nor modify the global cache, so I choosed this implementation finally.

saintube: OK.

pkg/scheduler/plugins/cpunormalization/least_allocated.go
```diff
@@ -0,0 +1,76 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file contains code derived from and/or modified from Kubernetes
```
eahydra, line 18: If not modified, you can just keep the kubernetes license. and add the file into .licenseignore to skip license-linter.

pkg/scheduler/plugins/cpunormalization/plugin.go
```diff
@@ -0,0 +1,241 @@
...
+var (
+	_ framework.EnqueueExtensions = &Plugin{}
+
+	_ framework.PreFilterPlugin = &Plugin{}
+	_ framework.FilterPlugin    = &Plugin{}
+	_ framework.ScorePlugin     = &Plugin{}
+)
+
+var nodeResourceStrategyTypeMap = map[config.ScoringStrategyType]scorer{
+	config.LeastAllocated: func(args *config.NodeResourcesFitArgs) *resourceAllocationScorer {
```
eahydra, line 51: Here why use NodeResourceFitArgs?  You can define the CPUNormalizationArgs in pkg/scheduler/apis

pkg/scheduler/plugins/cpunormalization/plugin.go
```diff
@@ -0,0 +1,241 @@
...
+
+// Plugin checks if a node has sufficient CPU resources under CPU normalization.
+// It works like the NodeResourcesFit plugin but with CPU requests of CPUSet Pods amplified on Nodes with CPU normalization.
+type Plugin struct {
+	handle framework.Handle
+	resourceAllocationScorer
+}
+
+func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error) {
+	pluginArgs, ok := args.(*config.NodeResourcesFitArgs)
```
eahydra, line 77: Please fix the args type. Now it does not work.

pkg/scheduler/plugins/cpunormalization/plugin.go
```diff
@@ -0,0 +1,241 @@
...
+	return s
+}
+
+func (p *Plugin) Name() string {
+	return Name
+}
+
+func (p *Plugin) EventsToRegister() []framework.ClusterEvent {
+	return []framework.ClusterEvent{
+		{Resource: framework.Pod, ActionType: framework.Delete},
```
eahydra, line 116: IMO the plugin does not need to implement the interface since the native plugin such as `NodeResourcesFit` has already implemented it.

pkg/scheduler/plugins/cpunormalization/resource_allocation.go
```diff
@@ -0,0 +1,147 @@
...
+	)
+
+	return score, nil
+}
+
+// calculateResourceAllocatableRequest returns 2 parameters:
+// - 1st param: quantity of allocatable resource on the node.
+// - 2nd param: aggregated quantity of requested resource on the node.
+// Note: if it's an extended resource, and the pod doesn't request it, (0, 0) is returned.
+func (r *resourceAllocationScorer) calculateResourceAllocatableRequest(nodeInfo *framework.NodeInfo, pod *corev1.Pod, resource corev1.ResourceName, nRatio float64) (int64, int64) {
```
eahydra, line 112: Plugins just handle CPU resources, so we can simplify functionality and remove unused code.

eahydra (changes requested):

pkg/scheduler/plugins/cpunormalization/plugin.go
zqzten: Currently the plugin has the same config as `NodeResourceFit` so I choose to reuse its args here. It's okay to define a specific `CPUNormalizationArgs` type if you think that necessary.

pkg/scheduler/plugins/cpunormalization/resource_allocation.go
zqzten: This only applies to the filter logic I think. We should keep the complete score logic of `NodeResourcesFit` to make sure our score differs from `NodeResourcesFit` only under CPU normalization cases. If we only socre CPU, our score would be very different from which `NodeResourcesFit` gives.

pkg/scheduler/plugins/cpunormalization/plugin.go
eahydra: It is necessary to define a new one. It is impossible to reuse NodeResourceFitArgs directly because of the constraints of the framework.

pkg/scheduler/plugins/cpunormalization/resource_allocation.go
eahydra: Ok. got it. You are right.

zqzten:
/unhold
@eahydra PTAL

pkg/scheduler/plugins/nodecpuamplification/least_allocated_test.go
```diff
@@ -0,0 +1,429 @@
+/*
+Copyright 2022 The Koordinator Authors.
```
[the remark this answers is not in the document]
zqzten: The test files are modified to utilize our own `testSharedLister` rather than the internal `cache.NewSnapshot`, and to use `NodeCPUAmplificationArgs` rather than `NodeResourcesFitArgs`, so I think it's okay to keep our own licence?

PR #1605, opened 2023-09-02
koord-manager: add quota profile controller

shaloulcy:

add quota profile controller. It will update the quota

### V. Checklist

### DIFF
--- a/pkg/quota-controller/profile/profile_controller.go
+++ b/pkg/quota-controller/profile/profile_controller.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"math"
 	"reflect"
+	"strconv"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/quota-controller/profile/profile_controller.go
```diff
@@ -0,0 +1,175 @@
...
+		if !reflect.DeepEqual(quota.Labels, oldQuota.Labels) || !reflect.DeepEqual(quota.Spec, oldQuota.Spec) {
+			err = r.Client.Update(context.TODO(), quota)
+			if err != nil {
+				klog.Errorf("failed update quota for profile %v, error: %v", req.NamespacedName, err)
+				return ctrl.Result{Requeue: true}, err
+			}
+		}
+	}
+
+	return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
hormes, line 158: Why keep retrying?

pkg/quota-controller/profile/profile_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	min := corev1.ResourceList{}
+	for _, key := range resourceKeys {
+		resourceName := corev1.ResourceName(key)
+		quantity, ok := totalResource[resourceName]
+		if ok {
+			min[resourceName] = quantity
+		}
+	}
+
+	quota.Spec.Min = min
```
hormes, line 125: From Sum(node Allocatable) ---> min , we need a fragmentation factor here, and this factor is configurable.

shaloulcy:
#1606

PR #1607, opened 2023-09-03
koord-scheduler: ignore non-existing resources in NUMA Node

eahydra:

At present, koordlet only reports cpu/memory in NodeResourceTopology. If a Pod requests resources from other resources, the Pod will not be able to be scheduled. So we should skip these non-existing resources.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/resource_manager.go
+++ b/pkg/scheduler/plugins/nodenumaresource/resource_manager.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/nodenumaresource/resource_manager.go
```diff
@@ -207,10 +212,17 @@ func (c *resourceManager) allocateResourcesByHint(node *corev1.Node, pod *corev1
 			break
 		}
 	}
-	requests = quotav1.RemoveZeros(requests)
-	if len(requests) > 0 {
-		resourceNames := quotav1.ResourceNames(requests)
-		return nil, fmt.Errorf("insufficient resources: %v", resourceNames)
+
+	var insufficientResources []string
```
hormes, line 216: Add a unittests to verify this？

eahydra: done

eahydra (commented):
/approve

PR #1611, opened 2023-09-04
koord-scheduler: optimize NUMA Topology Manager interface

eahydra:

Convert the Allocate method to Admit, which is only used to calculate the optimal topology Hint, and store the Best Hint of each node in the Store. In the subsequent Score, Reserve and other stages, the Hint of the node can be queried through the Store. Once a Hint is determined, it cannot be changed, otherwise it will conflict with future functions that support Reservation/Preemption on NUMA-Aware.

### V. Checklist


jasonliu747 (approved):
/lgtm

hormes (approved):

PR #1640, opened 2023-09-11
koordlet: report NRT zone resources mergeable

saintube:

1. statesinformer: revise NRT reporting, mergeable update ZoneList so as to cooperate with NRT reporting of the other components.

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
+++ b/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
@@ -57,8 +57,6 @@ import (
 
 const (
 	nodeTopoInformerName PluginName = "nodeTopoInformer"
-
-	NodeZoneType = "Node"
 )
 
 type nodeTopologyStatus struct {
@@ -100,8 +98,8 @@ func (n *nodeTopologyStatus) updateNRT(nrt *v1alpha1.NodeResourceTopology) {
 
 	nrt.TopologyPolicies = []string{string(n.TopologyPolicy)}
 
-	// TBD: merge with the existing
-	nrt.Zones = n.Zones
+	// trim useless zone name and merge with the existing zone list
+	nrt.Zones = util.MergeZoneList(nrt.Zones, n.Zones)
 }
 
 type nodeTopoInformer struct {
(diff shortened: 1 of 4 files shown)


pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
```diff
@@ -100,8 +98,8 @@ func (n *nodeTopologyStatus) updateNRT(nrt *v1alpha1.NodeResourceTopology) {
 
 	nrt.TopologyPolicies = []string{string(n.TopologyPolicy)}
 
-	// TBD: merge with the existing
-	nrt.Zones = n.Zones
+	// trim useless zone name and merge with the existing zone list
+	nrt.Zones = util.MergeZoneList(util.TrimZoneList(nrt.Zones, "fake-name"), n.Zones)
```
zwzhang0107, line 102: is the `fake-name` still necessary?

pkg/util/resource.go
```diff
@@ -118,3 +126,114 @@ func IsResourceDiff(old, new corev1.ResourceList, resourceName corev1.ResourceNa
...
+		}
+		for resourceName, quantityB := range resourceListB {
+			resourceListA[resourceName] = quantityB
+		}
+	}
+
+	return ZoneResourceListToZoneList(zoneResourcesA)
+}
+
+func IsZoneListResourceEqual(a, b v1alpha1.ZoneList) bool {
```
zwzhang0107, line 207: not used yet

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
saintube: Currently, the NRT is created firstly with no actual zone resources and updated later. It still keeps the `fake-name` in creation to avoid breaking up other components like the scheduler's assumption.

pkg/util/resource.go
saintube: It is prepared for the controller's usage.

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
saintube: Removed. There was no at least one zone assumption so the `fake-name` makes nonsense.

hormes (approved):

PR #1660, opened 2023-09-15
koord-manager: fix  LessThanOrEqual in elastic quota

shaloulcy:

fix LessThanOrEqual for elastic quota webhook

### V. Checklist

### DIFF
--- a/pkg/util/resource_test.go
+++ b/pkg/util/resource_test.go
@@ -1067,3 +1067,82 @@ func TestIsZoneListResourceEqual(t *testing.T) {
 		})
 	}
 }
+
+func TestLessThanOrEqualEnhanced(t *testing.T) {
+	tests := []struct {
+		name   string
+		a      corev1.ResourceList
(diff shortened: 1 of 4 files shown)


pkg/util/resource_test.go
```diff
@@ -371,3 +371,70 @@ func TestQuantityPtr(t *testing.T) {
...
+				corev1.ResourceMemory: *resource.NewQuantity(1024, resource.BinarySI),
+			},
+			b: corev1.ResourceList{
+				corev1.ResourceCPU:    *resource.NewQuantity(10, resource.DecimalSI),
+				corev1.ResourceMemory: *resource.NewQuantity(1024, resource.BinarySI),
+			},
+			expect: true,
+		},
+		{
+			name: "is equal",
```
eahydra, line 1091: same case name with case1.
And missed case a > b.

shaloulcy: done

PR #1669, opened 2023-09-19
koord-scheduler: optimize NodeNUMAResource scoring

eahydra:

There are many unreasonable aspects in the existing implementation:
1. Nodes with NUMA Topology Policy turned on will consider multiple resources when scoring, but if they are ordinary nodes, only the CPU dimension will be considered.
2. The calculation allocation logic when there is NUMATopologyHint is inconsistent with the normal situation, and there should not be this difference.
3. Although the logic for calculating the total allocable amount and the allocated amount will be different according to the NUMA Topology Policy, the difference is only in scope.

This PR eliminates these differences and the overall implementation is clearer.

### V. Checklist


jasonliu747 (commented):
/lgtm

hormes (approved):

PR #1677, opened 2023-09-22
koord-scheduler: NodeNUMAResource supports filtering/scoring with node-level amplification ratios

eahydra:

Based on #1602, implement amplified CPU filtering/scoring in plugin NodeNUMAResource.

After a detailed discussion with @zqzten, the author of #1602, the core goal of #1602 is to consider that if there is a LSE/LSR Pod on a node with a CPU amplification ratio, the corresponding CPU requests must be amplified to ensure the runtime QoS of the LSE/LSR Pod. Of course, it can also work to implement a new plugin separately, but it is more reasonable to implement it in the NodeNUMAResource plugin, because the NodeNUMAResource plugin supports allocating CPU cores to LSE/LSR Pods. Therefore, in the Filter stage, the requested amount and the allocated amount are amplified (prerequisite is LSE/LSR Pod) to ensure that the capacity is legal; the situation is similar in the scoring stage.

### V. Checklist


saintube (commented):
/lgtm

hormes (approved):

pkg/scheduler/plugins/nodenumaresource/plugin.go
```diff
@@ -317,6 +322,46 @@ func (p *Plugin) Filter(ctx context.Context, cycleState *framework.CycleState, p
...
 
+func (p *Plugin) filterAmplifiedCPUs(state *preFilterState, nodeInfo *framework.NodeInfo) *framework.Status {
+	quantity := state.requests[corev1.ResourceCPU]
+	podRequestMilliCPU := quantity.MilliValue()
+	if podRequestMilliCPU == 0 {
+		return nil
+	}
+
+	node := nodeInfo.Node()
+	ratios, err := extension.GetNodeResourceAmplificationRatios(node.Annotations)
```
zqzten, line 333: Nit: We can call the convenience method `cpuAmplificationRatio, err := extension.GetNodeResourceAmplificationRatio(node.Annotations, corev1.ResourceCPU)` here

pkg/scheduler/plugins/nodenumaresource/scoring.go
```diff
@@ -89,6 +92,37 @@ func (p *Plugin) Score(ctx context.Context, cycleState *framework.CycleState, po
 	return p.scorer.score(requested, allocatable, framework.NewResource(resourceOptions.requests))
 }
 
+func (p *Plugin) scoreWithAmplifiedCPUs(cycleState *framework.CycleState, state *preFilterState, pod *corev1.Pod, nodeInfo *framework.NodeInfo, topologyOptions TopologyOptions) (int64, *framework.Status) {
+	quantity := state.requests[corev1.ResourceCPU]
+	if quantity.IsZero() {
+		return 0, nil
```
zqzten, line 98: I think we should call the default scorer here rather than return 0 to keep the same behavior as the `NodeResourceFit` plugin.

PR #1691, opened 2023-09-29
koord-manager: support setup transformers

eahydra:
### V. Checklist


saintube:
Before we use the transformer to mutate the node cpu/mem allocatable in the informer cache, we need to implement the [`raw-allocatable`](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20230811-node-resource-amplification.md#api) API to let the node resource plugins know the real allocatable.

PR #1694, opened 2023-10-01
scheduler: use asynclog to reduce klog performance overhead

eahydra:

By default, klog outputs logs synchronously to stderr, which will affect performance when there are too many logs.
Therefore, we can reduce the delay caused by syscall and IO by outputting logs asynchronously, thus improving performance.

When enable `--async-log`,  also need to enable `-logtostderr` and disable `-alsologtostderr`. And also can set `--async-log-queue-length` to fine-tune performance.

### V. Checklist

### DIFF
--- a/pkg/util/asynclog/async_log.go
+++ b/pkg/util/asynclog/async_log.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


pkg/util/asynclog/async_log.go
```diff
@@ -0,0 +1,116 @@
...
+	"os"
+	"sync"
+	"sync/atomic"
+
+	"github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+)
+
+var (
+	enableAsync = pflag.Bool("async-log", false, "enable to async write log and must enable logtostderr")
```
hormes, line 31: When these two flags will be added to the program's binary file, and when the application should consider enabling asynchronous logging, a comment is needed to explain the usage precautions.

eahydra: update.

PR #1698, opened 2023-10-08
ci: optimize make generate and update to go 1.19

eahydra:
### V. Checklist


pkg/webhook/util/writer/atomic/atomic_writer.go
```diff
@@ -85,15 +85,18 @@ const (
 // The Write algorithm is:
 //
 //  1. The payload is validated; if the payload is invalid, the function returns
-//  2. The current timestamped directory is detected by reading the data directory
+//     2. The current timestamped directory is detected by reading the data directory
```
jasonliu747, line 88: 1和2之间可能需要换行，不然go fmt就会把第二行缩进

PR #1702, opened 2023-10-09
scheduler: modify coscheduling Less function with considering childScheduleCycle

### LINKED ISSUE #1701: [BUG] Coscheduling will cause an infinite loop in large-scale job scenarios.
**What happened**:

1. Create a big job, e.g. the Job has 255 Pods
2. The job enables the coscheduling.
3. And the cluster has insufficient resources to satisfied the Job.
4. Add enough resources into the cluster.
5. But the job still failed to schedule.

All pods of the job scheduled successfully.

Well...hard to reproduce.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

buptcozy:

fix #1701

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1703, opened 2023-10-09
koord-manager: enhance batch resource configuration and updating

saintube:

koord-manager: enhance batch resource configuration and updating:

1. Refactor the NRT updating in the BatchResource plugin. Add the PreUpdate stage for checking and updating additional objects alongside the framework's Node updating.
2. Support the `cpuCalculatePolicy` in the ColocationStrategy. Enable calculating `batch-cpu` according to high-priority pods' maximal of requests and usages. The default policy is `usage`, which allows the low-priority pods to reclaim the resources requested but unused by the high-priority pods. The new policy `maxUsageAndRequest` is helpful when the cpu resources are not strongly expected to be overcommitted between different priority bands, where neither used resources nor requested resources can be allocatable to the low-priority pods.
3. Support colocation strategy based on the node metadata. Add an annotation protocol for node-level configuration. Add two label protocols for label-preferred use cases.

### V. Checklist


pkg/slo-controller/noderesource/framework/extender_plugin.go
```diff
@@ -24,8 +24,17 @@ import (
...
+//     node resource reconcile loop:
+//  1. Calculate / Reset (determined by whether the config disables node resources, e.g. colocation disabled)
+//  2. NodePreUpdate
+//  3. NodePrepare: (it should be idempotent since it can be called multiple times in one loop)
+//  4. NodeSync
+//  5. NodeMetaSync
 var (
 	globalSetupExtender             = NewRegistry("Setup")
+	globalNodePreUpdateExtender     = NewRegistry("NodePreUpdate")
 	globalNodePrepareExtender       = NewRegistry("NodePrepare")
```
zwzhang0107, line 38: better to rename NodePreparePlugin.Execute to NodePreparePlugin.Prepare

pkg/slo-controller/noderesource/framework/extender_plugin.go
```diff
@@ -24,8 +24,17 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/slo-controller/metrics"
 )
 
+// node resource initialization:
+//  1. Setup
+//     node resource reconcile loop:
+//  1. Calculate / Reset (determined by whether the config disables node resources, e.g. colocation disabled)
+//  2. NodePreUpdate
+//  3. NodePrepare: (it should be idempotent since it can be called multiple times in one loop)
```
zwzhang0107, line 32: NodeStatusPrepare
or just NodePrepare if some label is included.

pkg/slo-controller/noderesource/framework/extender_plugin.go
```diff
@@ -24,8 +24,17 @@ import (
...
+//  4. NodeSync
+//  5. NodeMetaSync
```
zwzhang0107, line 34: only for need sync checking?
NodeStatusCheck
NodeMetaCheck

pkg/slo-controller/noderesource/framework/extender_plugin.go
```diff
@@ -24,8 +24,17 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/slo-controller/metrics"
 )
 
+// node resource initialization:
```
zwzhang0107, line 27: need more explanation about the running order of these plugins

e.g. there are several stages:
Setup, Reset if node not enabled

Setup
Calculate(run plugins for generate internal nr)
Update(run PreUpdatePlugin for nr injection or other crd update; run checkNeedSync plugin, run Prepare plugins, do update node obj)

apis/extension/node_colocation.go
```diff
@@ -0,0 +1,24 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+// NOTE: functions in this file can be overwritten for extension
+
```
eahydra, line 20: Maybe we should add some comments to describe what the labels means.

saintube: done

pkg/slo-controller/noderesource/framework/extender_plugin.go
saintube: done

hormes:
> Enable calculating batch-cpu according to high-priority pods' requests instead of usages.

Explain when you need to use request instead of usage?

apis/extension/node_colocation.go
```diff
@@ -0,0 +1,26 @@
...
+package extension
+
+const (
+	// LabelCPUReclaimRatio denotes the CPU reclaim ratio of a node. The value is a float number.
+	// It takes precedence to the CPUReclaimThresholdPercent in the slo-controller-config.
+	LabelCPUReclaimRatio = NodeDomainPrefix + "/cpu-reclaim-ratio"
+	// LabelMemoryReclaimRatio denotes the memory reclaim ratio of a node. The value is a float number.
+	// It takes precedence to the MemoryReclaimThresholdPercent in the slo-controller-config.
+	LabelMemoryReclaimRatio = NodeDomainPrefix + "/memory-reclaim-ratio"
```
hormes, line 25: Should these two fields be placed in annotation?

saintube:
> > Enable calculating batch-cpu according to high-priority pods' requests instead of usages.
> 
> Explain when you need to use request instead of usage?

Add code comments for different calculating policies.

apis/extension/node_colocation.go
saintube: These label protocols are for some label-preferred scenarios. A standard annotation protocol is newly added.

PR #1709, opened 2023-10-12
scheduler: make ReservationInfo.AddAssignedPod idempotent

eahydra:

When the scheduler starts, some Pods may be updated multiple times in a short period of time, causing the ReservationInfo to add the same Pod multiple times, and ultimately cannot be reused because the reservation's allocated is greater than or equal to the reservation's allocable.  

The root cause is that the forceSyncEventHandler of the frameworkext helper module requires that all EventHandlers must be idempotent.

So the PR fix the issue. Make the ReservationInfo method `AddAssignedPod` idempotent.

### V. Checklist


buptcozy (approved):
/approve

PR #1713, opened 2023-10-18
scheduler: reservation senses node affinity

eahydra:

Reservation plugin senses NodeAffinity and reduces the amount of calculation. For example, DaemonSet Pod is only scheduled to one node, so we only need to process the Reservation of one node.

### V. Checklist


buptcozy (approved):
/lgtm

PR #1720, opened 2023-10-24
apis: add core sched apis

saintube:

koordlet: add the apis for the Linux [Core Scheduling](https://docs.kernel.org/admin-guide/hw-vuln/core-scheduling.html) feature.

### V. Checklist


apis/slo/v1alpha1/pod.go
```diff
@@ -64,3 +64,19 @@ func GetPodMemoryQoSConfig(pod *corev1.Pod) (*PodMemoryQOSConfig, error) {
...
+	AnnotationPodEnableCoreSched = apiext.DomainPrefix + "enable-core-sched"
+
+	// AnnotationCoreSchedGroupID is the annotation key of the group ID of the Core Sched.
+	// Core Sched: https://docs.kernel.org/admin-guide/hw-vuln/core-scheduling.html
+	// Pods with the different core sched group IDs will not be running at the same SMT core at the same time, which
+	// means they will take different core sched cookies. And a pod with cookie is allowed to run on the same core with
+	// a pod who does not set any cookie (cookie=0).
+	// NOTE: If a pod sets the core sched enabled but misses the core sched ID, the pod will be assigned an individual
+	// and stateless internal ID according to the pod's UID.
+	AnnotationCoreSchedGroupID = apiext.NodeDomainPrefix + "/core-sched-group-id"
```
eahydra, line 81: Is there any format or specification requirement for the value corresponding to this annotation?

saintube: added

apis/slo/v1alpha1/pod.go
```diff
@@ -64,3 +65,49 @@ func GetPodMemoryQoSConfig(pod *corev1.Pod) (*PodMemoryQOSConfig, error) {
...
+	// AnnotationCoreSchedGroupID is the annotation key of the group ID of the Core Sched.
+	// Value is expected to be a valid UUID.
+	// Core Sched: https://docs.kernel.org/admin-guide/hw-vuln/core-scheduling.html
+	// NOTE: Pods with the different core sched group IDs will not be running at the same SMT core at the same time,
+	//       which means they will take different core sched cookies. If a pod sets the core sched disabled, it will
+	//       take the default core sched cookie (0) and will also be force-idled to run on the same SMT core
+	//       concurrently with the core-sched-enabled pods.
+	//       If a pod sets the core sched enabled but misses the core sched ID, the pod will be assigned an individual
+	//       and stateless internal ID according to the pod's UID.
+	AnnotationCoreSchedGroupID = apiext.NodeDomainPrefix + "/core-sched-group-id"
```
hormes, line 85: Why not consider merging the above two fields? Setting core-sched-group-id means enabling core-sched.

apis/slo/v1alpha1/nodeslo_types.go
```diff
@@ -25,10 +25,21 @@ import (
 
 // CPUQOS enables cpu qos features.
 type CPUQOS struct {
+	// applied policy for the CPU QoS, default = "groupIdentity"
+	Policy *CPUQOSPolicy `json:"policy,omitempty"`
```
hormes, line 29: Will these two strategies exist at the same time?

saintube: No, they can not co-exist.

apis/slo/v1alpha1/pod.go
saintube: The core scheduling can be enabled at node-level via NodeSLO where all pods are set cookies. So there may be a case that some pods disable the core sched (use the default cookie 0) while other pods enabled due to the NodeSLO spec.

eahydra, line 85: The annotation's prefix is NodeDomainPrefix, but only Pods can use the key, so is this a typo? Also I would recommend merging these two annotations.

saintube: merged into one annotation

jasonliu747 (approved):
/lgtm

apis/slo/v1alpha1/nodeslo_types.go
```diff
@@ -26,9 +26,27 @@ import (
 // CPUQOS enables cpu qos features.
 type CPUQOS struct {
 	// group identity value for pods, default = 0
+	// NOTE: It takes effect if CPUPolicy = "groupIdentity".
 	GroupIdentity *int64 `json:"groupIdentity,omitempty" validate:"omitempty,min=-1,max=2"`
+	// cpu.idle value for pods, default = 0.
+	// `1` means using SCHED_IDLE.
+	// CGroup Idle (introduced since mainline Linux 5.15): https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r
+	// NOTE: It takes effect if CPUPolicy = "coreSched".
+	Idle *int64 `json:"idle,omitempty" validate:"omitempty,min=0,max=1"`
```
zwzhang0107, line 35: SchedIdle

saintube: done

hormes (approved):

PR #1721, opened 2023-10-24
descheduler: check nodemetrics cr is expired or not when descheduling

lucming:
case:
The nodemetric cr has not been updated for a long time, e.g. the koordlet was deleted, but the load information written last time is particularly high, causing pods on this node being evicted all the time, even though the actual load was low. 

check nodemetrics cr is expired or not when descheduling
### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1726, opened 2023-10-30
scheduler: numa-aware scheduling supports selecting numa node by score

### LINKED ISSUE #1680: [proposal] scheduler numa-aware scheduling should support select hint by policy
**What is your proposal**:

scheduler numa-aware scheduling should support select hint by policy

**Why is this needed**:

Currently the NodeNUMAResource plugin uses TopologyManager to generate the best hint, but the hint is the narrowest, e.g. if the node's topology strategy is `SingleNUMANode`, the best hint is always the one with the smallest NUMA Node ID, it works like a bin-packing strategy. However, some scenarios require selecting the least allocated NUMA node to ensure runtime QoS.

eahydra:

Currently the NodeNUMAResource plugin uses TopologyManager to generate the best hint, but the hint is the narrowest, e.g. if the node's topology strategy is SingleNUMANode, the best hint is always the one with the smallest NUMA Node ID, it works like a bin-packing strategy. However, some scenarios require selecting the least allocated NUMA node to ensure runtime QoS.

Add a new scoring strategy to NodeNUMAResourceArgs to support NUMA-Node level scoring, so that we can select the numa node with the highest score, and then in score-phase, we can select the node with the highest score at the node level.

User can configure the NodeNUMAResourceArgs as the following:
```yaml
apiVersion: kubescheduler.config.k8s.io/v1beta2
kind: KubeSchedulerConfiguration
profiles:
  - pluginConfig:
      - name: NodeNUMAResource
        args:
          apiVersion: kubescheduler.config.k8s.io/v1beta2
          kind: NodeNUMAResourceArgs
          scoringStrategy:  # Here configure Node level scoring strategy
            type: LeastAllocated
            resources:
              - name: cpu
                weight: 1
              - name: memory
                weight: 1
              - name: "kubernetes.io/batch-cpu"
                weight: 1
              - name: "kubernetes.io/batch-memory"
                weight: 1
          numaScoringStrategy: # Here configure NUMA-Node level scoring strategy
            type: LeastAllocated
            resources:
              - name: cpu
                weight: 1
              - name: memory
                weight: 1
              - name: "kubernetes.io/batch-cpu"
                weight: 1
              - name: "kubernetes.io/batch-memory"
                weight: 1
```

We can  create a deployment affinities the special node with 10 replicas, 4C8G per replica, and use debug api `/apis/v1/plugins/NodeNUMAResource/nodes/:nodeName` to view effect. 
Use `LeastAllocated` strategy, we can get the result as following(just one node which has two numa-nodes):
```json
{
    ....
    "allocatedNUMANodeResources": [
       {"node":0,"resources":{"cpu":"25","memory":"53Gi"}},
       {"node":1,"resources":{"cpu":"28","memory":"58Gi"}}
    ]
}
```

Use `MostAllocated` strategy, we can get the result as following(just one node which has two numa-nodes):
```json
{
   ....
   "allocatedNUMANodeResources":[
        {"node":0,"resources":{"cpu":"53500m","memory":"113233253Ki"}}
   ]
}
```

fix #1680

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -103,8 +103,14 @@ type ScoringStrategy struct {
 type NodeNUMAResourceArgs struct {
 	metav1.TypeMeta
 
+	// DefaultCPUBindPolicy represents the default CPU bind policy.
+	// If it is empty and the Pod does not declare a binding policy,
+	// the core will not be bound to the LSE/LSR type Pod.
 	DefaultCPUBindPolicy CPUBindPolicy
-	ScoringStrategy      *ScoringStrategy
+	// ScoringStrategy is used to configure the scoring strategy of the Node-level.
+	ScoringStrategy *ScoringStrategy
+	// NUMAScoringStrategy is used to configure the scoring strategy of the NUMANode-level
+	NUMAScoringStrategy *ScoringStrategy
 }
 
 // CPUBindPolicy defines the CPU binding policy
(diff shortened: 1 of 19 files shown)


jasonliu747 (commented):
/lgtm

hormes (approved):

PR #1732, opened 2023-11-03
scheduler: revise numa-aware hints generator

eahydra:

Modify NUMA-Aware hints generator implementation.
Some specific scenarios, such as pods requesting memory and huge-pages at the same time, or requiring CPU binding, may conflict with generating hints from resources only, resulting in the fact that although hints are generated, they are actually unusable. So we need to consider these conditions when producing hints.

### V. Checklist


jasonliu747 (approved):
/lgtm

Issue #1727, opened 2023-10-30
[proposal] QoS supports of Hadoop YARN task in koordlet/slo-controller


zwzhang0107:
**What is your proposal**:
Koordinator is almost ready to support running Hadoop YARN tasks as Batch+BE QoS in K8s. Here are the [detailed design](https://github.com/koordinator-sh/koordinator/discussions/1297).

Since YARN tasks are running under a specified cgroup path(kubepods/besteffort/hadoop-yarn), koordinator should take the tasks into account during Batch resource calculation and QoS Management.

**Why is this needed**:
Resource usage of YARN tasks will be count as system usage if koordlet dose not collect it independently, which leads to the miscalculation of `Batch Allocatable` and the omission of QoS Management.

**Is there a suggested solution, if so, please add it**:
Define YARN tasks as out-of-band Host Application, see the [api design](https://github.com/koordinator-sh/koordinator/pull/1508/files#diff-00b2f1de4aa8c404e9eb52c7390c2bb47f04454be559c57577d61a4bb19b3cdd) for more details.
```yaml
  host-application-config: |
    {
      "applications": [
        {
          "name": "yarn-task",
          "priority": "koord-batch",
          "qos": "BE",
          "cgroupPath": {
            "base": "CgroupBaseTypeKubeBesteffort",
            "parentDir": "./",
            "relativePath": "hadoop-yarn/",
          }
        }
      ]
    }
```
- [x] add plugin in koordlet metricAdvisor for collecting the resource usage of host-applications by cgroup path.(@zwzhang0107)
- [x] koordlet statesInformer reports the hosts application resource usage  in [NodeMetric](https://github.com/koordinator-sh/koordinator/pull/1725). 
- [x] slo-controller treats resource usage of YARN tasks as batch prioirty during caculcating batch allocatable
- [x] slo-controller exclude yarrn allocated from node annotation when calculating k8s batch allocatable
- [x] CPUSuppress strategy should exclude YARN tasks during [system usage calculation](https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/qosmanager/plugins/cpusuppress/cpu_suppress.go#L166).
- [ ] add plugin in statesInformer for collecting YARN task info from yarn-copilot, which will be used for startegies like memory eviction.
[...]

PR #1739, opened 2023-11-13
slo-controller: supppor third party allocator for batch resource calculation

zwzhang0107:

- slo-controller treats resource usage of YARN tasks as batch prioirty during caculcating batch allocatable
- slo-controller exclude YARN allocated from node annotation when calculating k8s batch allocatable
- slo-controller add origin allocatable(extended resource) on node annotation for calculating YARN total

#1727

add node annotation with
```
node.koordinator.sh/thirdPartyAllocations: "{"allocations":[
  {
    "name":"hadoop-yarn",
    "priority":"koord-batch",
    "resources":{"kubernetes.io/batch-cpu":"10000","kubernetes.io/batch-memory":"10Gi"}
  }
]}"
```

batch-cpu/memory in node.Status will be excluded compared with `node.koordinator.sh/originExtendedAllocatable`

### V. Checklist


pkg/slo-controller/noderesource/plugins/batchresource/plugin.go
```diff
@@ -114,6 +114,38 @@ func (p *Plugin) Prepare(_ *configuration.ColocationStrategy, node *corev1.Node,
...
+		return err
+	}
+
+	if batchMilliCPU < 0 || batchMemory < 0 {
+		// batch resources are reset, no need to recalculate node status
+		return nil
+	}
+
+	// subtract third party allocated
+	thirdPartyBatchAllocated, err := slov1alpha1.GetThirdPartyAllocatedResByPriority(node.Annotations, extension.PriorityBatch)
```
saintube, line 134: do we need to watch the change in third-party allocation?

zwzhang0107: maybe later, now we just use the default ticker for updating.

FillZpp (approved):

PR #1753, opened 2023-11-22
scheduler: fix the resources required by the pod in the quota are sufficient but the pod cannot be scheduled

qinfustu:
fix the resources required by the pod in the quota are sufficient but the pod cannot be scheduled.

The resources required by the pod in the quota are sufficient but the pod cannot be scheduled.
e.q.
quota info
```yaml
apiVersion: scheduling.sigs.k8s.io/v1alpha1
kind: ElasticQuota
metadata:
  name: quota-sub
  namespace: default
  annotations:
    quota.scheduling.koordinator.sh/runtime: '{"cpu":"40","memory":"40Gi", "nvidia.com/gpu": "6"}'
  labels:
    quota.scheduling.koordinator.sh/parent: "quota-parent"
    quota.scheduling.koordinator.sh/is-parent: "false"
spec:
  max:
    cpu: 40
    memory: 40Gi
    nvidia.com/gpu: 8
  min:
    cpu: 40
    memory: 40Gi
    nvidia.com/gpu: 8
status:
  used:
    cpu: 10
    memory: 20Gi
    nvidia.com/gpu: 8
```
pod info
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: pod-example
  namespace: default
  labels:
    quota.scheduling.koordinator.sh/name: "quota-sub"
spec:
  containers:
  - command:
    - sleep
    - 365d
    image: busybox
    imagePullPolicy: IfNotPresent
    name: curlimage
    resources:
      limits:
        cpu: 2
        memory: 4Gi
      requests:
        cpu: 2
        memory: 4Gi
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
  restartPolicy: Always
```
At this time, the pod will not be scheduled due to insufficient quota resources(nvidia.com/gpu: runtime < used). In fact, the cpu and memory required by the pod are sufficient.
There are scenarios where runtime < used appears：
- Enable the borrowing / returning of resources between different quota. But disable "used > runtime revoke" logic.
- Operators adjust quota resource distribution.

App version: 1.20
Kubernetes version (use kubectl version): 1.19

### V. Checklist


pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -229,7 +229,7 @@ func (g *Plugin) PreFilter(ctx context.Context, cycleState *framework.CycleState
 	podRequest, _ := core.PodRequestsAndLimits(pod)
 	used := quotav1.Add(podRequest, state.used)
 
-	if isLessEqual, exceedDimensions := quotav1.LessThanOrEqual(used, state.runtime); !isLessEqual {
+	if isLessEqual, exceedDimensions := lessThanOrEqual(used, state.runtime, podRequest); !isLessEqual {
```
eahydra, line 232: I think there is no need to change this place. Just make sure that the resource dimension used is equal to pod request.  
That is, after line 230, change it to the following:
```go
used := quotav1.Add(podRequest, state.used)
used = quotav1.Mask(used, quotav1.ResourceNames(podRequest))
```
A more thorough modification is to find the intersection of used and request in the snapshotPostFilterState method.

qinfustu:
I am currently on paternity leave. Please help me follow up on PR merger, @eahydra. Thx~
Besides，I will commit the UT cases over the leave.

PR #1759, opened 2023-11-27
koordlet: kill container after calling eviction api success

### LINKED ISSUE #1758: [proposal] call eviction API before kill container
**What is your proposal**:

call eviction API before kill container

**Why is this needed**:

memoryEvict will kill container then call eviction api, this may disobey pod disruption budgets (PDB)

https://github.com/koordinator-sh/koordinator/blob/661dd71df888a7cb8e2b0891408b34225dc2ee85/pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go#L163

if eviction response OK, then we kill pod/container, otherwise skip. refer https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works

There is a bonus with call eviction API first, we can disable eviction by update rolebinding.

j4ckstraw:

fix #1758 

### V. Checklist

### DIFF
--- a/pkg/koordlet/qosmanager/framework/context.go
+++ b/pkg/koordlet/qosmanager/framework/context.go
@@ -62,7 +62,7 @@ func (r *Evictor) Start(stopCh <-chan struct{}) error {
 
 func (r *Evictor) EvictPodsIfNotEvicted(evictPods []*corev1.Pod, node *corev1.Node, reason string, message string) {
 	for _, evictPod := range evictPods {
-		r.evictPodIfNotEvicted(evictPod, node, reason, message)
+		r.EvictPodIfNotEvicted(evictPod, node, reason, message)
 	}
 }
 
@@ -74,16 +74,18 @@ func (r *Evictor) IsPodEvicted(pod *corev1.Pod) bool {
 	return evicted
 }
 
-func (r *Evictor) evictPodIfNotEvicted(evictPod *corev1.Pod, node *corev1.Node, reason string, message string) {
+func (r *Evictor) EvictPodIfNotEvicted(evictPod *corev1.Pod, node *corev1.Node, reason string, message string) bool {
 	_, evicted := r.podsEvicted.Get(string(evictPod.UID))
 	if evicted {
 		klog.V(5).Infof("Pod has been evicted! podID: %v, evict reason: %s", evictPod.UID, reason)
-		return
+		return true
 	}
 	success := r.evictPod(evictPod, reason, message)
 	if success {
 		_ = r.podsEvicted.SetDefault(string(evictPod.UID), evictPod.UID)
 	}
+
+	return success
 }
 
 func (r *Evictor) evictPod(evictPod *corev1.Pod, reason string, message string) bool {
--- a/pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go
+++ b/pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go
@@ -32,6 +32,7 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/qosmanager/helpers"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/resourceexecutor"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
+	"github.com/koordinator-sh/koordinator/pkg/util"
 )
 
 const (
@@ -50,6 +51,7 @@ type memoryEvictor struct {
 	metricCache           metriccache.MetricCache
 	evictor               *framework.Evictor
 	lastEvictTime         time.Time
+	onlyEvictByAPI        bool
 }
 
 type podInfo struct {
@@ -64,6 +66,7 @@ func New(opt *framework.Options) framework.QOSStrategy {
 		metricCollectInterval: opt.MetricAdvisorConfig.CollectResUsedInterval,
 		statesInformer:        opt.StatesInformer,
 		metricCache:           opt.MetricCache,
+		onlyEvictByAPI:        opt.Config.OnlyEvictByAPI,
 	}
 }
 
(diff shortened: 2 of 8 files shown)


pkg/koordlet/qosmanager/framework/context.go
```diff
@@ -62,20 +62,22 @@ func (r *Evictor) Start(stopCh <-chan struct{}) error {
 
 func (r *Evictor) EvictPodsIfNotEvicted(evictPods []*corev1.Pod, node *corev1.Node, reason string, message string) {
 	for _, evictPod := range evictPods {
-		r.evictPodIfNotEvicted(evictPod, node, reason, message)
+		r.EvictPodIfNotEvicted(evictPod, node, reason, message)
 	}
 }
 
-func (r *Evictor) evictPodIfNotEvicted(evictPod *corev1.Pod, node *corev1.Node, reason string, message string) {
+func (r *Evictor) EvictPodIfNotEvicted(evictPod *corev1.Pod, node *corev1.Node, reason string, message string) bool {
```
saintube, line 69: please add some UTs for an exposed method

j4ckstraw: OK

j4ckstraw: I failed to add a test which evict failed because of serviceaccount has no permission on pod/eviction api.
Help

saintube: A UT is enough, not an E2E. So I suppose the test should not require a role with `pod/eviction`. The kubelet eviction service can be mocked in that UT.

j4ckstraw: OK

zwzhang0107:
memory evict is just like oom killer, which dose not restricted by pdb.
so *by default*, memory evict should always kill containers even the eviction is failed for rpc timeout for example.

j4ckstraw:
> memory evict is just like oom killer, which dose not restricted by pdb. _by default_, memory evict should always kill containers even the eviction is failed for rpc timeout for example. so it is better to add flag to control the default logic.

How about add a koordlet feature-gate named `DoNotEvictPodIfCallEvictionAPIFailed`,  maybe it's too long, WDYT?

pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go
```diff
@@ -164,24 +166,28 @@ func (m *memoryEvictor) killAndEvictBEPods(node *corev1.Node, podMetrics map[str
...
-
-	var killedPods []*corev1.Pod
+	hasKillPods := false
 	for _, bePod := range bePodInfos {
 		if memoryReleased >= memoryNeedRelease {
 			break
 		}
 
+		ok := m.evictor.EvictPodIfNotEvicted(bePod.pod, node, resourceexecutor.EvictPodByNodeMemoryUsage, message)
+		if m.onlyEvictByAPI && !ok {
```
zwzhang0107, line 176: please add some error logs if the eviction fails, so does the cpu evict

pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go
```diff
@@ -164,24 +166,28 @@ func (m *memoryEvictor) killAndEvictBEPods(node *corev1.Node, podMetrics map[str
...
 		killMsg := fmt.Sprintf("%v, kill pod: %v", message, bePod.pod.Name)
 		helpers.KillContainers(bePod.pod, killMsg)
-		killedPods = append(killedPods, bePod.pod)
+		hasKillPods = true
 		if bePod.memUsed != 0 {
 			memoryReleased += int64(bePod.memUsed)
 		}
 	}
+	if hasKillPods {
+		m.lastEvictTime = time.Now()
```
zwzhang0107, line 188: evict by api should also be considered for calculating cooling time

pkg/koordlet/qosmanager/plugins/memoryevict/memory_evict.go
j4ckstraw: EvictPodIfNotEvicted->evictPod already has a log.
Do you need it here?

j4ckstraw:
> > > memory evict is just like oom killer, which dose not restricted by pdb. _by default_, memory evict should always kill containers even the eviction is failed for rpc timeout for example. so it is better to add flag to control the default logic.
> > 
> > 
> > How about add a koordlet feature-gate named `DoNotEvictPodIfCallEvictionAPIFailed`, maybe it's too long, WDYT?
> 
> `feature-gate` means new features, which will be iterated from alpha(default=false), beta(default=true), and ga. so an argument named `--only-evict-by-api`(default=false) is better, and only call eviction api when enabled.
> 
> also, please add a TODO comment for supporting fine-grained eviction arguments just like [kubelet](https://kubernetes.io/zh-cn/docs/concepts/scheduling-eviction/node-pressure-eviction/#eviction-thresholds)

Maybe you misunderstood me, I want to kill container on when calling eviction api returns ok, not eviction only by API.

zwzhang0107:
> want

Actually, after the eviction api returns ok, there is no need to kill container if you want to solve the grace termination or PDB problem.
It seems that defines a self-defined evictor args(--evictor="default/only-eviction/...", which means --only-evict-by-api) can solve your concerns?

j4ckstraw:
> > want
> 
> Actually, after the eviction api returns ok, there is no need to kill container if you want to solve the grace termination or PDB problem. It seems that defines a self-defined evictor args(--evictor="default/only-eviction/...", which means --only-evict-by-api) can solve your concerns?

Got you, I will file a new patch later.

j4ckstraw:
@zwzhang0107 PTAL

jasonliu747:
@j4ckstraw please solve the code conflicts, and we are good to go.

j4ckstraw:
@zwzhang0107 rebased

PR #1770, opened 2023-12-05
proposal: dynamic reservation provisioning

eahydra:

The Koordinator's Reservation CRD provides a mechanism for reserving resources on nodes. However, it currently lacks the capability for dynamic provisioning based on immediate workload demands. This proposal introduces Dynamic Reservation Provisioning (DRP), a new mechanism that enables the automatic creation of Reservation CRD instances through interaction with an external provisioner. DRP allows for the on-the-fly creation of Reservation CRD instances to meet the resource requirements of incoming Pods that carry a specific annotation.

### V. Checklist


docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Proposal
+
+### Overview
+
+The DRP mechanism will work as follows:
+
+1. Pod Annotation for Dynamic Reservation: Users can annotate Pods with `scheduling.koordinator.sh/reservation-template`, which includes the specification for a potential Reservation CRD.
+2. Scheduler Responsiveness: If an incoming Pod has this annotation and there is no available Reservation CRD instance that fits its requirements, the Koordinator scheduler will construct a virtual reservation object in memory. This object will emulate the behavior of a Reservation CRD instance to temporarily reserve resources for the Pod.
+3. Triggering Reservation Creation: In the PreBind phase of scheduling, the scheduler appends an annotation `scheduling.koordinator.sh/new-reservation-requirement` to the Pod, signaling an external provisioner to create an actual Reservation CRD instance based on the provided template.
```
hormes, line 94: Why doesn't the scheduler create this reservation directly?

eahydra: When I re-reviewed this process yesterday, I also considered whether it should be created directly by the scheduler? Discussing with @FillZpp  , the external provisioner may need to complete more complex custom logic in the Reservation creation process.

FillZpp: The provisioner(controller) may have its specific protocol on the Reservations it owned, e.g., labels, annotations, and also the Reservation might have to be owned by another owner reference.

Scheduler should not care about these things. It only helps to reserve resources, and let those provisioners create the Reservations they want.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,181 @@
...
+    scheduling.koordinator.sh/reservation-template: |
+      {
+        "owners": {
+          "labelSelector": {
+            "matchLabels": {
+              "tenancy-id": "1612285282502324"
+            }
+          }
+        },
+        "expectedSize": {
```
FillZpp, line 130: recommendedSize

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
FillZpp: The provisioner(controller) may have its specific protocol on the Reservations it owned, e.g., labels, annotations, and also the Reservation might have to be owned by another owner reference.

Scheduler should not care about these things. It only helps to reserve resources, and let those provisioners create the Reservations they want.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
eahydra: done

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
hormes: When creating externally, there are a series of questions that need to be answered, such as:
- If there is no response externally, what should be done? Rollback and how long will it take to do it?
- What to do if the reservation created externally is inconsistent with the resources occupied by the scheduler?

eahydra: Such exceptions are indeed possible, and should they occur, we can consider adopting the following fault tolerance mechanisms:

1. Expose any exceptions in Reservation creation through Events or Metrics.
2. The scheduler should have a default wait timeout, similar to the mechanism in Dynamic Volume Provisioning. If a Reservation is not created after waiting for 10 minutes, the process should be rolled back, and the Pod should be re-queued for another scheduling attempt.
3. Consider retaining the virtual reservation object; it could be reused in the next Pod scheduling attempt. It is recommended to keep it at first. Once the provisioner recovers from any exceptions, it can attempt to create a new Reservation object that can replace the virtual reservation object in memory.
4. In terms of the resource delivery success rate, if the creation of a Reservation fails, the Pod's scheduling could still be completed, and the virtual reservation object should be deleted.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,215 @@
...
+5. Completion of the Binding Process: After the scheduler receives notification of a new Reservation CRD instance, it replaces the virtual object with the actual reservation, adjusts the reservation allocable resources by an amount equal to the sum of the requests of the associated Pods, and continues to bind the Pod to the appropriate node.
+
+### Fault tolerance
+
+Once provisioner exception prevents reservation from being created, fault tolerance needs to be implemented:
+
+- The scheduler should have a default wait timeout, similar to the mechanism in Dynamic Volume Provisioning. If a Reservation is not created after waiting for 10 minutes, the process should be rolled back, and the Pod should be re-queued for another scheduling attempt.
+- Consider retaining the virtual reservation object; it could be reused in the next Pod scheduling attempt. It is recommended to keep it at first. Once the provisioner recovers from any exceptions, it can attempt to create a new Reservation object that can replace the virtual reservation object in memory.
```
hormes, line 107: It seems unreasonable. If the Reservation is not created externally, this resource reserved by the object should not be used by the Pod at this time.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,215 @@
...
+- In terms of the resource delivery success rate, if the creation of a Reservation fails, the Pod's scheduling could still be completed, and the virtual reservation object should be deleted.
```
hormes, line 108: It is also necessary to clarify the processing when the externally created Reservation is inconsistent with the object in memory.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,215 @@
...
+                "tenancy-id": "1612285282502324"
+              }
+            }
+          }
+        ],
+        "recommendedSize": {
+          "cpu": "4",
+          "memory": "8Gi"
+        },
+        "roundup": {
```
hormes, line 148: I think we can keep the strategy simple first and don’t need to add roundup for the time being.

docs/proposals/scheduling/20231205-dynamic-reservation-provisioning.md
```diff
@@ -0,0 +1,215 @@
...
+1. Introduce a means of defining dynamic reservation templates through Pod annotations.
+2. Define a protocol and process for the scheduler to interact with an external provisioner, which is responsible for creating actual Reservation CRD instances.
+
+### Non-Goals/Future Work
+
+1. None
+
+## User stories
+
+### Use Case 1: Just-in-Time Resource Allocation for High-Priority Jobs
```
hormes, line 65: These user stories don’t seem so intuitive and easy to understand.

PR #1782, opened 2023-12-18
scheduler: fix redundant argument when create a unschedulable status.

bogo-y:

Passing an empty string when creating a status could lead to `Status.reasons` has an element which is a empty string. It  causes a redundant `, ` in `Event.message` when there are more than two PostFilter plugins.

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1790, opened 2023-12-25
scheduler: Coscheduling skips check schedule cycle if pod has nominated node

xulinfei1996:
For a gang job, for example, there are four pods, three pods have been assumed, but the fourth cannot be scheduled. The fourth obtains resources through preemption. But in the next round of allocation, the gang's preFilter will be calibrated checking scheduleCycle, which makes makes pod fail to schedule, so we hope to modify the verification logic of gang's preFilter.

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1793, opened 2023-12-26
koordlet: revise system qos and numa aware cpuset

saintube:

- koordlet: revise the cpuset runtime hook for two cases below:
    1. If a pod allocates NUMA-aware resources excluding CPU, it should follow the non-NUMA cpuset.
    2. If a SYSTEM pod allocates NUMA-aware resources including CPU, it just follows the cpuset rule of NUMA resource since the NUMA-aware system-qos-resource is undefined.

Generally, a pod use NUMA-level cpuset should be:

1. QoS is not SYSTEM.
2. Pod annotations["scheduling.koordinator.sh/resource-status"].numaNodeResources[*]["cpu"] should be non-zero.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
+++ b/pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
@@ -26,7 +26,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 
-	ext "github.com/koordinator-sh/koordinator/apis/extension"
+	"github.com/koordinator-sh/koordinator/apis/extension"
 	"github.com/koordinator-sh/koordinator/pkg/features"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/protocol"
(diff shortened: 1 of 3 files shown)


pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
```diff
@@ -54,14 +54,31 @@ func (r *cpusetRule) getContainerCPUSet(containerReq *protocol.ContainerRequest)
...
+	// numa-aware
+	isNUMAAware := false
+	for _, numaNode := range podAlloc.NUMANodeResources {
+		// check if cpu resource is allocated in numa-level since there can be numa allocation without cpu
+		if numaNode.Resources != nil && !numaNode.Resources.Cpu().IsZero() {
+			isNUMAAware = true
+			break
+		}
+	}
+	if isNUMAAware {
```
zwzhang0107, line 80: cpu allocate result should be the first selection if scheduler has specified

saintube: done

PR #1796, opened 2023-12-26
slo-controller: Introduce node resource amplification plugin

zqzten:

This PR introduces node resource amplification plugin to koord-manager.

Currently, it is meant to implement below logic with bold text described in the [CPU normalization proposal](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20230831-cpu-normalization.md):

> On koord-manager side, the Node resource controller runs a reconcile loop to do following things for Nodes with CPU normalization enabled:
>
> 1. Get CPU normalization ratio of the Node from CPU normalization model according to the Node's CPU basic info from its NodeResourceTopology.
> 2. **Calculate the final CPU amplification ratio (this is covered in the Node Resource Amplification proposal) by following formula: CPU_Amplification_Ratio = CPU_Amplification_Ratio_Original * CPU_Normalization_Ratio. The CPU_Amplification_Ratio_Original defaults to 1 if not specified beforehand.**
> 3. **Annotate the** CPU normalization ratio and **new CPU amplification ratio to the Node**.

This makes CPU normalization works with the node resource amplification webhook and scheduler.

Note that currently we have not exposed resource amplification configurations to users, so there's no `CPU_Amplification_Ratio_Original` thus `CPU_Amplification_Ratio` would always be the same as `CPU_Normalization_Ratio`.

part of #1539 and #1570

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/plugins/resourceamplification/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/resourceamplification/plugin.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resourceamplification
+
+import (
+	"encoding/json"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/apis/configuration"
+	"github.com/koordinator-sh/koordinator/apis/extension"
+	"github.com/koordinator-sh/koordinator/pkg/slo-controller/noderesource/framework"
+)
+
+const PluginName = "ResourceAmplification"
+
+// Plugin calculates and updates final node resource amplification ratios automatically
+// based on user config (not implemented) and node cpu normalization ratio.
+type Plugin struct{}
+
+func (p *Plugin) Name() string {
+	return PluginName
+}
+
+func (p *Plugin) NeedSyncMeta(_ *configuration.ColocationStrategy, oldNode, newNode *corev1.Node) (bool, string) {
+	oldRatioStr := oldNode.Annotations[extension.AnnotationNodeResourceAmplificationRatio]
+	newRatioStr := newNode.Annotations[extension.AnnotationNodeResourceAmplificationRatio]
+
+	if oldRatioStr == "" && newRatioStr == "" {
+		return false, "ratio remains empty"
+	}
+	if oldRatioStr == "" {
+		return true, "old ratio is empty"
+	}
+	if newRatioStr == "" {
+		return true, "new ratio is empty"
+	}
+	if oldRatioStr == newRatioStr {
+		return false, "ratio remains unchanged"
+	}
+	return true, "ratio changed"
+}
+
+func (p *Plugin) Prepare(_ *configuration.ColocationStrategy, node *corev1.Node, nr *framework.NodeResource) error {
+	ratioStr, ok := nr.Annotations[extension.AnnotationNodeResourceAmplificationRatio]
+	if !ok {
+		klog.V(6).Infof("prepare node resource amplification ratio to unset, node %s", node.Name)
+		delete(node.Annotations, extension.AnnotationNodeResourceAmplificationRatio)
+		return nil
+	}
+
+	if node.Annotations == nil {
+		node.Annotations = map[string]string{}
+	}
+	node.Annotations[extension.AnnotationNodeResourceAmplificationRatio] = ratioStr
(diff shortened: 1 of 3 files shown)


hormes:
This will cause the normalization and node oversold ratios can not be used at the same time. How do we look at this problem?

zqzten:
> This will cause the normalization and node oversold ratios can not be used at the same time. How do we look at this problem?

It can, we amplify the CPU amplification ratio according to the CPU normalization ratio rather than replace it.

Here's a simple user story:

1. User enables CPU normalization, Node A has a powerful CPU so it gets a normalization ratio 1.5
2. Node A currently does not have a CPU amplification ratio, so it is set to be equal to normalization ratio 1.5
3. Some controller found Node A's CPU util is low for a long time and decide to oversell its CPU with ratio 2, so it updates its CPU amplification ratio to 2
4. The CPU normalization controller found amplification ratio changed, so it re-amplifies it to `2*1.5=3`

hormes:
> > This will cause the normalization and node oversold ratios can not be used at the same time. How do we look at this problem?
> 
> It can, we amplify the CPU amplification ratio according to the CPU normalization ratio rather than replace it.
> 
> Here's a simple user story:
> 
> 1. User enables CPU normalization, Node A has a powerful CPU so it gets a normalization ratio 1.5
> 2. Node A currently does not have a CPU amplification ratio, so it is set to be equal to normalization ratio 1.5
> 3. Some controller found Node A's CPU util is low for a long time and decide to oversell its CPU with ratio 2, so it updates its CPU amplification ratio to 2
> 4. The CPU normalization controller found amplification ratio changed, so it re-amplifies it to `2*1.5=3`, which makes CPU normalization and amplification work together.

 I saw this `AnnotationCPUAmplificationRatioBeforeNormalization` definition, so when the user modifies the node amplification coefficient, is he actually changing this value?

Or do we let the webhook do a little more and let the webhook multiply the normalization and node amplification coefficients and apply them to the allocatable? Only one user-configured amplification factor is retained on the node.

zqzten:
> I saw this `AnnotationCPUAmplificationRatioBeforeNormalization` definition, so when the user modifies the node amplification coefficient, is he actually changing this value?

Yes, we assume he is.

> Or do we let the webhook do a little more and let the webhook multiply the normalization and node amplification coefficients and apply them to the allocatable? Only one user-configured amplification factor is retained on the node.

This sounds neat but is not applicable currently, because the scheduler also uses the amplification ratio annotation to correctly handle CPUSet/NUMA cases.

I think there could be two ways under current circumstance:

1. Keep the current implementation which calculates final CPU amplification ratio automatically according to the original one and CPU normalization ratio.
2. Introduce a new annotation for user to set original CPU amplification ratio, and let koord-manager calculate the final CPU amplification ratio according to this new annotation and CPU normalization ratio. The webhook and scheduler still uses the old annotation to do amplification related works.

Maybe we can have a discussion at today's community meeting?

zqzten:
> I think there could be two ways under current circumstance:
> 
> 1. Keep the current implementation which calculates final CPU amplification ratio automatically according to the original one and CPU normalization ratio.
> 2. Introduce a new annotation for user to set original CPU amplification ratio, and let koord-manager calculate the final CPU amplification ratio according to this new annotation and CPU normalization ratio. The webhook and scheduler still uses the old annotation to do amplification related works.

As discussed at community meeting, we choose the way 2 but with one difference: we do not introduce the new annotation for user for now. Perhaps we would introduce an individual configuration for resource amplification (just like cpu normalization) in the future after the use case become clearer.

The code has been changed according to our decision, PTAL /cc @hormes

zqzten:
> I think there could be two ways under current circumstance:
> 
> 1. Keep the current implementation which calculates final CPU amplification ratio automatically according to the original one and CPU normalization ratio.
> 2. Introduce a new annotation for user to set original CPU amplification ratio, and let koord-manager calculate the final CPU amplification ratio according to this new annotation and CPU normalization ratio. The webhook and scheduler still uses the old annotation to do amplification related works.

As discussed at community meeting, we choose the way 2 but with one difference: we do not introduce the new annotation for user for now. Perhaps we would introduce an individual configuration for resource amplification (just like cpu normalization) in the future after the use case become clearer.

The code has been changed according to our decision, PTAL.
/cc @hormes

pkg/slo-controller/noderesource/plugins/resourceamplification/plugin.go
```diff
@@ -0,0 +1,112 @@
...
+		corev1.ResourceCPU: extension.Ratio(normRatio),
+	}
+	ratioBytes, _ := json.Marshal(ampRatios)
+	ratioStr := string(ratioBytes)
+	klog.V(6).Infof("calculate resource amplification ratio %s for node %s", ratioStr, node.Name)
+	return []framework.ResourceItem{
+		{
+			Name: PluginName,
+			Annotations: map[string]string{
+				extension.AnnotationNodeResourceAmplificationRatio: ratioStr,
```
eahydra, line 108: If AnnotationNodeResourceAmplificationRatio already exists and the value is {"memory": "1.2"}, then the memory will be erased here.

zqzten: In current design, the `AnnotationNodeResourceAmplificationRatio` is fully managed by this plugin. In the future, we shall introduce a new annotation or other configuration methods for users and this plugin would take these into consideration when calculating the final `AnnotationNodeResourceAmplificationRatio`.

zqzten: I've updated the comment to make this intention clearer.

eahydra: ok.

saintube (approved):
/lgtm

PR #1803, opened 2023-12-28
scheduler: FilterReservation should perceive preemptible resources

eahydra:

The original implementation of Reservation.FilterReservation only checks whether there is any remaining Reservation, but in fact it also needs to sense the preemption status, otherwise the preemption scenario cannot be satisfied.


buptcozy (approved):
/approve

PR #1779, opened 2023-12-16
docs: support NetworkQoS based on external plugin such as terway-qos

lucming:
### Ⅰ. Describe what this PR does
this is doc for netqos base on terway-qos. and there are some disscussion in https://github.com/koordinator-sh/koordinator/discussions/1764

### V. Checklist

- [ ] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [ ] All checks passed in `make test`

PR #1805, opened 2024-01-02
koordlet: netqos plugin based on tc

lucming:
this pr implements a netqos plugin base on tc,
and there are apis doc:  https://github.com/koordinator-sh/koordinator/pull/1779

### V. Checklist


zwzhang0107:
it will be better to spilt as an independent PR after the proposal #1779 merged

PR #1807, opened 2024-01-02
metrics: seperater metrics as internal and external for slo-controller and koordlet

zwzhang0107:

seperate koordlet/slo-controller metrics as internal and external
ExternalHTTPPath: for end users, such as batch resource util (also compatible with old URL /metircs for koordlet）
InternalHTTPPath: for cluster admin, such as performance of components.
DefaultHTTPPath: merge all internal, external and controller runtime metrics.

### V. Checklist

### DIFF
--- a/cmd/koord-manager/main.go
+++ b/cmd/koord-manager/main.go
@@ -24,21 +24,26 @@ import (
 	"os"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	_ "k8s.io/component-base/metrics/prometheus/clientgo" // load restclient and workqueue metrics
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	"github.com/koordinator-sh/koordinator/cmd/koord-manager/extensions"
 	"github.com/koordinator-sh/koordinator/cmd/koord-manager/options"
 	extclient "github.com/koordinator-sh/koordinator/pkg/client"
 	"github.com/koordinator-sh/koordinator/pkg/features"
+	"github.com/koordinator-sh/koordinator/pkg/slo-controller/metrics"
 	utilclient "github.com/koordinator-sh/koordinator/pkg/util/client"
 	utilfeature "github.com/koordinator-sh/koordinator/pkg/util/feature"
 	"github.com/koordinator-sh/koordinator/pkg/util/fieldindex"
+	metricsutil "github.com/koordinator-sh/koordinator/pkg/util/metrics"
 	_ "github.com/koordinator-sh/koordinator/pkg/util/metrics/leadership"
 	"github.com/koordinator-sh/koordinator/pkg/util/sloconfig"
 	"github.com/koordinator-sh/koordinator/pkg/webhook"
@@ -170,6 +175,11 @@ func main() {
 		klog.V(4).Infof("webhook framework feature gate not enabled")
 	}
 
+	if err := installHTTPHandler(mgr); err != nil {
+		setupLog.Error(err, "unable to install http handler")
+		os.Exit(1)
+	}
+
 	setupLog.Info("starting manager")
 	extensions.StartExtensions(ctx, mgr)
 	if err := mgr.Start(ctx); err != nil {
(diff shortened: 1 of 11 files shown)


cmd/koord-manager/main.go
```diff
@@ -186,3 +195,14 @@ func setRestConfig(c *rest.Config) {
 		c.Burst = *restConfigBurst
 	}
 }
+
+func installHTTPHandler(mgr ctrl.Manager) error {
```
saintube, line 199: does this need to move into the extensions/options directory for internal modifications?

saintube (commented):
Need to update the deploy templates under the `/config` since the metrics paths are going to change.

cmd/koord-manager/main.go
zwzhang0107: not necessary since metrics.InternalRegistry/ExternalRegistry are global

zwzhang0107:
> Need to update the deploy templates under the `/config` since the metrics paths are going to change.

the url path has not defined yet both in config and charts

eahydra:
Is it really necessary to split it into two metrics urls? Users also need to configure two prometheus configurations for slo-controller, and this indicator must be distinguished when adding indicators in future code. Moreover, only two components, koord-manager and koordlet, are involved here. Do scheduler/descheduler also need to be followed? I don't think such a distinction should be made, and it is not a common or recommended practice.

zwzhang0107:
> Is it really necessary to split it into two metrics urls? Users also need to configure two prometheus configurations for slo-controller, and this indicator must be distinguished when adding indicators in future code. Moreover, only two components, koord-manager and koordlet, are involved here. Do scheduler/descheduler also need to be followed? I don't think such a distinction should be made, and it is not a common or recommended practice.

@eahydra  Yes, it is necessary. any extended metric will be charged(ref https://help.aliyun.com/zh/ack/ack-managed-and-ack-dedicated/user-guide/dynamic-resource-overcommitment?spm=a2c4g.11186623.0.0.74d118ab3vDxaM#ef961016a2uds), and for end users they only need care about `external` metrics, this will cost a lot if we do not separate internal and external after we add new metrics about performance.

scheduler/descheduler should also split if they have the same senario in future.

eahydra:
> > Is it really necessary to split it into two metrics urls? Users also need to configure two prometheus configurations for slo-controller, and this indicator must be distinguished when adding indicators in future code. Moreover, only two components, koord-manager and koordlet, are involved here. Do scheduler/descheduler also need to be followed? I don't think such a distinction should be made, and it is not a common or recommended practice.
> 
> @eahydra Yes, it is necessary. any extended metric will be charged(ref https://help.aliyun.com/zh/ack/ack-managed-and-ack-dedicated/user-guide/dynamic-resource-overcommitment?spm=a2c4g.11186623.0.0.74d118ab3vDxaM#ef961016a2uds), and for end users they only need care about `external` metrics, this will cost a lot if we do not separate internal and external after we add new metrics about performance.
> 
> scheduler/descheduler should also split if they have the same senario in future.

This seems to be solved through the reable_config mechanism of prometheus. Only need to configure the metrics regex and add action=keep.

zwzhang0107:
> reable_config

this is not supported for 

> > > Is it really necessary to split it into two metrics urls? Users also need to configure two prometheus configurations for slo-controller, and this indicator must be distinguished when adding indicators in future code. Moreover, only two components, koord-manager and koordlet, are involved here. Do scheduler/descheduler also need to be followed? I don't think such a distinction should be made, and it is not a common or recommended practice.
> > 
> > 
> > @eahydra Yes, it is necessary. any extended metric will be charged(ref https://help.aliyun.com/zh/ack/ack-managed-and-ack-dedicated/user-guide/dynamic-resource-overcommitment?spm=a2c4g.11186623.0.0.74d118ab3vDxaM#ef961016a2uds), and for end users they only need care about `external` metrics, this will cost a lot if we do not separate internal and external after we add new metrics about performance.
> > scheduler/descheduler should also split if they have the same senario in future.
> 
> This seems to be solved through the reable_config mechanism of prometheus. Only need to configure the metrics regex and add action=keep.

prometheus.yaml is not configurable for end user by default in product like ACK

saintube:
> seperate koordlet/slo-controller metrics as internal and external /external-metrics: for end users, such as batch resource util /metrics: for cluster admin, such as performance of components.
> 
> ### Ⅱ. Does this pull request fix one issue?
> ### Ⅲ. Describe how to verify it
> ### Ⅳ. Special notes for reviews
> ### V. Checklist
> * [ ]  I have written necessary docs and comments
> * [ ]  I have added necessary unit tests and integration tests
> * [ ]  All checks passed in `make test`

It seems the external metrics are using the path `/metrics`.

zwzhang0107:
> > ### Ⅰ. Describe what this PR does
> > seperate koordlet/slo-controller metrics as internal and external /external-metrics: for end users, such as batch resource util /metrics: for cluster admin, such as performance of components.
> > ### Ⅱ. Does this pull request fix one issue?
> > ### Ⅲ. Describe how to verify it
> > ### Ⅳ. Special notes for reviews
> > ### V. Checklist
> > 
> > * [ ]   I have written necessary docs and comments
> > * [ ]   I have added necessary unit tests and integration tests
> > * [ ]   All checks passed in `make test`
> 
> It seems the external metrics are using the path `/metrics`.

updated

saintube (approved):
/lgtm

FillZpp (approved):
/lgtm

PR #1814, opened 2024-01-04
scheduler: change reservation default allocate policy to aligned

eahydra:

In the past, for the Default policy, the remaining resources of the Reservation were returned to NodeInfo to ensure that the filter could pass the resource verification. However, in the FilterReservation, it only checked whether the remaining resources of the Reservation were sufficient. This actually led to the logical inconsistency between the Filter and the FilterReservation. For example, if a Reservation A has no remaining resources, it can pass the verification in the Filter, but fails in the FilterReservation. This is very strange. If we try to unify the logic of Filter and FilterReservation, we will find that the Default and Aligned strategies are consistent.

In addition, if a Pod declares a ReservationAffinity but the matching Reservation cannot meet the requirements, the node will be filtered out even if there are remaining resources on the node that can meet the Pod's requirements.

There is actually a potential challenge here: whether the Restricted strategy still makes sense. For example, if a Pod declares a Reservation Affinity, it means that it must be allocated from the Reservation. Does this mean that the Pod actually takes the initiative in allocation, not the Reservation? If it is determined that the Pod has the upper hand, the original Reservation AllocatePolicy can be discarded. This can be considered as part of future work.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_allocator.go
+++ b/pkg/scheduler/plugins/deviceshare/device_allocator.go
@@ -406,8 +406,8 @@ func defaultAllocateDevices(
 
 	var allocations []*apiext.DeviceAllocation
 	resourceMinorPairs := scoreDevices(podRequestPerInstance, nodeDeviceTotal, freeDevices, requestCtx.allocationScorer)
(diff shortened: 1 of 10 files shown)


buptcozy (approved):
/lgtm

PR #1838, opened 2024-01-14
koordlet: fix by adding lock during podmap update

xujihui1985:
add write lock during podmap update, otherwise there is potential race condition

### V. Checklist


jasonliu747 (commented):
/lgtm
/approve

xujihui1985:
thanks

Issue #1780, opened 2023-12-18
[proposal] supports disable runtime quota


eahydra:
**What is your proposal**:

koord-scheduler supports disable runtime quota.

**Why is this needed**:

Runtime quota can solve the fairness problem at the Quota layer, but in some scenarios, such as AI model training scenarios, the job may not be able to run due to runtime quota. It is necessary to provide a switch to turn off this ability to deal with this scenario.

**Is there a suggested solution, if so, please add it**:

eahydra:
resolved by #1839

PR #1839, opened 2024-01-15
scheduler : supports disable runtime quota

xulinfei1996:
Runtime quota can solve the fairness problem at the Quota layer, but in some scenarios, such as AI model training scenarios, the job may not be able to run due to runtime quota. It is necessary to provide a switch to turn off this ability to deal with this scenario.

to https://github.com/koordinator-sh/koordinator/issues/1780

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -186,6 +186,9 @@ type ElasticQuotaArgs struct {
 
 	// EnableCheckParentQuota check parentQuotaGroups' used and runtime Quota in PreFilter
 	EnableCheckParentQuota *bool
+
+	// EnableRuntimeQuota if true, use max instead of runtime for all checks.
+	EnableRuntimeQuota bool
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/scheduler/plugins/elasticquota/plugin_helper.go
+++ b/pkg/scheduler/plugins/elasticquota/plugin_helper.go
@@ -18,7 +18,6 @@ package elasticquota
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"sort"
 	"strings"
@@ -171,8 +170,6 @@ func (g *Plugin) createDefaultQuotaIfNotPresent() {
 			Max: g.pluginArgs.DefaultQuotaGroupMax.DeepCopy(),
 		},
 	}
-	sharedWeight, _ := json.Marshal(defaultElasticQuota.Spec.Max)
-	defaultElasticQuota.Annotations[extension.AnnotationRuntime] = string(sharedWeight)
 	_, err := g.client.SchedulingV1alpha1().ElasticQuotas(g.pluginArgs.QuotaGroupNamespace).
 		Create(context.TODO(), defaultElasticQuota, metav1.CreateOptions{})
 	if err != nil {
@@ -201,8 +198,6 @@ func (g *Plugin) createSystemQuotaIfNotPresent() {
 			Max: g.pluginArgs.SystemQuotaGroupMax.DeepCopy(),
 		},
 	}
-	sharedWeight, _ := json.Marshal(systemElasticQuota.Spec.Max)
-	systemElasticQuota.Annotations[extension.AnnotationRuntime] = string(sharedWeight)
 	_, err := g.client.SchedulingV1alpha1().ElasticQuotas(g.pluginArgs.QuotaGroupNamespace).
 		Create(context.TODO(), systemElasticQuota, metav1.CreateOptions{})
 	if err != nil {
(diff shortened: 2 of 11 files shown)


pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
```diff
@@ -816,7 +816,10 @@ func (gqm *GroupQuotaManager) GetQuotaInformationForSyncHandler(quotaName string
 			nil, nil, nil, nil, fmt.Errorf("groupQuotaManager doesn't have this quota:%v", quotaName)
 	}
 
-	runtime = gqm.RefreshRuntimeNoLock(quotaName)
+	if !disableRuntimeQuota {
+		runtime = gqm.RefreshRuntimeNoLock(quotaName)
```
eahydra, line 820: Why does runtime quota need to be refreshed here?

pkg/scheduler/plugins/elasticquota/plugin_helper.go
```diff
@@ -172,7 +172,9 @@ func (g *Plugin) createDefaultQuotaIfNotPresent() {
 		},
 	}
 	sharedWeight, _ := json.Marshal(defaultElasticQuota.Spec.Max)
-	defaultElasticQuota.Annotations[extension.AnnotationRuntime] = string(sharedWeight)
+	if !*g.pluginArgs.DisableRuntimeQuota {
+		defaultElasticQuota.Annotations[extension.AnnotationRuntime] = string(sharedWeight)
```
eahydra, line 176: It's weird here. Why does it need to set runtime to System/Default?

pkg/scheduler/apis/config/types.go
```diff
@@ -186,6 +186,9 @@ type ElasticQuotaArgs struct {
 
 	// EnableCheckParentQuota check parentQuotaGroups' used and runtime Quota in PreFilter
 	EnableCheckParentQuota *bool
+
+	// DisableRuntimeQuota if true, use max instead of runtime for all checks.
+	DisableRuntimeQuota *bool
```
eahydra, line 191: Here we can define the field type as Bool instead of bool pointer. The code-gen can generate code converting to bool from pointers. Other fields can be changed together.

pkg/scheduler/plugins/elasticquota/plugin_helper.go
```diff
@@ -239,12 +243,15 @@ func (g *Plugin) createRootQuotaIfNotPresent() {
 	klog.Infof("create RootQuota successfully")
 }
 
-func (g *Plugin) snapshotPostFilterState(quotaInfo *core.QuotaInfo, state *framework.CycleState) *PostFilterState {
+func (g *Plugin) snapshotPostFilterState(quotaInfo *core.QuotaInfo, state *framework.CycleState, disableRuntimeQuota bool) *PostFilterState {
```
eahydra, line 246: It is recommended not to sense disableRuntimeQuota here. It is more appropriate to directly pass in a limit value, and calling it usedLimit is not very appropriate. You can call it limit directly.

pkg/scheduler/apis/config/types.go
eahydra, line 191: BTW maybe we can define the field named EnableRuntimeQuota, and by default is true. So other codes will like the following:
```go
var limit corev1.ResourceList
if pl.args.EnableRuntimeQuota {
    limit = pl.quotaManager.RefreshRuntime(quotaName)
}
limit = quotav1.Max(quotaInfo.GetMax(), limit)
```

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
buptcozy, line 820: 经过讨论，在所有用到quota.GetRuntime()的地方，封装成一层函数即可：

util.GetRuntime(quotaInfo , ignoreRuntime bool) {} 这样

xulinfei1996: 经讨论，这里不需要修改，已还原

pkg/scheduler/plugins/elasticquota/plugin_helper.go
xulinfei1996: d，已删除System/Default Quota cr上的runtime

pkg/scheduler/apis/config/types.go
xulinfei1996: done

pkg/scheduler/plugins/elasticquota/plugin_helper.go
xulinfei1996: done

pkg/scheduler/apis/config/types.go
xulinfei1996: done

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
xulinfei1996: done

pkg/scheduler/plugins/elasticquota/plugin_helper.go
```diff
@@ -298,3 +294,10 @@ func printResourceList(rl v1.ResourceList) string {
 	})
 	return strings.Join(res, ",")
 }
+
+func (g *Plugin) getQuotaInfoRuntime(quotaInfo *core.QuotaInfo) v1.ResourceList {
```
eahydra, line 298: 这个函数不太建议叫getQuotaInfoRuntime，因为返回的值不一定是runtime。之前的limit或者干脆就叫max那也说得过去。

xulinfei1996: done

eahydra (approved):
/lgtm
/approve

PR #1841, opened 2024-01-15
webhook: validate network qos config in webhook

lucming:
add some logic to verify netqos config is valid

### V. Checklist

### DIFF
--- a/pkg/webhook/cm/plugins/sloconfig/resource_qos_checker.go
+++ b/pkg/webhook/cm/plugins/sloconfig/resource_qos_checker.go
@@ -19,11 +19,22 @@ package sloconfig
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/storage"
 	"k8s.io/klog/v2"
 
 	"github.com/koordinator-sh/koordinator/apis/configuration"
+	"github.com/koordinator-sh/koordinator/apis/slo/v1alpha1"
+)
+
+const (
+	// InvalidPercentageValueMsg is an error message for value must in percentage range.
+	InvalidPercentageValueMsg string = `must be percentage value just in 0-100'`
 )
 
 var _ ConfigChecker = &ResourceQOSChecker{}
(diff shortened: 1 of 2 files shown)


zwzhang0107:
besides the format, there are more things need to check:
1. sum(NetworkQOS.request) < 100% or TotalNetworkBandwidth, at least leave some for system overhead
2. NetworkQOS.limit < TotalNetworkBandwidth
3. percent and absolute value cannot be used at the same time

jasonliu747 (commented):
/lgtm

pkg/webhook/cm/plugins/sloconfig/resource_qos_checker.go
```diff
@@ -47,7 +59,119 @@ func NewResourceQOSChecker(oldConfig, newConfig *corev1.ConfigMap, needUnmarshal
...
+		allErrs = append(allErrs, field.Invalid(fldPath, intOrPercent, "must be an integer just in 0-100 or quantity (e.g '50M')"))
+	}
+
+	return allErrs
+}
+
+// ValidateQuantityField validates that given value is quantity format, like 50M.
+func ValidateQuantityField(quantityStr string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, err := resource.ParseQuantity(quantityStr); err != nil {
```
eahydra, line 162: Why not consider returning Parse error directly? In addition, is the InvalidQuantityMsg of 50M appropriate, or should it be 50Mi?

lucming: this is just for type checking, maybe 50M and 50Mi are both acceptable?

eahydra: I understand that the error returned by ParseQuantity already contains these error messages. If I understand correctly, there is no need to customize this kind of error message.

lucming: Because I want to get the path to the invalid fields contained in the structure. such as:
```bash
ResourceQOSCfg.ClusterStrategy.LSRClass.IngressRequest: Invalid value: "50a": must be quantity format, like 50M
```
there are some similar checks in k8s: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/validation/validation.go#L460-L473

eahydra: Validation percentage makes sense, but `InvalidQuantityMsg` is not required.

eahydra (approved):
/lgtm
/approve

PR #1843, opened 2024-01-16
koordlet: support NetworkQoS plugin

l1b0k:

This pull request enhances the Koordlet by introducing a new NetQoS plugin. 
Previous discuss at #1764 #1779.

This update includes the incorporation of two distinct interfaces: one for Koordlet and another for [Terway-QoS](https://github.com/AliyunContainerService/terway-qos).  

This default is Koordlet ,  and it is not implemented yet.
The Terway-QoS plugin support  three level priority 'L0', 'L1', 'L2',  the mapping with koordinator qos is as follow.

Koordinator QoS | Kubernetes QoS | Terway QoS
-- | -- | --
SYSTEM | -- | L0
LSE | Guaranteed | L1
LSR | Guaranteed | L1
LS | Guaranteed/Burstable | L1
BE | BestEffort | L2

When specifying different levels of traffic limitations, only three levels of QoS configuration will be used: SYSTEM, LS, and BE.

### V. Checklist


zwzhang0107:
how about implement in runtime hooks so that we can reuse the reconcile and NRI hook(set network qos for during pod createing.

l1b0k:
i would take a look

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
```diff
@@ -0,0 +1,289 @@
...
+				EgressBandwidth:  egress,
+			},
+		})
+	}
+	outPods, err := json.Marshal(pods)
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(filepath.Join(rootPath, podConfig), outPods, 0644)
```
zwzhang0107, line 139: should use ResourceUpdateExecutor when updating config, which has cache an compare before update content

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
```diff
@@ -0,0 +1,289 @@
...
+	lock sync.RWMutex
+	node *Node
+}
+
+func (p *Plugin) Register(op hooks.Options) {
+	klog.V(5).Infof("register hook %v", "terwqy qos configure generator")
+	rule.Register(ruleNameForNodeQoS, description,
+		rule.WithParseFunc(statesinformer.RegisterTypeNodeSLOSpec, p.parseRuleForNodeSLO),
+		rule.WithUpdateCallback(p.update))
+	rule.Register(ruleNameForAllPods, description,
```
zwzhang0107, line 65: use hooks.Register(rmconfig.PreRunPodSandbox,) can make sure pod can get network qos before they start.

pkg/koordlet/runtimehooks/config.go
```diff
@@ -73,6 +74,11 @@ const (
 	// owner: @saintube @zwzhang0107
 	// alpha: v1.4
 	CoreSched featuregate.Feature = "CoreSched"
+
+	// TerwayQoS enables net QoS feature of koordlet.
+	// owner: @l1b0k
+	// alpha: v1.5
+	TerwayQoS featuregate.Feature = "TerwayQoS"
```
zwzhang0107, line 81: should also define NetworkQOS policy here https://github.com/koordinator-sh/koordinator/blob/main/apis/slo/v1alpha1/nodeslo_types.go#L245

zwzhang0107:
also, there is some work on slo-controller: convert network qos from configmap(slo-controller-config) to NodeSLO

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
```diff
@@ -0,0 +1,289 @@
...
+	slov1alpha1 "github.com/koordinator-sh/koordinator/apis/slo/v1alpha1"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/resourceexecutor"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/hooks"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/rule"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/util"
+)
+
+const (
+	rootPath   = "/var/lib/terway/qos"
```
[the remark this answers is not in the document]
l1b0k: The parameters are not of the runtime plugin and are not supported for configuration at the moment.

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
l1b0k: The PreRunPodSandbox does not contain Pod information, so aspects like the Pod QoS level cannot be determined. Additionally, the synchronization of QoS settings with Terway is asynchronous and not guaranteed to take effect before startup. This is not supported at the moment.

config/crd/bases/slo.koordinator.sh_nodeslos.yaml
```diff
@@ -1222,6 +1222,59 @@ spec:
                       cpuPolicy:
                         description: applied policy for the CPU QoS, default = "groupIdentity"
                         type: string
+                      networkQOS:
```
lucming, line 1225: why add this field? didn't see it referenced anywhere.

l1b0k: Added based on the previous comments https://github.com/koordinator-sh/koordinator/pull/1843#discussion_r1470551137

lucming: I think this means you should implement `netqospolicy` in a similar way to `cpupolicy`, something like this
```golang
type ResourceQOSPolicies struct {
	// applied policy for the CPU QoS, default = "groupIdentity"
	CPUPolicy    *CPUQOSPolicy `json:"cpuPolicy,omitempty"`
	// applied policy for the Net QoS, default = "tc"
	NETQOSPolicy *NETQOSPolicy `json:"NETQOSPolicy,omitempty"`
}

type NETQOSPolicy string

const (
	// NETQOSPolicyTerwayQos indicates implement netqos by terway-qos.
	NETQOSPolicyTerwayQos NETQOSPolicy = "terway-qos"
	// NETQOSPolicyTC indicates implement netqos by tc.
	NETQOSPolicyTC NETQOSPolicy = "tc"
)
```

l1b0k: If that's the understanding, does a runtime plugin need to support multiple implementations? 

Previously, I thought that by using a runtime plugin, different implementations were supported through multiple runtime plugins.

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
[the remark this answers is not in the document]
zwzhang0107: host: /var/lib/ --> koordlet-container:/host-var-lib/

/host-var-lib/terway/qos/

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
zwzhang0107: // TODO can parse qos from cgroup dir

zwzhang0107:
it will be great if there are more ut added to hit the patch target:

codecov/patch — 14.88% of diff hit (target 70.00%)

pkg/koordlet/runtimehooks/hooks/terwayqos/terwayqos.go
```diff
@@ -0,0 +1,428 @@
...
+		q, err := parseQoS(qos.LSClass.NetworkQOS, total)
+		if err != nil {
+			return err
+		}
+		node.L1RxBpsMin = q.IngressRequestBps
+		node.L1RxBpsMax = q.IngressLimitBps
+		node.L1TxBpsMin = q.EgressRequestBps
+		node.L1TxBpsMax = q.EgressLimitBps
+	}
+
```
lucming, line 294: LSRClass also belongs to l1

l1b0k: If all configurations are allowed, it will lead to chaos. I hope to keep only two configuration options.

l1b0k: What you are referring to as "mapping" is the assignment of priorities to Pods, which is accomplished in the getPodPrio method. 
The requests and limits for L1/L2 configuration are obtained from the LSClass and BEClass definitions. It should not cause any confusion.

zwzhang0107: it seems NetworkQOS in ResourceQOSStrategy.LSRClass should be ignored in terway qos plugin currently, since there are only limited level(l0-l2) in configurations

zwzhang0107:
PTAL 
/assign @fillzpp @jasonliu747

FillZpp (approved):
/lgtm

jasonliu747 (commented):
/lgtm

PR #1849, opened 2024-01-17
koordlet: eliminate cache GC is not started

saintube:

koordlet: eliminate the error "cache GC is not started" when the koordlet starts.

It fixes the cgroups updating failure of the runtime hooks' rule callbacks since they can be invoked just after the states informer's running and before the runtime hook runs.

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #1855, opened 2024-01-19
scheduler: ElasticQuota runtime is no longer calculated when not needed

eahydra:

Changes have been made in PR #1839. When `runtime` is not needed, `max` will be used as the upper limit of ElasticQuota. But it hasn't been done thoroughly yet. Since there is no need to use `runtime`, there is no need to calculate runtime in the PreFilter stage. In addition, `GetQuotaSummary` is an externally exposed interface. It only needs to read the most recent `runtime`, and there is no need to recalculate `runtime`. If we need to compute `runtime` more aggressively in the future, we should consider implementing a more efficient, cohesive method to support it.

### V. Checklist


buptcozy (approved):
/lgtm

PR #1858, opened 2024-01-19
descheduler: fix bug of nodeAnomalyDetector in LowNodeLoad plugin

bogo-y:
fix https://github.com/koordinator-sh/koordinator/issues/1854
We should not try to mark all source nodes as normal because only abnormal nodes have been evicted. This PR changes it so that only abnormal nodes will be marked as normal.

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1859, opened 2024-01-19
scheduler: remove the restriction that reservation cannot be preempted

### LINKED ISSUE #1856: [proposal] support Reservation preemption
**What is your proposal**:
koord-scheduler supports Reservation preemption.

**Why is this needed**:
In our scenario, Reservations also have different priorities. The lower priority Reservations may create earlier than the higher priority Reservations, so they are assigned first. When the higher priority Reservations come, there are not enough resource to assign them. So we expect that when a high priority Reservation cannot be assigned, it can preempt a lower priority Reservation.
 
eahydra:

Remove the restriction that reservation cannot be preempted. If users create reservation with priority, and mark the reservation is preemptible, the reservation can be preempted, otherwise, the reservation cannot be preempt since the priority is the highest by default.

fix #1856   implement #1887

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
@@ -18,7 +18,6 @@ package eventhandlers
 
 import (
 	"context"
-	"math"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


eahydra:
@buptcozy @xulinfei1996 Do you have any comments?

FillZpp (approved):
/lgtm

PR #1861, opened 2024-01-20
scheduler: improve Coscheduling PreFilter status

### LINKED ISSUE #1860: [proposal] The Coscheduling PreFilter method should not return Error if failed
**What is your proposal**:

If some Pods with the same PodGroup CRD object are created and scheduling fails, the coscheduling PreFilter returns `framework.Error` to stop preemption, but this is not the best option. It should return `framework.UnschedulableAndUnresolvable`.

**Why is this needed**:

`framework.Error` is special in the scheduler. It represents an internal service exception. We can determine whether the scheduler is healthy by identifying whether there are Errors, but a plugin that frequently returns Errors will prevent us from identifying exceptions.

eahydra:

If some Pods with the same PodGroup CRD object are created and scheduling fails, the coscheduling PreFilter returns framework.Error to stop preemption, but this is not the best option. It should return framework.UnschedulableAndUnresolvable.

framework.Error is special in the scheduler. It represents an internal service exception. We can determine whether the scheduler is healthy by identifying whether there are Errors, but a plugin that frequently returns Errors will prevent us from identifying exceptions.

fix #1860

### V. Checklist


buptcozy (approved):
/lgtm

PR #1868, opened 2024-01-24
koordlet: fix metric record err

ls-2018:
Data recording error

### DIFF
--- a/pkg/koordlet/metricsadvisor/collectors/noderesource/node_resource_collector.go
+++ b/pkg/koordlet/metricsadvisor/collectors/noderesource/node_resource_collector.go
@@ -131,7 +131,7 @@ func (n *nodeResourceCollector) collectNodeResUsed() {
 			nodeMetrics = append(nodeMetrics, metric...)
 		}
 		if info := deviceCollector.Infos(); info != nil {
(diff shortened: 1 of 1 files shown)


jasonliu747 (approved):

pkg/koordlet/metricsadvisor/collectors/noderesource/node_resource_collector.go
```diff
@@ -126,12 +126,12 @@ func (n *nodeResourceCollector) collectNodeResUsed() {
...
 	nodeMetrics = append(nodeMetrics, cpuUsageMetrics)
 
-	for _, deviceCollector := range n.deviceCollectors {
+	for deviceType, deviceCollector := range n.deviceCollectors {
 		if metric, _ := deviceCollector.GetNodeMetric(); metric != nil {
 			nodeMetrics = append(nodeMetrics, metric...)
 		}
 		if info := deviceCollector.Infos(); info != nil {
-			n.metricDB.Set(koordletutil.GPUDeviceType, info)
+			n.metricDB.Set(deviceType, info)
```
saintube, line 134: Nice catch, but perhaps `info.Type()` here.
The registered key in the `n.deviceCollectors` is exactly the collector name, though the GPU collector plugin's name is the same as its Type.

PR #1881, opened 2024-01-30
utils: add metric server client

zwzhang0107:

add metric server client util

### V. Checklist


pkg/util/metricserver/metrics_client.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright 2023 The Koordinator Authors.
```
jasonliu747, line 2: image

PR #1892, opened 2024-02-04
scheduler: improve NodeNUMAResource handling node cpu bind policy

eahydra:

If the node enables label `node.koordinator.sh/cpu-bind-policy` and configures the corresponding value to `FullPCPUsOnly` or `SpreadByPCPUs`, then the scheduler should ensure that the allocation result corresponding to `ResourceSpec.RequiredCPUBindPolicy` is consistent.

And the node policy may conflict with the `ResourceSpec.RequiredCPUBindPolicy` defined in the Pod.  For example: the node defined the policy with `FullPCPUsOnly`, and the Pod required `SpreadByPCPUs`,  this creates a conflict that should be filtered.

If the Pod does not declare the ResourceSpec with CPUBindPolicy(either required or preferred), and the node declares the CPUBindPolicy, the scheduler should allocate CPUs by node's CPUBindPolicy, even though the Pod's QoSClass is LS. Although the previous implementation could work, but the implementation was not complete enough.
As for whether koordlet or kubelet will set the cgroup according to the scheduler allocation result on the LS Pod, I think this is a strategic issue. 

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/cpu_topology.go
+++ b/pkg/scheduler/plugins/nodenumaresource/cpu_topology.go
@@ -75,7 +75,7 @@ func (b *CPUTopologyBuilder) Result() *CPUTopology {
 
 // IsValid checks if the topology is valid
 func (topo *CPUTopology) IsValid() bool {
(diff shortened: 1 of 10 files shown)


eahydra:
/hold cancel

PTAL

PR #1894, opened 2024-02-06
scheduler: add Reservation restricted options to control allocatable resources

eahydra:

When a Pod uses a Reservation with an allocation policy of Restricted, the maximum amount of resources that can be used by intersecting resource requests cannot exceed the amount of resources in the Reservation. But in fact, some scenarios require that even intersecting resource requests should be allowed to break this limit. This PR introduces annotation `scheduling.koordinator.sh/reservation-restricted-options` to solve this problem.

1. Create Reservation with the annotation `scheduling.koordinator.sh/reservation-restricted-options`.

```yaml
kind: Reservation
metadata:
  annotations:
    scheduling.koordinator.sh/reservation-restricted-options: |- 
      {"resources": ["cpu", "memory"]}
  name: test-reservation
spec:
  allocateOnce: false
  allocatePolicy: Restricted
  owners:
  - labelSelector:
      matchLabels:
        app: test
  template:
    metadata:
      creationTimestamp: null
    spec:
      containers:
      - image: mock
        imagePullPolicy: IfNotPresent
        name: mock
        resources:
          limits:
            cpu: "4"
            memory: 8Gi
            fakeResources: "4"
          requests:
            cpu: "4"
            memory: 8Gi
            fakeResources: "4"
  ttl: 0s
```

2. Create Pods that used the Reservation.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: test
    spec:
      containers:
      - command:
        - bash
        - -c
        - sleep 360000
        image: busybox:latest
        imagePullPolicy: Always
        name: main
        resources:
          limits:
            cpu: "1"
            fakeResources: "8"
            memory: 1Gi
```

3. Pod successfully scheduled.

### V. Checklist

### DIFF
--- a/apis/extension/reservation.go
+++ b/apis/extension/reservation.go
@@ -38,6 +38,9 @@ const (
 
 	// AnnotationReservationAffinity represents the constraints of Pod selection Reservation
 	AnnotationReservationAffinity = SchedulingDomainPrefix + "/reservation-affinity"
+
+	// AnnotationReservationRestrictedOptions represent the Reservation Restricted options
+	AnnotationReservationRestrictedOptions = SchedulingDomainPrefix + "/reservation-restricted-options"
 )
 
 type ReservationAllocated struct {
@@ -65,6 +68,14 @@ type ReservationAffinitySelector struct {
 	// Required. A list of reservation selector terms. The terms are ORed.
 	// Reuse corev1.NodeSelectorTerm to avoid defining too many repeated definitions.
 	ReservationSelectorTerms []corev1.NodeSelectorTerm `json:"reservationSelectorTerms,omitempty"`
+}
+
+type ReservationRestrictedOptions struct {
+	// Resources means that when the Pod intersects with these resources,
+	// it can only allocate the reserved amount at most.
+	// If the Reservation's AllocatePolicy is Restricted, and no resources configured,
+	// by default the resources equal all reserved resources by the Reservation.
+	Resources []corev1.ResourceName `json:"resources,omitempty"`
 }
 
 func GetReservationAllocated(pod *corev1.Pod) (*ReservationAllocated, error) {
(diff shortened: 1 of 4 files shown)


apis/extension/reservation.go
```diff
@@ -67,6 +70,14 @@ type ReservationAffinitySelector struct {
 	ReservationSelectorTerms []corev1.NodeSelectorTerm `json:"reservationSelectorTerms,omitempty"`
 }
 
+type ReservationRestrictedOptions struct {
+	// Resources means that when the Pod intersects with these resources,
+	// it can only allocate the reserved amount at most.
+	// If the Reservation's AllocatePolicy is Restricted, and no resources configured,
+	// by default the resources equal all reserved resources by the Reservation.
+	Resources []corev1.ResourceName `json:"resources,omitempty"`
+}
```
hormes, line 79: Since the default is full restriction, why not use a whitelist to specify the resources that can be relaxed?

hormes: According to the current definition, the intuition is that by default all resources are unrestricted and the free part of the machine can be used.

eahydra: Controlling restricted lists is easier to maintain.

eahydra: > According to the current definition, the intuition is that by default all resources are unrestricted and the free part of the machine can be used.

The restricted Reservation restricts every reserved resources by default.

hormes: Well, I got this, but it didn‘t quite fit with intuition.

hormes (approved):

PR #1932, opened 2024-03-04
koordlet: fix bug that cpu eviction does not take effect

wangxiaoq:

cpuset's value of sanbox container cloud always be set to all cpu ids of the machine by kubelet, if we use this value, the cpu eviction function will not take effect.

NONE

### V. Checklist


zwzhang0107:
looks great! it will be better to add some ut for this..

saintube (approved):
/lgtm

PR #1948, opened 2024-03-13
scheduler: improve coscheduling's PostFilter if not collect enough

eahydra:

If the PreFilter of Coscheduling return UnschedulableAndUnresolvable since do not collect enough pods of a PodGroup, the PostFilter of Coscheduling will mark the schedule cycle invalid. But there is no need to do this, because once the Pods are collected enough, scheduling can be quickly performed instead of having to wait until the next cycle to actually start scheduling.

### V. Checklist


buptcozy (approved):
/approve

PR #1956, opened 2024-03-15
descheduler: change the global configuration sequence for LowNodeLoadNodePool

ls-2018:
`The user's configuration in the pool should be used first`

### V. Checklist


eahydra:
Can you give more reasons for doing this? Maybe someone wants to deal with other node pools first. So maybe we should add fields like "order", "index" or "priority" to support user configuration via scenarios.

ls-2018:
I think I misunderstood. I misunderstood the configuration of this parameter.
It turns out that the parameters under LowNodeLoadArgs are at the same level as those in NodePools.

PR #1957, opened 2024-03-15
descheduler: change the order of check len(lowNodes) and len(sourceNodes)

bogo-y:
We firstly check len(lowNodes) after classifyNodes, which cloud lead to the sourceNodes lose a chance to be mark as abnormal. If we change the order, the sourceNodes cloud be evict more quickly when we just add a new node to a overloaded cluster.

### V. Checklist


eahydra (commented):
/lgtm

eahydra:
please add signature with command `git commit -s --amend --no-edit`

Issue #1959, opened 2024-03-18
[BUG] deviceshare plugin not handle AddPod\RemovePod correctly


buptcozy:
**What happened**:

in generally, when we execute AddPod logic here, the pod may be in scheduling status,it won't exist in nodeDeviceCache's used map, so there is a bug that  when the framework execute RunFilterPluginsWithNominatedPods with AddPod for high priority pods, the plugin can't reserve resource for hese high priority pods, In RDMA\VF\nv-switch scenario, it can cause high priority pods assign fail due to some resources is assigned by low priority pods. So we reused the "Reserve" logic to generate an assign placement and save it in nominator. We will clear the nominator cache In "Reserve" and "UnReserve", which means we will do clean job no matter assign success or not, this is the same process of the origin k8s framework nominate process.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

PR #1960, opened 2024-03-18
scheduler : fix deviceshare plugin of add\remove pod (#1959)

buptcozy:
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/nominator.go
+++ b/pkg/scheduler/plugins/deviceshare/nominator.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deviceshare
+
+import (
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+
+	schedulingv1alpha1 "github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1"
+)
+
+type Nominator struct {
+	sync.Mutex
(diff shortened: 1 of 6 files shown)


pkg/scheduler/plugins/deviceshare/nominator.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (nominator *Nominator) RemovePod(pod *corev1.Pod) {
+	nominator.Lock()
+	defer nominator.Unlock()
+
+	podNamespacedName := pod.Namespace + "/" + pod.Name
+	delete(nominator.nominateMap, podNamespacedName)
+}
+
+func (nominator *Nominator) GetPodAllocate(pod *corev1.Pod) map[schedulingv1alpha1.DeviceType]deviceResources {
```
xulinfei1996, line 54: GetPodAllocated or GetPodAllocation may be better

buptcozy: fixed

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -178,12 +179,40 @@ func (p *Plugin) AddPod(ctx context.Context, cycleState *framework.CycleState, p
...
+	//we will do clean job no matter assign success or not, this is the same process of the origin k8s framework
+	//nominate process.
+	if len(podAllocated) == 0 && p.nominator != nil {
+		if p.nominator.IsPodExist(podInfoToAdd.Pod) {
+			podAllocated = p.nominator.GetPodAllocated(podInfoToAdd.Pod)
+		} else {
+			assignFunc := func(state *preFilterState, nodeDeviceInfo *nodeDevice, pod *corev1.Pod, assignResult apiext.DeviceAllocations) {
+				//do nothing
+			}
+			result, _ := p.reserveInternal(ctx, cycleState, state, podInfoToAdd.Pod, nodeInfo, nd, assignFunc)
```
eahydra, line 200: It looks a little strange here. 

IMO if the PodToAdd is passed from RunFilterPluginsWithNominatedPods,  it means that the PodToAdd had been preempted other lower priority pods or same priority pods preempted lower priority pods, so that the podToAdd could be added into PodNominator of k8s scheduler framework. 

In the preempt phase, all pods of one Node are removed firstly, then try to call RunFilterPlugins to determine whether the Pod can allocate resources from the Node. If successful, we can store the result in the nominator.  When re-scheduling using RunFiltersWithNominatedPods, we can get the result of the last preemption of other Pods and add into the state of plugins to prevent resources that have been preempted from being used by other scheduled Pods.

PR #1962, opened 2024-03-20
descheduler: only include the overused resources when sorting removable pods

bogo-y:
We have global weights for resources in LowNodeLoad plugin, and we sort pods on abnormal nodes based on these weights and their usage. In most cases, a abnormal node only has one kind of  resource that is overutilized, so we may need to adjust the weights according to the actual situation before sorting the removable pods. I had a idea that we only include the overused resource for each pod through preparing weights before sorting. Could you give more suggestions?

### V. Checklist


eahydra (approved):
/lgtm
/approve

PR #1968, opened 2024-03-21
webhook: fix priority admission conflict

j4ckstraw:
when we set priorityclass koord-xxx with preemptionPolicy=Never, priority admission refuse to create pod with error like this:
```text
pods "test" is forbidden: the string value of PreemptionPolicy (Never)
must not be provided in pod spec; priority admission controller computed
PreemptLowerPriority from the given PriorityClass name
```

we overwrite PriorityClassName/Priority in cluster_colocation_profile, so as well PreemptionPolicy.

see
https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/priority/admission.go#L188

### V. Checklist

### DIFF
--- a/pkg/webhook/pod/mutating/cluster_colocation_profile.go
+++ b/pkg/webhook/pod/mutating/cluster_colocation_profile.go
@@ -210,6 +210,7 @@ func (h *PodMutatingHandler) doMutateByColocationProfile(ctx context.Context, po
 		}
 		pod.Spec.PriorityClassName = profile.Spec.PriorityClassName
 		pod.Spec.Priority = pointer.Int32(priorityClass.Value)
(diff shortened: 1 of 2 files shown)


j4ckstraw:
We change priorityclass preemptionPolicy from PreemptLowerPriority to Never for some reason, then new pod refused by priority admission.

priority admission check if priority and preemptionPolicy is corresponding to priorityClass [link](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/priority/admission.go#L188
)

while koordinator only overwrite PriorityClassName and Priority in cluster colocation profile [link](https://github.com/koordinator-sh/koordinator/blob/main/pkg/webhook/pod/mutating/cluster_colocation_profile.go#L212), preemptionPolicy overwrite is also needed.

eahydra (approved):
/lgtm
/approve

PR #1972, opened 2024-03-26
koordlet: add NRI reconnect

kangclzjc:
Currently, if containerd restart or upgrade, koordlet will lost connection between koordlet and containerd. This PR will reconnect to containerd.

Restart containerd and see koordlet hooks still work.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/config.go
+++ b/pkg/koordlet/runtimehooks/config.go
@@ -18,6 +18,7 @@ package runtimehooks
 
 import (
 	"flag"
+	"math"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/runtime"
@@ -113,6 +114,11 @@ type Config struct {
 	RuntimeHookHostEndpoint         string
 	RuntimeHookDisableStages        []string
 	RuntimeHooksNRI                 bool
+	RuntimeHooksNRIConnectTimeout   time.Duration
+	RuntimeHooksNRIBackOffDuration  time.Duration
+	RuntimeHooksNRIBackOffCap       time.Duration
+	RuntimeHooksNRIBackOffFactor    float64
+	RuntimeHooksNRIBackOffSteps     int
 	RuntimeHooksNRISocketPath       string
 	RuntimeHookReconcileInterval    time.Duration
 }
@@ -127,6 +133,11 @@ func NewDefaultConfig() *Config {
 		RuntimeHookHostEndpoint:         "/var/run/koordlet/koordlet.sock",
 		RuntimeHookDisableStages:        []string{},
 		RuntimeHooksNRI:                 true,
+		RuntimeHooksNRIConnectTimeout:   6 * time.Second,
+		RuntimeHooksNRIBackOffDuration:  1 * time.Second,
+		RuntimeHooksNRIBackOffCap:       1<<62 - 1,
+		RuntimeHooksNRIBackOffSteps:     math.MaxInt32,
+		RuntimeHooksNRIBackOffFactor:    2,
 		RuntimeHooksNRISocketPath:       "nri/nri.sock",
 		RuntimeHookReconcileInterval:    10 * time.Second,
 	}
@@ -139,6 +150,11 @@ func (c *Config) InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.RuntimeHooksPluginFailurePolicy, "runtime-hooks-plugin-failure-policy", c.RuntimeHooksPluginFailurePolicy, "stop running other hooks once someone failed")
 	fs.StringVar(&c.RuntimeHookConfigFilePath, "runtime-hooks-config-path", c.RuntimeHookConfigFilePath, "config file path for runtime hooks")
 	fs.StringVar(&c.RuntimeHookHostEndpoint, "runtime-hooks-host-endpoint", c.RuntimeHookHostEndpoint, "host endpoint of runtime proxy")
+	fs.DurationVar(&c.RuntimeHooksNRIConnectTimeout, "runtime-hooks-nri-connect-timeout", c.RuntimeHooksNRIConnectTimeout, "nri server connect time out, it should be a little more than default plugin registration timeout(5 seconds) which is defined in containerd config")
+	fs.DurationVar(&c.RuntimeHooksNRIBackOffDuration, "runtime-hooks-nri-backoff-duration", c.RuntimeHooksNRIBackOffDuration, "nri server backoff duration")
+	fs.DurationVar(&c.RuntimeHooksNRIBackOffCap, "runtime-hooks-nri-backoff-cap", c.RuntimeHooksNRIBackOffCap, "nri server backoff cap")
+	fs.IntVar(&c.RuntimeHooksNRIBackOffSteps, "runtime-hooks-nri-backoff-steps", c.RuntimeHooksNRIBackOffSteps, "nri server backoff steps")
+	fs.Float64Var(&c.RuntimeHooksNRIBackOffFactor, "runtime-hooks-nri-backoff-factor", c.RuntimeHooksNRIBackOffFactor, "nri server reconnect backoff factor")
 	fs.Var(cliflag.NewStringSlice(&c.RuntimeHookDisableStages), "runtime-hooks-disable-stages", "disable stages for runtime hooks")
 	fs.BoolVar(&c.RuntimeHooksNRI, "enable-nri-runtime-hook", c.RuntimeHooksNRI, "enable/disable runtime hooks nri mode")
 	fs.DurationVar(&c.RuntimeHookReconcileInterval, "runtime-hooks-reconcile-interval", c.RuntimeHookReconcileInterval, "reconcile interval for each plugins")
--- a/pkg/koordlet/runtimehooks/config_test.go
+++ b/pkg/koordlet/runtimehooks/config_test.go
@@ -18,6 +18,7 @@ package runtimehooks
 
 import (
 	"flag"
+	"math"
 	"testing"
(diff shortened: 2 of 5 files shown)


pkg/koordlet/runtimehooks/runtimehooks.go
```diff
@@ -57,12 +57,14 @@ func (r *runtimeHook) Run(stopCh <-chan struct{}) error {
...
-		if err := r.nriServer.Start(); err != nil {
-			// if NRI is not enabled or container runtime not support NRI, we just skip NRI server start
-			klog.Errorf("nri mode runtime hook server start failed: %v", err)
-		} else {
-			klog.V(4).Infof("nri mode runtime hook server has started")
-		}
+		go func() {
+			if err := r.nriServer.Start(); err != nil {
+				// if NRI is not enabled or container runtime not support NRI, we just skip NRI server start
+				klog.Errorf("nri mode runtime hook server start failed: %v", err)
```
saintube, line 63: How about Warningf since the error is tolerated? Or we may distinguish the case in which the NRI should be started successfully but not.

kangclzjc: Yes, we could use Warningf as this error is not fatal.

pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -209,6 +221,20 @@ func (p *NriServer) UpdateContainer(pod *api.PodSandbox, container *api.Containe
...
+	for {
+		stub, err := stub.New(p, append(opts, stub.WithOnClose(p.onClose))...)
+		if err != nil {
+			klog.Errorf("failed to create plugin stub: %v", err)
+		}
+
+		p.stub = stub
+		err = p.Start()
+		if err != nil {
+			time.Sleep(p.options.ReconnectInterval)
```
kangclzjc, line 233: Do we also need a Reconnection limit to control retry times ?

saintube: Agree. Perhaps a retry with backoff.

pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -209,6 +231,21 @@ func (p *NriServer) UpdateContainer(pod *api.PodSandbox, container *api.Containe
 }
 
 func (p *NriServer) onClose() {
-	p.stub.Stop()
-	klog.V(6).Infof("NRI server closes")
+	for attempt := 1; attempt <= p.options.ReconnectionOption.LimitTimes; attempt++ {
```
kangclzjc, line 234: TODO: NRI reconnect duration, status persist

pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -104,19 +113,32 @@ func NewNriServer(opt Options) (*NriServer, error) {
 }
 
 func (p *NriServer) Start() error {
-	go func() {
```
kangclzjc, line 107: TODO: add precheck and postcheck of NRI enable/disable

kangclzjc:
@saintube PTAL.

pkg/koordlet/runtimehooks/runtimehooks.go
```diff
@@ -101,13 +104,22 @@ func NewRuntimeHook(si statesinformer.StatesInformer, cfg *Config) (RuntimeHook,
 		Executor:            e,
 	}
 
+	backOff := wait.Backoff{
+		Duration: cfg.RuntimeHooksNRIReconnectInitInterval,
+		Factor:   cfg.RuntimeHooksNRIReconnectMul,
+		Jitter:   0.1,
+		Steps:    cfg.RuntimeHooksNRIReconnectLimit,
```
saintube, line 111: The meaning of the `Backoff.Steps` is different from the definition of the argument `RuntimeHooksNRIReconnectLimit`.

kangclzjc: in wait.ExponentialBackoff, Steps is the same as Limit 
image
Steps will minus one in each round

pkg/koordlet/runtimehooks/config.go
```diff
@@ -139,6 +149,11 @@ func (c *Config) InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.RuntimeHooksPluginFailurePolicy, "runtime-hooks-plugin-failure-policy", c.RuntimeHooksPluginFailurePolicy, "stop running other hooks once someone failed")
 	fs.StringVar(&c.RuntimeHookConfigFilePath, "runtime-hooks-config-path", c.RuntimeHookConfigFilePath, "config file path for runtime hooks")
 	fs.StringVar(&c.RuntimeHookHostEndpoint, "runtime-hooks-host-endpoint", c.RuntimeHookHostEndpoint, "host endpoint of runtime proxy")
+	fs.DurationVar(&c.RuntimeHooksNRIConnectTimeout, "runtime-hooks-nri-connect-timeout", c.RuntimeHooksNRIConnectTimeout, "nri server connect time out")
```
hormes, line 152: Should we align directly to the definitions in Backoff, such as Duration/Factor/Steps/Cap?

kangclzjc: Yeah, we can align these to Duration/Factor/Steps/Cap, but c.RuntimeHooksNRIConnectTimeout is the timeout that NRI related, it is not related with any backoff parameters.

pkg/koordlet/runtimehooks/runtimehooks.go
saintube: The steps determine the maximum backoff duration, not the number of retries.
However, it is alright since the argument has changed to the steps.

pkg/koordlet/runtimehooks/config.go
```diff
@@ -150,10 +150,10 @@ func (c *Config) InitFlags(fs *flag.FlagSet) {
...
 	fs.DurationVar(&c.RuntimeHooksNRIConnectTimeout, "runtime-hooks-nri-connect-timeout", c.RuntimeHooksNRIConnectTimeout, "nri server connect time out")
-	fs.DurationVar(&c.RuntimeHooksNRIReconnectInitInterval, "runtime-hooks-nri-reconnect-init-interval", c.RuntimeHooksNRIReconnectInitInterval, "nri server reconnection init interval")
-	fs.DurationVar(&c.RuntimeHooksNRIReconnectMaxDuration, "runtime-hooks-nri-reconnect-max-duration", c.RuntimeHooksNRIReconnectMaxDuration, "nri server reconnection max duration, when hit the limit, will not reconnect")
-	fs.IntVar(&c.RuntimeHooksNRIReconnectLimit, "runtime-hooks-nri-reconnect-max-times", c.RuntimeHooksNRIReconnectLimit, "nri server reconnect max limit")
-	fs.Float64Var(&c.RuntimeHooksNRIReconnectMul, "runtime-hooks-nri-reconnect-mul", c.RuntimeHooksNRIReconnectMul, "nri server reconnect backoff mul")
+	fs.DurationVar(&c.RuntimeHooksNRIBackOffDuration, "runtime-hooks-nri-backoff-duration", c.RuntimeHooksNRIBackOffDuration, "nri server backoff duration")
+	fs.DurationVar(&c.RuntimeHooksNRIBackOffCap, "runtime-hooks-nri-backoff-cap", c.RuntimeHooksNRIBackOffCap, "nri server backoff cap, when hit the limit, will not reconnect")
```
saintube, line 154: please also notice the definition of the backOff.Cap

pkg/koordlet/runtimehooks/config_test.go
```diff
@@ -36,6 +36,11 @@ func Test_NewDefaultConfig(t *testing.T) {
 		RuntimeHookHostEndpoint:         "/var/run/koordlet/koordlet.sock",
 		RuntimeHookDisableStages:        []string{},
 		RuntimeHooksNRI:                 true,
+		RuntimeHooksNRIConnectTimeout:   2 * time.Second,
+		RuntimeHooksNRIBackOffDuration:  1 * time.Second,
+		RuntimeHooksNRIBackOffCap:       10 * time.Minute,
```
saintube, line 41: should unlimitedly retry by default

kangclzjc: Sure, let's make nri reconnect unlimitedly.

saintube (commented):
/lgtm

saintube:
PTAL cc @zwzhang0107 @hormes

pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -104,19 +108,34 @@ func NewNriServer(opt Options) (*NriServer, error) {
...
+	err := p.options.Validate()
+	if err != nil {
+		return err
+	}
+	success := time.After(p.options.NriConnectTimeout)
```
zwzhang0107, line 115: this will make stub.Run always success after `NriConnectTimeout`

kangclzjc: yes, and once stub.Run successfully, it will never return. So if after this timeout, we couldn't get any response, we should regard it as success.

> this will make stub.Run always success after `NriConnectTimeout`

zwzhang0107: so maybe we don't need to do `errorChan <- err`, and just return error directly?

kangclzjc: maybe we can't return error directrly here since we should handle the error in other goroutine. And since the error return may need some time in the goroutine, we may need one channel to handle the error which is generated in the other goroutine.

PR #1973, opened 2024-03-26
slo-controller: refactor codes for reading

hormes:

Here we just adjusted the naming of the code and added a diagram to explain the calculation logic of Batch resources. The code logic has not been changed.


zwzhang0107:
we also have a `node.koordinator.sh/system-qos-resource` inside koordinator reserve, which is used for pod with system qos.

hormes:
> we also have a `node.koordinator.sh/system-qos-resource` inside koordinator reserve, which is used for pod with system qos.

It is in koordinator, the reason I won't elaborate here is to make it easier to understand. There will be a user document later to explain in detail some of the details and how to use it.

PR #1980, opened 2024-03-29
scheduler: assure Fairness and DeadlockFree in Coscheduling

ZiMengSheng:

[](scheduler: supply TestFairness for CoScheduling QueueSort)

### V. Checklist


pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -76,6 +76,7 @@ type Gang struct {
 	// the pod has been scheduled in this cycle, so we should reject it. With `totalChildrenNum`'s help, when the last pod comes to make all
 	// `childrenScheduleRoundMap`'s values equal to `scheduleCycle`, Gang's `scheduleCycle` will be added by 1, which means a new schedule cycle.
 	ScheduleCycle            int
+	LastScheduleTime         time.Time
```
hormes, line 79: Please add comments here for readability
// Pods with the same priority are scheduled according to the scheduling time of the previous round to ensure fairness. For a Gang, its last round time as a whole is recorded as LastScheduleTime, that is, all Pods below it have the same time, which is LastScheduleTime.

PR #1982, opened 2024-04-02
apis & slo-controller: allow specifying node-wise total bandwidth via annotation

### LINKED ISSUE #1883: [proposal] Get node bandwidth automatically for network qos
**What is your proposal**:
In cloud environment, node bandwidth is limited according to the instance type by providers. Define in slo-controller-config is complex since they may various instance types in cluster.

We need to define a protocol on Node, such as annotation called `node-bandwith`. koordinator users can implements their own controller to patch the annotation on node according to the instance type.

koordinator also need to check whether the network qos config is valid for each node and alert if there is something unreasonable.

**Why is this needed**:

sjtufl:

Introduce an extension API (through annotation) to specify the total bandwidth for a node. This parameter takes precedence over ones configured in cluster strategy or node strategy, and will take effect in NodeSLO.

fixes #1883 (part 1).

### DIFF
--- a/apis/extension/node_qos.go
+++ b/apis/extension/node_qos.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


sjtufl:
@zwzhang0107  PTAL

zwzhang0107:
remember to commit with -s

sjtufl:
> remember to commit with -s

done

apis/extension/node_qos.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	// AnnotationNodeBandwidth specifies the total bandwidth of the node, which can be
+	// set by cluster administrator or third party components. The value should be a valid
+	// resource.Quantity. Unit: bps.
+	AnnotationNodeBandwidth = NodeDomainPrefix + "/bandwidth"
```
hormes, line 29: net-bandwidth?

sjtufl: renamed to `network-bandwidth`

sjtufl:
@zwzhang0107 @hormes PTAL, just renamed api to `network-bandwidth`

jasonliu747 (approved):
/lgtm

Issue #1442, opened 2023-07-04
[proposal] Support the resource management for Mid-tier pods.


saintube:
**What is your proposal**:

Support the resource management for Mid-tier pods in the koordlet:

- [ ] Identify the extended resources of Mid-tier.
    - [ ] In the runtime hook, Set the necessary cgroups for Mid+BE pods/containers, just like the BatchResource hook.
    - [x] In the metrics module, export the node allocatable and pod requests/limits about Mid resources. #1874
- [x] Distinguish pods with priority classes (non-Prod) apart from the QoS. #1469

**Why is this needed**:

The #1361 adds the Mid-tier resource, allowing a pod with Mid priority class to allocate `kubernetes.io/mid-cpu` and `kubernetes.io/mid-memory` resources. However, the resource management of Mid-tier is not fully supported in the koordlet. For example, the koordlet does not set the related cgroups like `cpu.cfs_quota_us` for a Mid+BE pod.

j4ckstraw:
https://github.com/koordinator-sh/koordinator/pull/1874 export mid resource metric

ZiMengSheng:
Mid-tier resource management has landed: `mid-cpu`/`mid-memory` extended resources (`apis/extension/resource.go`), Mid resource metrics export, and the `batchresource` runtime hook managing cgroups for Mid+BE pods. All checklist items here are done. Closing as completed.

PR #1984, opened 2024-04-03
koordlet: add mid runtimehooks

j4ckstraw:

fix https://github.com/koordinator-sh/koordinator/issues/1442 

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/hooks/extendedresource/extended_resource.go
+++ b/pkg/koordlet/runtimehooks/hooks/extendedresource/extended_resource.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package batchresource
+package extendedresourcee
 
 import (
 	"fmt"
--- a/pkg/koordlet/runtimehooks/config.go
+++ b/pkg/koordlet/runtimehooks/config.go
@@ -25,10 +25,10 @@ import (
 	"k8s.io/component-base/featuregate"
 
 	"github.com/koordinator-sh/koordinator/pkg/features"
-	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/hooks/batchresource"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/hooks/coresched"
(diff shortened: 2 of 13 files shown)


j4ckstraw:
I have no idea how to deal with the BatchResource featuregate, What do you think? @zwzhang0107

pkg/koordlet/runtimehooks/hooks/extendedresource/extended_resource.go
```diff
@@ -379,6 +380,7 @@ func (p *plugin) SetContainerMemoryLimit(proto protocol.HooksProtocol) error {
 	return nil
 }
 
-func isPodQoSBEByAttr(labels map[string]string, annotations map[string]string) bool {
-	return apiext.GetQoSClassByAttrs(labels, annotations) == apiext.QoSBE
+func isPodQoSMatchedByAttr(labels map[string]string, annotations map[string]string) bool {
+	podQoS := apiext.GetQoSClassByAttrs(labels, annotations)
+	return podQoS == apiext.QoSBE || podQoS == apiext.QoSLS
```
zwzhang0107, line 385: which cgroup will Mid+LS pods use
/besteffort may be not appropriate
more things need to be clarify if using /burstable, such as the fairness of cpushares value between Prodand Mid

j4ckstraw: This function is just a filter which means it allow LS and BE.

zwzhang0107: Mid+BE in /besteffort seems reasonable, while Mid+LS should better use /burstable(but this needs some hack works since kubelet is not allowed yet)

I think now we can let koordlet supports Mid+BE first.

So here are the things we should fix:
1. keep podQOSConditions still as apiext.QoSBE
2. define independent GetContainerMidMilliCPURequest and Memory instead of using GetContainerBatchMilliCPURequest

Moreover, there are several QoS policies for BE, we need to clarify whether they should support Mid+BE and how to support, for example the evict order of pods. it will be better if we have some docs first.

PR #1990, opened 2024-04-08
scheduler: remove reservation nominator useless code

xulinfei1996:
If Reservation is Active, Failed or Succeed, it must be scheduled after it triggers preemption, so its nominated information will be deleted in this round schedule process. As in the framework, it will exec DeleteNominatedReservePod in RunReservePluginsReserve.
In a conclusion, it is no need to DeleteReservePod in ReservationHandler OnUpdate handler.

### V. Checklist


saintube:
> If Reservation is Active, Failed or Succeed, it must be scheduled after it triggers preemption, so its nominated information will be deleted in this round schedule process.

AFAIK, the reservation [would not get scheduled](https://github.com/koordinator-sh/koordinator/blob/main/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go#L220) (enqueue the SchedulingQueue) when its phase becomes Available, Failed, or Succeeded. Besides, deleting an unreserved reservation pod in the nominator seems idempotent. So I'm wondering if removing the deletion of the nominated reservation is correct and required.

PR #1992, opened 2024-04-08
scheduler: add new pod estimate with loadaware plugin

zwForrest:

When scheduling a new pod, it is necessary to consider the resource utilization of the new pod with the loadaware plugin, to prevent the node utilization from becoming too high after the new pod is scheduled.

This can prevent nodes with excessive utilization  entering score phase.

The resource utilization  for the new pod here is calculated using the original estimated method.

### V. Checklist


pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -210,8 +215,14 @@ func (p *Plugin) filterNodeUsage(node *corev1.Node, nodeMetric *slov1alpha1.Node
 			continue
 		}
 
-		used := nodeUsage.ResourceList[resourceName]
-		usage := int64(math.Round(float64(used.MilliValue()) / float64(total.MilliValue()) * 100))
+		var usage int64
+		used := getResourceValue(resourceName, nodeUsage.ResourceList[resourceName])
+		if resourceName == corev1.ResourceCPU {
```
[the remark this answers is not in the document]
zwForrest: Because an existing function `getResourceValue` is used here, which handles CPU separately within the function, so a split was also made here.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -42,8 +42,8 @@ import (
 
 const (
 	Name                                    = "LoadAwareScheduling"
-	ErrReasonUsageExceedThreshold           = "node(s) %s usage exceed threshold"
-	ErrReasonAggregatedUsageExceedThreshold = "node(s) %s aggregated usage exceed threshold"
+	ErrReasonUsageExceedThreshold           = "node(s) %s usage exceed threshold after adding new pod"
```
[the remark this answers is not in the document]
zwForrest: ok

pkg/scheduler/plugins/loadaware/load_aware.go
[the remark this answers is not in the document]
eahydra: IMO there is no need to distinguish whether it is CPU.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -170,7 +175,7 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 	return nil
 }
 
-func (p *Plugin) filterNodeUsage(node *corev1.Node, nodeMetric *slov1alpha1.NodeMetric, filterProfile *usageThresholdsFilterProfile) *framework.Status {
+func (p *Plugin) filterNodeUsage(node *corev1.Node, nodeMetric *slov1alpha1.NodeMetric, filterProfile *usageThresholdsFilterProfile, estimatedUsed map[corev1.ResourceName]int64) *framework.Status {
```
eahydra, line 178: It's recommended here to consider calling the function `estimatedAssignedPodUsed` to get usage of assigned pods, thus the used is equal to totalUsed(from nodeMetrics) +  estimatedUsed(currentPod) + estimatedAssignedPodUsed.

zwForrest: ok

hormes:
What is the status of this issue?

zwForrest:
> What is the status of this issue?

The corresponding changes have been updated.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,9 +146,29 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
 
+	allocatable, err := p.estimator.EstimateNode(node)
+	if err != nil {
+		klog.ErrorS(err, "Failed to EstimateNode", "node", node.Name)
+		return framework.NewStatus(framework.Error, err.Error())
+	}
 	filterProfile := generateUsageThresholdsFilterProfile(node, p.args)
-	if len(filterProfile.ProdUsageThresholds) > 0 && extension.GetPodPriorityClassWithDefault(pod) == extension.PriorityProd {
-		status := p.filterProdUsage(node, nodeMetric, filterProfile.ProdUsageThresholds)
+	prodPod := len(filterProfile.ProdUsageThresholds) > 0 && extension.GetPodPriorityClassWithDefault(pod) == extension.PriorityProd && p.args.ScoreAccordingProdUsage
```
eahydra, line 155: Why add new condition `&& p.args.ScoreAccordingProdUsage` here?

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,9 +146,29 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 		return nil
 	}
 
+	allocatable, err := p.estimator.EstimateNode(node)
+	if err != nil {
+		klog.ErrorS(err, "Failed to EstimateNode", "node", node.Name)
+		return framework.NewStatus(framework.Error, err.Error())
```
eahydra, line 152: We should not return framework.Error here, we should return framework.Unschedulable or UnschedulableAndUnresolvable to ensure that Pod is assigned to nodes as much as possible.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,9 +146,29 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+	}
 	filterProfile := generateUsageThresholdsFilterProfile(node, p.args)
-	if len(filterProfile.ProdUsageThresholds) > 0 && extension.GetPodPriorityClassWithDefault(pod) == extension.PriorityProd {
-		status := p.filterProdUsage(node, nodeMetric, filterProfile.ProdUsageThresholds)
+	prodPod := len(filterProfile.ProdUsageThresholds) > 0 && extension.GetPodPriorityClassWithDefault(pod) == extension.PriorityProd && p.args.ScoreAccordingProdUsage
+	podMetrics := buildPodMetricMap(p.podLister, nodeMetric, prodPod)
+
+	estimatedUsed, err := p.estimator.EstimatePod(pod)
+	if err != nil {
+		return framework.NewStatus(framework.Error, err.Error())
```
eahydra, line 160: Also here, do not return error.

eahydra:
I'm sorry I've been away from X for a while.

I reviewed this PR and reviewed the implementation of the loadaware plugin. I noticed that there are many areas that need to be improved in the code I implemented in the past. Especially in this PR, you can see that the logic of filter and score are very similar. We should calculate the usage first, and then decide whether to score or filter. When calculating usage, we can decide whether to use estimated.

These codes should be wrapped as a function that can be used in filter and score stage.
https://github.com/koordinator-sh/koordinator/blob/038308b7359feceea12e17fb327ad3acd521e0e1/pkg/scheduler/plugins/loadaware/load_aware.go#L291-L327

pkg/scheduler/plugins/loadaware/load_aware.go
zwForrest: I changed this return to nil; if there is an error, filter will not work.

pkg/scheduler/plugins/loadaware/load_aware.go
zwForrest: I made a mistake here, I thought this parameter determines whether the Prod Pod needs to be calculated separately. But it is only used for score stage. I will delete this condition

saintube:
@zwForrest Thanks for your contributions! Please resolve the conflicts and we will continue the review.

zwForrest:
> @zwForrest Thanks for your contributions! Please resolve the conflicts and we will continue the review.

Done, The error is caused by other code.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,75 +145,52 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+	prodPod := len(filterProfile.ProdUsageThresholds) > 0 && extension.GetPodPriorityClassWithDefault(pod) == extension.PriorityProd
+
+	var nodeUsage *slov1alpha1.ResourceMap
+	var usageThresholds map[corev1.ResourceName]int64
+	if prodPod {
+		usageThresholds = filterProfile.ProdUsageThresholds
 	} else {
-		var usageThresholds map[corev1.ResourceName]int64
+		if nodeMetric.Status.NodeMetric != nil {
+			if filterProfile.AggregatedUsage != nil {
```
saintube, line 161: Here the cyclomatic complexity is too high. Please revise the logic for readability.
e.g. The `nodeUsage` is exactly a pre-processing variable that only gets used in the function `GetEstimatedUsed`. Perhaps it can be moved into the function.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -289,11 +235,38 @@ func (p *Plugin) Score(ctx context.Context, state *framework.CycleState, pod *co
...
+		if nodeMetric.Status.NodeMetric != nil {
+			if scoreWithAggregation(p.args.Aggregated) {
+				nodeUsage = getTargetAggregatedUsage(nodeMetric, &p.args.Aggregated.ScoreAggregatedDuration, p.args.Aggregated.ScoreAggregationType)
+			} else {
+				nodeUsage = &nodeMetric.Status.NodeMetric.NodeUsage
+			}
+		}
+	}
+	estimatedUsed, err := p.GetEstimatedUsed(nodeName, nodeMetric, pod, nodeUsage, prodPod)
+	if err != nil {
```
saintube, line 249: expect a log message while estimating abnormally
same below

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,75 +145,52 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+		total := getResourceValue(resourceName, allocatable[resourceName])
+		if total == 0 {
 			continue
 		}
-
-		used := nodeUsage.ResourceList[resourceName]
-		usage := int64(math.Round(float64(used.MilliValue()) / float64(total.MilliValue()) * 100))
-		if usage >= threshold {
+		usage := int64(math.Round(float64(estimatedUsed[resourceName]) / float64(total) * 100))
+		if usage > value {
```
saintube, line 191: Also a cyclomatic complexity problem. Same above.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -146,75 +145,52 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 		return nil
 	}
 
+	allocatable, err := p.estimator.EstimateNode(node)
+	if err != nil {
```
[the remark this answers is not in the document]
zwForrest: ok, i wil add one error message

pkg/scheduler/plugins/loadaware/load_aware.go
zwForrest: I think it is necessary here.

pkg/scheduler/plugins/loadaware/load_aware.go
zwForrest: ok

pkg/scheduler/plugins/loadaware/load_aware.go
zwForrest: ok, I will move the nodemetric check to after obtaining this resource.

saintube:
image
@zwForrest Please remember to correct the failed UT cases.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -140,117 +140,47 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 		}
 		return framework.NewStatus(framework.Error, err.Error())
 	}
+	if nodeMetric.Status.NodeMetric == nil {
+		klog.Warningf("nodeMetrics(%s) should not be nil.\n", node.Name)
```
saintube, line 144: klog / glog does not need a tailing `\n`.
Same below.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -325,13 +280,28 @@ func (p *Plugin) Score(ctx context.Context, state *framework.CycleState, pod *co
...
+	for resourceName, value := range usageThresholds {
+		if value == 0 {
+			continue
+		}
+		total := getResourceValue(resourceName, allocatable[resourceName])
+		if total == 0 {
+			continue
+		}
+		usage := int64(math.Round(float64(estimatedUsed[resourceName]) / float64(total) * 100))
+		if usage > value {
```
saintube, line 296: to reduce the cyclomatic complexity and improve the readability, consider this pattern:

```go
if usage <= value {
  continue
}
// ...
```

zwForrest:
> @zwForrest Please check your logic to pass the UTs.

PR #1974, opened 2024-03-27
koordlet: add pod resctrl

kangclzjc:
### Ⅰ. Describe what this PR does
This PR will enhance pod level resctrl which adjust pod level LLC/MB.

### Ⅱ. Does this pull request fix one issue?
#1831 

### Ⅲ. Describe how to verify it
create a pod with annotation like below. And check /sys/fs/resctrl/koordlet-podid/schemata
```
{
  "llc": {
    "schemata": {
      "range": [20,80],
    },
    "schemataPerCache": [
      {
        "cacheid" : 0,
        "range": [20,50]
      },
      {
        "cacheid" : 1,
        "range": [20,60]
      },
    ],
  },
  "mb": {
    "schemata": {
      "percent": 20,
    },
    "schemataPerCache": [
      {
        "cacheid": 0,
        "percent": 20
      },
      {
        "cacheid": 1,
        "percent": 40
      },
    ],
  }
}
```
### V. Checklist

- [ ] I have written necessary docs and comments
- [ ] I have added necessary unit tests and integration tests
- [ ] All checks passed in `make test`

PR #2005, opened 2024-04-15
koordlet: add resctrl qos collector

### LINKED ISSUE #1832: [proposal] QoS class level LLC/MBA metrics collector
**What is your proposal**:
Add a QoS class level LLC/MBA metrics collector. This collector leverages NodeSLO and PodsInformer to gather real-time data on fixed class LLC and MBA. This proposal is part of #1798 

**Why is this needed**:
We propose bolstering resctrl capabilities by integrating real-time QoS class level LLC/MBA monitoring. This richer data landscape will empower Koordinator to make informed decisions regarding the dynamic allocation of LLC, MBA, and potentially even other resources, tailoring them to the specific needs of diverse workloads.

Rouzip:
Add resctrl qos collector.

fixes #1832

After enable resctrl flag in config: 
``` bash
curl http://localhost:9316/metrics|grep resctrl
```

### V. Checklist


Rouzip:
Sorry for the late pr, any comments is welcome.

pkg/koordlet/util/system/resctrl.go
```diff
@@ -560,6 +575,21 @@ func CheckAndTryEnableResctrlCat() error {
 	return nil
 }
 
+func InitCatGroupIfNotExist(group string) error {
```
saintube, line 578: It seems never to be used.

pkg/koordlet/resourceexecutor/resctrl.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
+)
+
+const ErrResctrlDir = "resctrl path or file not exist"
+const CacheIdIndex = 2
+
+func NewResctrlReader() ResctrlReader {
+	if vendorId, err := system.GetVendorIDByCPUInfo(system.GetCPUInfoPath()); err != nil {
+		// FIXME: should we panic there?
+		klog.V(0).ErrorS(err, "get cpu vendor error")
```
saintube, line 38: Please notice many virtual machines may not define a known vendor. For the scenarios, the QoS plugin and collector of the Resctrl QoS should abort the resctrl setting since the VM environment is unsupported. In this function, perhaps we should throw the error and let the caller abort. It is common where a Kubernetes cluster manages both bare-metal and VM nodes, and the koordlet should not panic for this to fail other resource management functions.

pkg/koordlet/resourceexecutor/resctrl.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func NewResctrlRDTReader() ResctrlReader {
+	return &ResctrlRDTReader{}
+}
+
+func (rr *ResctrlRDTReader) ReadResctrlL3Stat(parent string) (map[CacheId]uint64, error) {
+	l3Stat := make(map[CacheId]uint64)
+	monDataPath := system.GetResctrlMonDataPath(parent)
+	fd, err := os.Open(monDataPath)
```
saintube, line 83: missing `fd.Close()`

pkg/koordlet/util/system/resctrl.go
Rouzip: This function is used in #1974 as the util function, I can remove this function in this pr.

pkg/koordlet/metrics/resctrl.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	ResourceTypeLLC = "llc"
+	ResourceTypeMB  = "mb"
+
+	ResctrlResourceType = "resctrl_resource_type"
+	ResctrlCacheId      = "resctrl_cacheid"
+	ResctrlQos          = "resctrl_qos"
```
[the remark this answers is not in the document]
Rouzip: 👌

pkg/koordlet/resourceexecutor/resctrl.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
+)
+
+const ErrResctrlDir = "resctrl path or file not exist"
+const CacheIdIndex = 2
+
+func NewResctrlReader() ResctrlReader {
```
saintube, line 35: What if the resctrl is enabled during the koordlet's running? Perhaps we can check the support status inside the reading methods.

pkg/koordlet/util/system/resctrl.go
```diff
@@ -90,6 +97,15 @@ func isKernelSupportResctrl() (bool, error) {
 	return isCatFlagSet && isMbaFlagSet, nil
 }
 
+func IsVendorSupportResctrl() bool {
+	vendorID, err := GetVendorIDByCPUInfo(GetCPUInfoPath())
+	if err != nil {
+		klog.V(4).Infof("GetVendorIDByCPUInfo error: %v", err)
+		return false
+	}
+	return vendorID == AMD_VENDOR_ID || vendorID == INTEL_VENDOR_ID
```
saintube, line 106: ARM can support resctrl via MPAM.

pkg/koordlet/metricsadvisor/framework/options.go
```diff
@@ -27,5 +27,6 @@ type Options struct {
 	StatesInformer statesinformer.StatesInformer
 	MetricCache    metriccache.MetricCache
 	CgroupReader   resourceexecutor.CgroupReader
+	ResctrlReader  resourceexecutor.ResctrlReader
```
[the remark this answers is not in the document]
Rouzip: ok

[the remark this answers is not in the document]
Rouzip: > How about placing the ResctrlReader inside the ResctrlCollector plugin since it is not a common reader?

ok

pkg/koordlet/resourceexecutor/resctrl.go
Rouzip: In my design, if both the CPU and OS support resctrl, data will not be actually [collected](https://github.com/koordinator-sh/koordinator/blob/e98f7eb7fc2b533580bf6399ff8b4642e7b7143f/pkg/koordlet/metricsadvisor/collectors/resctrl/resctrl_collector.go#L97-L114) after enabling the resctrl collector plugin without mounting; data will be reported after `mount`.

pkg/koordlet/util/system/resctrl.go
Rouzip: I plan to add support for ARM architecture in future prs.

pkg/koordlet/resourceexecutor/resctrl.go
saintube: OK

pkg/koordlet/util/system/resctrl.go
Rouzip: > OK. We can leave a TODO. However, please do not fail for the ARM platform since we did support it.

Since the ARM platform has numerous [vendor IDs](https://github.com/util-linux/util-linux/blob/0284eb3a8a6505dd9745b042089046ad368bfe74/sys-utils/lscpu-arm.c#L294-L315), I included an ARM vendor map to verify vendor support for various CPUs.

Rouzip:
hello, PTAL 😊 @saintube

pkg/koordlet/resourceexecutor/resctrl.go
```diff
@@ -0,0 +1,147 @@
...
+	if vendorId, err := system.GetVendorIDByCPUInfo(system.GetCPUInfoPath()); err != nil {
+		klog.V(0).ErrorS(err, "get cpu vendor error, stop start resctrl collector")
+		return &fakeReader{}
+	} else {
+		switch vendorId {
+		case system.INTEL_VENDOR_ID:
+			return NewResctrlRDTReader()
+		case system.AMD_VENDOR_ID:
+			return &ResctrlAMDReader{}
+		default:
```
saintube, line 45: So ARM is unsupported by the collector? I'm still wondering if we can adapt the resctrl interfaces and forget the vendor-specific initialization. We can throw the error if any environment does not support some of the resctrl interfaces. This may be a better backward-compatible solution.

pkg/koordlet/util/system/resctrl.go
```diff
@@ -103,18 +123,30 @@ func IsSupportResctrl() (bool, error) {
...
-		klog.Infof("IsSupportResctrl result, cpuSupport: %v, kernelSupport: %v",
+		klog.V(4).Infof("isResctrlAvailableByKernelCmd result, cpuSupport: %v, kernelSupport: %v",
 			cpuSupport, kernelCmdSupport)
 		isSupportResctrl = cpuSupport
 		isInit = true
 	}
 	return isSupportResctrl, nil
 }
 
+func IsResctrlCollectorAvailableByCpuInfo() (bool, error) {
```
saintube, line 134: This should be checked once the agent starts and after the path flags are initialized. But the collector.Enabled() can be called for multiple times. Recommend to move the invocation inside the IsSupportResctrl.

pkg/koordlet/resourceexecutor/resctrl.go
Rouzip: Currently, I cannot test on an ARM platform. According to my understanding of this [introduction](https://openanolis.cn/sig/ARM_ARCH_SIG/doc/475502679126360261), ARM's operations rely on ACPI, which contrasts with the x86 method. Any guidance and additional resources on MPAM would be appreciated.

saintube: We can rely on the resctrl files instead of the vendor specification.

Rouzip: > We can depend on the resctrl interfaces instead of the vendor specification. In case any platform does not fully support the resctrl, throw the error.

Fix now.

pkg/koordlet/util/system/resctrl.go
Rouzip: > This should be checked once the agent starts and after the path flags are initialized. But the collector.Enabled() can be called for multiple times. Recommend to move the invocation inside the IsSupportResctrl.

Any ideas?

saintube: It is not about checking which flags, but when to call this check.

Rouzip: > It is not about checking which flags, but when to call this check.

Ok. I will fix it soon.

Rouzip: done

pkg/koordlet/util/system/resctrl.go
```diff
@@ -103,18 +125,33 @@ func IsSupportResctrl() (bool, error) {
...
+func IsSupportResctrlCollector() (bool, error) {
+	var err error
+	collectorOnceFunc.Do(func() {
+		path := GetCPUInfoPath()
+		mbm, err1 := isResctrlMBMAvailableByCpuInfo(path)
+		cqm, err2 := isResctrlCQMAvailableByCpuInfo(path)
+		isSupportResctrlCollector = mbm && cqm
+		err = multierr.Append(err1, err2)
+	})
+	return isSupportResctrlCollector, err
```
saintube, line 145: If IsSupportResctrlCollector is invoked more than once, the err becomes nil.

saintube:
@Rouzip Thanks for your great contributions. Since it is a large patch, we need to make some tests before it is merged.

pkg/koordlet/util/system/resctrl.go
Rouzip: > If IsSupportResctrlCollector is invoked more than once, the err becomes nil.

Yeah, so print errors during initialization, then just use the results later on.

Rouzip:
> @Rouzip Thanks for your great contributions. Since it is a large patch, we need to make some tests before it is merged.

😊 Anything I can do?

saintube:
> > @Rouzip Thanks for your great contributions. Since it is a large patch, we need to make some tests before it is merged.
> 
> 😊 Anything I can do?



> > @Rouzip Thanks for your great contributions. Since it is a large patch, we need to make some tests before it is merged.
> 
> 😊 Anything I can do?

We will verify this patch on some test environments later. It would also be appreciated if you could add more UTs to increase the coverage to no less than the target of 70%.

saintube (approved):
/lgtm
PTAL /cc @zwzhang0107 @hormes

pkg/koordlet/metricsadvisor/collectors/resctrl/resctrl_collector.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"go.uber.org/atomic"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/pkg/features"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/metriccache"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/metrics"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/metricsadvisor/framework"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/qosmanager/plugins/resctrl"
```
zwzhang0107, line 30: it's better not not import qos plugin
these const can be moved to utils

pkg/koordlet/resourceexecutor/resctrl.go
```diff
@@ -0,0 +1,160 @@
...
+
+func NewResctrlRDTReader() ResctrlReader {
+	return &ResctrlRDTReader{}
+}
+
+func NewResctrlQoSReader() ResctrlReader {
+	return &ResctrlAMDReader{}
+}
+
+func (rr *ResctrlBaseReader) ReadResctrlL3Stat(parent string) (map[CacheId]uint64, error) {
```
zwzhang0107, line 92: more comments about ReadResctrlL3Stat/ReadResctrlMBStat can make the reader easy understand

pkg/koordlet/metricsadvisor/framework/config.go
```diff
@@ -48,7 +50,9 @@ func NewDefaultConfig() *Config {
 		PSICollectorInterval:             10 * time.Second,
 		CPICollectorTimeWindow:           10 * time.Second,
 		ColdPageCollectorInterval:        5 * time.Second,
+		ResctrlCollectorInterval:         1 * time.Second,
```
zwzhang0107, line 53: now 1 second seems to short and not necessary now, how about 10 seconds by default?

pkg/koordlet/metrics/resctrl.go
```diff
@@ -0,0 +1,58 @@
...
+	ResctrlCacheId      = "cache_id"
+	ResctrlQos          = "qos"
+	ResctrlMbType       = "mb_type"
+)
+
+var (
+	QosResctrl = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "qos_resctrl",
+		Help:      "qos resctrl collected by koordlet",
```
zwzhang0107, line 39: detailed help msg will be helpful since resctrl is really an advanced metrics.

pkg/koordlet/metrics/resctrl.go
```diff
@@ -0,0 +1,58 @@
...
+	ResctrlResourceType = "resource_type"
+	ResctrlCacheId      = "cache_id"
+	ResctrlQos          = "qos"
+	ResctrlMbType       = "mb_type"
+)
+
+var (
+	QosResctrl = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "qos_resctrl",
```
zwzhang0107, line 38: "qos" is already included in labels, maybe define as two metric such as "llc_occupancy" and "mbm_occupancy". so that we don't need to set "MetricPropertyResctrlMbType="") during record metrics.

so does the `MetricPropertiesFunc`

pkg/koordlet/metriccache/metric_types.go
```diff
@@ -102,6 +106,12 @@ const (
 
 	MetricPropertyCPIResource MetricProperty = "cpi_resource"
 
+	MetricPropertyNodeQos MetricProperty = "node_qos"
```
zwzhang0107, line 109: MetricPropertyQos MetricProperty = "qos"

pkg/koordlet/metricsadvisor/collectors/resctrl/resctrl_collector.go
Rouzip: The QoS of resctrl needs to align with that in qosmanager, so the qos plugin is imported.

Rouzip: Any ideas is welcome~ @zwzhang0107
cc @saintube

ZiMengSheng: /assign zwzhang0107

Issue #1895, opened 2024-02-06
[proposal] improve coscheduling's log


eahydra:
**What is your proposal**:

The logs of CoScheduling Plugin  needs to be optimized to remove unnecessary interference.

E.g.,
https://github.com/koordinator-sh/koordinator/blob/bdaf28494f888eb35ad1ef4904bc3bc54c63bca3/pkg/scheduler/plugins/coscheduling/core/gang.go#L189-L196

**Why is this needed**: 

**Is there a suggested solution, if so, please add it**:

googs1025:
Hello, could you please tell me more details? Is it about adjusting the log level and removing unnecessary log outputs?

googs1025:
@eahydra should we use `klog.V(4).ErrorS()`, This ensures log level printing and also reflects error messages

PR #2009, opened 2024-04-17
scheduler: improve gang log

googs1025:
I modified the current log level output method. In order to reduce log printing, I only print when --v=4 is used, and use klog.V().ErrorS for the err!=nil part.

fix: https://github.com/koordinator-sh/koordinator/issues/1895#issuecomment-2060218803

### V. Checklist


hormes:
Why do I need to downgrade error logs to V4?

googs1025:
> Why do we need to downgrade error logs to V4?

Indeed, this level needs to be confirmed. However, setting the log level can effectively reduce the number of log prints.

pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -145,7 +146,7 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
 	matchPolicy := util.GetGangMatchPolicyByPod(pod)
 	if matchPolicy != extension.GangMatchPolicyOnlyWaiting && matchPolicy != extension.GangMatchPolicyWaitingAndRunning &&
 		matchPolicy != extension.GangMatchPolicyOnceSatisfied {
-		klog.Errorf("pod's annotation AnnotationGangMatchPolicy illegal, gangName: %v, value: %v",
+		klog.V(4).Infof("pod's annotation AnnotationGangMatchPolicy illegal, gangName: %v, value: %v",
```
eahydra, line 149: We should determine if matchPolicy is an empty string and if not, log an error.

pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -124,19 +124,20 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
...
 			gang.Name, pod.Annotations[extension.AnnotationGangTotalNum])
 		totalChildrenNum = int64(minRequiredNumber)
 	} else if totalChildrenNum != 0 && totalChildrenNum < int64(minRequiredNumber) {
-		klog.Errorf("pod's annotation totalNumber cannot less than minRequiredNumber, gangName: %v, totalNumber: %v,minRequiredNumber: %v",
+
+		klog.V(4).Infof("pod's annotation totalNumber cannot less than minRequiredNumber, gangName: %v, totalNumber: %v,minRequiredNumber: %v",
 			gang.Name, pod.Annotations[extension.AnnotationGangTotalNum], minRequiredNumber)
 		totalChildrenNum = int64(minRequiredNumber)
 	}
 	gang.TotalChildrenNum = int(totalChildrenNum)
```
eahydra, line 136: We encourage cleaner code, which of course includes improved logging.

```suggestion
var expectedTotalNum int64
s := pod.Annotations[extension.AnnotationGangTotalNum]
if s != "" {
   var err error
   expectedTotalNum, err = strconv.ParseInt(s)
   if err != nil {
      klog.V(4).ErrorS(err, "illegal annotation", "key", extension.AnnotationGangTotalNum, "value": s)
   }
}
if expectedTotalNum < int64(minRequiredNumber) {
   expectedTotalNum = int64(minRequiredNumber)
}
gang.TotalChildrenNum = int(expectedTotalNum)
```

pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -157,15 +158,15 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
 
 	waitTime, err := time.ParseDuration(pod.Annotations[extension.AnnotationGangWaitTime])
 	if err != nil || waitTime <= 0 {
-		klog.Errorf("pod's annotation GangWaitTimeAnnotation illegal, gangName: %v, value: %v",
+		klog.V(4).ErrorS(err, "pod's annotation GangWaitTimeAnnotation illegal, gangName: %v, value: %v",
```
eahydra, line 161: All these logs have same problems, should improve these codes to make more clean.

googs1025:
Sorry for being late. I've been busy with other work recently. I'll go through the comments and make some revisions during the holidays.

hormes (approved):

jasonliu747 (approved):
/lgtm

PR #2011, opened 2024-04-18
scheduler: optimize scheduleCycle logic when notEnoughChildren

ZiMengSheng:

If the PreFilter of Coscheduling return UnschedulableAndUnresolvable since do not collect enough pods of a PodGroup, the PostFilter of Coscheduling will mark the schedule cycle invalid. But there is no need to do this, because once the Pods are collected enough, scheduling can be quickly performed instead of having to wait until the next cycle to actually start scheduling.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -55,14 +55,16 @@ const (
 	PodGroupNotFound Status = "PodGroup not found"
 	Success          Status = "Success"
 	Wait             Status = "Wait"
+
+	stateKey = "CoScheduling"
 )
 
 // Manager defines the interfaces for PodGroup management.
 type Manager interface {
-	PreFilter(context.Context, *corev1.Pod) error
+	PreFilter(context.Context, *framework.CycleState, *corev1.Pod) (err error)
 	Permit(context.Context, *corev1.Pod) (time.Duration, Status)
 	PostBind(context.Context, *corev1.Pod, string)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -221,32 +221,28 @@ func (pgMgr *PodGroupManager) ActivateSiblings(pod *corev1.Pod, state *framework
...
 // iii.Check whether the Gang is OnceResourceSatisfied
 // iv.Check whether the Gang has met the scheduleCycleValid check, and reject the pod if negative(only Strict mode ).
 // v.Try update scheduleCycle, scheduleCycleValid, childrenScheduleRoundMap as mentioned above.
-func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) error {
+func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod, pluginName string) (err error) {
 	if !util.IsPodNeedGang(pod) {
 		return nil
 	}
+	preFilterState := &stateData{skipReject: false, skipSetCycleInvalid: false}
+	state.Write(framework.StateKey(pluginName), preFilterState)
```
buptcozy, line 235: 这个既然已经在gang的plugin里了，这个pluginName是不是用const的字符串就行了，感觉不用还当个参数传递进来，怪怪的

buptcozy (approved):
/lgtm

PR #2015, opened 2024-04-22
koordlet: revise node topology reporting for the kubelet cpu manager

saintube:

- CPU topology reporting: always update the NodeResourceTopology's annotation `node.koordinator.sh/node-cpu-allocs` even if the Kubelet cpu manager policy is none or it failed to read the checkpoint file, instead of only updating when it succeeded in parsing the checkpoint file and at least one pod is managed by the static policy. It also helps when we want the koordlet to ignore the cpu manager cpuset in some special cases.
- Command args: Remove the misused argument `cgroup-kube-dir`. Add an argument to parameterize the host directory `/var/lib/kubelet`.

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
+++ b/pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
@@ -51,6 +51,7 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
 	koordletutil "github.com/koordinator-sh/koordinator/pkg/koordlet/util"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/kubelet"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
 	"github.com/koordinator-sh/koordinator/pkg/util"
 	"github.com/koordinator-sh/koordinator/pkg/util/cpuset"
 )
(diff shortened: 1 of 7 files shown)


pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
```diff
@@ -827,6 +815,29 @@ func (s *nodeTopoInformer) calTopologyZoneList(nodeCPUInfo *metriccache.NodeCPUI
...
 
+func (s *nodeTopoInformer) calKubeletAllocatedCPUs(sharePoolCPUs map[int32]*extension.CPUInfo) ([]extension.PodCPUAlloc, error) {
+	// Users can specify the kubelet RootDirectory on the host in the koordlet DaemonSet,
+	// inside koordlet it is mounted to the path /var/lib/kubelet by default.
+	stateFilePath := kubelet.GetCPUManagerStateFilePath(system.Conf.VarLibKubeletRootDir)
+	data, err := os.ReadFile(stateFilePath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to read state file, err: %v", err)
+	}
+	if err != nil || len(data) <= 0 { // os.IsNotExist(err)
```
zwzhang0107, line 826: when will be not exist error

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
```diff
@@ -367,23 +356,21 @@ func (s *nodeTopoInformer) calcNodeTopo() (*nodeTopologyStatus, error) {
...
 	}
 
-	annotations := map[string]string{
+	nodeTopoStatus.Annotations = map[string]string{
 		extension.AnnotationCPUBasicInfo:            string(cpuBasicInfoJSON),
 		extension.AnnotationNodeCPUTopology:         string(cpuTopologyJSON),
 		extension.AnnotationNodeCPUSharedPools:      string(lsCPUSharePoolsJSON),
 		extension.AnnotationKubeletCPUManagerPolicy: string(cpuManagerPolicyJSON),
 		extension.AnnotationNodeBECPUSharedPools:    string(beCPUSharePoolsJSON),
+		extension.AnnotationNodeCPUAllocs:           string(podAllocsJSON),
```
zwzhang0107, line 365: which kind of error should be ignored or considered(not report)

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
saintube: noted below:
```go
// NOTE: This happens when the cpu_manager_state or the kubelet allocated cpus are expected to ignore for
// some special cases (e.g. pods' cpuset.cpus are not managed by the koordlet).
```

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
saintube: We'd better not support the partial update for these annotations. Instead, the koordlet fails the whole NRT reporting when failed to get some of the cpu topology info.

pkg/koordlet/statesinformer/impl/states_noderesourcetopology.go
zwzhang0107: it seems a hack to set wrong`kubelet root dir`.
we need to define which scenario koordinator supports.

PR #2038, opened 2024-05-08
all: migrate to 1.28.7

ZiMengSheng:
### V. Checklist


docker/koordlet.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$TARGETPLATFORM golang:1.18 as builder
+FROM --platform=$TARGETPLATFORM golang:1.20 as builder
```
saintube, line 1: Because the koordlet depends on the glibc (feature Libpfm imported) and the golang:1.20 uses glibc >= 2.34, the Ubuntu image should be upgraded.

cmd/koord-manager/main.go
```diff
@@ -70,7 +71,7 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true, "Whether you need to enable leader election.")
 	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "koordinator-system",
 		"This determines the namespace in which the leader election configmap will be created, it will use in-cluster namespace if empty.")
-	flag.StringVar(&leaderElectResourceLock, "leader-elect-resource-lock", resourcelock.ConfigMapsLeasesResourceLock,
+	flag.StringVar(&leaderElectResourceLock, "leader-elect-resource-lock", "configmapsleases",
```
saintube, line 74: this option is not available at 1.28, we may switch to leases as the default

saintube (commented):
For the upgradation of the client-go and Docker image, some compatible issues should be noticed for the koordlet and koord-manager.

.github/workflows/e2e-k8s-1.28.yaml
```diff
@@ -0,0 +1,106 @@
+name: E2E-K8S-1.24
```
saintube, line 1: do not forget update the workflow name

.github/workflows/e2e-k8s-1.28.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - main
+      - release-*
+  pull_request: {}
+  workflow_dispatch: {}
+
+env:
+  # Common versions
+  GO_VERSION: '1.20'
+  KIND_ACTION_VERSION: 'v1.5.0'
+  KIND_VERSION: 'v0.20.0'
```
saintube, line 15: perhaps upgrade to 0.21.0 here or in the later PR, since the v0.20.0 mainly supports up to k8s 1.27.
see https://github.com/kubernetes-sigs/kind/releases

saintube (commented):
/lgtm

hormes (approved):

Issue #1831, opened 2024-01-12
[proposal] support pod level resctrl enhancement


kangclzjc:
**What is your proposal**:
Koordinator support LLC and MBA configuration and adjustment by config map based on class level. As Koordinator already support NRI in 0.3.0 release, we want to add pod-level LLC/MBA configuration/adjustment. This is part of #1798 

**Why is this needed**:
If there exists noisy neighbor, this feature can limit noisy neighbor's LLC and MBA.

**Is there a suggested solution, if so, please add it**:

ZiMengSheng:
Closing this stale proposal — pod-level resctrl has had no maintainer discussion or assignee since 2024, and the current koordlet roadmap is focused elsewhere. Node-level RDT/resctrl support remains available. Happy to revisit a refreshed proposal if interest returns.

PR #2052, opened 2024-05-21
koordlet: add resctrl updater

kangclzjc:
Add new schemata updater for resctrl runtime hook. It's part of the https://github.com/koordinator-sh/koordinator/pull/1974. Resctrl runtime hook will leverage this to create the resctrl group and write schemata.

### V. Checklist

### DIFF
--- a/pkg/koordlet/util/system/resctrl.go
+++ b/pkg/koordlet/util/system/resctrl.go
@@ -256,11 +256,18 @@ func (r *ResctrlSchemataRaw) L3Number() int {
 }
 
 func (r *ResctrlSchemataRaw) CacheIds() []int {
-	ids := []int{}
+	ids1 := []int{}
 	for id := range r.L3 {
-		ids = append(ids, id)
+		ids1 = append(ids1, id)
 	}
-	return ids
+	ids2 := []int{}
+	for id := range r.MB {
+		ids2 = append(ids2, id)
+	}
+	if len(ids1) >= len(ids2) {
+		return ids1
+	}
+	return ids2
 }
 
 func (r *ResctrlSchemataRaw) L3String() string {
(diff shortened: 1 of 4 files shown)


kangclzjc:
@saintube PTAL.

pkg/koordlet/util/system/resctrl.go
```diff
@@ -256,11 +256,18 @@ func (r *ResctrlSchemataRaw) L3Number() int {
 }
 
 func (r *ResctrlSchemataRaw) CacheIds() []int {
```
saintube, line 258: Would you be able to separate the L3 number and the MB number? Only using the larger one may cause the wrong output in the DeepCopy and String methods.

kangclzjc: Yes, then we can split this one.

saintube: OK. Please leave a TODO here and resolve it later.
Currently, there are some methods invoking the `CacheIds()` like `ResctrlSchemataRaw.DeepCopy()` and `MBString()`, it only supports the case where the L3 number and the MB number are the same.

kangclzjc: yes, we can only support the case which L3 number == MB number. We leave a TODO here and resolve it later.

kangclzjc: > OK. Please leave a TODO here and resolve it later. Currently, there are some methods invoking the `CacheIds()` like `ResctrlSchemataRaw.DeepCopy()` and `MBString()`, it only supports the case where the L3 number and the MB number are the same.

Add a TODO here. We could fix it in other PR.

saintube (approved):
/lgtm

PR #2055, opened 2024-05-21
 koordlet: skip the container which is not running in cpuBurst applyCFSQuotaBurst

yangfeiyu20102011:
Some containers are not running, maybe they have failed, just skip them.
It can also reduce the confusing err msgs.

### V. Checklist

### DIFF
--- a/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
+++ b/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
@@ -355,6 +355,11 @@ func (b *cpuBurst) applyCFSQuotaBurst(burstCfg *slov1alpha1.CPUBurstConfig, podM
 			continue
 		}
 
+		if containerStat.State.Running == nil {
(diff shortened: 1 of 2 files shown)


pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
```diff
@@ -355,6 +355,11 @@ func (b *cpuBurst) applyCFSQuotaBurst(burstCfg *slov1alpha1.CPUBurstConfig, podM
 			continue
 		}
 
+		if containerStat.State.Running == nil {
+			klog.Infof("skip container %s/%s/%s, because it is not running", pod.Namespace, pod.Name, containerStat.Name)
```
saintube, line 359: how about `klog.V(6).Infof` for skip message

yangfeiyu20102011: > how about `klog.V(6).Infof` for skip message

ok, I have resolved it.

saintube (approved):
/lgtm

PR #2057, opened 2024-05-22
koordlet: add taskids in statesinformer

kangclzjc:
Add taskids for each pod as an attributes. It's part of the https://github.com/koordinator-sh/koordinator/pull/1974. Resctrl runtime hook will leverage this to write new taskids to resctrl group.

### V. Checklist

### DIFF
--- a/pkg/koordlet/statesinformer/impl/states_pods.go
+++ b/pkg/koordlet/statesinformer/impl/states_pods.go
@@ -31,6 +31,7 @@ import (
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/metrics"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/pleg"
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/resourceexecutor"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/statesinformer"
 	koordletutil "github.com/koordinator-sh/koordinator/pkg/koordlet/util"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
@@ -57,6 +58,7 @@ type podsInformer struct {
 	nodeInformer *nodeInformer
 
 	callbackRunner *callbackRunner
+	cgroupReader   resourceexecutor.CgroupReader
 }
 
 func NewPodsInformer() *podsInformer {
@@ -85,6 +87,8 @@ func (s *podsInformer) Setup(ctx *PluginOption, states *PluginState) {
 	s.nodeInformer = nodeInformer
 
 	s.callbackRunner = states.callbackRunner
+
+	s.cgroupReader = resourceexecutor.NewCgroupReader()
 }
 
 func (s *podsInformer) Start(stopCh <-chan struct{}) {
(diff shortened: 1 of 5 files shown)


kangclzjc:
@saintube PTAL.

pkg/koordlet/statesinformer/impl/states_pods.go
```diff
@@ -141,6 +145,70 @@ func (s *podsInformer) GetAllPods() []*statesinformer.PodMeta {
 	return pods
 }
 
+func (s *podsInformer) getTaskIds(podMeta *statesinformer.PodMeta) {
+	pod := podMeta.Pod
```
saintube, line 149: We'd better filter the pods that should not be running to reduce unnecessary mem allocation and reading. e.g. the terminated pods

pkg/koordlet/statesinformer/impl/states_pods.go
```diff
@@ -155,10 +223,15 @@ func (s *podsInformer) syncPods() error {
 	for i := range podList.Items {
 		pod := &podList.Items[i]
 		podMeta := &statesinformer.PodMeta{
-			Pod:       pod, // no need to deep-copy from unmarshalled
-			CgroupDir: genPodCgroupParentDir(pod),
+			Pod:              pod, // no need to deep-copy from unmarshalled
+			CgroupDir:        genPodCgroupParentDir(pod),
+			ContainerTaskIds: make(map[string][]int32),
```
saintube, line 228: perhaps postpone the initialization inside the `if s.config.EnablePodTaskIds` or `getTaskIds`

pkg/koordlet/statesinformer/impl/states_pods_test.go
```diff
@@ -531,3 +536,242 @@ func Test_recordPodResourceMetrics(t *testing.T) {
...
+
+			testingPrepareContainerCgroupCPUTasks(t, helper, tt.fields.containerParentDir, tt.fields.containerTasksStr)
+
+			system.CommonRootDir = ""
+			if tt.fields.invalidPath {
+				system.Conf.CgroupRootDir = "invalidPath"
+			}
+
+			c := NewDefaultConfig()
+			c.KubeletSyncInterval = 3 * time.Second
```
saintube, line 759: is is necessary to adjust this parameter

pkg/koordlet/statesinformer/impl/states_pods.go
```diff
@@ -141,6 +145,70 @@ func (s *podsInformer) GetAllPods() []*statesinformer.PodMeta {
...
+		if err != nil && resourceexecutor.IsCgroupDirErr(err) {
+			klog.V(5).Infof("failed to read container task ids whose cgroup path %s does not exists, err: %s",
+				containerDir, err)
+			continue
+		} else if err != nil {
+			klog.Warningf("failed to get pod container cgroup task ids for container %s/%s/%s, err: %s",
+				pod.Namespace, pod.Name, container.Name, err)
+			continue
+		}
+		if podMeta.ContainerTaskIds != nil {
```
saintube, line 180: how about moving this check to the front

pkg/koordlet/statesinformer/impl/states_pods.go
kangclzjc: Can we add a check to check pod phase and only getTaskIds when pod is in running status?
```go
// These are the valid statuses of pods.
const (
	// PodPending means the pod has been accepted by the system, but one or more of the containers
	// has not been started. This includes time before being bound to a node, as well as time spent
	// pulling images onto the host.
	PodPending PodPhase = "Pending"
	// PodRunning means the pod has been bound to a node and all of the containers have been started.
	// At least one container is still running or is in the process of being restarted.
	PodRunning PodPhase = "Running"
	// PodSucceeded means that all containers in the pod have voluntarily terminated
	// with a container exit code of 0, and the system is not going to restart any of these containers.
	PodSucceeded PodPhase = "Succeeded"
	// PodFailed means that all containers in the pod have terminated, and at least one container has
	// terminated in a failure (exited with a non-zero exit code or was stopped by the system).
	PodFailed PodPhase = "Failed"
	// PodUnknown means that for some reason the state of the pod could not be obtained, typically due
	// to an error in communicating with the host of the pod.
	// Deprecated: It isn't being set since 2015 (74da3b14b0c0f658b3bb8d2def5094686d0e9095)
	PodUnknown PodPhase = "Unknown"
)
```

pkg/koordlet/statesinformer/impl/states_pods.go
kangclzjc: yes, we can postpone this.

pkg/koordlet/statesinformer/impl/states_pods.go
kangclzjc: Yes, let's do it.

pkg/koordlet/statesinformer/impl/states_pods.go
saintube: Yes. We usually focus on the pods whose phases are Pending or Running, while the others mean the pods are terminated. However, some pods or containers may not be running in the phases due to the containerStatuses and the inconsistency.

pkg/koordlet/statesinformer/impl/states_pods_test.go
saintube: ok

kangclzjc:
@saintube PTAL.

pkg/koordlet/statesinformer/impl/states_pods.go
```diff
@@ -159,6 +227,11 @@ func (s *podsInformer) syncPods() error {
 			CgroupDir: genPodCgroupParentDir(pod),
 		}
 		newPodMap[string(pod.UID)] = podMeta
+		if s.config.EnablePodTaskIds && pod.Status.Phase != corev1.PodSucceeded && pod.Status.Phase != corev1.PodFailed {
```
saintube, line 230: recommend to use `util.IsPodTerminated`

saintube (approved):
/lgtm

PR #2066, opened 2024-05-27
koord-descheduler: balance prod pods between nodes in LowNodeLoad

### LINKED ISSUE #2043: [proposal] deschedule production pods between nodes
**What is your proposal**:
When the loads of the two nodes are almost the same, the production application and batch application loads of this node are also similar. We expect that the production application and the batch application will be balanced between nodes. Currently, the loadaware scheduling plugin can evaluate pod loads based on Production and Batch pod, but descheduler does not have the ability to balance based on production applications. This way the node can reduce hotspots caused by production application load.

**Why is this needed**:
reduce hotspots caused by production application load.

zwForrest:
Balance the usage of prod pods among nodes. Ensure that both the resource usage of prod pods on nodes and the overall node usage are balanced. This helps reduce the impact of node usage on prod applications.

fix #2043 

### V. Checklist

### DIFF
--- a/pkg/descheduler/apis/config/types_loadaware.go
+++ b/pkg/descheduler/apis/config/types_loadaware.go
@@ -37,6 +37,7 @@ type LowNodeLoadArgs struct {
 
 	// NumberOfNodes can be configured to activate the strategy only when the number of under utilized nodes are above the configured value.
 	// This could be helpful in large clusters where a few nodes could go under utilized frequently or for a short period of time.
+	// This parameter includes the sum of nodes with low node utilization, low prod utilization, and both.
 	// By default, NumberOfNodes is set to zero.
 	NumberOfNodes int32
 
@@ -65,11 +66,17 @@ type LowNodeLoadArgs struct {
 	// A resource consumption above (resp. below) this window is considered as overutilization (resp. underutilization).
 	UseDeviationThresholds bool
 
-	// HighThresholds defines the target usage threshold of resources
+	// HighThresholds defines the target usage threshold of node resources
 	HighThresholds ResourceThresholds
 
-	// LowThresholds defines the low usage threshold of resources
+	// LowThresholds defines the low usage threshold of node resources
 	LowThresholds ResourceThresholds
+
+	// ProdHighThresholds defines the target usage threshold of Prod resources
+	ProdHighThresholds ResourceThresholds
+
+	// ProdLowThresholds defines the low usage threshold of Prod resources
+	ProdLowThresholds ResourceThresholds
 
 	// ResourceWeights indicates the weights of resources.
 	// The weights of resources are both 1 by default.
@@ -97,11 +104,17 @@ type LowNodeLoadNodePool struct {
 	// A resource consumption above (resp. below) this window is considered as overutilization (resp. underutilization).
 	UseDeviationThresholds bool
 
-	// HighThresholds defines the target usage threshold of resources
+	// HighThresholds defines the target usage threshold of node resources
 	HighThresholds ResourceThresholds
 
-	// LowThresholds defines the low usage threshold of resources
+	// LowThresholds defines the low usage threshold of node resources
 	LowThresholds ResourceThresholds
+
+	// ProdHighThresholds defines the target usage threshold of Prod resources
+	ProdHighThresholds ResourceThresholds `json:"prodHighThresholds,omitempty"`
+
+	// ProdLowThresholds defines the low usage threshold of Prod resources
+	ProdLowThresholds ResourceThresholds `json:"prodLowThresholds,omitempty"`
 
 	// ResourceWeights indicates the weights of resources.
 	// The weights of resources are both 1 by default.
--- a/pkg/descheduler/framework/plugins/loadaware/low_node_load.go
+++ b/pkg/descheduler/framework/plugins/loadaware/low_node_load.go
@@ -56,6 +56,7 @@ type LowNodeLoad struct {
 	nodeMetricLister     koordslolisters.NodeMetricLister
 	args                 *deschedulerconfig.LowNodeLoadArgs
 	nodeAnomalyDetectors *gocache.Cache
+	prodAnomalyDetectors *gocache.Cache
 }
 
 // NewLowNodeLoad builds plugin from its arguments while passing a handle
@@ -107,6 +108,7 @@ func NewLowNodeLoad(args runtime.Object, handle framework.Handle) (framework.Plu
 	koordSharedInformerFactory.WaitForCacheSync(context.TODO().Done())
 
 	nodeAnomalyDetectors := gocache.New(loadLoadUtilizationArgs.DetectorCacheTimeout.Duration, loadLoadUtilizationArgs.DetectorCacheTimeout.Duration)
+	prodAnomalyDetectors := gocache.New(loadLoadUtilizationArgs.DetectorCacheTimeout.Duration, loadLoadUtilizationArgs.DetectorCacheTimeout.Duration)
 
 	return &LowNodeLoad{
 		handle:               handle,
@@ -114,6 +116,7 @@ func NewLowNodeLoad(args runtime.Object, handle framework.Handle) (framework.Plu
 		args:                 loadLoadUtilizationArgs,
 		podFilter:            podFilter,
 		nodeAnomalyDetectors: nodeAnomalyDetectors,
+		prodAnomalyDetectors: prodAnomalyDetectors,
 	}, nil
 }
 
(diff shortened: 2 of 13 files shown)


pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +320,97 @@ func evictPodsFromSourceNodes(
...
+		podFilter, nodeIndexer, continueEviction, evictionReasonGenerator)
 
-	keysAndValues := []interface{}{
+	prodTotalAvailableUsages := newAvailableUsage(resourceNames)
+	for _, resourceName := range resourceNames {
+		if quantity, ok := prodTotalAvailableUsages[resourceName]; ok {
+			if _, totalOk := prodAvailableUsages[resourceName]; totalOk {
+				quantity.Add(*prodAvailableUsages[resourceName])
+			}
+			if _, prodOk := prodBothAvailableUsage[resourceName]; prodOk {
```
songtao98, line 357: `prodBothAvailableUsage` rootly was calculated from the same nodes where `bothTotalAvailableUsage` was calculated, that is, `bothDestinationNodes`. If we migrate pods from `targetNodes` to `bothDestinationNodes` in the first `balancePods`, some `bothDestinationNodes` might changed and cannot be as a Destination, even if there is still space in `prodBothAvailableUsage`.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +320,97 @@ func evictPodsFromSourceNodes(
...
-				case corev1.ResourceMemory, corev1.ResourceEphemeralStorage, corev1.ResourceStorage:
-					quantity = resource.NewQuantity(0, resource.BinarySI)
-				default:
-					quantity = resource.NewQuantity(0, resource.DecimalSI)
-				}
-				totalAvailableUsages[resourceName] = quantity
+	totalAvailableUsages, targetNodes := targetAvailableUsage(destinationNodes, resourceNames, false)
+	prodAvailableUsages, prodTargetNodes := targetAvailableUsage(prodDestinationNodes, resourceNames, true)
+	bothTotalAvailableUsage, bothTotalNodes := targetAvailableUsage(bothDestinationNodes, resourceNames, false)
+	prodBothAvailableUsage, prodBothTotalNodes := targetAvailableUsage(bothDestinationNodes, resourceNames, true)
```
songtao98, line 326: We need to keep in mind that actually `bothTotalNodes` are same with `prodBothTotalNodes`.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -285,22 +419,54 @@ func evictPodsFromSourceNodes(
...
+				case corev1.ResourceMemory, corev1.ResourceEphemeralStorage, corev1.ResourceStorage:
+					quantity = resource.NewQuantity(0, resource.BinarySI)
+				default:
+					quantity = resource.NewQuantity(0, resource.DecimalSI)
+				}
+				totalAvailableUsages[resourceName] = quantity
+			}
+			if prod {
+				quantity.Add(*destinationNode.thresholds.prodHighResourceThreshold[resourceName])
+				quantity.Sub(*destinationNode.prodUsage[resourceName])
```
songtao98, line 454: The prod availble usage should be `min(prodHighThreshold - prodUsage, totalHighThreshold - totalUsage).` Otherwise a node may be overutilized on a total-dimension after prod pod migration.

songtao98 (commented):
There are several tiny coding style suggestions in your PR. However, we need to discuss about the available resources calculation logic again.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
songtao98: need to re-calculate `prodTotalAvailableUsages`

pkg/descheduler/framework/plugins/loadaware/low_node_load.go
```diff
@@ -161,45 +162,58 @@ func (pl *LowNodeLoad) processOneNodePool(ctx context.Context, nodePool *desched
...
 		klog.V(4).InfoS("No nodes are underutilized, nothing to do here, you might tune your thresholds further", "nodePool", nodePool.Name)
 		return nil
 	}
 
 	resetNodesAsNormal(lowNodes, pl.nodeAnomalyDetectors)
+	resetNodesAsNormal(prodLowNodes, pl.nodeAnomalyDetectors)
+	resetNodesAsNormal(bothLowNodes, pl.nodeAnomalyDetectors)
 
-	if len(lowNodes) <= int(pl.args.NumberOfNodes) {
+	allLowNodes := len(lowNodes) + len(prodLowNodes) + len(bothLowNodes)
```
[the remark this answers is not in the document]
zwForrest: Corresponding explanatory information has been added to parameter `NumberOfNodes`

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
zwForrest: The corresponding calculation method has been updated.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
zwForrest: ok

songtao98:
@zwForrest 
The PR's title would be better using `koord-descheduler: balance prod pods between nodes in LowNodeLoad`.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
zwForrest: The corresponding calculation method has been updated.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +312,115 @@ func evictPodsFromSourceNodes(
...
+	klog.V(4).InfoS("node pool availableUsage", "onlyNodeTotal", totalAvailableUsages, "onlyProdOnly", prodAvailableUsages,
+		"bothLowNodesTotal", bothTotalAvailableUsage, "bothLowProdTotal", prodBothAvailableUsage)
+
+	nodeTotalAvailableUsages := newAvailableUsage(resourceNames)
+	for _, resourceName := range resourceNames {
+		if quantity, ok := nodeTotalAvailableUsages[resourceName]; ok {
+			if _, totalOk := totalAvailableUsages[resourceName]; totalOk {
+				quantity.Add(*totalAvailableUsages[resourceName])
+			}
+			if _, prodOk := bothTotalAvailableUsage[resourceName]; prodOk {
```
songtao98, line 328: Why it is named prodOk? Maybe bothOk?

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +312,115 @@ func evictPodsFromSourceNodes(
...
+		podFilter, nodeIndexer, continueEviction, evictionReasonGenerator)
+
+	// bothLowNode will be used by nodeHigh and prodHigh nodes, needs sub resources used by pods on nodeHigh.
+	for _, resourceName := range resourceNames {
+		if quantity, ok := nodeTotalAvailableUsages[resourceName]; ok {
+			if bothTotalAvailableUsage[resourceName].Cmp(*quantity) > 1 {
+				bothTotalAvailableUsage[resourceName].Sub(*quantity)
+			} else {
+				// all resource used by nodeHigh pods
+				bothTotalAvailableUsage[resourceName].Sub(*bothTotalAvailableUsage[resourceName])
```
songtao98, line 352: Is the logic here right? I think if bothTotalAvialbleUsage is bigger than nodeTotalAvailableUsage, it means that bothTotalAvailableUsage is not used. So nothing need to do here. This makes it Zero.

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +312,115 @@ func evictPodsFromSourceNodes(
...
-			quantity.Sub(*destinationNode.usage[resourceName])
 		}
 	}
 
-	keysAndValues := []interface{}{
+	prodTotalAvailableUsages := newAvailableUsage(resourceNames)
+	for _, resourceName := range resourceNames {
+		if quantity, ok := prodTotalAvailableUsages[resourceName]; ok {
+			if _, totalOk := prodAvailableUsages[resourceName]; totalOk {
+				quantity.Add(*prodAvailableUsages[resourceName])
```
songtao98, line 361: These variables' name is not clear, consider rename them? For example, totalOk is used to judge prodAvailableUsage?

pkg/descheduler/framework/plugins/loadaware/utilization_util.go
```diff
@@ -243,40 +312,114 @@ func evictPodsFromSourceNodes(
...
+	targetNodes = append(targetNodes, bothTotalNodes...)
+	balancePods(ctx, nodePoolName, sourceNodes, targetNodes, nodeTotalAvailableUsages, dryRun, nodeFit, false, resourceWeights, podEvictor,
+		podFilter, nodeIndexer, continueEviction, evictionReasonGenerator)
+
+	// bothLowNode will be used by nodeHigh and prodHigh nodes, needs sub resources used by pods on nodeHigh.
+	for _, resourceName := range resourceNames {
+		if quantity, ok := nodeTotalAvailableUsages[resourceName]; ok {
+			// A part of bothTotalAvailableUsage has been used,
+			// then the remaining part of nodeTotalAvailableUsage can be utilized.
+			if bothTotalAvailableUsage[resourceName].Cmp(*quantity) > 1 {
```
songtao98, line 350: I think it should be `if bothTotalAvailableUsage[resourceName].Cmp(*quantity) > 0`. Check the definition of Cmp function?

songtao98 (changes requested):

pkg/descheduler/apis/config/types_loadaware.go
```diff
@@ -65,12 +66,18 @@ type LowNodeLoadArgs struct {
...
-	// LowThresholds defines the low usage threshold of resources
+	// LowThresholds defines the low usage threshold of node resources
 	LowThresholds ResourceThresholds
 
+	// ProdHighThresholds defines the target usage threshold of Prod resources
+	ProdHighThresholds ResourceThresholds
+
+	// ProdLowThresholds defines the low usage threshold of Prod resources
+	ProdLowThresholds ResourceThresholds
+
```
hormes, line 80: After adding the new threshold, is the goal of the descheduling to satisfy both requirements at the same time? In other words, are they in an AND relationship?

songtao98: ps: If `ProdLowThresholds` is nil, set as `MinResourcePercentage` or `MaxResourcePercentage` in `newThreshold` function. This makes sure all nodes are under-utilized so nothing will do by this plugin. And the same logic is used to `LowThreshold`, too.

hormes (approved):

PR #2067, opened 2024-05-27
koordlet: fix nri reconnect params

saintube:

- Fix the koordlet NRI plugin parameters which may leak during the reconnections.
- Fix a flaky test `TestNriServer_Start` in the koordlet NRI module.
- Add arguments for the koordlet NRI plugin.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/nri/server.go
+++ b/pkg/koordlet/runtimehooks/nri/server.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/containerd/nri/pkg/api"
 	"github.com/containerd/nri/pkg/stub"
+	"go.uber.org/atomic"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


pkg/koordlet/runtimehooks/nri/server.go
```diff
@@ -248,13 +261,17 @@ func (p *NriServer) RemovePodSandbox(pod *api.PodSandbox) error {
 func (p *NriServer) onClose() {
 	//TODO: consider the pod status during restart
 	retryFunc := func() (bool, error) {
-		stub, err := stub.New(p, append(opts, stub.WithOnClose(p.onClose))...)
+		if p.stopped.Load() { // if set to stopped, no longer reconnect
```
kangclzjc, line 264: So, currently, nri server never stop, right ?

saintube: Yes. Currently, this method is used by the testing.

PR #2070, opened 2024-05-28
manager: only sync label when label key exists

ZiMengSheng:
### V. Checklist


pkg/slo-controller/noderesource/plugins/gpudeviceresource/plugin.go
```diff
@@ -183,13 +183,17 @@ func (p *Plugin) calculate(node *corev1.Node, device *schedulingv1alpha1.Device)
 	klog.V(5).InfoS("calculate gpu resources", "node", node.Name, "resources", gpuResources)
 
 	// calculate labels about gpu driver and model
-	if device.Labels != nil {
+	updatedLabels := map[string]string{}
+	if gpuModel, ok := device.Labels[extension.LabelGPUModel]; ok {
```
saintube, line 187: Does the controller need to handle the case where the node labels should be removed?

ZiMengSheng: No, doesn't need to.

FillZpp (approved):
/lgtm

PR #2071, opened 2024-05-28
koordlet: reduce unnecessary call of Err() for better performance

yangfeiyu20102011:
image

func AddSeries is called frequently, it's within multiple loops.
it.Err() will take a lot of time when the data is huge, we'd better reduce calling it.


saintube:
/lgtm
PTAL @zwzhang0107

saintube (approved):

PR #2076, opened 2024-05-31
koord-scheduler: add a parameter to mark whether scheduling is allowed on node with expired nodemetric

lucming:
add `enableScheduleWhenNodeMetricsExpired` parameter for `loadaware` scheduler plugin to mark whether scheduling is allowed on node with expired nodemetric.   

it be used to address the following issues：
1. not allow scheduling when `nodemetric` expired：avoid pods being scheduled to nodes with high load.
2. allow scheduling when `nodemetric` expired：the actual load of the node is very low but the nodemetric is very high. It may be blocked by the `filter` phase,  causing the node to be unable to schedule a new pod anymore. FYI:  https://github.com/koordinator-sh/koordinator/pull/1563

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/v1beta3/types.go
+++ b/pkg/scheduler/apis/config/v1beta3/types.go
@@ -36,6 +36,8 @@ type LoadAwareSchedulingArgs struct {
 	// When NodeMetrics expired, the node is considered abnormal.
 	// Default is 180 seconds.
 	NodeMetricExpirationSeconds *int64 `json:"nodeMetricExpirationSeconds,omitempty"`
+	// EnableScheduleWhenNodeMetricsExpired Indicates whether nodes with expired nodeMetrics are allowed to schedule pods.
+	EnableScheduleWhenNodeMetricsExpired *bool `json:"enableScheduleWhenNodeMetricsExpired,omitempty"`
 	// ResourceWeights indicates the weights of resources.
 	// The weights of CPU and Memory are both 1 by default.
 	ResourceWeights map[corev1.ResourceName]int64 `json:"resourceWeights,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -143,6 +144,9 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 
 	if p.args.FilterExpiredNodeMetrics != nil && *p.args.FilterExpiredNodeMetrics &&
 		p.args.NodeMetricExpirationSeconds != nil && isNodeMetricExpired(nodeMetric, *p.args.NodeMetricExpirationSeconds) {
+		if p.args.EnableScheduleWhenNodeMetricsExpired != nil && !*p.args.EnableScheduleWhenNodeMetricsExpired {
```
saintube, line 147: To avoid regression, could you please add a UT for the new argument?

pkg/scheduler/apis/config/v1beta3/types.go
```diff
@@ -36,6 +36,8 @@ type LoadAwareSchedulingArgs struct {
 	// When NodeMetrics expired, the node is considered abnormal.
 	// Default is 180 seconds.
 	NodeMetricExpirationSeconds *int64 `json:"nodeMetricExpirationSeconds,omitempty"`
+	// EnableScheduleWhenNodeMetricsExpired Indicates whether nodes with expired nodeMetrics are allowed to schedule pods.
+	EnableScheduleWhenNodeMetricsExpired *bool `json:"enableScheduleWhenNodeMetricsExpired,omitempty"`
```
saintube, line 40: shall we add an argument for the scoring phase?

saintube (commented):
PTAL cc @ZiMengSheng

pkg/scheduler/apis/config/v1beta3/types.go
lucming: the socre phase will returns 0 when nodeMetric expired, it still has a chance to schedule, so it doesn't seem to have much of an effect.

saintube: OK

PR #2085, opened 2024-06-06
scheduler: refine reservation msg if matched filtered out by other plugins

saintube:

- Add a prefix to define the Reservation-level error reasons when scheduling failed.
- Append the failure reasons when other plugins filter out a pod specifying reservation affinities.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin_test.go
+++ b/pkg/scheduler/plugins/reservation/plugin_test.go
@@ -1834,10 +1834,18 @@ func TestFilterReservation(t *testing.T) {
 }
 
 func TestPostFilter(t *testing.T) {
+	testFilterStatus := framework.NewStatus(framework.Unschedulable, "node(s) didn't match the requested node name")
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/reservation/plugin_test.go
```diff
@@ -1867,41 +1871,88 @@ func TestPostFilter(t *testing.T) {
...
+				filteredNodeStatusMap: framework.NodeToStatusMap{
+					"test-node-0": {},
+					"test-node-1": testFilterStatus,
+				},
+			},
+			want: nil,
+			want1: framework.NewStatus(framework.Unschedulable,
+				"3 Reservation(s) didn't match affinity rules",
+				"1 Reservation(s) is unschedulable",
+				"2 Reservation(s) by node node(s) didn't match the requested node name",
```
zwzhang0107, line 1954: 格式微调下：
2 Reservation(s) for node reason that node(s) didn't match the requested node name

saintube: done

PR #2100, opened 2024-06-12
koord-scheduler: Pod updates should not update the timestamp

zwForrest:
Pod updates should not update the timestamp in loadaware cache. This would cause pods that have been in the cache for a while to estimate resource during scheduling instead of using the actual resource reported on nodemetric. This discrepancy arises from the timestamp change without timely updating of nodemetric.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/loadaware/pod_assign_cache.go
+++ b/pkg/scheduler/plugins/loadaware/pod_assign_cache.go
@@ -61,9 +61,14 @@ func (p *podAssignCache) assign(nodeName string, pod *corev1.Pod) {
 		m = make(map[types.UID]*podAssignInfo)
 		p.podInfoItems[nodeName] = m
 	}
-	m[pod.UID] = &podAssignInfo{
-		timestamp: timeNowFn(),
-		pod:       pod,
+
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -61,9 +61,14 @@ func (p *podAssignCache) assign(nodeName string, pod *corev1.Pod) {
 		m = make(map[types.UID]*podAssignInfo)
 		p.podInfoItems[nodeName] = m
 	}
-	m[pod.UID] = &podAssignInfo{
-		timestamp: timeNowFn(),
-		pod:       pod,
+
+	if _, ok := m[pod.UID]; ok {
```
saintube, line 65: Thanks for your contributions. Could you please add a UT for this case to avoid regression?

zwForrest: Ok

saintube (commented):
/lgtm
PTAL /cc @ZiMengSheng

PR #2112, opened 2024-06-19
koordlet: add allpods reconciler func

kangclzjc:
Add Allpodslevel reconciler function in case some runtime hooks may need to consume all pods at once. It's part of the https://github.com/koordinator-sh/koordinator/pull/1974.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/reconciler/reconciler.go
+++ b/pkg/koordlet/runtimehooks/reconciler/reconciler.go
@@ -447,32 +447,33 @@ func (c *reconciler) reconcilePodCgroup(stopCh <-chan struct{}) {
 			}
 
 			for _, r := range globalCgroupReconcilers.allPodsLevel {
-				var reconcileFn reconcileFunc4AllPods
 				currentPods := make([]protocol.HooksProtocol, 0)
+				fns := make(map[string]reconcileFunc4AllPods)
 				for _, podMeta := range podsMeta {
-					if fn, ok := r.fn4AllPods[r.filter.Filter(podMeta)]; ok {
+					key := r.filter.Filter(podMeta)
+					if fn, ok := r.fn4AllPods[key]; ok {
 						podCtx := protocol.HooksProtocolBuilder.Pod(podMeta)
 						currentPods = append(currentPods, podCtx)
-						if reconcileFn == nil {
-							reconcileFn = fn
-						}
+						fns[key] = fn
 					}
 				}
 
 				if len(currentPods) > 0 {
-					if reconcileFn == nil {
+					if len(fns) == 0 {
 						klog.V(5).Infof("calling reconcile function %v aborted, condition %s not registered",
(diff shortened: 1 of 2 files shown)


kangclzjc:
I will add test later.

kangclzjc:
@saintube PTAL.

pkg/koordlet/runtimehooks/reconciler/reconciler.go
```diff
@@ -385,6 +445,34 @@ func (c *reconciler) reconcilePodCgroup(stopCh <-chan struct{}) {
...
+
+			for _, r := range globalCgroupReconcilers.allPodsLevel {
+				var reconcileFn reconcileFunc4AllPods
+				currentPods := make([]protocol.HooksProtocol, 0)
+				for _, podMeta := range podsMeta {
+					if fn, ok := r.fn4AllPods[r.filter.Filter(podMeta)]; ok {
+						podCtx := protocol.HooksProtocolBuilder.Pod(podMeta)
+						currentPods = append(currentPods, podCtx)
+						if reconcileFn == nil {
+							reconcileFn = fn
```
saintube, line 457: It seems the last reconcileFn would be picked. Why not multiple funcs?

kangclzjc: because there only exist one reconcileFn for each item in globalCgroupReconcilers.allPodsLevel. Here we want to get the reconcileFn and all pods.
```
allPodsLevel:          map[string]*cgroupReconciler{},
...
fn4AllPods  map[string]reconcileFunc4AllPods
```

saintube: The definition above seems to map the reconcileFunc4AllPods by the condition type. So it could be multiple reconcileFns for all condition types. If the reconcileFns is multiple, the next pod should not rely on the condition type of the previous pods. If the reconcileFns is single, the reconcileFn can be retrieved outside the pod loop.

kangclzjc: Yes, you're right. There could be muliple reconcileFns.

pkg/koordlet/runtimehooks/reconciler/reconciler.go
```diff
@@ -385,6 +445,35 @@ func (c *reconciler) reconcilePodCgroup(stopCh <-chan struct{}) {
 					}
 				}
 			}
+
+			for _, r := range globalCgroupReconcilers.allPodsLevel {
+				currentPods := make([]protocol.HooksProtocol, 0)
+				fns := make(map[string]reconcileFunc4AllPods)
```
kangclzjc, line 451: @saintube , I add a map here to store condition => reconcileFn and add a test too, PTAL.

saintube:
PTAL /cc @zwzhang0107 @hormes

PR #2128, opened 2024-07-08
koord-manager: support manually configured default cpu normalization settings

yangfeiyu20102011:
In some dedicated clusters, we don't need complex CPU models; instead, we prefer manually configured default  CPU normalization settings for greater flexibility.

### V. Checklist

### DIFF
--- a/apis/configuration/slo_controller_config.go
+++ b/apis/configuration/slo_controller_config.go
@@ -275,6 +275,8 @@ type CPUNormalizationStrategy struct {
 	// Enable defines whether the cpu normalization is enabled.
 	// If set to false, the node cpu normalization ratio will be removed.
 	Enable *bool `json:"enable,omitempty"`
+	// DefaultRatio defines the default cpu normalization.
+	DefaultRatio *float64 `json:"defaultRatio,omitempty"`
 	// RatioModel defines the cpu normalization ratio of each CPU model.
 	// It maps the CPUModel of BasicInfo into the ratios.
 	RatioModel map[string]ModelRatioCfg `json:"ratioModel,omitempty"`
--- a/pkg/slo-controller/noderesource/plugins/cpunormalization/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/cpunormalization/plugin.go
@@ -174,10 +174,11 @@ func (p *Plugin) Calculate(_ *configuration.ColocationStrategy, node *corev1.Nod
 		return nil, fmt.Errorf("failed to get CPUBasicInfo in cpu normalization calculation, err: info is missing")
 	}
 
-	ratio, err := getCPUNormalizationRatioFromModel(basicInfo, strategy)
+	ratio, err := getCPUNormalizationRatio(basicInfo, strategy)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get ratio in cpu normalization calculation, err: %s", err)
 	}
+
 	if err = isCPUNormalizationRatioValid(ratio); err != nil {
 		return nil, fmt.Errorf("failed to validate ratio in cpu normalization calculation, ratio %v, err: %s", ratio, err)
 	}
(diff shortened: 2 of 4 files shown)


apis/configuration/slo_controller_config.go
```diff
@@ -275,6 +275,9 @@ type CPUNormalizationStrategy struct {
 	// Enable defines whether the cpu normalization is enabled.
 	// If set to false, the node cpu normalization ratio will be removed.
 	Enable *bool `json:"enable,omitempty"`
+	// Ratio defines the cpu normalization.
+	// If Ratio is set, directly use it.
+	Ratio *float64 `json:"ratio,omitempty"`
```
saintube, line 280: To not break the design of the cpu normalization, perhaps it is a `DefaultRatio` when the RatioModel is missing or the cpu model is unknown.

yangfeiyu20102011: In some cases, there are some dedacated node pools in a cluster. Pods will exhibit their own performance characteristics, and we can set different ratios to meet their respective needs.
The default ratio may not meet the above requirements.

saintube: It seems the requirements. For your dedicated node pools, you can set the defaultRatio with nodeSelector targeting the node pool's labels. The main difference between the `defaultRatio` and the `ratio` of this commit is that the `RatioModels` take precedence over the `defaultRatio to keep the "cpu normalization" as the goal. In contrast, the `ratio` is just a special field to bypass the models.

yangfeiyu20102011: > It seems the requirements can be satisfied. For your dedicated node pools, you can set the defaultRatio with nodeSelector targeting the node pool's labels. The main difference between the `defaultRatio` and the `ratio` of this commit is that the `RatioModels` take precedence over the `defaultRatio` to keep the "cpu normalization" as the goal. In contrast, the `ratio` is just a special field to bypass the models.

Good idea, I misunderstood your meaning at first. I have changed it to `DefaultRatio`. 
When  the RatioModel is missing or the cpu model is unknown, use the `DefaultRatio`. 
 thanks. cc @saintube

pkg/slo-controller/noderesource/plugins/cpunormalization/plugin.go
```diff
@@ -176,8 +176,14 @@ func (p *Plugin) Calculate(_ *configuration.ColocationStrategy, node *corev1.Nod
 
 	ratio, err := getCPUNormalizationRatioFromModel(basicInfo, strategy)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get ratio in cpu normalization calculation, err: %s", err)
+		if strategy.DefaultRatio != nil {
```
saintube, line 179: recommend to move the default logic inside the function getCPUNormalizationRatio

pkg/slo-controller/noderesource/plugins/cpunormalization/plugin.go
```diff
@@ -174,10 +174,11 @@ func (p *Plugin) Calculate(_ *configuration.ColocationStrategy, node *corev1.Nod
 		return nil, fmt.Errorf("failed to get CPUBasicInfo in cpu normalization calculation, err: info is missing")
 	}
 
-	ratio, err := getCPUNormalizationRatioFromModel(basicInfo, strategy)
+	ratio, err := getCPUNormalizationRatio(basicInfo, strategy)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get ratio in cpu normalization calculation, err: %s", err)
+		return nil, fmt.Errorf("failed to get cpu normalization ratio for node %s, err: %w", node.Name, err)
```
saintube, line 179: No need to wrap the nodeName in the error again, since the framework can print it.

saintube (commented):
/lgtm

yangfeiyu20102011:
It still needs an approval. Thanks.
cc @saintube @FillZpp

saintube:
PTAL /cc @hormes

saintube (approved):
/lgtm

PR #2136, opened 2024-07-17
scheduler: optimize when only 1 reservation/node

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/reservation/nominator.go
```diff
@@ -152,6 +152,10 @@ func (pl *Plugin) NominateReservation(ctx context.Context, cycleState *framework
 		return nil, framework.AsStatus(fmt.Errorf("not implemented frameworkext.FrameworkExtender"))
 	}
 
+	if len(reservationInfos) == 1 && state.hasAffinity {
```
saintube, line 155: move up to L145?

PR #2137, opened 2024-07-17
all: support priority and preemption policy transformer

saintube:

Support features PriorityTransformer and PreemptionPolicyTransformer:

- PriorityTransformer: transform the priority value of the pod according to its priority class, allowing us to adapt the pods existing in the cluster that have not set priorities.
- PreemptionPolicyTransformer: transform the preemption policy of the pod according to a preemption label and the global default policy, allowing us to take over the pods existing in the cluster that may set the default preemptible policy. It helps when the cluster has many kinds of pods set the PreemptLowPriority by default priority classes but we want to enable the preemption only for specific pods.

### V. Checklist

### DIFF
--- a/apis/extension/preemption.go
+++ b/apis/extension/preemption.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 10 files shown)


apis/extension/preemption.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	// LabelPodPreemptionPolicy is used to set the PreemptionPolicy of a Pod when the feature
+	// PreemptionPolicyTransformer is enabled.
+	LabelPodPreemptionPolicy = SchedulingDomainPrefix + "/preemption-policy"
+	// LabelPodSetPreemptionPolicy determines whether the pod should set the DefaultPreemptionPolicy override
+	// the pod.spec.preemptionPolicy when the feature PreemptionPolicyTransformer enabled and the pod does not set
+	// the LabelPodPreemptionPolicy.
+	// Default use true when this label is not set.
+	LabelPodSetPreemptionPolicy = SchedulingDomainPrefix + "/set-preemption-policy"
```
hormes, line 29: As discussed, this switch is not actually needed. When we need to remove the transformer logic, we can just make a global configuration change for the scheduler.

saintube: fixed

PR #2141, opened 2024-07-25
scheduler: elastic quota ignore terminating pod

shaloulcy:

pod is deleted gracefully, it may be stuck long time. some user want ignore the pods when using elastic quota

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
+++ b/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
```diff
@@ -1054,3 +1084,19 @@ func (gqm *GroupQuotaManager) doUpdateOneGroupSharedWeightNoLock(quotaName strin
 
 	gqm.updateOneGroupSharedWeightNoLock(quotaInfo)
 }
+
+func isTerminatingPod(pod *v1.Pod) bool {
```
hormes, line 1088: It is strange to add feature-gate in isTerminatingPod. Why not change the name of this function to shouldBeIgnored and also change the previous related functions.

shaloulcy: fix it

saintube (approved):
/lgtm

PR #2149, opened 2024-08-01
koord-manager: fix invalid assignments in the resource amplification …

yangfeiyu20102011:
NodeResourceAmplificationPlugin func handleUpdate

In the else branch, the assignment to **originalResources** does not take effect in the subsequent process, which might be a typographical error.
It will print  "admit for node-x, skip resource cpu as it is not found in original resources",
and resource will not be amplified.
This result is unexpected.

This PR fixes the issue and adds some useful information.

### V. Checklist


saintube:
PTAL /cc @zqzten

zqzten:
Yeah that's a bug I think. /cc @hormes can make a double check on that.

Also, this case is not covered in test now, it would be nice if you can also add a test for it to ensure the fix.

hormes (approved):

PR #2153, opened 2024-08-05
koordlet: add some information to improve log readability

yangfeiyu20102011:
A module may have multiple hooks, and the current logs cannot distinguish which specific hook was registered or invoked.
It is essential to add some necessary descriptive information.
image

### V. Checklist


jasonliu747 (approved):
/lgtm

PR #2159, opened 2024-08-06
【WIP】koordlet: perform operations only under the specified cpuburst strategy and improve config logging

yangfeiyu20102011:
image
Under the specified cpuBurstOnly strategy, the plugin cpuburst still periodically updates the CFS values, which is not expected. Determine the corresponding cpuburst strategy at the beginning and then perform the appropriate actions.
image

### V. Checklist

### DIFF
--- a/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
+++ b/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
@@ -241,15 +241,21 @@ func (b *cpuBurst) start() {
 		// merge burst config from pod and node
 		cpuBurstCfg := genPodBurstConfig(podMeta.Pod, &b.nodeCPUBurstStrategy.CPUBurstConfig)
 		if cpuBurstCfg == nil {
-			klog.Warningf("pod %v/%v burst config illegal, burst config %v",
-				podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
+			klog.Warningf("pod %v/%v burst config illegal, because burst config is nil",
+				podMeta.Pod.Namespace, podMeta.Pod.Name)
 			continue
 		}
-		klog.V(5).Infof("get pod %v/%v cpu burst config: %v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
-		// set cpu.cfs_burst_us for pod and containers
-		b.applyCPUBurst(cpuBurstCfg, podMeta)
-		// scale cpu.cfs_quota_us for pod and containers
-		b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)
+		klog.V(5).Infof("get pod %v/%v cpu burst config: %#v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
+
+		if cpuBurstEnabled(cpuBurstCfg.Policy) {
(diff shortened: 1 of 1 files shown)


pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
```diff
@@ -241,15 +241,21 @@ func (b *cpuBurst) start() {
...
 			continue
 		}
-		klog.V(5).Infof("get pod %v/%v cpu burst config: %v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
-		// set cpu.cfs_burst_us for pod and containers
-		b.applyCPUBurst(cpuBurstCfg, podMeta)
-		// scale cpu.cfs_quota_us for pod and containers
-		b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)
+		klog.V(5).Infof("get pod %v/%v cpu burst config: %#v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
+
+		if cpuBurstEnabled(cpuBurstCfg.Policy) {
```
saintube, line 250: please note that the strategy should support reseting the related cgroups when the policy disables

yangfeiyu20102011: > please note that the strategy should support reseting the related cgroups when the policy disables

When both CPUNormalization and CPUBurst are enabled, CFS will be repeatedly changed by the plugins.
How about adding a flag variable, when `CPUBurstPolicy` changed，call `b.applyCPUBurst(cpuBurstCfg, podMeta)` and `b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)` as before. @saintube 

```
func (b *cpuBurst) start(){
...
var flag bool
if nodeSLO.Spec.CPUBurstStrategy changed {
flag = true
}
...
if cpuBurstEnabled(cpuBurstCfg.Policy) || flag {
			// set cpu.cfs_burst_us for pod and containers
			b.applyCPUBurst(cpuBurstCfg, podMeta)
}

if cfsQuotaBurstEnabled(cpuBurstCfg.Policy) || flag {
			// scale cpu.cfs_quota_us for pod and containers
			b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)
}

}
```

saintube: Maybe we can add a new policy `"ignored"` to describe the behavior that stops the cgroups reconciliation, so the existing policies are still backward-compatible.

yangfeiyu20102011: > Maybe we can add a new policy `"ignored"` to describe the behavior that stops the cgroups reconciliation, so the existing policies are still backward-compatible.

But someone may still want to use `cpuBurstOnly`,  `applyCPUBurst()` is also needed. @saintube

saintube: Yes. The user may switch the policy from `auto` to `cpuBurstOnly`. He/She should reset the cfs_quota. 
Adding a separate attribute is a good idea not to break the backward compatibility of the current policies. Possibly, we can define a new field in the CPUBurstStrategy to adjust the configurations dynamically.

PR #2160, opened 2024-08-06
koord-descheduler: fixes namespace object limiter

songtao98:
Currently, if user uses evictAllBarePods to evict a bare pod, the namespace object limiter cannot get its ownerRef to add a limiterKey for this pod's namespace. This lead to the namespace limiter for bare pods does not work.

This PR fixes namespace object limiter for koord-descheduler's MigrationController. 

### V. Checklist


pkg/descheduler/controllers/migration/controller.go
```diff
@@ -899,7 +904,7 @@ func (r *Reconciler) track(limit rate.Limit, limiterKey, processScope string, li
 	}
 	limiter := limiters[limiterKey]
 	if limiter == nil {
-		limiter = rate.NewLimiter(limit, maxMigratingReplicas)
+		limiter = rate.NewLimiter(limit, 1)
```
hormes, line 907: Why set burst to 1?

songtao98: If maxMigratingReplicas is set more than 1, suppose it is set as 2, in 5 minutes. The limiter at the beginning will have 2 tokens, then descheduler can evict 2 pods at the same time and of course cost these 2 tokens. Then at the 150th seconds from the beginning(i.e. 2.5min later), the limiter will generate 1 token, then descheduler can evict another Pod. If now you check "how many pods has been evicted in that 5 minutes", you will find the result is 3. That's exceeds the maxMigratingReplicas.

So I changed the burst to 1. Every 2.5 minutes the limiter generate 1 token. Pods will be evicted one by one and evenly through the timewindow. 

The disadvantage is that if there are not to many pods to be evicted, it will make the eviction slower. Suppose there are only 2 pods to be evicted in 5 minutes, they cannot be evicted together. But in my opinion it's acceptable and better than exceeding the maxMigratingReplicas.

songtao98: I added a new arg `Burst` into `objectLimiter` struct in the newest version of this PR. This allows users to config the number of evictions that can exceed the limit within a time window. The default burst is set as 1 which means **strictly** limited.

PR #2168, opened 2024-08-14
koord-descheduler: support node selector for each descheduler profile

songtao98:
The `DeschedulerConfiguration` supports users to config multiple profiles with different plugin enabled with different pluginConfigs. There are user cases such as to config `HighNodeUtilization` plugin with different threshold in different group of nodes, e.g. nodepools. However, most of the plugins only supports to filter pods through specific filters.

When dealing with nodes, it is the descheduler framework to decide which nodes can be plugins' checking target through a global `nodeSelector`. This PR enables configuration of `nodeSelector` further into each profile. Note that the descheduler framework will first select nodes at a global level through the outer `nodeSelector`, then select nodes according to each profile's inside `nodeSelector`. 

### V. Checklist

### DIFF
--- a/pkg/descheduler/apis/config/types.go
+++ b/pkg/descheduler/apis/config/types.go
@@ -77,6 +77,7 @@ type DeschedulerProfile struct {
 	Name         string
 	PluginConfig []PluginConfig
 	Plugins      *Plugins
+	NodeSelector *metav1.LabelSelector
 }
 
 type Plugins struct {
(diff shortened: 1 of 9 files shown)


pkg/descheduler/apis/config/types.go
```diff
@@ -77,6 +77,7 @@ type DeschedulerProfile struct {
 	Name         string
 	PluginConfig []PluginConfig
 	Plugins      *Plugins
+	NodeSelector *metav1.LabelSelector
```
hormes, line 80: Shouldn't this be a plugin configuration?

songtao98: Generally speaking Yes. And actually I recommend plugins to support configuration of multiple `NodeSelector` like what `LowNodeLoad` do with `NodePool`. This PR is intended to solve the kubernetes descheduler plugin adaption problem for multiple node configuration.

songtao98: This will be fixed by #2200

PR #2169, opened 2024-08-15
scheduler: remove deepCopy of reservationInfo

ZiMengSheng:

就是为了去掉这个，但是不想加锁，所以这样操作下
image

### V. Checklist


pkg/scheduler/plugins/reservation/snapshot.go
```diff
@@ -0,0 +1,94 @@
+package reservation
```
saintube, line 1: missing a license header

PR #2193, opened 2024-09-04
koordlet: fix wrong msg in calculateBESuppressCPU

yangfeiyu20102011:
Printed information does not meet expectations.
In the CPU normalization scenario, Allocatable is much greater than Capacity.
According to the logic here, the printed information should be Capacity.

### V. Checklist


saintube (approved):
/lgtm

PR #2206, opened 2024-09-12
koordlet: support change pod CPUQOS by annotations

j4ckstraw:
### V. Checklist

### DIFF
--- a/apis/slo/v1alpha1/pod.go
+++ b/apis/slo/v1alpha1/pod.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -26,6 +27,8 @@ import (
 
 const (
 	AnnotationPodCPUBurst = apiext.DomainPrefix + "cpuBurst"
+
+	AnnotationPodCPUQoS = apiext.DomainPrefix + "cpuQOS"
 
 	AnnotationPodMemoryQoS = apiext.DomainPrefix + "memoryQOS"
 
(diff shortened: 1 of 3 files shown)


zwzhang0107:
seems we need some UT for this

apis/slo/v1alpha1/pod.go
```diff
@@ -27,6 +27,8 @@ import (
 const (
 	AnnotationPodCPUBurst = apiext.DomainPrefix + "cpuBurst"
 
+	AnnotationPodCPUQoS = apiext.DomainPrefix + "CPUQOS"
```
zwzhang0107, line 30: apiext.DomainPrefix + "cpuQOS"

pkg/koordlet/runtimehooks/hooks/groupidentity/interceptor.go
```diff
@@ -17,14 +17,39 @@ limitations under the License.
...
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 
 	ext "github.com/koordinator-sh/koordinator/apis/extension"
+	slov1alpha1 "github.com/koordinator-sh/koordinator/apis/slo/v1alpha1"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/runtimehooks/protocol"
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/util"
 )
 
+func getPodCPUQoSConfig(annotations map[string]string) (*slov1alpha1.CPUQOSCfg, error) {
```
zwzhang0107, line 32: better to mv this func to api/slo/v1aplha1/pod.go

pkg/koordlet/runtimehooks/hooks/groupidentity/interceptor.go
```diff
@@ -35,6 +60,27 @@ func (b *bvtPlugin) SetPodBvtValue(p protocol.HooksProtocol) error {
 	podQOS := ext.GetQoSClassByAttrs(req.Labels, req.Annotations)
 	podKubeQOS := util.GetKubeQoSByCgroupParent(req.CgroupParent)
 	podBvt := r.getPodBvtValue(podQOS, podKubeQOS)
+
+	// pod annotatations take precedence to the default CPUQoS
+	cfg, err := getPodCPUQoSConfig(req.Annotations)
```
zwzhang0107, line 65: better to use DefaultCPUQOS() as default and merge with annotation, besides, cfg.Enable == nil means disabled by default

pkg/koordlet/runtimehooks/hooks/groupidentity/interceptor.go
j4ckstraw: this function is different to exist functions like `GetPodMemoryQoSConfig`, which receive parameter from annotation, so I define it in local file. If we move it to `apis/slo/v1aplha1/pod.go`, how about named `GetPodCPUQoSConfigFromAnnotations`?

j4ckstraw: or `GetPodCPUQoSConfigByAttrs`?

Issue #2184, opened 2024-08-30
[proposal] Reservation Affinity supports the reservation name


saintube:
**What is your proposal**:

Some use cases of the Resource Reservation regard the reservation objects as nodes, thus bringing the need for the pod to set the reservation name like the nodeName to schedule to the target reservation. In this proposal, we want to support that the pod can set the reservation name in the annotation of the reservation affinity to allocate the specific reservation object. Note that the scheduler should perform the admission when the pod requests unmatched resources.

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

saintube:
linked to #2186

PR #2207, opened 2024-09-18
scheduler: support reservation name, taints and tolerations

### LINKED ISSUE #2183: [proposal] Reservation supports taints and tolerations
**What is your proposal**:

Some use cases of the Resource Reservation regard the reservation objects as nodes, thus bringing the need to apply taints to the reservations. In this proposal, we want to support that the reservation can set taints where only pods with corresponding tolerations can allocate their resources.

**Why is this needed**:

saintube:

- Support the reservation name and taints & tolerations in the reservation scheduling.

fixes #2183, #2184.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -25,6 +25,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -174,6 +175,7 @@ type stateData struct {
 type schedulingStateData struct {
 	preemptLock          sync.RWMutex
 	hasAffinity          bool
+	reservationName      string
 	podRequests          corev1.ResourceList
 	podRequestsResources *framework.Resource
 	preemptible          map[string]corev1.ResourceList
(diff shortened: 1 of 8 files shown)


pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -634,6 +696,12 @@ func (pl *Plugin) makePostFilterReasons(state *stateData, filteredNodeStatusMap
 	// to prepare the exact messages by ourselves.
 	var reasons []string
 	var b strings.Builder
+	if nameUnmatched > 0 {
+		b.WriteString(strconv.Itoa(nameUnmatched))
+		b.WriteString(" Reservation(s) didn't match the requested reservation name")
```
zwzhang0107, line 701: will this message get cut due to out of length

zwzhang0107:
add some comments that PreferNoSchedule/NoExecute is not supported yet for reservation.toleration

pkg/util/reservation/reservation.go
```diff
@@ -566,11 +599,20 @@ func GetReservationRestrictedResources(allocatableResources []corev1.ResourceNam
...
 
 // IsReservationReason checks if the error reason is at the reservation-level.
 func IsReservationReason(reason string) bool {
 	return strings.HasPrefix(reason, ErrReasonPrefix)
 }
+
+// DoNotScheduleTaintsFilter can filter out the node taints that reject scheduling Pod on a Node.
+func DoNotScheduleTaintsFilter(t *corev1.Taint) bool {
+	// PodToleratesNodeTaints is only interested in NoSchedule and NoExecute taints.
+	return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
```
zwzhang0107, line 617: PreferNoSchedule?

pkg/scheduler/plugins/reservation/plugin.go
saintube: yes, however the msg probably not longer than the max length 1024

pkg/util/reservation/reservation.go
saintube: comments added

PR #2237, opened 2024-10-25
webhook: Support setting quota admission to zero for special usage scenarios, such as temporarily pausing pod submissions.

TaoYang526:

Currently admission will be set to max when the parsed value of it is zero, here including 3 situations:
(1) not configured.
(2) configured with no content.
(3) configured with specified-zero resources.
Situation (1) and (2) are suitable to use max since we better have a default limitation for pods. But for Situation (3), we have a usage scenario that new pod submissions should be pausing temporarily and keep running pods going, the most simplest approach is to set admission to zero without updating min/max (may be unable to recover it later for some verifications like checkMinQuotaSum). And I think it's reasonable to use configured admission even if it's zero.

If possible, we can change the condition that chooses to use max for admission, from "admission is zero" to "admission is empty", to exclude the situation (3) which may be depended for some special usage scenarios.

Dose that make sense? Hope to hear your thoughts!

UT

### V. Checklist

### DIFF
--- a/pkg/webhook/pod/validating/evaluate_quota_test.go
+++ b/pkg/webhook/pod/validating/evaluate_quota_test.go
@@ -250,6 +250,82 @@ func TestEvalucateQuota(t *testing.T) {
 			wantErr:     true,
 			wantUsed:    corev1.ResourceList{},
 		},
+		{
+			name:      "admission not set, use max by default",
+			operation: admissionv1.Create,
+			newPod: elasticquota.MakePod("ns1", "pod1").Label("quota.scheduling.koordinator.sh/name", "quota1").
+				Container(corev1.ResourceList{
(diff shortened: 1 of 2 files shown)


shaloulcy:
lgtm

saintube (approved):
/lgtm

TaoYang526:
Thanks @shaloulcy @saintube for the review.
Just update the branch by rebasing it from main to fix the conflict.

TaoYang526:
@saintube Could you please help to merge this PR if no more changes are needed?

saintube:
PTAL /cc @zwzhang0107 @hormes

saintube (approved):

TaoYang526:
Thanks @saintube for the review and commit!

PR #2246, opened 2024-11-04
scheduler: gang scheduling add closeHistoryEvaluate annotation

LY-today:
**What is your proposal**:
Add user wishes comments to the logic in the gang policy to prevent the gang policy from directly invalidating after the pod is destroyed and rebuilt.
`if gang.getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied && gang.isGangOnceResourceSatisfied() {
		return nil
	}`

**Why is this needed**:
The gang policy immediately becomes invalid after the pod is destroyed and rebuilt. The newly created pod becomes running even when the total number of pods does not meet the MinRequiredNumber. However, some users think this does not meet their expectations. They believe that the semantics of the gang policy are still followed in this scenario. , so it is recommended to add user intention judgment logic

Annotation can be added to the podGroup, for example: gang.scheduling.koordinator.sh/close-history-evaluate. At this time, the rebuilt pod is destroyed. When the total number of pods does not meet the MinRequiredNumber, it is still pending, maintaining gang semantics.

Checklist

- [ ✅] I have written necessary docs and comments
- [ ✅] I have added necessary unit tests and integration tests
- [ ✅] All checks passed in `make test`

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -283,8 +283,9 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, state *framework.Cy
 		return fmt.Errorf("gang has not init, gangName: %v, podName: %v", gang.Name,
 			util.GetId(pod.Namespace, pod.Name))
 	}
+
 	// resourceSatisfied means pod will directly pass the PreFilter
-	if gang.getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied && gang.isGangOnceResourceSatisfied() {
+	if gang.getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied && gang.isGangOnceResourceSatisfied() && !gang.CloseHistoryEvaluate {
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


LY-today:
@saintube PR has been submitted, please check

pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -283,8 +283,9 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, state *framework.Cy
 		return fmt.Errorf("gang has not init, gangName: %v, podName: %v", gang.Name,
 			util.GetId(pod.Namespace, pod.Name))
 	}
+
 	// resourceSatisfied means pod will directly pass the PreFilter
-	if gang.getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied && gang.isGangOnceResourceSatisfied() {
+	if gang.getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied && gang.isGangOnceResourceSatisfied() && !gang.CloseHistoryEvaluate {
```
saintube, line 288: what is the difference between the CloseHistoryEvaluate and the GangMatchPolicyWaitingAndRunning?

LY-today: Can GangMatchPolicyWaitingAndRunning solve the above problem?

saintube: FYI: https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220901-gang-scheduling.md#gang-match-policy

PR #2250, opened 2024-11-05
api: supply id for minor meaningless device

ferris-cx:

The RDMA device is mounted inside the container based on the scheduling assignment result.

The scheduling framework already supports joint allocation of Gpus and RDMA devices, but several semantics need to be tested in practice, including preference and samePCIE. In order to allow RDMA devices to access the container, we need to fix the lack of BDF addresses in the results assigned by the scheduling algorithm to RDMA devices

1. In the k8s cluster, prepare one or more servers that support rdma network adapters as cluster nodes. Install the new version of the koordlet component, koord-manager, and the revamped multus-cni on each node, and check the node status. The number of resources is displayed as the actual number of RDMA nics on the node.

2. Write a pod.YAML to apply for RDMA network card resources, kubectl apply-f pod. On pod note (device-allocated), check whether the rdma device bdf address (busID field) is included in the rdma allocation result, and check the accuracy.

3. When the pod is in the running state, enter the container and run the ifconfig command to check the number of network cards. If the case is correct, multiple network card lists will be output, such as net1, net2, net3, and so on.

1. Deploy components that support RDMA, including koordlet, koord-manager, koord-scheduler, and multus-cni.

2. Due to the complete end-to-end passthrough, multi-CNI plug-in is also required, which complies with CNI specifications and supports multi-NIC allocation. Assigning the RDMA network card PF/VF to the Pod mentioned here requires the device ID to be injected into the component, otherwise it will not work. This change will be maintained separately in the multi-cni project or another PR

### V. Checklist

### DIFF
--- a/apis/extension/device_share.go
+++ b/apis/extension/device_share.go
@@ -80,14 +80,32 @@ const (
         "koordinator.sh/gpu-memory": "16Gi"
       }
     }
+  ],
+  "rdma": [
+    {
+      "minor": 0,
+      "id": "0000:09:00.0",
+      "resources": {
+        "koordinator.sh/rdma": 100,
+      }
+    },
+    {
+      "minor": 1,
+      "id": "0000:10:00.0",
+      "resources": {
+        "koordinator.sh/rdma": 100,
+      }
+    }
   ]
 }
 */
 type DeviceAllocations map[schedulingv1alpha1.DeviceType][]*DeviceAllocation
 
 type DeviceAllocation struct {
-	Minor     int32                      `json:"minor"`
-	Resources corev1.ResourceList        `json:"resources"`
+	Minor     int32               `json:"minor"`
+	Resources corev1.ResourceList `json:"resources"`
+	// ID is the well known identifier for device, because for some device, such as rdma, Minor is meaningless
+	ID        string                     `json:"id,omitempty"`
 	Extension *DeviceAllocationExtension `json:"extension,omitempty"`
 }
 
--- a/pkg/scheduler/plugins/deviceshare/plugin.go
+++ b/pkg/scheduler/plugins/deviceshare/plugin.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	k8sfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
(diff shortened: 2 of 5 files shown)


pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -545,12 +546,43 @@ func (p *Plugin) preBindObject(ctx context.Context, cycleState *framework.CycleS
 		return nil
 	}
 
+	var deviceAllocs *apiext.DeviceAllocations
```
ZiMengSheng, line 549: can we consider add move this logic to Reserve() phase?

apis/extension/device_share.go
```diff
@@ -89,6 +89,7 @@ type DeviceAllocation struct {
 	Minor     int32                      `json:"minor"`
 	Resources corev1.ResourceList        `json:"resources"`
 	Extension *DeviceAllocationExtension `json:"extension,omitempty"`
+	BusID     string                     `json:"busID,omitempty"`
```
ZiMengSheng, line 92: I prefer define it as this:
```
type DeviceAllocation struct {
	Minor     int32                      `json:"minor"`
	Resources corev1.ResourceList        `json:"resources"`
	Extension *DeviceAllocationExtension `json:"extension,omitempty"`
	Topology  *DeviceTopology            `json:"topology,omitempty"`
}

type DeviceTopology struct {
	// BusID is the domain:bus:device.function formatted identifier of PCI/PCIE device
	BusID string `json:"busID,omitempty"`
}
```

ZiMengSheng, line 92: I prefer define it as this:
```
type DeviceAllocation struct {
	Minor     int32                      `json:"minor"`
	Resources corev1.ResourceList        `json:"resources"`
	Extension *DeviceAllocationExtension `json:"extension,omitempty"`
	Topology  *DeviceTopology            `json:"topology,omitempty"`
}

type DeviceTopology struct {
	// BusID is the domain:bus:device.function formatted identifier of PCI/PCIE device
	BusID string `json:"busID,omitempty"`
}
```

ZiMengSheng, line 92: I prefer define it as this:
```
type DeviceAllocation struct {
	Minor     int32                      `json:"minor"`
	Resources corev1.ResourceList        `json:"resources"`
	Extension *DeviceAllocationExtension `json:"extension,omitempty"`
	Topology  *DeviceTopology            `json:"topology,omitempty"`
}

type DeviceTopology struct {
	// BusID is the domain:bus:device.function formatted identifier of PCI/PCIE device
	BusID string `json:"busID,omitempty"`
}
```

ZiMengSheng: and use ID store BDF

ferris-cx: OK！BusId is one of the attributes of the topology structure and can be encapsulated in the topology structure！

ferris-cx: Was the ID used for something else?

ZiMengSheng: Ok，cancel this

pkg/scheduler/plugins/deviceshare/plugin.go
ferris-cx: ok

apis/extension/device_share.go
```diff
@@ -63,21 +63,19 @@ const (
 // DeviceAllocations would be injected into Pod as form of annotation during Pre-bind stage.
 /*
 {
-  "gpu": [
+  "rdma": [
```
songtao98, line 66: How about explain both gpu and rdma here and in the proposal? As a plain reader I might be confused what this annotation will look like when device type is gpu.

ZiMengSheng: Nice Suggesstion

PR #2253, opened 2024-11-07
koord-manager: consider NodeReserved when calculate mid resource.

tan90github:
…n calculate mid resource.

Just like batch resources, we should consider SafetyMargin, NodeReserved, and SystemUsed when calculating mid resources.
```
Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
Unallocated[Mid] = max(NodeCapacity - NodeSafetyMargin - SystemUsed - Allocated[Prod], 0)
```
ref node resource model:
https://github.com/koordinator-sh/koordinator/blob/main/docs/images/node-resource-model.png

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/plugins/midresource/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/midresource/plugin.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
@@ -126,9 +127,9 @@ func (p *Plugin) degradeCalculate(node *corev1.Node, message string) []framework
 	return p.Reset(node, message)
 }
 
-// Unallocated[Mid] = max(NodeAllocatable - Allocated[Prod], 0)
-func (p *Plugin) getUnallocated(node *corev1.Node, podList *corev1.PodList) corev1.ResourceList {
-	allocated := corev1.ResourceList{}
+// Unallocated[Mid] = max(NodeCapacity - NodeReserved - Allocated[Prod], 0)
+func (p *Plugin) getUnallocated(nodeName string, podList *corev1.PodList, nodeCapacity, nodeReserved corev1.ResourceList) corev1.ResourceList {
+	prodPodAllocated := corev1.ResourceList{}
 	for i := range podList.Items {
 		pod := &podList.Items[i]
 		priorityClass := extension.GetPodPriorityClassWithDefault(pod)
@@ -142,19 +143,27 @@ func (p *Plugin) getUnallocated(node *corev1.Node, podList *corev1.PodList) core
 			continue
 		}
 		podRequest := util.GetPodRequest(pod, corev1.ResourceCPU, corev1.ResourceMemory)
-		allocated = quotav1.Add(allocated, podRequest)
+		prodPodAllocated = quotav1.Add(prodPodAllocated, podRequest)
 	}
 
-	return quotav1.SubtractWithNonNegativeResult(node.Status.Allocatable, allocated)
+	midUnallocated := quotav1.Max(quotav1.Subtract(quotav1.Subtract(nodeCapacity, nodeReserved), prodPodAllocated), util.NewZeroResourceList())
+	cpuMsg := fmt.Sprintf("midUnallocatedCPU[core]:%v = max(nodeCapacity:%v - nodeReserved:%v - prodPodAllocated:%v, 0)",
+		midUnallocated.Cpu(), nodeCapacity.Cpu(), nodeReserved.Cpu(), prodPodAllocated.Cpu())
+	memMsg := fmt.Sprintf("midUnallocatedMem[GB]:%v = max(nodeCapacity:%v - nodeReserved:%v - prodPodAllocated:%v, 0)",
+		midUnallocated.Memory().ScaledValue(resource.Giga), nodeCapacity.Memory().ScaledValue(resource.Giga),
+		nodeReserved.Memory().ScaledValue(resource.Giga), prodPodAllocated.Memory().ScaledValue(resource.Giga))
+
+	klog.V(6).Infof("calculated mid unallocated for node %s, cpu(core) %v, memory(GB) %v", nodeName, cpuMsg, memMsg)
+	return midUnallocated
 }
 
 func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *corev1.Node, podList *corev1.PodList,
 	resourceMetrics *framework.ResourceMetrics) []framework.ResourceItem {
 	// Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
-	// Unallocated[Mid] = max(NodeAllocatable - Allocated[Prod], 0)
+	// Unallocated[Mid] = max(NodeCapacity - NodeReserved - Allocated[Prod], 0)
 
-	var allocatableMilliCPU, allocatableMemory, prodReclaimableMilliCPU int64
-	var prodReclaimableMemory string = "0"
+	var allocatableMilliCPU, allocatableMemory int64
+	prodReclaimableCPU, prodReclaimableMemory := resource.NewQuantity(0, resource.DecimalSI), resource.NewQuantity(0, resource.BinarySI)
 	prodReclaimableMetic := resourceMetrics.NodeMetric.Status.ProdReclaimableMetric
 
 	if prodReclaimableMetic == nil || prodReclaimableMetic.Resource.ResourceList == nil {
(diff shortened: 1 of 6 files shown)


tan90github:
PTAL @saintube thanks!

saintube:
> ```
> Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
> Unallocated[Mid] = max(NodeCapacity - NodeSafetyMargin - NodeReserved - SystemUsed - Allocated[Prod], 0)
> ```

@tan90github Is it necessary to substitute the SystemUsed for the Unallocated[Mid]? The calculation of the mid-reclaimable should have included the prediction of the system usage.

tan90github:
- When mid-reclaimable is not zero, it indeed includes system usage, so I shouldn't consider systemUsed.This is my mistake.
- But when mid-reclaimable is zero, I think we should take into account these system occupations. 
- By the way, it seems that the systemUsed used in calculating midReclaimable doesn't include hostApplication  currently. How should we consider it? The same issue is also present in nodereserved.

What's your opinion? @saintube thx.

saintube:
> * When mid-reclaimable is not zero, it indeed includes system usage, so I shouldn't consider systemUsed.This is my mistake.
> * But when mid-reclaimable is zero, I think we should take into account these system occupations.
> * By the way, it seems that the systemUsed used in calculating midReclaimable doesn't include hostApplication  currently. How should we consider it? The same issue is also present in nodereserved.
> 
> What's your opinion? @saintube thx.

@tan90github Yes, it is indeed an issue that the Mid reclaimable may be capped to zero and then ignore the system usage. IMO, we can take the system usage into account by substituting the max(systemUsed, nodeReserved).
Overall, we should derive the "unallocated" from the capacity (excluding the reserved) and the request, detouring the usage, since it is the duty of the reclaimable to concisely estimate and predict the unused allocatable. And double-substituting the usage can make the calculation of the Mid-tier confusing.

tan90github:
> Yes, it is indeed an issue that the Mid reclaimable may be capped to zero and then ignore the system usage. IMO, we can take the system usage into account by substituting the max(systemUsed, nodeReserved).
> Overall, we should derive the "unallocated" from the capacity (excluding the reserved) and the request, detouring the usage, since it is the duty of the reclaimable to concisely estimate and predict the unused allocatable. And double-substituting the usage can make the calculation of the Mid-tier confusing.

```
unallocated = nodeCapacity - max(systemUsed, nodeReserved) - sum(request[prod])
```
@saintube  is that right?

saintube:
> ```
> unallocated = nodeCapacity - max(systemUsed, nodeReserved) - sum(request[prod])
> ```
> 
> @saintube is that right?

@tan90github Yep.

saintube:
> I noticed that there is a `prediction-safety-margin-percent`parameter in prodReclaimable to control the safety threshold, and there is also a `midUnallocatedPercent` in Unallocated to control it, thus the midAllocatable resources are safe enough. Therefore, I did not consider NodeSafetyMargin when calculating Unallocated. Now the calculation formula for mid resources is as follows:
> 
> ```
> // Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
> // Unallocated[Mid] = max(NodeCapacity - NodeReserved - Allocated[Prod], 0)
> ```
> 
> cc @saintube

@tan90github Agreed.

tan90github:
> > I noticed that there is a `prediction-safety-margin-percent`parameter in prodReclaimable to control the safety threshold, and there is also a `midUnallocatedPercent` in Unallocated to control it, thus the midAllocatable resources are safe enough. Therefore, I did not consider NodeSafetyMargin when calculating Unallocated. Now the calculation formula for mid resources is as follows:
> > ```
> > // Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
> > // Unallocated[Mid] = max(NodeCapacity - NodeReserved - Allocated[Prod], 0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > cc @saintube
> 
> @tan90github Agreed.

I modified the code, PTAL@saintube thx.

pkg/slo-controller/noderesource/plugins/util/util.go
```diff
@@ -270,8 +275,17 @@ func GetNodeCapacity(node *corev1.Node) corev1.ResourceList {
 
 // GetNodeSafetyMargin gets node-level safe-guarding reservation with the node's allocatable
 func GetNodeSafetyMargin(strategy *configuration.ColocationStrategy, nodeCapacity corev1.ResourceList) corev1.ResourceList {
-	cpuReserveQuant := util.MultiplyMilliQuant(nodeCapacity[corev1.ResourceCPU], getReserveRatio(*strategy.CPUReclaimThresholdPercent))
-	memReserveQuant := util.MultiplyQuant(nodeCapacity[corev1.ResourceMemory], getReserveRatio(*strategy.MemoryReclaimThresholdPercent))
+	defaultstrategy := sloconfig.DefaultColocationStrategy()
```
saintube, line 278: is it necessary to merge the default here?

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
```diff
@@ -163,18 +172,34 @@ func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *cor
...
+		prodReclaimableCPU = prodReclaimable.Cpu()
+		prodReclaimableMemory = prodReclaimable.Memory()
+		allocatableMilliCPU = prodReclaimableCPU.MilliValue()
+		allocatableMemory = prodReclaimableMemory.Value()
 	}
 
-	nodeAllocatable := node.Status.Allocatable
+	nodeMetric := resourceMetrics.NodeMetric
+
+	hostAppHPUsed := resutil.GetHostAppHPUsed(resourceMetrics)
```
saintube, line 183: To calculate the allocatable of the Mid-tier, it should ignore the host used of the Mid priority.

pkg/slo-controller/noderesource/plugins/util/util.go
tan90github: @saintube The reason for this change is that when I run the test unit `Test_getNodeReservation`, if `CPUReclaimThresholdPercent` or `MemoryReclaimThresholdPercent` not configured, the code will panic since `CPUReclaimThresholdPercent` is nil. I will add the relevant test unit later.

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
tan90github: fixed

pkg/slo-controller/noderesource/plugins/util/util.go
saintube: Perhaps the default config should be prepared in the UT instead of the main function.

pkg/slo-controller/noderesource/plugins/util/util.go
```diff
@@ -181,11 +187,12 @@ func GetPodMetricUsage(info *slov1alpha1.PodMetricInfo) corev1.ResourceList {
 	return GetResourceListForCPUAndMemory(info.PodUsage.ResourceList)
 }
 
-func GetHostAppHPUsed(resourceMetrics *framework.ResourceMetrics) corev1.ResourceList {
+func GetHostAppHPUsed(resourceMetrics *framework.ResourceMetrics, pluginName string) corev1.ResourceList {
 	hostAppHPUsed := util.NewZeroResourceList()
 	for _, hostAppMetric := range resourceMetrics.NodeMetric.Status.HostApplicationMetric {
-		if hostAppMetric.Priority == extension.PriorityBatch || hostAppMetric.Priority == extension.PriorityFree {
-			// only consider higher priority usage for batch allocatable
+		if (pluginName == midPluginName && hostAppMetric.Priority == extension.PriorityMid) ||
```
saintube, line 193: it is not recommended to hard-coding the plugin name in the util func

saintube: we can simply compare if the priority class is higher

pkg/slo-controller/noderesource/plugins/util/util.go
tan90github: fixed.

pkg/slo-controller/noderesource/plugins/util/util.go
tan90github: fixed.

tan90github:
Thanks for your reviews, @saintube . All problems have been fixed.

saintube (commented):
/lgtm
PTAL /cc @zwzhang0107

tan90github:
PTAL thx. @zwzhang0107 @hormes cc @saintube

tan90github:
Can we merge this PR now? I forgot to ask about it in yesterday's bi-weekly meeting. The PR has been LGTM for a while. Current mid-resource calculation logic is flawed (it doesn't take into account things like nodeReserved), and this PR addresses that issue, making the mid-resource calculation complete. If there are any further suggestions about this PR,  feel free to let me know. @saintube @zwzhang0107 thx.

PR #2273, opened 2024-11-18
koordlet: collect metric for host application memory usage with page cache

j4ckstraw:
### Ⅰ. Describe what this PR does

collect host application memory usage with page cache.

rely on https://github.com/koordinator-sh/koordinator/pull/2278 

### V. Checklist

- [x] I have written necessary docs and comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `make test`

PR #2278, opened 2024-11-22
koordlet: add GetNodeMetricSpec in statesInformer interface

j4ckstraw:
add func GetNodeMetricSpec, so we can get nodeMetricSpec conveniently by statesInformer interface

add GetNodeMetricSpec in statesInformer interface
prepare for https://github.com/koordinator-sh/koordinator/pull/2273

### V. Checklist

### DIFF
--- a/pkg/koordlet/prediction/prediction_test.go
+++ b/pkg/koordlet/prediction/prediction_test.go
@@ -43,7 +43,12 @@ func (m *mockStatesInformer) HasSynced() bool {
 func (m *mockStatesInformer) GetNode() *v1.Node {
 	return m.node
 }
+
 func (m *mockStatesInformer) GetNodeSLO() *slov1alpha1.NodeSLO {
+	return nil
+}
+
+func (m *mockStatesInformer) GetNodeMetricSpec() *slov1alpha1.NodeMetricSpec {
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


saintube (commented):
/lgtm

j4ckstraw:
> @j4ckstraw BTW, in which cases does the state informer need to export the NodeMetric? As it is an aggregated result, we usually fetch the metric resource from the metric cache instead.

To support memoryCollectPolicy in https://github.com/koordinator-sh/koordinator/pull/2273#issuecomment-2485534706

saintube:
> > @j4ckstraw BTW, in which cases does the state informer need to export the NodeMetric? As it is an aggregated result, we usually fetch the metric resource from the metric cache instead.
> 
> To support memoryCollectPolicy in [#2273 (comment)](https://github.com/koordinator-sh/koordinator/pull/2273#issuecomment-2485534706)

@j4ckstraw This usage could be problematic. The metric advisor is the node metrics' data resource, so its collection should not rely on the node metrics to avoid circular reference.

j4ckstraw:
> > > @j4ckstraw BTW, in which cases does the state informer need to export the NodeMetric? As it is an aggregated result, we usually fetch the metric resource from the metric cache instead.
> > 
> > 
> > To support memoryCollectPolicy in [#2273 (comment)](https://github.com/koordinator-sh/koordinator/pull/2273#issuecomment-2485534706)
> 
> @j4ckstraw This usage could be problematic. The metric advisor is the node metrics' data resource, so its collection should avoid depending on the node metrics due to circular reference. Perhaps, we can introduce the `GetNodeMetricSpec` to prevent misuse.

OK, I'll change the pr.

j4ckstraw:
PTAL @saintube

saintube (approved):

PR #2282, opened 2024-11-28
koordlet: fix reconciler description mismatch.

tan90github:
Different conditions within the same reconciler correspond to different descriptions, similar to fn, but the current description has not been updated according to the condition; this PR fixes that issue.

### V. Checklist


tan90github:
PTAL @saintube thx.

saintube (approved):
/lgtm

PR #2289, opened 2024-12-05
proposal: min-excess control for quota

TaoYang526:

We have leveraged multi-level quotas capability of elastic-quota to manage products and their tenants in our cluster, there is a requirement to support a mixed payment type for some products: tenants can reserve specified resource they want, meanwhile they are allowed to use more resource on demand, both reserved and on-demand resource need to be limited on the product and tenant levels. For example, tenant1 has 100 reserved CU, if he use 120 CU actually, 100 CU is reserved resource and 20 CU is on-demand resource, we can control the limitation of reserved resource via min concept of elastic-quota, but there's no approach to control the limitation of on-demand resource for now.

Therefore, I propose to introduce a new concept "min-excess" to limit on-demand resource usage under control, "min-excess" represents the limitation of excess part out of min, meanwhile we need to record the "min-excess-used" resource. For the leaf quotas, "min-excess-used" resource equals (used - min), and for parent quotas, it is the sum of the "min-excess-used" resource from all the leaf quotas belong to them.

Take a usage scenario for example:
```
|- product  min: 300, max: 500, used: 300, min-excess: 100, min-excess-used: 100
   |- tenant1  min: 100, max: 300, used: 150, min-excess-used: 50
   |- tenant2  min: 100, max: 300, used: 150, min-excess-used: 50
   |- tenant3  min: 100, max: 300, used: 0
```
Both tenant1 and tenant2 have utilized 50 units of on-demand resource, resulting in an actual usage of 100 units for the product-level on-demand resource. With the min-excess=100 configuration on the product quota, now there's a cap on the on-demand resource usage at the product quota level, tenant1 and tenant2 won't get more resource, meanwhile tenant3 can still start workloads within the reserved resource.

This feature is disabled by default, with almost no resource usage, and can be enabled by the --feature-gates argument.

### V. Checklist

### DIFF
--- a/apis/extension/elastic_quota.go
+++ b/apis/extension/elastic_quota.go
@@ -55,6 +55,7 @@ const (
 	AnnotationNonPreemptibleRequest = QuotaKoordinatorPrefix + "/non-preemptible-request"
 	AnnotationNonPreemptibleUsed    = QuotaKoordinatorPrefix + "/non-preemptible-used"
 	AnnotationAdmission             = QuotaKoordinatorPrefix + "/admission"
+	AnnotationMinExcess             = QuotaKoordinatorPrefix + "/min-excess"
 )
 
 func GetParentQuotaName(quota *v1alpha1.ElasticQuota) string {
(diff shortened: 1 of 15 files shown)


TaoYang526:
Closed this since CustomLimiter interface is a better solution according to offline discussions. Details please refer to #2363.

PR #2291, opened 2024-12-09
slo-controller: fix mid resource calculate formula

lijunxin559:

When calculating *Allocatable[mid]* resources, due to possible oversold, *ProdReclaimableMetric* will be greater than *NodeAllocatable* * *thresholdRatio*, so the calculated *Allocatable[mid]* value accidentally includes the oversold part. 
In particular, when *thresholdRatio = 1*, *Allocatable[mid] = NodeAllocatable*, which is not the original intention of the koordinator resource model design.

I suggest taking the current free resources of this node into account when calculating Allocatable[mid], that is, *Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio, FreeResourceOnNode) + Unallocated[Mid] * midUnallocatedRatio*

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/plugins/midresource/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/midresource/plugin.go
@@ -159,14 +159,14 @@ func (p *Plugin) getUnallocated(nodeName string, podList *corev1.PodList, nodeCa
 
 func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *corev1.Node, podList *corev1.PodList,
 	resourceMetrics *framework.ResourceMetrics) []framework.ResourceItem {
-	// Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
+	// Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio, NodeUnused) + Unallocated[Mid] * midUnallocatedRatio
 	// Unallocated[Mid] = max(NodeCapacity - NodeReserved - Allocated[Prod], 0)
 
 	var allocatableMilliCPU, allocatableMemory int64
 	prodReclaimableCPU, prodReclaimableMemory := resource.NewQuantity(0, resource.DecimalSI), resource.NewQuantity(0, resource.BinarySI)
-	prodReclaimableMetic := resourceMetrics.NodeMetric.Status.ProdReclaimableMetric
+	prodReclaimableMetric := resourceMetrics.NodeMetric.Status.ProdReclaimableMetric
 
-	if prodReclaimableMetic == nil || prodReclaimableMetic.Resource.ResourceList == nil {
+	if prodReclaimableMetric == nil || prodReclaimableMetric.Resource.ResourceList == nil {
 		klog.V(4).Infof("no valid prod reclaimable, so use default zero value")
 		allocatableMilliCPU = 0
 		allocatableMemory = 0
@@ -185,7 +185,7 @@ func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *cor
 	nodeCapacity := resutil.GetNodeCapacity(node)
 
 	systemUsed := resutil.GetResourceListForCPUAndMemory(nodeMetric.Status.NodeMetric.SystemUsage.ResourceList)
-	// resource usage of host applications with prod priority will be count as host system usage since they consumes the
+	// resource usage of host applications with prod priority will be count as host system usage since they consume the
 	// node reserved resource.
 	systemUsed = quotav1.Add(systemUsed, hostAppHPUsed)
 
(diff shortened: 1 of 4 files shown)


saintube:
@lijunxin559 Thanks for your contribution. 

> Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio, FreeResourceOnNode) + Unallocated[Mid] * midUnallocatedRatio

I suppose the `FreeResourceOnNode` equals `Unallocated[Mid]`, so what does the formula mean? If the node has no unallocated Prod resources, the allocatable Mid in the new formula will be zero, which obeys the expectation of the Prod overcommitment.
Perhaps, to avoid the Prod allocatable being overcommitted twice (i.e. overcommit the unused overcommitted Prod requests), we need to minimize it with the ProdAllocated. WDYT?

lijunxin559:
> @lijunxin559 Thanks for your contribution.
> 
> > Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio, FreeResourceOnNode) + Unallocated[Mid] * midUnallocatedRatio
> 
> I suppose the `FreeResourceOnNode` equals `Unallocated[Mid]`, so what does the formula mean? If the node has no unallocated Prod resources, the allocatable Mid in the new formula will be zero, which obeys the expectation of the Prod overcommitment. To avoid the Prod allocatable being overcommitted twice (i.e. overcommit the unused overcommitted Prod requests) in the node resource controller, we need to minimize it with the ProdAllocated and Node capacity. And we'd better revise the new peak predictor in the koordlet, e.g. the PriorityPeakPredictor should minimize the node capacity as the limit of the peak. WDYT?

Morning, thanks for your reply ~ (^^)

*Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio* is  the original code logic. I think *Unallocated[Mid]* is intended to express the part of resources that are not allocated to prod. I just added a limit on the actual idleness of a node, which does not conflict with the original formula.

Yes, the predictor improvement you mentioned can also avoid secondary overselling to a certain extent, but I think the logic of dividing the predictor and the calculation of mid resources into two parts may be clearer. The predictor can focus on modeling and predicting the idle resources of the pod and ensure the reliability of the prediction, regardless of whether this part of the allocated resources actually exists or not. And the secondary utilization of this part of the idle resources is separated separately (mid and batch were also done before), so that so that the logic is decoupled and the subsequent use of the predictor function is more scalable. Right?

saintube:
> Morning, thanks for your reply ~ (^^)
> 
> _Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio_ is the original code logic. I think _Unallocated[Mid]_ is intended to express the part of resources that are not allocated to prod. I just added a limit on the actual idleness of a node, which does not conflict with the original formula.
> 
> Yes, the predictor improvement you mentioned can also avoid secondary overselling to a certain extent, but I think the logic of dividing the predictor and the calculation of mid resources into two parts may be clearer. The predictor can focus on modeling and predicting the idle resources of the pod and ensure the reliability of the prediction, regardless of whether this part of the allocated resources actually exists or not. And the secondary utilization of this part of the idle resources is separated separately (mid and batch were also done before), so that the logic is decoupled and the subsequent use of the predictor function is more scalable. Right?

@lijunxin559 OK. So the `FreeResourceOnNode` means the `NodeCapacity - NodeUsage`. It seems reasonable.

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
```diff
@@ -159,14 +159,14 @@ func (p *Plugin) getUnallocated(nodeName string, podList *corev1.PodList, nodeCa
 
 func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *corev1.Node, podList *corev1.PodList,
 	resourceMetrics *framework.ResourceMetrics) []framework.ResourceItem {
-	// Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio) + Unallocated[Mid] * midUnallocatedRatio
+	// Allocatable[Mid]' := min(Reclaimable[Mid], NodeAllocatable * thresholdRatio, FreeResourceOnNode) + Unallocated[Mid] * midUnallocatedRatio
```
saintube, line 162: `FreeResourceOnNode` is ambiguous. How about `NodeUnused` or barely `NodeCapacity - NodeUsed`?

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
```diff
@@ -219,3 +220,11 @@ func (p *Plugin) calculate(strategy *configuration.ColocationStrategy, node *cor
 		},
 	}
 }
+
+func getFreePhysicalResourceOnNode(node *corev1.Node, nodeMetrics *slov1alpha1.NodeMetric) *corev1.ResourceList {
+	// Capacity - nodeUsed
+	nodeCapacity := resutil.GetNodeCapacity(node)
+	nodeUsed := nodeMetrics.Status.NodeMetric.NodeUsage.ResourceList
```
saintube, line 227: need to check if the NodeUsage is valid

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
lijunxin559: You are right . NodeUnused is better.

pkg/slo-controller/noderesource/plugins/midresource/plugin.go
lijunxin559: Fixed 
I proposed to aggressively set nodeUnUsed to 0 when it is invalid, which is safer

pkg/slo-controller/noderesource/plugins/util/util.go
```diff
@@ -103,12 +103,15 @@ func CalculateBatchResourceByPolicy(strategy *configuration.ColocationStrategy,
 }
 
 func CalculateMidResourceByPolicy(strategy *configuration.ColocationStrategy, nodeCapacity, unallocated corev1.ResourceList, allocatableMilliCPU, allocatableMemory int64,
-	prodReclaimableCPU, prodReclaimableMemory *resource.Quantity, nodeName string) (*resource.Quantity, *resource.Quantity, string, string) {
+	prodReclaimableCPU, prodReclaimableMemory *resource.Quantity, nodeUnused *corev1.ResourceList, nodeName string) (*resource.Quantity, *resource.Quantity, string, string) {
```
saintube, line 106: `nodeUnused *corev1.ResourceList` -> `nodeUnused corev1.ResourceList`

saintube (approved):
/lgtm

jasonliu747 (approved):

PR #2315, opened 2025-01-07
scheduler: revise ReservationFilterPlugin and fix preempting pods resources

saintube:

Koord-Scheduler: 

1. Revise the ReservationFilterPlugin interface to make it extensible and provide more accurate filtering results. Please see the section "Special notes" for the details.
2. Support node reservation API in a Reservation object.
3. Fix a bug about counting reservation's "pods" resources during the preemption.

Rename the ReservationFilterPlugin method `FilterReservation` to `FilterNominateReservation` since only the NominateReservation calls it during the PreScore or the Reserve phase.

Re-define the method `FilterReservation` that is invoked in the Filter phase:

```go
type ReservationFilterPlugin interface {
	framework.Plugin
	FilterReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeInfo *framework.NodeInfo) *framework.Status
	FilterNominateReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeName string) *framework.Status
}
```

Compared with the current FilterReservation which runs in the PreScore or Reserve, the new method is invoked during the Filter phase of the reservation plugin. It should show more concise reasons when the scheduling fails, and it is helpful for internal extensions.

In this version, we keep the FilterNominateReservation and will move the current implementation to the Filter phase in the future.

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/interface.go
+++ b/pkg/scheduler/frameworkext/interface.go
@@ -64,7 +64,8 @@ type FrameworkExtender interface {
 	// DEPRECATED: use RunReservationExtensionRestoreReservation instead.
 	RunReservationExtensionFinalRestoreReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, states PluginToNodeReservationRestoreStates) *framework.Status
 
-	RunReservationFilterPlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeName string) *framework.Status
+	RunReservationFilterPlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeInfo *framework.NodeInfo) *framework.Status
+	RunNominateReservationFilterPlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeName string) *framework.Status
 	RunReservationScorePlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfos []*ReservationInfo, nodeName string) (PluginToReservationScores, *framework.Status)
 
 	RunNUMATopologyManagerAdmit(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeName string, numaNodes []int, policyType apiext.NUMATopologyPolicy, exclusivePolicy apiext.NumaTopologyExclusive, allNUMANodeStatus []apiext.NumaNodeStatus) *framework.Status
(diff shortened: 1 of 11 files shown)


pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -474,6 +479,14 @@ func (pl *Plugin) filterWithReservations(ctx context.Context, cycleState *framew
 	allInsufficientResourcesByNode := sets.NewString()
 	var allInsufficientResourceReasonsByReservation []string
 	for _, rInfo := range matchedReservations {
+		// Before nominating a reservation in PreScore or Reserve, check the reservation by multiple plugins to make
+		// the Filter phase give a more accurate result. It is extensible to support more policies.
+		status := extender.RunReservationFilterPlugins(ctx, cycleState, pod, rInfo, nodeInfo)
```
ZiMengSheng, line 484: Will we implement different RunReservationFilterPlugins in different scenarios? Because I see all currently inplementation just return nil.

pkg/scheduler/frameworkext/interface.go
```diff
@@ -64,7 +64,8 @@ type FrameworkExtender interface {
 	// DEPRECATED: use RunReservationExtensionRestoreReservation instead.
 	RunReservationExtensionFinalRestoreReservation(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, states PluginToNodeReservationRestoreStates) *framework.Status
 
-	RunReservationFilterPlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeName string) *framework.Status
+	RunReservationFilterPlugins(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, reservationInfo *ReservationInfo, nodeInfo *framework.NodeInfo) *framework.Status
```
ZiMengSheng, line 67: Whether we should modifiy some plugin filter to avoid aome redundant Reservation Filter Logic

pkg/scheduler/plugins/reservation/plugin.go
saintube: As described in the "Special Notes" section, the existing FilterReservation plugins will be refactored into the re-defined method. Currently, we leave the methods for internal extensions and future usage.

pkg/scheduler/frameworkext/interface.go
saintube: Yes, but in another PR.
For example, the core logic of the nodenumaresource plugin and the deviceshare plugin, i.e. tryAllocateFromReservation, includes both Filter and Allocate logic, so it would take more work if we want to refactor them cleanly.

pkg/scheduler/frameworkext/interface.go
```diff
@@ -120,10 +121,14 @@ type ReservationRestorePlugin interface {
 }
 
 // ReservationFilterPlugin is an interface for Filter Reservation plugins.
-// These plugins will be called during the Reserve phase to determine whether the Reservation can participate in the Reserve
+// FilterReservation will be called in the Filter phase for determining which reservations are available.
+// FilterNominateReservation will be called in the PreScore or the Reserve phase to nominate a reservation whether it
+// can participate the Reserve.
+// TODO: merge into one method.
```
ZiMengSheng, line 127: Looking forward a merged method.

hormes (approved):

PR #2323, opened 2025-01-17
webhook: support elasticquota enable update resource key

lijunxin559:

Currently, the consistency check for resource types in elastic quota's add, delete and update is insufficient and too rigid. For example, maxResourceKey requires parent-child consistency, which can lead to the following problems:
1. There is no unified logic to check the various resource types of *min/max*, which can easily lead to inconsistencies
2. Unable to flexibly support changing resource types, especially when adding new resources, limited by the initial creation status

However, considering the original intention of quota design and current computing logic, it does not affect related calculations while supporting resource type checks(also already added more unitTest cases). Among them:
1. The setting intentions of *min* and *max* are different and the calculation logic is not dependent. The resource types need to be checked separately
2. *Guaranteed* is affected by *min* calculation, so there is no need to check separately

I propose to add a feature called *ElasticQuotaEnableUpdateResourceKey* to support changes to ElasticQuota Resource Key, with the following modifications:
1. Support *min* and *max* resourceKey checks during elasticQuota add and update: *checkSubAndParentGroupQuotaKey()*
2. Support AnnotationSharedWeight updated: when SharedWeight has a key which is not included in max, delete it; when SharedWeight does not have the key of max, it defaults to be filled in the quantity of max
3. Maintain compatibility of resourceKeys update logic: *updateResourceKeyNoLock()* 
4. Propose standard process operations for elastic quota resourceKey: add resourceKey from parent node to child node, delete resourceKey from child node to parent node

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
+++ b/pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
@@ -563,22 +563,31 @@ func (gqm *GroupQuotaManager) updateOneGroupSharedWeightNoLock(quotaInfo *QuotaI
 	gqm.runtimeQuotaCalculatorMap[quotaInfo.ParentName].updateOneGroupSharedWeight(quotaInfo)
 }
 
+// updateResourceKeyNoLock based on quotaInfo.CalculateInfo.Max of self
+// Note: RootQuotaName need to be updated as allResourceKeys
 func (gqm *GroupQuotaManager) updateResourceKeyNoLock() {
 	// collect all dimensions
-	resourceKeys := make(map[v1.ResourceName]struct{})
+	allResourceKeys := make(map[v1.ResourceName]struct{})
 	for quotaName, quotaInfo := range gqm.quotaInfoMap {
-		if quotaName == extension.DefaultQuotaName || quotaName == extension.SystemQuotaName {
+		if quotaName == extension.RootQuotaName || quotaName == extension.DefaultQuotaName || quotaName == extension.SystemQuotaName {
 			continue
 		}
+		resourceKeys := make(map[v1.ResourceName]struct{})
 		for resName := range quotaInfo.CalculateInfo.Max {
+			allResourceKeys[resName] = struct{}{}
 			resourceKeys[resName] = struct{}{}
+		}
+		// update right now
+		if runtimeQuotaCalculator, ok := gqm.runtimeQuotaCalculatorMap[quotaName]; ok && runtimeQuotaCalculator != nil && !reflect.DeepEqual(resourceKeys, runtimeQuotaCalculator.resourceKeys) {
+			runtimeQuotaCalculator.updateResourceKeys(resourceKeys)
 		}
 	}
 
-	if !reflect.DeepEqual(resourceKeys, gqm.resourceKeys) {
-		gqm.resourceKeys = resourceKeys
-		for _, runtimeQuotaCalculator := range gqm.runtimeQuotaCalculatorMap {
-			runtimeQuotaCalculator.updateResourceKeys(resourceKeys)
+	// keep special ones same
+	if !reflect.DeepEqual(allResourceKeys, gqm.resourceKeys) {
+		gqm.resourceKeys = allResourceKeys
+		if runtimeQuotaCalculator, ok := gqm.runtimeQuotaCalculatorMap[extension.RootQuotaName]; ok && runtimeQuotaCalculator != nil {
+			runtimeQuotaCalculator.updateResourceKeys(allResourceKeys)
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


pkg/features/scheduler_features.go
```diff
@@ -82,6 +82,7 @@ var defaultSchedulerFeatureGates = map[featuregate.Feature]featuregate.FeatureSp
 	ElasticQuotaIgnoreTerminatingPod:          {Default: false, PreRelease: featuregate.Alpha},
 	ElasticQuotaImmediateIgnoreTerminatingPod: {Default: false, PreRelease: featuregate.Alpha},
 	ElasticQuotaGuaranteeUsage:                {Default: false, PreRelease: featuregate.Alpha},
+	ElasticQuotaEnableUpdateResourceKey:       {Default: false, PreRelease: featuregate.Alpha},
```
saintube, line 85: would you implement the feature in the scheduler?

lijunxin559: At present, the feature is mainly used to affect the verification behavior of elasticquota, and the scheduler does not use it. Should it be placed in a more appropriate position?

saintube: So the ElasticQuotaEnableUpdateResourceKey in the defaultSchedulerFeatureGates is redundant. Please add it after the scheduler's logic is implemented.

lijunxin559: Got it. The features of the scheduler need to have specific functional bindings.
That means we directly support the incremental update feature of elasticquota, right?

lijunxin559: deleted ElasticQuotaEnableUpdateResourceKey in defaultSchedulerFeatureGates

pkg/webhook/elasticquota/quota_topology_check.go
```diff
@@ -68,7 +68,7 @@ func (qt *quotaTopology) validateQuotaSelfItem(quota *v1alpha1.ElasticQuota) err
 }
 
 // validateQuotaTopology checks the quotaInfo's topology with its parent and its children.
-// oldQuotaInfo is null wben validate a new create request, and is the current quotaInfo when validate a update request.
+// oldQuotaInfo is null when validate a new create request, and is the current quotaInfo when validate a update request.
 func (qt *quotaTopology) validateQuotaTopology(oldQuotaInfo, newQuotaInfo *QuotaInfo, oldNamespaces []string) error {
```
shaloulcy, line 72: webhook需要加个逻辑，修改shared weight。假如修改了max，shared  weight里面没有这个key，需要添加进去

shaloulcy:
group_quota_manager.go里面的UpdateQuota、RefreshRuntime 需要加单侧覆盖这种情况

pkg/webhook/elasticquota/quota_topology_check.go
lijunxin559: Yes, you are right.
AnnotationSharedWeight updated logic has been fixed  in fillQuotaDefaultInformation(): When SharedWeight has a key with max that does not exist, delete it; when SharedWeight does not have the key of max, it defaults to filling in the quantity of max
and added in min/max key checked in validateQuotaSelfItem()

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
```diff
@@ -563,22 +563,34 @@ func (gqm *GroupQuotaManager) updateOneGroupSharedWeightNoLock(quotaInfo *QuotaI
...
 	}
 
-	if !reflect.DeepEqual(resourceKeys, gqm.resourceKeys) {
-		gqm.resourceKeys = resourceKeys
-		for _, runtimeQuotaCalculator := range gqm.runtimeQuotaCalculatorMap {
-			runtimeQuotaCalculator.updateResourceKeys(resourceKeys)
+	// keep special ones same
+	if !reflect.DeepEqual(allResourceKeys, gqm.resourceKeys) {
+		gqm.resourceKeys = allResourceKeys
+		quotas := []string{extension.RootQuotaName, extension.DefaultQuotaName, extension.SystemQuotaName}
```
shaloulcy, line 589: DefaultQuotaName/SystemQuotaName没有runtimeQuotaCalculator，不需要更新

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
```diff
@@ -563,22 +563,34 @@ func (gqm *GroupQuotaManager) updateOneGroupSharedWeightNoLock(quotaInfo *QuotaI
 	gqm.runtimeQuotaCalculatorMap[quotaInfo.ParentName].updateOneGroupSharedWeight(quotaInfo)
 }
 
+// updateResourceKeyNoLock based on quotaInfo.CalculateInfo.Max of self
+// Note: RootQuotaName need to be updated as allResourceKeys
 func (gqm *GroupQuotaManager) updateResourceKeyNoLock() {
```
shaloulcy, line 568: 这个函数要不补个单侧吧

lijunxin559: added

pkg/scheduler/plugins/elasticquota/core/group_quota_manager.go
lijunxin559: fixed

saintube (approved):
Good job! @lijunxin559

PR #2325, opened 2025-01-20
koordlet: fix prodReclaimablePredictor result to avoid influence of o…

lijunxin559:
…versold

When calculating Allocatable[mid] resources, due to possible oversold, ProdReclaimableMetric will be greater than NodeAllocatable * thresholdRatio, so the calculated Allocatable[mid] value accidentally includes the oversold part. However, our previous attempts at modifying the computational model in PR #2291  were not sufficient as they would erase the role of the prodPod estimation model, resulting in the loss of the more stable mid resource characteristics after modification. Therefore, further modifications to the prodPod are needed.

Therefore, I optimized the behavior of ProdReclaimablePredictor by adjusting the values based on the node's runtime information when returning the prediction results, thereby affecting the collectMetric results. And added necessary related tests has proved that the modified calculations are reasonable.

### V. Checklist

### DIFF
--- a/pkg/koordlet/metrics/resource_summary.go
+++ b/pkg/koordlet/metrics/resource_summary.go
@@ -35,6 +35,12 @@ var (
 		Help:      "the node reclaimable of different priorities resources updated by koordinator",
 	}, []string{NodeKey, PriorityKey, ResourceKey, UnitKey})
 
+	NodeResourcePriorityReclaimableStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "node_resource_priority_reclaimable_status",
+		Help:      "status of node reclaimable of different priorities resources updated by koordinator",
+	}, []string{NodeKey, PriorityKey})
+
 	ContainerResourceRequests = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Subsystem: KoordletSubsystem,
 		Name:      "container_resource_requests",
@@ -50,6 +56,7 @@ var (
 	ResourceSummaryCollectors = []prometheus.Collector{
 		NodeResourceAllocatable,
 		NodeResourcePriorityReclaimable,
+		NodeResourcePriorityReclaimableStatus,
 		ContainerResourceRequests,
 		ContainerResourceLimits,
 	}
@@ -74,6 +81,15 @@ func RecordNodeResourcePriorityReclaimable(resourceName string, unit string, pri
 	labels[ResourceKey] = resourceName
 	labels[UnitKey] = unit
 	NodeResourcePriorityReclaimable.With(labels).Set(value)
+}
+
+func RecordNodeResourcePriorityReclaimableStatus(priority string, value float64) {
+	labels := genNodeLabels()
+	if labels == nil {
+		return
+	}
+	labels[PriorityKey] = priority
+	NodeResourcePriorityReclaimableStatus.With(labels).Set(value)
 }
 
 func RecordContainerResourceRequests(resourceName string, unit string, status *corev1.ContainerStatus, pod *corev1.Pod, value float64) {
--- a/pkg/koordlet/prediction/peak_predictor.go
+++ b/pkg/koordlet/prediction/peak_predictor.go
@@ -33,6 +33,10 @@ import (
 // PredictorType defines constants for different types of predictors.
 type PredictorType int
 
+type PredictorContext struct {
+	Node *v1.Node
+}
+
 const (
 	// ProdReclaimablePredictor represents the type of a reclaimable production predictor.
 	ProdReclaimablePredictor PredictorType = iota
@@ -40,7 +44,7 @@ const (
 
 // PredictorFactory is an interface for creating predictors of different types.
 type PredictorFactory interface {
-	New(PredictorType) Predictor
+	New(predictorType PredictorType, context PredictorContext) Predictor
 }
 
 type Predictor interface {
@@ -65,19 +69,22 @@ func NewPredictorFactory(predictServer PredictServer, coldStartDuration time.Dur
 }
 
 // New creates a new instance of a predictor based on the given type.
-func (f *predictorFactory) New(t PredictorType) Predictor {
+func (f *predictorFactory) New(t PredictorType, context PredictorContext) Predictor {
 	switch t {
 	case ProdReclaimablePredictor:
 		podPredictor := &podReclaimablePredictor{
 			predictServer:       f.predictServer,
+			node:                context.Node,
 			coldStartDuration:   f.coldStartDuration,
 			safetyMarginPercent: f.safetyMarginPercent,
 			podFilterFn:         isPodReclaimableForProd,
 			reclaimable:         util.NewZeroResourceList(),
+			unReclaimable:       util.NewZeroResourceList(),
 			pods:                make(map[string]bool),
 		}
 		priorityPredictor := &priorityReclaimablePredictor{
 			predictServer:         f.predictServer,
+			node:                  context.Node,
 			safetyMarginPercent:   f.safetyMarginPercent,
 			priorityClassFilterFn: isPriorityClassReclaimableForProd,
 			reclaimRequest:        util.NewZeroResourceList(),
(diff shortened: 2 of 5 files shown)


pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -210,10 +240,16 @@ func (p *podReclaimablePredictor) AddPod(pod *v1.Pod) error {
 }
 
 // GetResult returns the predicted resource list for the added pods.
+// The result is the sum of the reclaimable resources of the added pods.
 func (p *podReclaimablePredictor) GetResult() (v1.ResourceList, error) {
-	metrics.RecordNodePredictedResourceReclaimable(string(v1.ResourceCPU), metrics.UnitCore, p.GetPredictorName(), float64(p.reclaimable.Cpu().MilliValue())/1000)
-	metrics.RecordNodePredictedResourceReclaimable(string(v1.ResourceMemory), metrics.UnitByte, p.GetPredictorName(), float64(p.reclaimable.Memory().Value()))
-	return p.reclaimable, nil
+	if p.node == nil || p.node.Status.Capacity == nil {
```
saintube, line 245: Please notice that the node allocatable can be updated or mutated either by machine-level operations or the webhook with the [resource amplification ratio](https://koordinator.sh/blog/release-v1.4.0/#4-cpu-normalization). We can retrieve the original node allocatable from the annotation `node.koordinator.sh/raw-allocatable` if it exists. Perhaps the node should be an argument of the `GetResult()`.
And, we prefer the allocatable instead of the capacity since there does not count the node reservation in contrast to the node resource calculation.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -285,21 +325,24 @@ func (n *priorityReclaimablePredictor) GetResult() (v1.ResourceList, error) {
...
 	}
 
 	// reclaimable[P] := max(request[P] - peak[P], 0)
-	reclaimable := quotav1.Max(quotav1.Subtract(n.reclaimRequest, reclaimPredict), util.NewZeroResourceList())
-	metrics.RecordNodePredictedResourceReclaimable(string(v1.ResourceCPU), metrics.UnitCore, n.GetPredictorName(), float64(reclaimable.Cpu().MilliValue())/1000)
-	metrics.RecordNodePredictedResourceReclaimable(string(v1.ResourceMemory), metrics.UnitByte, n.GetPredictorName(), float64(reclaimable.Memory().Value()))
-	return reclaimable, nil
+	reclaimable := quotav1.Max(quotav1.Subtract(p.reclaimRequest, unReclaimable), util.NewZeroResourceList())
+	// fixReclaimable[P] := min(nodeCapacity[P]-unReclaimable[P],reclaimable[P])
+	fixReclaimable := quotav1.SubtractWithNonNegativeResult(p.node.Status.Capacity, unReclaimable)
```
saintube, line 341: same above

pkg/koordlet/statesinformer/impl/states_nodemetric.go
```diff
@@ -360,8 +360,8 @@ func (r *nodeMetricInformer) collectMetric() (*slov1alpha1.NodeMetricInfo, []*sl
 		Start:     &startTime,
 		End:       &endTime,
 	}
-	prodPredictor := r.predictorFactory.New(prediction.ProdReclaimablePredictor)
-
+	node := r.nodeSLOInformer.callbackRunner.statesInformer.GetNode()
```
saintube, line 363: why not retrieve from the nodeInformer

pkg/koordlet/prediction/peak_predictor.go
lijunxin559: Thank you for the reminder about the amplification factor. It has been added
I think it would be better to include the node in the predictor instead of directly using it as a parameter for GetResult()
Yes, allocatable is better

pkg/koordlet/metrics/resource_summary.go
```diff
@@ -35,6 +35,12 @@ var (
 		Help:      "the node reclaimable of different priorities resources updated by koordinator",
 	}, []string{NodeKey, PriorityKey, ResourceKey, UnitKey})
 
+	NodeResourcePriorityReclaimableErr = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
saintube, line 38: Need to keep the variable name same as the metric name.

pkg/koordlet/metrics/resource_summary.go
```diff
@@ -76,6 +83,16 @@ func RecordNodeResourcePriorityReclaimable(resourceName string, unit string, pri
 	NodeResourcePriorityReclaimable.With(labels).Set(value)
 }
 
+func RecordNodeResourcePriorityReclaimableErr(priority string, err string, value float64) {
+	labels := genNodeLabels()
+	if labels == nil {
+		return
+	}
+	labels[PriorityKey] = priority
+	labels[ErrorKey] = err
```
saintube, line 92: Picking the error message as a label value is not a good choice. Perhaps it should be a status about succeeded or failed.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -33,14 +33,18 @@ import (
 // PredictorType defines constants for different types of predictors.
 type PredictorType int
 
+type ProdReclaimablePredictorOptions struct {
```
saintube, line 36: If an option is necessary, consider using an interface of Options and method ApplyFor(), or a strongly-typed struct for all predictors. We need to distinguish options for different priorities, and we prefer not to use a weakly-type argument for the initialization.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -210,10 +240,21 @@ func (p *podReclaimablePredictor) AddPod(pod *v1.Pod) error {
 }
 
 // GetResult returns the predicted resource list for the added pods.
+// The result is the sum of the reclaimable resources of the added pods.
 func (p *podReclaimablePredictor) GetResult() (v1.ResourceList, error) {
```
saintube, line 244: As described, the node allocatable can be updated sometimes. So we'd better pass the latest object.

pkg/koordlet/metrics/resource_summary.go
lijunxin559: In this case, it only indicates whether it is currently available and not convenient for checking errors. If it is deemed unnecessary, I can remove it

pkg/koordlet/prediction/peak_predictor.go
lijunxin559: I don't think one round of calculation will take too long, right? The time interval from starting to retrieve info to obtaining this calculation result should be tolerable

Of course, we can retrieve it at the end and pass the node parameter into the GetResult(), but won't this make the interface lose its flexibility? I suggest including all the content related to this calculation logic in the interface

pkg/koordlet/prediction/peak_predictor.go
lijunxin559: yes , you are right, ProdReclaimablePredictorOptions -> PredictorContext

pkg/koordlet/metrics/resource_summary.go
lijunxin559: fixed

pkg/koordlet/metrics/resource_summary.go
```diff
@@ -76,6 +83,15 @@ func RecordNodeResourcePriorityReclaimable(resourceName string, unit string, pri
 	NodeResourcePriorityReclaimable.With(labels).Set(value)
 }
 
+func RecordNodeResourcePriorityReclaimableStatus(priority string, value float64) {
```
saintube, line 86: It seems passing a boolean is enough. No need to define a StatusSucceedBool.

pkg/koordlet/prediction/peak_predictor.go
saintube: It's OK. I missed the fact that a predictor is instantiated once a round.

pkg/koordlet/metrics/resource_summary.go
lijunxin559: ok, done

saintube (commented):
/lgtm

saintube:
PTAL /cc @zwzhang0107 @hormes @jasonliu747

PR #2349, opened 2025-02-26
koordlet: support init container for CPUSetAllocator

zwzhang0107:

Set cpuset for init container in CPUSetAllocator.

https://github.com/koordinator-sh/koordinator/issues/2350

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/reconciler/reconciler.go
+++ b/pkg/koordlet/runtimehooks/reconciler/reconciler.go
@@ -484,6 +484,32 @@ func (c *reconciler) reconcilePodCgroup(stopCh <-chan struct{}) {
 					}
 				}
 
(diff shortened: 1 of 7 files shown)


pkg/koordlet/runtimehooks/reconciler/reconciler.go
```diff
@@ -484,6 +484,31 @@ func (c *reconciler) reconcilePodCgroup(stopCh <-chan struct{}) {
 					}
 				}
 
+				for _, initContainerStat := range podMeta.Pod.Status.InitContainerStatuses {
```
saintube, line 487: perhaps the completed/failed init containers should be filtered out, and sidecar containers should be handled

zwzhang0107: we leave a TODO here

Issue #2367, opened 2025-03-15
[proposal] add metrics for reservation


googs1025:
**What is your proposal**:
add metrics for reservation

**Why is this needed**:
When using the reservation feature, our SRE wants to focus on the metrics about the reservation status. We don’t want too many reservation failures. We need metrics monitoring

**Is there a suggested solution, if so, please add it**:

like this below: 
```go
        ReservationFailedTotal = metrics.NewCounterVec(
		&metrics.CounterOpts{
			Subsystem: schedulermetrics.SchedulerSubsystem,
			Name:      "reservation_failed_total",
			Help:      `The total number of failed Reservation`,
		}, []string{"name"},
	)

	ReservationSucceededTotal = metrics.NewCounterVec(
		&metrics.CounterOpts{
			Subsystem: schedulermetrics.SchedulerSubsystem,
			Name:      "reservation_successed_total",
			Help:      `The total number of successed Reservation`,
		}, []string{"name"},
	)
```

googs1025:
In addition, I think whether other features also need to increase metrics, this may be discussed. 🤔

PR #2368, opened 2025-03-17
feature(reservation): add ReservationFailedTotal ReservationSucceededTotal metrics for reservation feature

googs1025:
add ReservationFailedTotal ReservationSucceededTotal metrics for reservation feature

fix https://github.com/koordinator-sh/koordinator/issues/2367

### V. Checklist


googs1025:
I have to increase more ut

pkg/scheduler/metrics/metrics.go
```diff
@@ -34,8 +34,26 @@ var (
 			StabilityLevel: metrics.STABLE,
 		}, []string{"profile"})
 
+	ReservationFailedTotal = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem: schedulermetrics.SchedulerSubsystem,
```
googs1025, line 39: not sure if use scheduler prefix? 🤔

saintube: It's OK for a metric prefix.
However, we should handle the metrics GC if using the label `name`. An option is to use `utilmetrics.NewGCCounterVec`.

googs1025: Do we have relevant documentation? 🤔

googs1025: done

pkg/scheduler/metrics/metrics.go
```diff
@@ -56,3 +101,7 @@ func RegisterMetrics(extraMetrics ...metrics.Registerable) {
 		legacyregistry.MustRegister(metric)
 	}
 }
+
+func RegisterGCMetrics(extraMetrics ...prometheus.Collector) {
```
googs1025, line 105: I understand this is just a transition period? We have other metrics that have not been renamed as GC Metrics, so two registration methods are required

pkg/scheduler/metrics/metrics.go
```diff
@@ -19,33 +19,78 @@ package metrics
...
-		&metrics.CounterOpts{
-			Subsystem:      schedulermetrics.SchedulerSubsystem,
-			Name:           "scheduling_timeout",
-			Help:           "The currently scheduled Pod exceeds the maximum acceptable time interval",
-			StabilityLevel: metrics.STABLE,
-		}, []string{"profile"})
-
-	metricsList = []metrics.Registerable{
-		SchedulingTimeout,
+	SchedulingTimeout = utilmetrics.NewGCCounterVec("scheduling_timeout", prometheus.NewCounterVec(
```
saintube, line 33: This metric does not need a GC since it just labels different profiles very few in a cluster.
We should add GC logic for metrics yielding unlimited label pairs.

pkg/scheduler/metrics/metrics.go
```diff
@@ -19,33 +19,78 @@ package metrics
...
+		prometheus.CounterOpts{
+			Subsystem: schedulermetrics.SchedulerSubsystem,
+			Name:      "scheduling_timeout",
+			Help:      "The currently scheduled Pod exceeds the maximum acceptable time interval",
+		}, []string{"profile"}))
+
+	ReservationSucceededTotal = utilmetrics.NewGCCounterVec("reservation_successed_total", prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: schedulermetrics.SchedulerSubsystem,
+			Name:      "reservation_successed_total",
```
saintube, line 43: @googs1025 
I'm wondering about the usage of the metric `reservation_succeeded_total{name="xxx"}`. The phases `Succeeded` and `Failed` should be irreversible for reservations, so the metric value would be at most 1 when it is increased.
Perhaps, it should be splited into two metrics `pending_reservations` and `reservation_status_phase{reservation="", phase="Pending|Available|Succeeded|Failed"}`.

pkg/scheduler/metrics/metrics.go
saintube: The method `GetCounterVec()` has already returned a registable metric for the registry, so I think it is not a new Register func.

pkg/scheduler/metrics/metrics.go
googs1025: got this

pkg/scheduler/metrics/metrics.go
googs1025: There are some differences between standard metrics and gc metrics. It should be better for us to use two registration methods.
```go
// RegisterStandardAndGCMetrics registers both standard and garbage collection metrics.
func RegisterStandardAndGCMetrics(standardMetrics []metrics.Registerable, gcMetrics []prometheus.Collector) {
	for _, metric := range standardMetrics {
		legacyregistry.MustRegister(metric)
	}
	for _, metric := range gcMetrics {
		legacyregistry.RawMustRegister(metric)
	}
}
```

pkg/scheduler/metrics/metrics.go
googs1025: Should we first support `Success` `Failed`? Because this does not involve state transfer, especially `Failed` phase is the most concerned state 🤔

pkg/util/reservation/reservation.go
```diff
@@ -288,7 +289,13 @@ func SetReservationUnschedulable(r *schedulingv1alpha1.Reservation, msg string)
 }
 
 func SetReservationExpired(r *schedulingv1alpha1.Reservation) {
-	r.Status.Phase = schedulingv1alpha1.ReservationFailed
+	// Check if the Reservation is already in failed phase,
+	// if not, set the phase  and increment metric
+	wasFailed := r.Status.Phase == schedulingv1alpha1.ReservationFailed
```
saintube, line 294: do you need to unset the metrics for other phases?

googs1025: If I understand correctly, I don't think it is necessary. 🤔  Because `Failed` and `Successed` both represent final states. We will only record this final state. This final state will only be recorded once and will not be converted, which is simpler. Because Failed and Successed will not be converted to each other, there is no need to unset the metrics. That is, for example: `scheduler_reservation_status_phase{name="normalReservation",phase="Failed"} 1` and `scheduler_reservation_status_phase{name="normalReservation",phase="Succeeded"} 1`will only have one appear (because Failed Successed is a final state and cannot be converted).

saintube: OK. In this assumption, we can skip unset for other phases.

There is another question. What if the reservation has already failed when the scheduler starts? Perhaps we should set the metric according to the current state, instead of capturing the updating event.

googs1025: I don't quite understand what you mean. 

But I can explain the current logic:

Currently, the reservation phase is changed and metrics are added during the first state transition.

Consider a scenario: if reservationA is already in the final state (success or failure), the scheduler is restarted and the controller coordination process enters the SetReservationExpired or SetReservationSucceeded method. In this case, the "metrics" indicator will not be added again.

saintube: There are two questions above:

1. When to record the reservation phase: The pod metric like `kube_pod_status_phase` will remain consistent after the kube-state-metric fail-over. So, we'd better record the reservation phase regardless of whether it just changed.
2. Does the metric need garbage collection: As described, it does not since we have the assumption that for a reservation it cannot both transfer to Failed and Succeeded. However, it is still needed if we record the Pending phase and the Available phase.

pkg/util/reservation/reservation.go
```diff
@@ -289,6 +290,8 @@ func SetReservationUnschedulable(r *schedulingv1alpha1.Reservation, msg string)
 
 func SetReservationExpired(r *schedulingv1alpha1.Reservation) {
 	r.Status.Phase = schedulingv1alpha1.ReservationFailed
+	metrics.RecordReservationPhase(r.Name, string(schedulingv1alpha1.ReservationFailed), 1)
```
googs1025, line 293: When entering the `SetReservationExpired` (similar to `SetReservationSucceeded` and `SetReservationAvailable`), we will set the previous state to 0 and increase the current state to 1. This is reentrant, so there is no problem when the coordination is repeated.

googs1025: cc @saintube

pkg/scheduler/frameworkext/reservation_info.go
```diff
@@ -84,6 +85,7 @@ func NewReservationInfo(r *schedulingv1alpha1.Reservation) *ReservationInfo {
 	allocatable := reservationutil.ReservationRequests(r)
 	reserved := util.GetNodeReservationFromAnnotation(r.Annotations)
 	resourceNames := quotav1.ResourceNames(allocatable)
+	metrics.RecordReservationPhase(r.Name, string(schedulingv1alpha1.ReservationPending), 1)
```
googs1025, line 88: in `NewReservationInfo`. I thinks we can set pend phase

saintube: I'm afraid not. This function can be called when creating a temporary ReservationInfo for some calculation.

Please check the reservation controller. I guess there is a good place to record each stage, as we do not care about the metric timeliness.

googs1025: Sorry, do you mean only `pending` phase needs to be set in the controller or other phase also need to be set in the controller? 🤔

saintube: For all phases. I think it would make the metric recording more clean.

googs1025: changed implement to controller

pkg/scheduler/metrics/metrics.go
```diff
@@ -34,25 +39,88 @@ var (
...
+		reservationNameKey:  name,
+		reservationPhaseKey: phase,
+	}
+	ReservationStatusPhase.WithSet(labels, value)
+}
+
+// RecordReservationPhases records all possible phases of a reservation as metrics.
+// For each phase, it sets the value to 1.0 if it matches the current phase of the reservation,
+// otherwise, it sets the value to 0.0.
+func RecordReservationPhases(reservation *schedulingv1alpha1.Reservation) {
```
saintube, line 76: Please move this function to the directory reservation/controller and leave just the `RecordReservationPhase` here. It makes the metrics package clean and does not rely on other struct definitions to reduce risks of cyclic dependency.

pkg/scheduler/metrics/metrics.go
```diff
@@ -34,25 +39,88 @@ var (
...
+		RegisterStandardAndGCMetrics(metricsList, gcMetricsList)
 	})
 }
 
-// RegisterMetrics registers a list of metrics.
-// This function is exported because it is intended to be used by out-of-tree plugins to register their custom metrics.
-func RegisterMetrics(extraMetrics ...metrics.Registerable) {
-	for _, metric := range extraMetrics {
+// RegisterStandardAndGCMetrics registers both standard and garbage collection metrics.
+func RegisterStandardAndGCMetrics(standardMetrics []metrics.Registerable, gcMetrics []prometheus.Collector) {
```
saintube, line 119: This naming is more clear than the first version.

pkg/scheduler/metrics/metrics.go
googs1025: done

pkg/scheduler/plugins/reservation/controller/controller.go
```diff
@@ -201,14 +193,20 @@ func (c *Controller) sync(reservationName string) (result, error) {
 
 func (c *Controller) expireReservation(reservation *schedulingv1alpha1.Reservation) error {
 	reservationutil.SetReservationExpired(reservation)
-	_, err := c.koordClientSet.SchedulingV1alpha1().Reservations().UpdateStatus(context.TODO(), reservation, metav1.UpdateOptions{})
-	return err
+	return c.updateReservationStatus(reservation)
 }
 
 func (c *Controller) syncStatus(reservation *schedulingv1alpha1.Reservation) error {
 	if reservation.Status.NodeName == "" {
```
saintube, line 200: This will skip the pending reservations metrics, but we need it.

googs1025: add `c.pendingReservation(reservation)`

pkg/scheduler/plugins/reservation/controller/controller.go
```diff
@@ -201,13 +193,24 @@ func (c *Controller) sync(reservationName string) (result, error) {
 
 func (c *Controller) expireReservation(reservation *schedulingv1alpha1.Reservation) error {
 	reservationutil.SetReservationExpired(reservation)
-	_, err := c.koordClientSet.SchedulingV1alpha1().Reservations().UpdateStatus(context.TODO(), reservation, metav1.UpdateOptions{})
-	return err
+	return c.updateReservationStatus(reservation)
+}
+
+func (c *Controller) pendingReservation(reservation *schedulingv1alpha1.Reservation) error {
+	reservation.Status.Phase = schedulingv1alpha1.ReservationPending
```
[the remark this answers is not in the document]
googs1025: oh, got what you mean, change to as below: 
```go
if reservation.Status.NodeName == "" {
    RecordReservationPhases(reservation)
    return nil
}
```

saintube (commented):
/lgtm
PTAL /cc @ZiMengSheng @hormes

saintube (approved):
/lgtm

pkg/scheduler/plugins/reservation/controller/controller.go
```diff
@@ -203,14 +195,21 @@ func (c *Controller) sync(reservationName string) (result, error) {
...
-	return err
+	return c.updateReservationStatus(reservation)
 }
 
 func (c *Controller) syncStatus(reservation *schedulingv1alpha1.Reservation) error {
 	if reservation.Status.NodeName == "" {
+		RecordReservationPhases(reservation)
 		return nil
 	}
+	if isReservationNeedExpiration(reservation) {
```
saintube, line 206: The pending reservation should consider expiration, so please move the expiration logic ahead of the `if reservation.Status.NodeName == ""`.

googs1025: thanks for reminding. my bad

googs1025: done

pkg/scheduler/plugins/elasticquota/metrics.go
```diff
@@ -51,14 +51,16 @@ var (
...
+	metricsList = []metrics.Registerable{
 		ElasticQuotaSpecMetric,
 		ElasticQuotaStatusMetric,
 		UpdateElasticQuotaStatusLatency,
-	)
+	}
+)
+
+func init() {
+	koordschedulermetrics.RegisterStandardAndGCMetrics(metricsList, nil)
```
saintube, line 63: Recommend to separate the definitions for registering regular metrics and GC metrics, so there is no need to adapt.

googs1025: separate to:

```go
// RegisterMetrics registers a list of metrics.
// This function is exported because it is intended to be used by out-of-tree plugins to register their custom metrics.
func RegisterMetrics(extraMetrics ...metrics.Registerable) {
	for _, metric := range extraMetrics {
		legacyregistry.MustRegister(metric)
	}
}

// RegisterGCMetrics registers garbage collection metrics.
func RegisterGCMetrics(gcMetrics ...prometheus.Collector) {
	for _, metric := range gcMetrics {
		legacyregistry.RawMustRegister(metric)
	}
}
```

PR #2390, opened 2025-04-03
apis: add default Pending phase when creating reservation

googs1025:
### Ⅱ. Does this pull request fix one issue?None
- add default Pending phase when creating reservation

### V. Checklist

### DIFF
--- a/apis/scheduling/v1alpha1/reservation_types.go
+++ b/apis/scheduling/v1alpha1/reservation_types.go
@@ -115,6 +115,7 @@ type ReservationStatus struct {
 	// The `phase` indicates whether is reservation is waiting for process, available to allocate or failed/expired to
 	// get cleanup.
 	// +optional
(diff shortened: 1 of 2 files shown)


apis/scheduling/v1alpha1/reservation_types.go
```diff
@@ -115,6 +115,7 @@ type ReservationStatus struct {
 	// The `phase` indicates whether is reservation is waiting for process, available to allocate or failed/expired to
 	// get cleanup.
 	// +optional
+	// +kubebuilder:default="Pending"
```
googs1025, line 118: I found that when we create a reservation, there is no pending phase until the availble phase. A better way is to set the initial status to pending when creating it. 🤔

googs1025:
I have used 
`make manifests` and `make generate` to generate crd code

jasonliu747 (approved):
/lgtm

PR #2395, opened 2025-04-08
scheduler: Add Pod QosClass Sort for coscheduling

dongjiang1989:

Campare Pod QosClass, When QosClass not equal, the higher Qos is at the front of the queue

### V. Checklist


pkg/scheduler/plugins/coscheduling/coscheduling.go
```diff
@@ -170,6 +170,12 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
 		return childScheduleCycle1 < childScheduleCycle2
 	}
 
+	kubeQoS1 := extension.GetPodQoSClassWithDefault(podInfo1.Pod)
```
googs1025, line 173: Why do we need to obtain Qos type judgment in coscheduling? 🤔 
If I understand correctly, pod have a priority defined by their own priority. 
In QoS, it is used as a basis for judging whether to evict, and there is no need to sort QoS.

dongjiang1989: Hmmm.... Before comparing `Pod.Name`, it may make more sense to compare the `QosClass` set by the user.

saintube:
PTAL /cc @ZiMengSheng

pkg/scheduler/plugins/coscheduling/coscheduling.go
ZiMengSheng: Pod.Name is only used when sorting Pods in the same gangGroup, providing certainty of Pod dequeueing in the same gangGroup.

PR #2399, opened 2025-04-09
scheduler: reduce loadaware lock overhead, improve reservation args

saintube:
### V. Checklist


ZiMengSheng:
UT Fail

saintube:
> UT Fail

fixed

PR #2409, opened 2025-04-17
webhook: check quota that only allow max >= used for specific resources

zheng-weihao:

Adding an optional annotation `quota.scheduling.koordinator.sh/max-strict-check-resource-keys` for elastic quota that record resource keys that updates for quota spec should be checked max >= used in json format.

Example: `quota.scheduling.koordinator.sh/max-strict-check-resource-keys: '["cpu", "memory"]'`

NONE

### V. Checklist

### DIFF
--- a/apis/extension/elastic_quota.go
+++ b/apis/extension/elastic_quota.go
@@ -55,7 +55,7 @@ const (
 	AnnotationNonPreemptibleRequest      = QuotaKoordinatorPrefix + "/non-preemptible-request"
 	AnnotationNonPreemptibleUsed         = QuotaKoordinatorPrefix + "/non-preemptible-used"
 	AnnotationAdmission                  = QuotaKoordinatorPrefix + "/admission"
-	AnnotationCheckMaxGeUsedResourceKeys = QuotaKoordinatorPrefix + "/check-max-ge-used-resource-keys"
+	AnnotationMaxStrictCheckResourceKeys = QuotaKoordinatorPrefix + "/max-strict-check-resource-keys"
 )
 
 func GetParentQuotaName(quota *v1alpha1.ElasticQuota) string {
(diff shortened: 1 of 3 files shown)


pkg/webhook/elasticquota/quota_topology_check.go
```diff
@@ -70,6 +70,20 @@ func (qt *quotaTopology) validateQuotaSelfItem(quota *v1alpha1.ElasticQuota) err
...
+	if err != nil {
+		return fmt.Errorf("%v quota.Annotation[%v]'s value is invalid: %w", quota.Name, extension.AnnotationCheckMaxGeUsedResourceKeys, err)
+	}
+	for _, key := range resourceKeys {
+		if usedVal, exist := quota.Status.Used[key]; !exist {
+			continue // check should be skipped if current used never counts
+		} else if maxVal, exist := quota.Spec.Max[key]; !exist {
+			return fmt.Errorf("resourceKey %v of quota %v is included in used, which is not included in max but should check max >= used", key, quota.Name)
+		} else if maxVal.Cmp(usedVal) == -1 {
+			return fmt.Errorf("resourceKey %v of quota %v used,%v > max,%v", key, quota.Name, usedVal.String(), maxVal.String())
```
saintube, line 84: should be `used %v, > max %v`?

apis/extension/elastic_quota.go
```diff
@@ -28,33 +28,34 @@ import (
...
+	AnnotationResourceKeys               = QuotaKoordinatorPrefix + "/resource-keys"
+	AnnotationTotalResource              = QuotaKoordinatorPrefix + "/total-resource"
+	AnnotationUnschedulableResource      = QuotaKoordinatorPrefix + "/unschedulable-resource"
+	AnnotationQuotaNamespaces            = QuotaKoordinatorPrefix + "/namespaces"
+	AnnotationGuaranteed                 = QuotaKoordinatorPrefix + "/guaranteed"
+	AnnotationAllocated                  = QuotaKoordinatorPrefix + "/allocated"
+	AnnotationNonPreemptibleRequest      = QuotaKoordinatorPrefix + "/non-preemptible-request"
+	AnnotationNonPreemptibleUsed         = QuotaKoordinatorPrefix + "/non-preemptible-used"
+	AnnotationAdmission                  = QuotaKoordinatorPrefix + "/admission"
+	AnnotationCheckMaxGeUsedResourceKeys = QuotaKoordinatorPrefix + "/check-max-ge-used-resource-keys"
```
saintube, line 58: To revise the naming, how about `restricted-max-resource-keys`? WDYT

zheng-weihao: I think `restricted-max` does not exactly express the constraint when setting the annotation.
Since `ge` for "greater than or equal to" is commonly used, `check-max-ge-used-resource-keys` may be a good but not excellent choise.

pkg/webhook/elasticquota/quota_topology_check.go
zheng-weihao: solved

apis/extension/elastic_quota.go
saintube: This naming is kind of verbose...

zheng-weihao: How about `max-strict-check-resource-keys`

saintube: OK

zheng-weihao: Done @saintube

saintube (commented):
/lgtm
PTAL /cc @shaloulcy

PR #2419, opened 2025-04-22
scheduler: split quota PreFilter and fix preemption

saintube:

- koord-scheduler
  - Split ElasticQuota PreFilter to AfterPreFilter, so that the snapshotted quota stats can be adjusted before the real check.
  - Fix the quota recursive check and non-preemptible check with preemption. 
  - Reduce quotaInfo lock overhead by RWMutex.

### V. Checklist


saintube:
/hold
Looking for a more elegant way to support preemption.

PR #2424, opened 2025-04-29
scheduler: avoid string concatenation and podLister in loadaware func…

lijunxin559:
Optimization of loadaware performance

1. avoid string concatenation 
2. fetch pod/Priority  by reusing nodeMetrics/podMetrics/Priorit but not podLister

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/loadaware/helper.go
+++ b/pkg/scheduler/plugins/loadaware/helper.go
@@ -17,14 +17,13 @@ limitations under the License.
 package loadaware
 
 import (
-	"fmt"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
-	corev1listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/loadaware/helper.go
```diff
@@ -145,37 +144,32 @@ func generateUsageThresholdsFilterProfile(node *corev1.Node, args *schedulingcon
...
-		pod, err := podLister.Pods(podMetric.Namespace).Get(podMetric.Name)
-		if err != nil {
-			continue
-		}
-		if filterProdPod && extension.GetPodPriorityClassWithDefault(pod) != extension.PriorityProd {
+		if filterProdPod && podMetric.Priority != extension.PriorityProd {
 			continue
 		}
-		name := getPodNamespacedName(podMetric.Namespace, podMetric.Name)
+		name := types.NamespacedName{
```
googs1025, line 163: what is the different between `getPodNamespacedName` func 🤔

lijunxin559: no difference, but reduce unnecessary codes

saintube: @googs1025 It improves the performance by asynchronously constructing the key.

saintube: We do not need to render the string. Instead, use the struct as the key.

saintube (commented):
/lgtm

PR #2425, opened 2025-04-29
scheduler: optimize secondary device well planned

ZiMengSheng:

In a production environment, sufficient VF resources are usually planned for the GPU. Under this condition, you can skip the resource check of the VF in the Filter stage and only do it in the Reserve stage.

### V. Checklist


saintube (approved):
/lgtm
Recommended to add some description about this API.

PR #2428, opened 2025-04-29
scheduler: optimize filterNodeDevice

ZiMengSheng:
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_cache.go
+++ b/pkg/scheduler/plugins/deviceshare/device_cache.go
@@ -452,7 +452,7 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
 }
 
 type nodeDeviceCache struct {
-	lock sync.Mutex
+	lock sync.RWMutex
 	// nodeDeviceInfos stores nodeDevice for each node.
 	nodeDeviceInfos map[string]*nodeDevice
 }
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/deviceshare/utils.go
```diff
@@ -386,6 +386,12 @@ func parsePodDeviceShareExtensions(pod *corev1.Pod, podRequests map[schedulingv1
 
 	state.hints = hints
 	state.hintSelectors = hintSelectors
+	for _, selectors := range hintSelectors {
+		if selectors[0] != nil {
```
saintube, line 390: can we assert`len(selector) > 0`?

ZiMengSheng: yes, 
image

saintube, line 390: add a check to avoid nil pointer reference?

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -464,11 +464,17 @@ func newNodeDeviceCache() *nodeDeviceCache {
 }
 
 func (n *nodeDeviceCache) getNodeDevice(nodeName string, needInit bool) *nodeDevice {
+	if !needInit {
```
googs1025, line 467: Can we move this after the lock call so that we don't have to call the lock method repeatedly? 🤔

saintube (approved):
/lgtm

PR #2429, opened 2025-04-30
scheduler: remove Duplicate goroutines and remove Deprecated poll func

googs1025:
### V. Checklist


pkg/scheduler/plugins/coscheduling/controller/podgroup_test.go
```diff
@@ -181,8 +181,8 @@ func Test_Run(t *testing.T) {
 					kubeClient.CoreV1().Pods(p.Namespace).UpdateStatus(ctx, p, metav1.UpdateOptions{})
 				}
 			}
-			go ctrl.Start()
```
googs1025, line 184: already add goroutine here.

https://github.com/koordinator-sh/koordinator/blob/01a456b1d70423efbfabe20bf42d3497e4cf1c9b/pkg/scheduler/plugins/coscheduling/controller/podgroup.go#L100-L103

jasonliu747 (approved):
/lgtm

saintube (approved):

PR #2435, opened 2025-05-09
descheduler: return the unmatched cases first

googs1025:

We should return the unmatched cases first,and leave the filter method for last, because it is computationally expensive.

### V. Checklist


songtao98:
@googs1025 yes, I think this is a nice optimization in theory. However, have you already meet any performance bottleneck? It would be great if some real cases can be provided :)

/lgtm

googs1025:
> @googs1025 yes, I think this is a nice optimization in theory. However, have you already meet any performance bottleneck? It would be great if some real cases can be provided :)


Not yet, I just noticed a fix from the code level 😄

PR #2437, opened 2025-05-11
chore: qos manager removes repeat goroutine

googs1025:
- We will start gorotuine for each plugin in Run, no need to start it repeatedly

### V. Checklist


pkg/koordlet/qosmanager/qosmanager.go
```diff
@@ -118,7 +118,7 @@ func (r *qosManager) Run(stopCh <-chan struct{}) error {
 			klog.V(4).Infof("qos strategy %v is not enabled, skip running", name)
 			continue
 		}
-		go strategy.Run(stopCh)
+		strategy.Run(stopCh)
```
googs1025, line 121: We will start gorotuine for each plugin in `Run`, no need to start it repeatedly

saintube:
/lgtm
PTAL /cc @zwzhang0107

PR #2444, opened 2025-05-15
chore: fix wrong event type

googs1025:

- if evicting pod successfully, we should use `Normal` type

```go
// Valid values for event types (new types could be added in future)
const (
	// Information only and will not cause any problems
	EventTypeNormal string = "Normal"
	// These events are to warn that something might go wrong
	EventTypeWarning string = "Warning"
)
```

### V. Checklist

### DIFF
--- a/pkg/quota-controller/profile/profile_controller.go
+++ b/pkg/quota-controller/profile/profile_controller.go
@@ -214,7 +214,7 @@ func (r *QuotaProfileReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if !quotaExist {
 		err = r.Client.Create(context.TODO(), quota)
 		if err != nil {
-			r.Recorder.Eventf(profile, "Warning", ReasonCreateQuotaFailed, "failed to create quota, err: %s", err)
+			r.Recorder.Eventf(profile, corev1.EventTypeWarning, ReasonCreateQuotaFailed, "failed to create quota, err: %s", err)
 			klog.Errorf("failed create quota for profile %v, error: %v", req.NamespacedName, err)
 			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 		}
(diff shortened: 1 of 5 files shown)


pkg/koordlet/qosmanager/framework/context.go
```diff
@@ -95,7 +95,7 @@ func (r *Evictor) evictPod(evictPod *corev1.Pod, reason string, message string)
 	if err := util.EvictPodByVersion(context.TODO(), r.kubeClient, evictPod.Namespace, evictPod.Name, metav1.DeleteOptions{
 		GracePeriodSeconds: nil,
 		Preconditions:      metav1.NewUIDPreconditions(string(evictPod.UID))}, r.evictVersion); err == nil {
-		r.eventRecorder.Eventf(evictPod, corev1.EventTypeWarning, helpers.EvictPodSuccess, podEvictMessage)
+		r.eventRecorder.Eventf(evictPod, corev1.EventTypeNormal, helpers.EvictPodSuccess, podEvictMessage)
```
saintube, line 98: keep as Warning for eviction?

googs1025: I understand that the event for a successful eviction should be Normal, because Warning seems to be an unexpected problem. 🤔 But if Warning is what we want, I will roll it back

saintube: I prefer to keep it unchanged since it was set as a warning, as the eviction was notable even if it was successful.

saintube:
@googs1025 Do you still work on this?

googs1025:
> @googs1025 Do you still work on this?

Will go on after KubeCon finished

pkg/koordlet/qosmanager/framework/context.go
googs1025: removed. 😄

googs1025:
> @googs1025 Do you still work on this?

done

PR #2448, opened 2025-05-19
e2e: disable LoadAware Filter to fix flaky test

### LINKED ISSUE #2438: [BUG] e2e fails due to LoadAware scheduling threshold is too low.
**What happened**:

The E2E cases "BatchResource" sometimes failed with the timeout message that `node(s) cpu usage exceed threshold`. It indicates the Batch pod created during the test failed to schedule due to the LoadAware plugin's Filter. It is a flaking test since the machine's CPU usage of the GitHub Action pipeline is not deterministic to exceed or not exceed the filtering threshold.

The result of the E2E case should be stable, instead of failing occasionally.

In this case, the major admission should be the filtering of the extended resources like "kubernetes.io/batch-cpu" instead of the load-aware scheduling.
So during the case "BatchResource", we can set the filter threshold for the testing nodes to dismiss the load-aware scheduling.

- LoadAware scheduling arguments of the E2E cases: https://github.com/koordinator-sh/koordinator/blob/main/config/manager/scheduler-config.yaml#L32
- Node-level LoadAware thresholds: https://github.com/koordinator-sh/koordinator/blob/main/apis/extension/load_aware.go#L29

- https://github.com/koordinator-sh/koordinator/actions/runs/14928230417/job/42025615469?pr=2435
- https://github.com/koordinator-sh/koordinator/actions/runs/14956883806/job/42013736905?pr=2437
- https://github.com/koordinator-sh/koordinator/actions/runs/14529036074/job/40854284030

**Environment**:
- App version: v1.6.0
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

ZiMengSheng:

fixes #2438

### V. Checklist

### DIFF
--- a/config/manager/scheduler-config.yaml
+++ b/config/manager/scheduler-config.yaml
@@ -39,8 +39,8 @@ data:
               cpu: 1
               memory: 1
             usageThresholds:
-              cpu: 65
-              memory: 95
+              cpu: 0
+              memory: 0
             estimatedScalingFactors:
               cpu: 85
               memory: 70


saintube:
disable by usageThresholds?

Issue #2413, opened 2025-04-18
[BUG] 非 Koordinator 调度的 Pod 被非预期驱逐的问题


ZiMengSheng:
**What happened**:

现象：pod 是 default-scheduler 调度的，但是被 koord-scheduler 驱逐了...

 原因：仔细排查了下发起抢占 pod 用的 quota 确实没创建。不过这个逻辑有点危险，我想了下 canPreempt 里加个 feature gate 只限定驱逐 koordinator scheduler 调度的 pod 会不会合理一点？当然 disable default group 也可以达到这个效果。

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version:
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

PR #2449, opened 2025-05-19
scheduler: add ElasticQuota arg to control preemption of default quota (#2413)

LennonChin:

koord-scheduler

- The quota of pods that do not actively set through annotation will be recognized as the default quota. In some scenarios, many important pods are not scheduled by Koordinator, and their quotas are also recognized as default quota, this will cause the eviction operation to identify these important pods as victim pods, which is risky. We introduced a new argument `DisableDefaultQuotaPreemption` in ElasticQuota, when `DisableDefaultQuotaPreemption` is set to true, these pods can be avoided from being evicted.

unit tests passed.

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/v1/defaults.go
+++ b/pkg/scheduler/apis/config/v1/defaults.go
@@ -69,9 +69,10 @@ var (
 
 	defaultQuotaGroupNamespace = "koordinator-system"
 
-	defaultMonitorAllQuotas       = pointer.Bool(false)
-	defaultEnableCheckParentQuota = pointer.Bool(false)
-	defaultEnableRuntimeQuota     = pointer.Bool(true)
+	defaultMonitorAllQuotas              = pointer.Bool(false)
+	defaultEnableCheckParentQuota        = pointer.Bool(false)
+	defaultEnableRuntimeQuota            = pointer.Bool(true)
+	defaultDisableDefaultQuotaPreemption = pointer.Bool(true)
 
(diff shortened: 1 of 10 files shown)


pkg/scheduler/apis/config/v1/defaults.go
```diff
@@ -69,9 +69,10 @@ var (
 
 	defaultQuotaGroupNamespace = "koordinator-system"
 
-	defaultMonitorAllQuotas       = pointer.Bool(false)
-	defaultEnableCheckParentQuota = pointer.Bool(false)
-	defaultEnableRuntimeQuota     = pointer.Bool(true)
+	defaultMonitorAllQuotas              = pointer.Bool(false)
+	defaultEnableCheckParentQuota        = pointer.Bool(false)
+	defaultEnableRuntimeQuota            = pointer.Bool(true)
+	defaultDisableDefaultQuotaPreemption = pointer.Bool(false)
```
saintube, line 75: how about setting the default as `true`

LennonChin: > how about setting the default as `true`

The default setting here is `false` to keep existing users in use so that there is no inconsistency with what is expected after the upgrade. Of course, it is also possible to set the default value to `true`.

saintube: The current behavior of preempting pods, when they did not set a valid quota, is dangerous instead of by design.

saintube (approved):
/lgtm

Issue #2346, opened 2025-02-25
[proposal] Support different isolation level of GPU sharing


zqzten:
**What is your proposal**:
Introduce some mechanism to differentiate a GPU share Pod's GPU resource isolation level.

Currently, there're two common levels:
* Isolated: Pod can only use (or see) the GPU resourece it requested. e.g. The GPU has 16GiB memory and the Pod request 8GiB, it can only use up to 8GiB.
* Shared: Pod can use (or see) the GPU resource up to the full card which assigned to it. e.g. The GPU has 16GiB memory and the Pod request 8GiB, it can use up to 16GiB.

On scheduling side, scheduler should make sure all the Pods which share the same GPU have the same isolation level.
On device side, koordlet should select the right GPU share method based on the isolation level.

**Why is this needed**:
It's common that user want to use different isolation levels for different applications in one cluster. The isolated is somewhat like k8s guaranteed and the shared is like burstable.

**Is there a suggested solution, if so, please add it**:

zqzten:
Currently it might be overkill to put Pods with different GPU isolation levels on the same Node. Distribute them to different Nodes is a cleaner way, which can be done without additional works by applying Node selector and the existing GPU isolation provider label. Let's hold this topic until we encounter further requirements.

PR #2455, opened 2025-05-23
proposal: GPU Sharing Isolation Level

zqzten:
This is the proposal of #2346

### DIFF
--- a/docs/proposals/scheduling/20250523-gpu-sharing-isolation-level.md
+++ b/docs/proposals/scheduling/20250523-gpu-sharing-isolation-level.md
@@ -1,0 +1,74 @@
+---
+title: GPU Sharing Isolation Level
+authors:
+  - "@zqzten"
+reviewers:
+  - "@ZiMengSheng"
+  - "@hormes"
+creation-date: 2025-05-23
+last-updated: 2025-05-23
+---
+
+# GPU Sharing Isolation Level
+
+## Table of Contents
+
+<!-- TOC -->
+* [GPU Sharing Isolation Level](#gpu-sharing-isolation-level)
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+  * [Proposal](#proposal)
+    * [User Story](#user-story)
+    * [API](#api)
+    * [Workflow](#workflow)
+<!-- TOC -->
+
+## Summary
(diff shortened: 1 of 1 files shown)


ZiMengSheng:
This will generate additional GPU allocation algorithm complexity, and the code processing needs to have FG or XXX to ensure:
In scenarios where this isolation level distinction is not required, scheduling performance will not be affected by this complexity

docs/proposals/scheduling/20250523-gpu-sharing-isolation-level.md
```diff
@@ -0,0 +1,79 @@
...
+
+For the first ones, user labels the pods to set their GPU isolation level to **Isolated**, so that they would get guaranteed GPU share resources, different pods sharing the same GPU card would not disturb each other. Thus, the stability of the workload is ensured.
+
+For the second ones, user labels the pods to set their GPU isolation level to **Shared**, so that they would get burstable GPU share resources, every pod sharing the same GPU card can take use of the full card. Thus, the GPU utilization is maximized. A typical use case is running two pods with different models on the same GPU card, a model router would route requests to one of the model, so that there would be only one active inferencing process running on the GPU at the same time. By sharing the same GPU card, both of the pods could utilize the computing power of full card.
+
+### API
+
+- Introduce a new Pod label `koordinator.sh/gpu-share-isolation-level` whose value is one of `Isolated` and `Shared`. In the future, we might introduce more isolation levels such as `CoreIsolated` and `MemoryIsolated` for more use cases.
```
hormes, line 63: This will complicate resource ops, as we need to look at the usage patterns of each GPU. Can this pattern be put at the node level to simplify management complexity?

zqzten: It can, but may reduce its flexibility. Sometimes user has few GPU machines and cannot afford to divide it at node level. Maybe we can provide best practice to users instead, that if conditions permit, it is recommended to divide different workloads with different usage patterns to different node pools.

hormes: At least we need a place to see the current status of a GPU card.

zqzten:
After reconsideration, currently it might be overkill to put Pods with different GPU isolation levels on the same Node. Distribute them to different Nodes is a cleaner way, which can be done without additional works by applying Node selector and the existing GPU isolation provider label. So let's hold this proposal until we encounter further requirements.
/close

Issue #2443, opened 2025-05-15
[BUG] ElasticQuota status.used err


LY-today:
**What happened**:
集群，Q资源貌似充足，但pod pending，等待长时间后可以running，属于偶发现象

**What you expected to happen**:
资源充足时，不长时间pending

**How to reproduce it (as minimally and precisely as possible)**:
同上

LY-today:
![Image](https://github.com/user-attachments/assets/fdb509d3-78d2-40bf-b1c1-48d00a1cf93f)
create pod 后预期应该cpu是9，但没有更新
delete pod 后预期应该cpu是5，更新正常

LY-today:
集群有剩余资源，quota也有资源，但是pod 申请 1C 2Mi 却pending。看到annotation更新正常

EQ信息
```
$ kubectl get eq quota-example -o yaml
apiVersion: scheduling.sigs.k8s.io/v1alpha1
kind: ElasticQuota
metadata:
  annotations:
    quota.scheduling.koordinator.sh/request: '{"cpu":"6","memory":"8Mi"}'
    quota.scheduling.koordinator.sh/runtime: '{"cpu":"6","memory":"8Mi"}'
    quota.scheduling.koordinator.sh/shared-weight: '{"cpu":"40","memory":"40Gi"}'
  creationTimestamp: "2025-05-15T13:34:12Z"
  generation: 28
  labels:
    quota.scheduling.koordinator.sh/is-parent: "false"
    quota.scheduling.koordinator.sh/parent: koordinator-root-quota
  name: quota-example
  namespace: default
  resourceVersion: "211631"
  uid: 4d464379-74ce-4dc6-a3b2-f32f0a462bd3
spec:
  max:
    cpu: "40"
    memory: 40Gi
  min:
    cpu: "10"
    memory: 20Mi
status:
  used:
    cpu: "5"
    memory: 6Mi
```

pod信息：
```
[...]

PR #2457, opened 2025-05-26
fix: reActivate the pending pod when Q is changed

LY-today:
issues：https://github.com/koordinator-sh/koordinator/issues/2443

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/controller.go
+++ b/pkg/scheduler/plugins/elasticquota/controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


ZiMengSheng:
After the Pod Add event is processed, the EQ update event should be triggered. At this time, the Pod that failed due to EQ should be fished out of UnschedulableQ. The phenomenon described in the issue may occur because the EQ event occurs during the scheduling failure process, and then the Pod is directly added to UnschedulableQ when the scheduling fails.

It is recommended to turn on the SchedulingQueuingHints FG provided by the upstream Kubernetes, which will handle the event when the scheduling fails, giving the Pod a chance to enter the BackoffQ.

ZiMengSheng:
this can be fixed by enable SchedulingQueueHints FG

PR #2462, opened 2025-05-30
koord-manager: support enhanced validation for pod

TaoYang526:

This PR introduces a new pod enhanced validator that provides configurable validation rules for pods with hot reload capability through ConfigMap updates. The validator supports namespace whitelisting and required labels validation, particularly useful for enforcing quota-related labels to ensure resource governance.

Configuration Example:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: pod-enhanced-validator-config
  namespace: koordinator-system
data:
  enable: "true"
  rules: |
    [
      {
        "name": "require-quota-label",
        "requiredLabels": ["quota.scheduling.koordinator.sh/name"],
        "namespaceWhitelist": ["kube-system", "koordinator-system"]
      },
      {
        "name": "require-app-labels",
        "requiredLabels": ["product", "app"],
        "namespaceWhitelist": ["kube-system"]
      }
    ]
```

### V. Checklist


TaoYang526:
@shaloulcy @saintube  Could you please help to review this PR? Thanks.

pkg/webhook/pod/validating/webhooks.go
```diff
@@ -53,5 +59,13 @@ func (b *podValidateBuilder) Build() admission.Handler {
 	quotaAccessor := quotaevaluate.NewQuotaAccessor(h.Client)
 	h.QuotaEvaluator = quotaevaluate.NewQuotaEvaluator(quotaAccessor, 16, make(chan struct{}))
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.EnablePodEnhancedValidator) {
+		kubeClient := extclient.GetGenericClientWithName(ClientNamePodEnhancedValidator).KubeClient
```
saintube: why not use the controller manager's client

pkg/webhook/pod/validating/pod_enhanced_validator.go
```diff
@@ -0,0 +1,309 @@
...
+			Enable: false,
+			Rules:  []ValidationRule{},
+		},
+	}
+}
+
+// Start initializes and starts the informer for hot reload
+func (m *PodEnhancedValidator) Start(ctx context.Context) error {
+	// create informer for hot reload
+	factory := informers.NewSharedInformerFactoryWithOptions(
```
saintube: I'm not sure if we need to run an individual configMap informer for this pod validation and maintain a cached config, which introduces non-negligible memory overhead and more complicated logic.
Maybe we can reuse the client of the controller runtime and just Get the cm during the validation.

TaoYang526: Thanks for the suggestion, yes, it's better to reuse the client of the controller manager instead of introducing an individual informer, but retrieving and parsing the ConfigMap during the validation still introduces non-negligible overhead. I think we can periodically reconcile the configuration that expected not to be updated frequently. Thoughts?

pkg/webhook/pod/validating/webhooks.go
TaoYang526: Since we can't get informer from that client before, here will be changed to reuse that client later.

saintube:
@TaoYang526 Please fix the UT failure.
https://github.com/koordinator-sh/koordinator/actions/runs/15431913931/job/43431172149?pr=2462

TaoYang526:
> @TaoYang526 Please fix the UT failure. https://github.com/koordinator-sh/koordinator/actions/runs/15431913931/job/43431172149?pr=2462

@saintube  Done, already fixed the race condition caused by shared scheme in fake client when executing UT in parallel.

pkg/webhook/pod/validating/webhooks.go
```diff
@@ -53,5 +58,12 @@ func (b *podValidateBuilder) Build() admission.Handler {
 	quotaAccessor := quotaevaluate.NewQuotaAccessor(h.Client)
 	h.QuotaEvaluator = quotaevaluate.NewQuotaEvaluator(quotaAccessor, 16, make(chan struct{}))
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.EnablePodEnhancedValidator) {
+		h.PodEnhancedValidator = NewPodEnhancedValidator(b.mgr.GetClient())
```
saintube: Please note that the controller manager and the informer have not started before the validator's build. So this may cause the validator to get an unsynchronized config.

TaoYang526: PodEnhancedValidator will fetch the ConfigMap directly via client#Get instead of using an informer. Therefore, this concern should be resolved, correct?

saintube: Perhaps it needs to get initialized before the validation.

TaoYang526: Make sense to me, I have refactored the PodEnhancedValidator to use lazy loading instead of explicit startup. Please help to review it again, Thanks!
Related commit: https://github.com/koordinator-sh/koordinator/pull/2462/commits/9aa5897d4fd95bfe456d11698764bc308cd679c4

saintube (approved):
/lgtm

TaoYang526:
Thanks @saintube for the review and commit!

PR #2463, opened 2025-06-03
koordlet: add psi qos reconciler

Wheat2018:

add psi qos manager, which now supports 4 operators:

1. PSIExport: Collects PSI metrics for Pods and reports them via Pod Conditions.
2. MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
3. GroupShare: Groups Pods and allows CPU weight sharing within a group.
4. BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods. 

### V. Checklist


saintube:
++ /cc @songtao98 for PSI collector

apis/slo/v1alpha1/nodeslo_types.go
```diff
@@ -426,6 +426,69 @@ type SystemStrategy struct {
...
+	CPU *PSIThreshold `json:"cpu,omitempty"`
+	// Memory PSI threshold
+	Memory *PSIThreshold `json:"memory,omitempty"`
+	// IO PSI threshold
+	IO *PSIThreshold `json:"io,omitempty"`
+}
+
+type PSIThreshold struct {
+	// Avg10 indicates the average 10-second PSI threshold, range [0,10000] indicating [0%,100%].
+	Avg10 int64 `json:"avg10,omitempty" validate:"min=0,max=10000"`
```
saintube, line 454: To enable the CRD validation, it needs some kubebuilder tags like:
```
// +kubebuilder:validation:Minimum=0
```
https://book.kubebuilder.io/reference/markers

pkg/koordlet/qosmanager/plugins/psi/cgroup/io.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (iops *Iops) SetPromise(newIops int64) error {
+	return fmt.Errorf("not implemented")
+}
+
+func slope(old, new int64, interval time.Duration) int64 {
+	return int64(float64(new-old) / interval.Seconds())
+}
+
+func max(a, b int64) int64 {
```
saintube, line 213: math.MaxInt64?

pkg/util/sloconfig/nodeslo_config.go
```diff
@@ -432,6 +432,47 @@ func DefaultSystemStrategy() *slov1alpha1.SystemStrategy {
 	}
 }
 
+func DefaultPSIStrategy() *slov1alpha1.PSIStrategy {
+	return &slov1alpha1.PSIStrategy{
+		PSIExport: &slov1alpha1.PSIExportConfig{
+			Enable: pointer.Bool(true),
```
saintube, line 438: For an alpha feature, please disable it by default.

pkg/koordlet/qosmanager/plugins/psi/psi_reconcile.go
```diff
@@ -0,0 +1,190 @@
...
+var _ framework.QOSStrategy = &psiReconcile{}
+
+type psiReconcile struct {
+	reconcileInterval time.Duration
+	statesInformer    statesinformer.StatesInformer
+	manager           *PsiManager
+}
+
+func (p *psiReconcile) Enabled() bool {
+	return features.DefaultKoordletFeatureGate.Enabled(features.BlkIOReconcile) && p.reconcileInterval > 0
```
saintube, line 48: > features.DefaultKoordletFeatureGate.Enabled(features.BlkIOReconcile)

Please add a new feature gate.

pkg/koordlet/qosmanager/plugins/psi/cgroup/helper.go
```diff
@@ -0,0 +1,422 @@
...
+		return nil, fmt.Errorf("failed to parse %s: %w", parts[0], err)
+	}
+	period, err := parseInt64(parts[1])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse %s: %w", parts[1], err)
+	}
+	return &CpuQuota{Quota: quota, Period: period}, nil
+}
+
+func WriteCpuMax(cgroupPath string, max *CpuQuota) error {
```
saintube, line 141: please reuse the resource executor and the cgroup resource:

- https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/resourceexecutor/reader.go#L29
- https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/resourceexecutor/updater.go
- https://github.com/koordinator-sh/koordinator/blob/main/pkg/koordlet/util/system/cgroup_resource.go#L207

saintube (commented):
It's a delicate job, but it's also complex :) So please add a document describing its design and how to use it when you have time.

AutuSnow:
I have thoroughly read this code, and the general logic of the code is correct. However, a major problem is that it is duplicating the creation of wheels. pkg/coordlet/utl/system/has already implemented operations on resources, while pkg/coordlet/resourceexecutor/has implemented operations on the cgroup system, And this qosmanager seems to violate the existing plugin architecture, directly using runc/libcontainer/cgroups.ReadFile/WriteFile Operating files instead of obtaining them through framework.options,I'm not sure if what I said is correct, could you help me correct it @saintube

pkg/koordlet/qosmanager/plugins/psi/psi_manager.go
```diff
@@ -0,0 +1,191 @@
...
+		go func(op operator.Operator) {
+			if err := op.Exec(pm.pods, node); err != nil {
+				errCh <- fmt.Errorf("failed to perform %s: %w", op.Name(), err)
+			} else {
+				errCh <- nil
+			}
+		}(op)
+	}
+	for range pm.operators {
+		err = errors.Join(<-errCh)
```
AutuSnow, line 180: This will only retain the last err, which should be 
err=errors.Join (err,<- errCh)

pkg/koordlet/qosmanager/plugins/psi/psi_manager.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, pc := range pm.pods {
+		if err := pc.Load(pm.mask); err != nil {
+			klog.ErrorS(err, "Failed to load cgroup", "pod", klog.KObj(pc.Pod))
+		}
+	}
+
+	errCh := make(chan error)
+	for _, op := range pm.operators {
+		go func(op operator.Operator) {
+			if err := op.Exec(pm.pods, node); err != nil {
```
AutuSnow, line 172: Simultaneously starting multiple goroutines to execute operators while holding podsLock. Will there be a data race when multiple operators read and write data from the pm.pods map simultaneously.

pkg/koordlet/qosmanager/plugins/psi/cgroup/io.go
```diff
@@ -0,0 +1,218 @@
...
+	iops.Base.Throttle = newIops
+	return nil
+}
+
+func (iops *Iops) SetPromise(newIops int64) error {
+	return fmt.Errorf("not implemented")
+}
+
+func slope(old, new int64, interval time.Duration) int64 {
+	return int64(float64(new-old) / interval.Seconds())
```
AutuSnow, line 210: I'm a bit worried about this place. Will there be a situation of/0? Do we need to do interval<=0 for protection?

pkg/koordlet/qosmanager/plugins/psi/operator/memory_suppress.go
```diff
@@ -0,0 +1,172 @@
...
+	spot := math.Max(current/burst, ms.MinSpot)
+
+	if spot >= ms.MaxSpot { // accumulation of pressure in max spot for killing pod
+		ms.maxSpotCounter[r.Pod.UID] += r.Pressure().Rate()
+		if ms.maxSpotCounter[r.Pod.UID] > 0.5*float64(ms.KillPeriods) {
+			klog.InfoS("Pod in max pressure spot", "pod", klog.KObj(r.Pod), "spot", spot,
+				"accumulation", fmt.Sprintf("%.2f/%d", ms.maxSpotCounter[r.Pod.UID], ms.KillPeriods))
+		}
+		if ms.maxSpotCounter[r.Pod.UID] > float64(ms.KillPeriods) {
+			klog.InfoS("Pod in max pressure spot excceed max periods, kill it", "pod", klog.KObj(r.Pod), "spot", spot,
```
AutuSnow, line 127: The 'kill it' output of this seems to be inconsistent with the behavior it performs..

Wheat2018: The "kill it" means to remove the `memory.high` limit. By eliminating the soft upper limit protection on memory, it allows the memory of process to exceed `memory.max` and thus be OOM killed. The PSI manager never actively kills processes but rather lets the processes' own active memory allocation behavior lead to their own termination.

Wheat2018:
Since this project was proposed, I have experienced some changes in my work, so I may not have the energy to provide support for code improvement at the moment. If someone is interested in this project and willing to be a co-collaborator to advance it, I can offer whatever help I can. Thankss

AutuSnow:
> * MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
> * GroupShare: Groups Pods and allows CPU weight sharing within a group.
> * BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods.

If I have time, I will handle this PR. Of course, I would prefer if you could agree, and there is a prerequisite for enabling PSI detection, which is kernel version>=4.20

Wheat2018:
> > * MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
> > * GroupShare: Groups Pods and allows CPU weight sharing within a group.
> > * BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods.
> 
> If I have time, I will handle this PR. Of course, I would prefer if you could agree, and there is a prerequisite for enabling PSI detection, which is kernel version>=4.20

yep, in fact, the code only supports the use of PSI Manager under the cgroup v2 hierarchy, which means the kernel needs to be >= 5.8.0.

AutuSnow:
> > > * MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
> > > * GroupShare: Groups Pods and allows CPU weight sharing within a group.
> > > * BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods.
> > 
> > 
> > If I have time, I will handle this PR. Of course, I would prefer if you could agree, and there is a prerequisite for enabling PSI detection, which is kernel version>=4.20
> 
> yep, in fact, the code only supports the use of PSI Manager under the cgroup v2 hierarchy, which means the kernel needs to be >= 5.8.0.

This is not necessarily the case. Various pirated systems nowadays may encounter unexpected situations, and conducting pre check is also an effective way to reduce code failures

Wheat2018:
> > > > * MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
> > > > * GroupShare: Groups Pods and allows CPU weight sharing within a group.
> > > > * BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods.
> > > 
> > > 
> > > If I have time, I will handle this PR. Of course, I would prefer if you could agree, and there is a prerequisite for enabling PSI detection, which is kernel version>=4.20
> > 
> > 
> > yep, in fact, the code only supports the use of PSI Manager under the cgroup v2 hierarchy, which means the kernel needs to be >= 5.8.0.
> 
> This is not necessarily the case. Various pirated systems nowadays may encounter unexpected situations, and conducting pre check is also an effective way to reduce code failures

That makes sense.

AutuSnow:
> > > > > * MemorySuppress: Applies pressure to Pod memory allocation, increasing with the growth of allocated memory.
> > > > > * GroupShare: Groups Pods and allows CPU weight sharing within a group.
> > > > > * BudgetBalance: Balances CPU usage among Pods over time, beneficial for burstable Pods.
> > > > 
> > > > 
> > > > If I have time, I will handle this PR. Of course, I would prefer if you could agree, and there is a prerequisite for enabling PSI detection, which is kernel version>=4.20
> > > 
> > > 
> > > yep, in fact, the code only supports the use of PSI Manager under the cgroup v2 hierarchy, which means the kernel needs to be >= 5.8.0.
> > 
> > 
> > This is not necessarily the case. Various pirated systems nowadays may encounter unexpected situations, and conducting pre check is also an effective way to reduce code failures
> 
> That makes sense.

In order to better promote this development, can you provide an LSSues proposal where we can discuss details and add it to my to-do list

saintube:
Proposal: #2840 
cc @AutuSnow

ZiMengSheng:
Thanks for the substantial work here! PSI-based QoS is being re-chartered as a design-first effort in #2840 (PSI QoS strategy proposal), so implementation should follow that proposal's API and reconciliation design once it lands. Closing this implementation to avoid diverging from that direction — please feel free to rebase the work onto the new proposal once it's accepted. Thanks for the contribution!

PR #2423, opened 2025-04-25
proposal: heterogeneous GPU device reporting

ZhuZhezz:
### Ⅰ. Describe what this PR does
This is the proposal of #2418

Issue #2335, opened 2025-02-11
[Umbrella] Fine-grained device scheduling support Huawei Ascend NPU


zqzten:
This is an umbrella issue tracking the work of an e2e solution for supporting fine-grained device scheduling for Huawei Ascend NPU.

- [ ] support fine-grained device scheduling for NPU (full card, honor the HCCS topology)
- [ ] support fine-grained device scheduling for vNPU (NPU dynamic virtualizing)
- [ ] support NPU device reporting
- [ ] support NPU/vNPU allocating

PR #2467, opened 2025-06-05
scheduler: fine-grained device scheduling support Huawei Ascend NPU (full card)

zqzten:

This is the first time that we introduce a heterogeneous GPU (other than NVIDIA) to fine-grained device scheduling, so this PR also contains some fundamental works which would help us integrate other heterogeneous devices easier in the future.

Let's first explain the fundamental works:

As we all know, there're three major phases in the lifecycle of fine-grained device scheduling: **reporting, scheduling, allocating**.

**For reporting**, currently koordlet only supports reporting NVIDIA GPUs, but #2423 is working on heterogeneous GPU reporting including Huawei NPU, so **this PR only introduces the missing vendor label to device/node (which is necessary for the following works) and leaves the rest works to the implementation of #2423**.

**For scheduling**, current device share scheduling is rather vendor agnostic, so **this PR does not modify the core scheduling logic**.

**For allocating**, currently koordlet only supports allocate NVIDIA GPUs, if we want it to support other heterogeneous GPUs, we might consider integrate codes from device plugins provided by vendor of these GPUs which would soon make koordlet hard to maintain. So, can we just reuse these existing device plugins? The answer is YES.

**This PR introduces a device plugin adaption mechanism to device share scheduling**, which makes koord-scheduler able to write allocation results in some format that third party vendor's device plugins recognize, so that user can directly use them as allocators with koord-scheduler without modification. This is especially useful when third party device plugin natively supports fine-grained device allocation or virtualization.

---

Now let's focus on the adaption works for Huawei Ascend NPU. Note that this PR only covers full card scheduling, partial card (vNPU) scheduling would be covered in latter PRs.

Unlike NVIDIA GPUs, Huawei Ascend NPU uses the accurate number of AI cores of an NPU to represent the computing power resource instead of a ratio. So this PR introduces a new resource `huawei.com/npu-core` which represents the number of AI cores of Huawei NPU to device share scheduling, this resource is reported and recognized by the device plugin of Huawei NPU and would also be used in vNPU scheduling.

So from the user's perspective, requesting X full NPUs would be:

```yaml
huawei.com/npu-core: X*(the number of AI cores of the requested NPU model)
koordinator.sh/gpu-memory-ratio: X*100
```

Other combinations are not supported due to obvious reasons.

Another special point of Huawei NPU full card scheduling is that during multi card scheduling, it is required to allocate NPUs according to some hardware topology ([ref](https://www.hiascend.com/document/detail/zh/mindx-dl/600/clusterscheduling/ref/affinityschedulesd/dl_affinity_0005.html)). This can be achieved by the existing GPU partition feature of device share scheduling so this PR does not cover this.

part of #2335

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
+++ b/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deviceshare
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+
+	apiext "github.com/koordinator-sh/koordinator/apis/extension"
+	schedulingv1alpha1 "github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1"
+	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext"
+)
+
+const (
+	// AnnotationBindTimestamp represents the bind time (unix nano) of the pod which can be used by
+	// custom device plugins which are adapted for koord-scheduler to determine pod.
+	//
+	// Background: Device plugins cannot get pod manifest (including annotations) from kubelet so they have to
+	// use the bind time to determine the target pod when multiple pods are assigned to the same node at the same time.
+	AnnotationBindTimestamp = apiext.SchedulingDomainPrefix + "/bind-timestamp"
+
+	// AnnotationPredicateTime represents the bind time (unix nano) of the pod which is used by
+	// Huawei NPU device plugins to determine pod.
+	AnnotationPredicateTime = "predicate-time"
+	// AnnotationHuaweiNPUCore represents the NPU/vNPU allocation result for the pod which is used by
+	// Huawei NPU device plugins to allocate NPU(s)/vNPU for pod.
+	AnnotationHuaweiNPUCore = "huawei.com/npu-core"
+)
+
+// DevicePluginAdapter adapt koord-scheduler's device allocation result to the format that third party vendor's
+// device plugin recognizes, so that user can directly use them as allocators with koord-scheduler without modification.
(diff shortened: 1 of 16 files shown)


ZiMengSheng:
bravo

pkg/slo-controller/noderesource/plugins/gpudeviceresource/plugin.go
```diff
@@ -185,7 +188,12 @@ func (p *Plugin) calculate(node *corev1.Node, device *schedulingv1alpha1.Device)
 		totalKoordGPU.Add(d.Resources[extension.ResourceGPUCore])
 	}
 
-	gpuResources[extension.ResourceGPU] = *totalKoordGPU
+	gpuVendor := device.Labels[extension.LabelGPUVendor]
```
ZiMengSheng, line 191: Some comments are needed here, such as how huawei.com/npu-core on the Node comes from, and why the controller does not need to report

pkg/scheduler/plugins/deviceshare/devicehandler_gpu.go
```diff
@@ -70,6 +71,9 @@ func calcDesiredRequestsAndCountForGPU(podRequests corev1.ResourceList) (corev1.
 	if coreExists {
 		requests[apiext.ResourceGPUCore] = *resource.NewQuantity(gpuCore.Value()/desiredCount, resource.DecimalSI)
 	}
+	if huaweiNPUCoreExists {
```
ZiMengSheng, line 74: Why does DeviceShare need to support NPUCore allocation? Shouldn't NPU Core just trigger Huawei's DP?

pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext"
+)
+
+const (
+	AnnotationBindTimestamp = apiext.SchedulingDomainPrefix + "/bind-timestamp"
+
+	AnnotationPredicateTime = "predicate-time"
+	AnnotationHuaweiNPUCore = "huawei.com/npu-core"
+)
+
```
ZiMengSheng, line 48: Here we may need more comments to explain why we need to bind these annotations and what the coordination mechanism is.

pkg/slo-controller/noderesource/plugins/gpudeviceresource/plugin.go
zqzten: The `huawei.com/npu-core` resource is still reported by controller, though Huawei NPU device plugin would also report it, we keep it for potential future impls without device plugins, and there's no need to introduce special logic for it. The count reported by controller and dp is the same so there's no conflict.

P.S. The only special case for Huawei NPU is that the controller would not report `koordinator.sh/gpu` resource, which has been commented.

pkg/scheduler/plugins/deviceshare/devicehandler_gpu.go
zqzten: It would be needed in vNPU scheduling, just like the `koordinator.sh/gpu-core` resource for NVIDIA GPUs, so we want to keep the behavior consistent.

PR #2484, opened 2025-06-17
fix: GetLocalStorageInfo maybe hang

dabaooline:
… volumes use lvm command simultaneously.

    fix: GetLocalStorageInfo maybe hang, such as other component creating volumes use lvm command simultaneously.

   NONE

If creating an lvm volume takes a lot of time and executing other lvm commands at the same time, the command may hang.


jasonliu747 (commented):
/lgtm

Issue #2480, opened 2025-06-13
[proposal] Improve Coscheduling Capabilities


ZiMengSheng:
**What is your proposal**:

Gang 是 AI 场景下调度器的必备能力，在社区 scheduler 这种 Pod 作为一等公民的框架上增加扩展，使它原生支持Gang 是一件非常有挑战性的事情。Koordinator 已经提供了 Coscheduling 插件和基础能力，但是在用户体验和一些边界情况下仍然有不足之处，需要优化，包括但不限于：
1. Pod 成员不足的时候会被 Coscheduling PreEnqueue 给拦住，Pod 上就没有任何 Event 了，所以这里需要有一个控制器为这种情况输出有效信息
2. Koordinator Coscheduling 采用 PreEnqueue 使得 ActiveQ 中只有一个代表 Pod。但是在调度失败时，需要切换已经处于 Waiting 状态的代表 Pod 到当前失败 Pod，这种切换在 PostFilte 阶段完成，会和抢占有所冲突。所以这里需要一种方式使得切换 代表 Pod 的逻辑一定会被执行且不会跟抢占冲突。一种可能的方式是提供一种新的 BeforePostFilter 来放这个逻辑。

Gang is a must-have capability for schedulers in AI scenarios. It is very challenging to add extensions to the community scheduler framework, where Pod is a first-class citizen, to make it natively support Gang. Koordinator has provided Coscheduling plug-ins and basic capabilities, but there are still deficiencies in user experience and some edge cases, which need to be optimized, including but not limited to:
1. When there are insufficient Pod members, it will be blocked by Coscheduling PreEnqueue, and there will be no Event on the Pod, so there needs to be a controller to output valid information for this situation
2. Koordinator Coscheduling uses PreEnqueue so that there is only one representative Pod in ActiveQ. However, when scheduling fails, it is necessary to switch the representative Pod that is already in the Waiting state to the current failed Pod. This switch is completed in the PostFilter stage, which will conflict with preemption. Therefore, a way is needed here to ensure that the logic of switching the representative Pod will be executed and will not conflict with preemption. One possible way is to provide a new BeforePostFilter to put this logic.

**Why is this needed**:

Improve user experience and logical completeness

**Is there a suggested solution, if so, please add it**:
 
As described in Proposal

ZiMengSheng:
Coscheduling Capacity:
- https://github.com/koordinator-sh/koordinator/pull/2417
[...]

PR #2490, opened 2025-06-25
scheduler: put Coscheduling postFilter logic in afterPostFilter

ZiMengSheng:

Abstract the AfterPostFilter extension point and put the Coscheduling PostFilter logic in it to achieve the following:

1. The Reject sibling Pod logic of Coscheduling should not be executed before the Job-level preemption logic. The Job-level preemption plug-in should only initiate preemption for the currently unscheduled member Pods, so there should be no difference between the resource view seen during the preemption process and the resource view seen during the scheduling process
2. Under the premise of successful preemption, it is necessary to ensure that ReplaceRepresentative will be executed and the sibling Pod will fail quickly, which requires that the relevant logic in Coscheduling PostFilter will be executed

fixes part of #2480 

After the Job-level preemption plug-in successfully preempts, it needs to do resource accounting for the Pod in the WaitingPod.

### V. Checklist


pkg/scheduler/plugins/coscheduling/coscheduling_test.go
```diff
@@ -352,10 +352,7 @@ func TestPostFilter(t *testing.T) {
 				}
 			}
 			// reject waitingPods
-			_, code := gp.PostFilter(context.Background(), cycleState, tt.pod, nil)
```
saintube, line 355: Keep this and test both?

PR #2492, opened 2025-06-25
descheduler: fix podFitsAnyNodeWithThreshold when node not fit pod

dabaooline:

fix podFitsAnyNodeWithThreshold when node not fit pod revert the added resource.

None


dabaooline:
I don't think podFitsAnyNodeWithThreshold is a perfect filtering method. It  don't  pre reduce the resources of Fit's Pod.

ZiMengSheng:
@songtao98 PTAL

songtao98:
@dabaooline nice catch! If exceeded then the resource need to give back to the node. Could you add more Unit Tests as possible?

dabaooline:
> @dabaooline nice catch! If exceeded then the resource need to give back to the node. Could you add more Unit Tests as possible?

ok,I will try.

dabaooline:
@songtao98 PTAL. The ut has been added.

PR #2494, opened 2025-06-26
scheduler: set PodGroup's OccupiedBy field  correctly

googs1025:
Currently, PodGroup OccupiedBy field is not set correctly.

- set PodGroup's OccupiedBy field  correctly

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/controller/podgroup.go
+++ b/pkg/scheduler/plugins/coscheduling/controller/podgroup.go
@@ -315,14 +315,14 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup) 
 }
 
 func fillOccupiedObj(pg *schedv1alpha1.PodGroup, pod *v1.Pod) {
+	if len(pod.OwnerReferences) == 0 {
+		return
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/coscheduling/controller/podgroup.go
```diff
@@ -315,14 +315,14 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup)
...
 
 func fillOccupiedObj(pg *schedv1alpha1.PodGroup, pod *v1.Pod) {
+	if len(pod.OwnerReferences) == 0 {
+		return
+	}
 	var refs []string
 	for _, ownerRef := range pod.OwnerReferences {
 		refs = append(refs, fmt.Sprintf("%s/%s", pod.Namespace, ownerRef.Name))
 	}
-	if len(pg.Status.OccupiedBy) == 0 {
```
googs1025, line 322: This seems to be a bug. status OccupiedBy is always empty, so we can't set it

saintube:
Left the approval for @ZiMengSheng

PR #2496, opened 2025-06-26
scheduler: fine-grained device scheduling support Huawei Ascend vNPU

zqzten:

This is the following PR of #2467, some backgrounds can be found there.

The mechanism of Huawei Ascend vNPU is very different from NVIDIA vGPU, details can be found [here](https://www.hiascend.com/document/detail/zh/mindx-dl/600/clusterscheduling/clusterscheduling/cpaug/mxdlug_021.html), in short, there're mainly two differences:
* Besides AI core and memory, vNPU also virtualizes other NPU resources including AI CPU and DVPP modules.
* Only some combinations of vNPU resources are valid, such combinations are called **templates**.

In order to support vNPU, this PR introduces GPU shared resource templates to fine-grained device scheduling, the available templates can be configured in a specific ConfigMap, and scheduler would check if Pod requests match any of the template which is available for the GPU model of the Node.

For example, a Pod can request a vNPU by specifying following resourcs:
```yaml
koordinator.sh/gpu.shared: "1"
huawei.com/npu-core: "2"
huawei.com/npu-cpu: "2"
huawei.com/npu-dvpp: "25"
koordinator.sh/gpu-memory: 6Gi
```
This would match the `vir02` template of Ascend 310P.

Note that we abstract resources of DVPP modules (including VDEC, JPEGD, PNGD, etc.) to a ratio resource called `huawei.com/npu-dvpp` to simplify the request for DVPP modules. For DVPP, there're three different kinds of templates:
* Request full DVPP modules: this would match `huawei.com/npu-dvpp: "100"`
* Request no DVPP module: in this case, `huawei.com/npu-dvpp` resource would not be requested
* Does not care DVPP modules: this would match `huawei.com/npu-dvpp: X` in which `X` stands for the ratio of DVPP modules (which is equal to the ratio of AI cores) of the template.

Currently it is still a little complicated for users to request vNPU (though accurate), latter PR would introduce some mechanism which can automatically choose a template which meets user's requirement, similar to the [use case of Volcano](https://www.hiascend.com/document/detail/zh/mindx-dl/600/clusterscheduling/clusterscheduling/cpaug/cpaug_0020.html).

part of #2335

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/loadaware/utilization_util.go
+++ b/pkg/descheduler/framework/plugins/loadaware/utilization_util.go
@@ -746,9 +746,11 @@ func podFitsAnyNodeWithThreshold(nodeIndexer podutil.GetPodsAssignedToNodeFunc, 
 					thresholds = nodeThreshold.highResourceThreshold
 				}
 				exceeded := false
+				preReducedResources := make([]corev1.ResourceName, 0, len(thresholds))
 				for resourceName, threshold := range thresholds {
 					if used := usage[resourceName]; used != nil {
 						used.Add(podMetric.ResourceList[resourceName])
+						preReducedResources = append(preReducedResources, resourceName)
 						if used.Cmp(*threshold) > 0 {
 							exceeded = true
 							break
@@ -758,6 +760,12 @@ func podFitsAnyNodeWithThreshold(nodeIndexer podutil.GetPodsAssignedToNodeFunc, 
 				}
 				if exceeded {
 					klog.V(4).InfoS("Pod may cause node over-utilized", "pod", klog.KObj(pod), "node", klog.KObj(node))
+					// revert the change
+					for _, resourceName := range preReducedResources {
+						if used := usage[resourceName]; used != nil {
+							used.Sub(podMetric.ResourceList[resourceName])
+						}
+					}
 					continue
 				}
 			}
--- a/pkg/descheduler/framework/plugins/loadaware/utilization_util_test.go
+++ b/pkg/descheduler/framework/plugins/loadaware/utilization_util_test.go
@@ -333,10 +333,43 @@ func TestPodFitsAnyNodeWithThreshold(t *testing.T) {
 
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
-
+			copiedNodesUsage := sumNodeUsage(tt.nodeUsages)
 			if got := podFitsAnyNodeWithThreshold(getPodsAssignedToNode, tt.pod, tt.nodes, tt.nodeUsages, tt.nodeThresholds, false, tt.podMetric); got != tt.want {
 				t.Errorf("PodFitsAnyNode() = %v, want %v", got, tt.want)
+			} else {
+				currentTotalUsage := sumNodeUsage(tt.nodeUsages)
+				if !tt.want {
+					// if the pod not fits, the usage should revert to the original state
+					for resourceName, quantity := range copiedNodesUsage.usage {
+						assert.Equal(t, quantity.Value(), currentTotalUsage.usage[resourceName].Value(), "usage should not change")
+					}
+					for resourceName, quantity := range copiedNodesUsage.prodUsage {
+						assert.Equal(t, quantity.Value(), currentTotalUsage.prodUsage[resourceName].Value(), "prodUsage should not change")
+					}
+				}
 			}
 		})
 	}
+}
+
+func sumNodeUsage(nodeUsages map[string]*NodeUsage) *NodeUsage {
+	totalUsage := &NodeUsage{
+		usage:     make(map[corev1.ResourceName]*resource.Quantity),
+		prodUsage: make(map[corev1.ResourceName]*resource.Quantity),
+	}
+	for _, usage := range nodeUsages {
+		for resourceName, quantity := range usage.usage {
+			if totalUsage.usage[resourceName] == nil {
+				totalUsage.usage[resourceName] = resource.NewQuantity(0, quantity.Format)
+			}
+			totalUsage.usage[resourceName].Add(*quantity)
+		}
+		for resourceName, quantity := range usage.prodUsage {
+			if totalUsage.prodUsage[resourceName] == nil {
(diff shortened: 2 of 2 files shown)


pkg/scheduler/plugins/deviceshare/allocator_gpu.go
```diff
@@ -114,6 +122,63 @@ func removeZeroDevice(originalResources deviceResources) deviceResources {
...
+
+	return defaultAllocateDevices(nodeDevice, requestCtx, requestCtx.gpuRequirements.requestsPerGPU, desiredCount, maxDesiredCount, schedulingv1alpha1.GPU, nil)
+}
+
+func (a *GPUAllocator) allocateByTemplate(requestCtx *requestContext, nodeDevice *nodeDevice, desiredCount int, maxDesiredCount int, allocateContext *AllocateContext) ([]*apiext.DeviceAllocation, *framework.Status) {
+	if !requestCtx.gpuRequirements.gpuShared {
+		return nil, nil
+	}
+	vendor, model := requestCtx.node.Labels[apiext.LabelGPUVendor], requestCtx.node.Labels[apiext.LabelGPUModel]
+	if vendor != apiext.GPUVendorHuawei {
```
ZiMengSheng: why this is hardcoded?

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
```diff
@@ -114,6 +122,63 @@ func removeZeroDevice(originalResources deviceResources) deviceResources {
...
+			return nil, framework.NewStatus(framework.Error, err.Error())
+		}
+		var templateName string
+		for name, template := range templates {
+			if quotav1.Equals(requestsPerGPU, template) {
+				templateName = name
+				break
+			}
+		}
+		if templateName == "" {
```
ZiMengSheng: Can this logic be placed in the Prefilter stage to reduce the performance overhead of node-level allocation?

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
zqzten: Because we want to schedule by template only for vendors which enforce it, if we don't filter here, this would cause other irrelevant vendors meet the unexpected `no matched GPU shared resource template` error.

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
zqzten: No, because different nodes with different GPU models may have different templates, so the template match can only be done at node level.

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
ZiMengSheng: What I mean is whether the information of "which manufacturers need to be allocated according to the template" can be configured

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
ZiMengSheng: I know, but some pre-computation is possible. For example, can we calculate the template name that can be matched for each vendor-model in advance according to the Pod application in PreFilter? Here, we only need to map the template name according to the vendor-model?

pkg/scheduler/plugins/deviceshare/allocator_gpu.go
zqzten: Got it. However, after investigation, I found that configuring "which resources should be matched for template" is more suitable so I implemented this instead.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -72,25 +72,29 @@ var (
...
+	skip                                bool
+	podRequests                         map[schedulingv1alpha1.DeviceType]corev1.ResourceList
+	hints                               apiext.DeviceAllocateHints
+	hintSelectors                       map[schedulingv1alpha1.DeviceType][2]labels.Selector
+	hasSelectors                        bool
+	jointAllocate                       *apiext.DeviceJointAllocate
+	primaryDeviceType                   schedulingv1alpha1.DeviceType
+	podFitsSecondaryDeviceWellPlanned   bool
+	gpuRequirements                     *GPURequirements
+	enforceGPUSharedResourceTemplate    bool
```
ZiMengSheng: Can these two parameters be put into preFilterState.GPURequirements? If not, then the preFilterState.Clone() L113 function needs to clone these two parameters as well.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -655,9 +659,16 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error)
...
+		args.GPUSharedResourceTemplatesConfig.ConfigMapNamespace, args.GPUSharedResourceTemplatesConfig.ConfigMapName,
+		handle.SharedInformerFactory())
+
 	return &Plugin{
-		handle:                             extendedHandle,
-		nodeDeviceCache:                    deviceCache,
+		handle:                          extendedHandle,
+		nodeDeviceCache:                 deviceCache,
+		gpuSharedResourceTemplatesCache: gpuSharedResourceTemplatesCache,
+		gpuSharedResourceTemplatesMatchedResources: args.GPUSharedResourceTemplatesConfig.MatchedResources,
```
ZiMengSheng, line 671: If GPUSharedResourceTemplatesConfig is not set, the scheduler panics. This may be the reason why e2e keeps failing.

ZiMengSheng: image

You need to modify the file config/rbac/scheduler_role.yaml so that the scheduler in the e2e environment can perceive the configmapv

pkg/scheduler/plugins/deviceshare/plugin.go
zqzten: Moved to preFilterState.GPURequirements

pkg/scheduler/plugins/deviceshare/plugin.go
zqzten: RBAC issue, solved.

PR #2497, opened 2025-06-27
scheduler: use RWMutex instead of Mutex

googs1025:
- use RWMutex instead of Mutex

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/controller/controller.go
+++ b/pkg/scheduler/plugins/reservation/controller/controller.go
@@ -68,7 +68,7 @@ type Controller struct {
 	numWorker                  int
 	gcDuration                 time.Duration
 
-	lock sync.Mutex
+	lock sync.RWMutex
 	pods map[string]map[types.UID]*corev1.Pod
 }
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/reservation/controller/controller.go
```diff
@@ -270,7 +270,7 @@ func (c *Controller) updateReservationStatus(reservation *schedulingv1alpha1.Res
 
 func isReservationNeedExpiration(r *schedulingv1alpha1.Reservation) bool {
 	// 1. failed or succeeded reservations does not need to expire
-	if r.Status.Phase == schedulingv1alpha1.ReservationFailed || r.Status.Phase == schedulingv1alpha1.ReservationSucceeded {
+	if reservationutil.IsReservationFailed(r) || reservationutil.IsReservationSucceeded(r) {
```
googs1025, line 273: just use `reservationutil`

pkg/scheduler/plugins/reservation/controller/pod_eventhandler.go
```diff
@@ -100,14 +100,14 @@ func (c *Controller) deletePod(pod *corev1.Pod) {
 }
 
 func (c *Controller) getPods(nodeName string) map[types.UID]*corev1.Pod {
```
googs1025, line 102: I'm not sure if this is too complicated, why do we need to make a deep copy again? 🤔

saintube: You mean to return the `c.pods[nodeName]`? It is not thread-safe.

googs1025: got this. 😄

PR #2503, opened 2025-07-02
manager: colocationprofile support appending label suffix

saintube:

ClusterColocationProfile supports appending label suffixes.

It helps when to inject different label values with their labels and some suffixes. 

For example, we want to label the pod with its quota name and priority class.

### V. Checklist


zwzhang0107:
need an example

saintube:
> need an example

Added.

PR #2510, opened 2025-07-04
statesinformer: sort podsMetricInfo and hostAppMetricInfo by name for consistent output

googs1025:

When we use `kubectl get nodemetrics <nodeName> -oyaml` every time, we will encounter the problem of PodMetricInfo reordering (each sync will be different). This makes it difficult to troubleshoot the problem. We can sort by name every time we report.

None

### V. Checklist


pkg/koordlet/statesinformer/impl/states_nodemetric.go
```diff
@@ -387,6 +388,9 @@ func (r *nodeMetricInformer) collectMetric() (*slov1alpha1.NodeMetricInfo, []*sl
 		}
 		podsMetricInfo = append(podsMetricInfo, podMetric)
 	}
+	sort.Slice(podsMetricInfo, func(i, j int) bool {
+		return podsMetricInfo[i].Name < podsMetricInfo[j].Name
```
saintube, line 395: probably namespaced name

googs1025: done

pkg/koordlet/statesinformer/impl/states_nodemetric.go
```diff
@@ -387,6 +388,11 @@ func (r *nodeMetricInformer) collectMetric() (*slov1alpha1.NodeMetricInfo, []*sl
 		}
 		podsMetricInfo = append(podsMetricInfo, podMetric)
 	}
+	sort.Slice(podsMetricInfo, func(i, j int) bool {
+		iNamespaceName := podsMetricInfo[i].Namespace + "/" + podsMetricInfo[i].Name
```
saintube: It does not need a string concatenation. We can compare the ns and then compare the name.

googs1025: done

Issue #2491, opened 2025-06-25
[BUG] After deleting an ElasticQuota, its associated metrics can still be queried.


qinfustu:
**What happened**:

After deleting an ElasticQuota, its associated metrics can still be queried.

​​Reproduction Steps:​​
1. Fetch metrics for a specific ElasticQuota.
2. Delete the ElasticQuota CR.
3. Query the same ElasticQuota metrics again — they still persist.

**What you expected to happen**:

After deleting an ElasticQuota, its associated metrics should no longer be queryable. The metrics collection logic fails to listen for ElasticQuota deletion events, preventing the removal of associated metrics.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version: v1.16
- Kubernetes version (use `kubectl version`): 1.28
- Install details (e.g. helm install args): helm
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

ZiMengSheng:
Yep, maybe we need to manually delete the metric in the QuotaDelete event, like "ElasticQuotaSpecMetric.DeleteLabelValues(nodeName)"

qinfustu:
@ZiMengSheng I'm attempting to fix this issue.

PR #2516, opened 2025-07-08
scheduler: fix after deleting an ElasticQuota, its associated metrics can still be queried

qinfustu:

After deleting an ElasticQuota, its associated metrics can still be queried

Fix https://github.com/koordinator-sh/koordinator/issues/2491

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/controller_test.go
+++ b/pkg/scheduler/plugins/elasticquota/controller_test.go
@@ -396,34 +396,8 @@ func Test_syncElasticQuotaMetrics(t *testing.T) {
 		Allocated:             MakeResourceList().CPU(7).Mem(50).Obj(),
 		Guaranteed:            MakeResourceList().CPU(8).Mem(50).Obj(),
 	}
-	syncElasticQuotaMetrics(eq, summary)
-	metricsCh := make(chan prometheus.Metric, 10)
-	go func() {
-		ElasticQuotaSpecMetric.Collect(metricsCh)
-		ElasticQuotaStatusMetric.Collect(metricsCh)
-		close(metricsCh)
-	}()
-	var ms []prometheus.Metric
-loopChan:
-	for {
-		select {
-		case metric, ok := <-metricsCh:
-			if !ok {
-				break loopChan
-			}
-			ms = append(ms, metric)
-		}
-	}
-
-	var dtoMetrics []*dto.Metric
-	for _, v := range ms {
-		m := dto.Metric{}
-		assert.NoError(t, v.Write(&m))
-		dtoMetrics = append(dtoMetrics, &m)
-	}
-	sort.Slice(dtoMetrics, func(i, j int) bool {
-		return dtoMetrics[i].String() < dtoMetrics[j].String()
-	})
+	syncElasticQuotaMetrics(eq, summary, false)
+	dtoMetrics := getMetrics(t)
 	expect := []string{
 		`{"label":[{"name":"field","value":"allocated"},{"name":"is_parent","value":"true"},{"name":"name","value":"test-eq"},{"name":"parent","value":"root"},{"name":"resource","value":"cpu"},{"name":"tree","value":"tree-1"}],"gauge":{"value":7000}}`,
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/elasticquota/metrics.go
```diff
@@ -61,20 +61,24 @@ func init() {
...
+			recordResourceMetric(gaugeVec, quantity.MilliValue(), string(resourceName), field, labels, isDelete)
 		default:
-			recordResourceMetric(gaugeVec, quantity.Value(), string(resourceName), field, labels)
+			recordResourceMetric(gaugeVec, quantity.Value(), string(resourceName), field, labels, isDelete)
 		}
 	}
 }
 
-func recordResourceMetric(gaugeVec *metrics.GaugeVec, value int64, resourceName, field string, labels map[string]string) {
+func recordResourceMetric(gaugeVec *metrics.GaugeVec, value int64, resourceName, field string, labels map[string]string, isDelete bool) {
```
saintube: Actually, the deletion does not need a value. How about wrapping a `deleteResourceMetric`?

qinfustu: We need to retrieve labels from the QuotaInfoSummary. Adding a deleteResourceMetric method will lead to significant code duplication, such as the logic in syncElasticQuotaMetrics.

saintube: I think it's OK to add concise code with a different function.

qinfustu: I've made the changes, please review again. thx.

pkg/scheduler/plugins/elasticquota/controller_test.go
```diff
@@ -364,8 +364,52 @@ func Test_updateElasticQuotaStatusIfChanged(t *testing.T) {
...
-
+func Test_deleteElasticQuotaMetrics(t *testing.T) {
+	eq, summary := getEqAndSummary()
+	syncElasticQuotaMetrics(eq, summary)
+	dtoMetrics := getMetrics(t)
+	assert.Equal(t, 22, len(dtoMetrics))
+	deleteElasticQuotaMetrics(eq, summary)
+	dtoMetricsD := getMetrics(t)
+	assert.Equal(t, 0, len(dtoMetricsD))
+}
```
saintube, line 376: L367, L376, and L445 need blank lines between functions.

pkg/scheduler/plugins/elasticquota/quota_handler.go
```diff
@@ -104,7 +104,10 @@ func (g *Plugin) OnQuotaDelete(obj interface{}) {
 		klog.Errorf("quota is nil")
 		return
 	}
-
+	summary, _ := g.GetQuotaSummary(quota.Name, false)
+	if summary != nil {
```
ZiMengSheng, line 108: if summary == nil ?

pkg/scheduler/plugins/elasticquota/metrics.go
```diff
@@ -75,6 +75,22 @@ func RecordElasticQuotaMetric(gaugeVec *metrics.GaugeVec, resources corev1.Resou
...
 	labels["resource"] = resourceName
 	labels["field"] = field
-
 	gaugeVec.With(labels).Set(float64(value))
 }
+
+func DeleteElasticQuotaMetric(gaugeVec *metrics.GaugeVec, resources corev1.ResourceList, field string, labels map[string]string) {
+	for resourceName := range resources {
+		switch resourceName {
+		case corev1.ResourceCPU:
```
ZiMengSheng: Emm, is there any difference between the cpu and default?

pkg/scheduler/plugins/elasticquota/quota_handler.go
qinfustu: If the summary is nil, the labels cannot be accessed. Actually, the summary is deleted after this point.

pkg/scheduler/plugins/elasticquota/metrics.go
qinfustu: 😳😳😳

PR #2519, opened 2025-07-10
webhook: add more info when deleting quota failed

googs1025:

- show child quota and pod count on delete failure.
This can reduce our troubleshooting time when deletion fails, as there may not be any other info on the quota. 🤔 

### V. Checklist


pkg/webhook/elasticquota/quota_topology.go
```diff
@@ -186,7 +187,18 @@ func (qt *quotaTopology) ValidDeleteQuota(quota *v1alpha1.ElasticQuota) error {
 		return fmt.Errorf("failed list pods for quota %v, err: %v", quota.Name, err)
 	}
 	if len(podList.Items) > 0 {
-		return fmt.Errorf("delete quota failed, quota %v has child pods", quotaName)
+		podCount := len(podList.Items)
```
googs1025, line 190: change to:  `delete quota failed, quota example-quota has 3 child pods: podA, podB, ...`

ZiMengSheng:
@saintube PTAL

saintube (approved):
/lgtm

PR #2531, opened 2025-07-15
scheduler: loadaware support force estimated duration and other improvements

zheng-weihao:
#### New Features

For some workload, the process has to loading configuration and data in bootstraping for a while before it executes operators that consume a lot of cpu and memory.

The solution is to add args and pod annotations that can force loadaware to use estimated usage in the duration after the pod is scheduled or initialized (all init containers are executed successfully)

LoadAwareSchedulingArgs
- `forceEstimationSecondsAfterPodScheduled` default force estimated duration after scheduled, < 0 means disabled by default
- `forceEstimationSecondsAfterInitialized` default force estimated duration after initialized,  < 0 means disabled by default
- `allowCustomizeEstimationFromMetadata` whether plugin allow to customize pod estimation with its annotations

pod metadata.annotations
- `scheduling.koordinator.sh/force-load-estimation-seconds-after-pod-scheduled` customization will be skipped if value < 0 or invalid (not base 10 integer format)
- `scheduling.koordinator.sh/force-load-estimation-seconds-after-initialized` customization will be skipped if value < 0 or invalid (not base 10 integer format)
- `scheduling.koordinator.sh/load-estimated-scaling-factors` represents pod customized estimated scaling factors, factors from plugin args will only be used when resource name does not exist in annotation value, customization will be skipped if json format invalid

#### Improvements

1. Try to use pod scheduled condition time as timestamp when pod is added to cache to avoid estimate all pods when scheduler starts in several minutes (base on NodeMetric's `spec.metricCollectPolicy.reportIntervalSeconds` and force estimation settings)
2. Store estimated usage in cache to avoid redundant calculation when estimate assigned pod used for node.
3. Fix plugin ut and avoid manipulating assigned pod cache data directly in test code which duplicate the cache maintaining code.

NONE

### V. Checklist

### DIFF
--- a/apis/extension/load_aware.go
+++ b/apis/extension/load_aware.go
@@ -30,12 +30,12 @@ const (
 	AnnotationCustomUsageThresholds = SchedulingDomainPrefix + "/usage-thresholds"
 	// AnnotationCustomEstimatedScalingFactors represents the user-defined factor when estimating resource usage.
 	AnnotationCustomEstimatedScalingFactors = SchedulingDomainPrefix + "/load-estimated-scaling-factors"
-	// AnnotationCustomForceEstimationSecondsAfterPodScheduled represents the user-defined
+	// AnnotationCustomEstimatedSecondsAfterPodScheduled represents the user-defined
 	// force estimation seconds after pod scheduled.
-	AnnotationCustomForceEstimationSecondsAfterPodScheduled = SchedulingDomainPrefix + "/force-load-estimation-seconds-after-pod-scheduled"
-	// AnnotationCustomForceEstimationSecondsAfterInitialized represents the user-defined
+	AnnotationCustomEstimatedSecondsAfterPodScheduled = SchedulingDomainPrefix + "/load-estimated-seconds-after-pod-scheduled"
+	// AnnotationCustomEstimatedSecondsAfterInitialized represents the user-defined
 	// force estimation seconds after initialized.
-	AnnotationCustomForceEstimationSecondsAfterInitialized = SchedulingDomainPrefix + "/force-load-estimation-seconds-after-initialized"
+	AnnotationCustomEstimatedSecondsAfterInitialized = SchedulingDomainPrefix + "/load-estimated-seconds-after-initialized"
 )
 
 // CustomUsageThresholds supports user-defined node resource utilization thresholds.
@@ -81,8 +81,8 @@ func GetCustomEstimatedScalingFactors(pod *corev1.Pod) map[corev1.ResourceName]i
 	return nil
 }
 
-func GetCustomForceEstimationSecondsAfterPodScheduled(pod *corev1.Pod) int64 {
-	if s := pod.Annotations[AnnotationCustomForceEstimationSecondsAfterPodScheduled]; s != "" {
+func GetCustomEstimatedSecondsAfterPodScheduled(pod *corev1.Pod) int64 {
+	if s := pod.Annotations[AnnotationCustomEstimatedSecondsAfterPodScheduled]; s != "" {
 		if i, err := strconv.ParseInt(s, 10, 64); err == nil {
 			return i
 		}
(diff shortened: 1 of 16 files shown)


pkg/scheduler/apis/config/v1beta3/types.go
```diff
@@ -56,6 +56,14 @@ type LoadAwareSchedulingArgs struct {
 	// EstimatedScalingFactors indicates the factor when estimating resource usage.
 	// The default value of CPU is 85%, and the default value of Memory is 70%.
 	EstimatedScalingFactors map[corev1.ResourceName]int64 `json:"estimatedScalingFactors,omitempty"`
+	// ForceEstimationSecondsAfterPodScheduled indicates the force estimation duration
+	// after pod condition PodScheduled transition to True in seconds.
+	ForceEstimationSecondsAfterPodScheduled *int64 `json:"forceEstimationSecondsAfterPodScheduled,omitempty"`
+	// ForceEstimationSecondsAfterInitialized indicates the force estimation duration
+	// after pod condition Initialized transition to True in seconds.
+	ForceEstimationSecondsAfterInitialized *int64 `json:"forceEstimationSecondsAfterInitialized,omitempty"`
```
saintube: What is the difference between the `forceEstimationSecondsAfterPodScheduled` and the `forceEstimationSecondsAfterInitialized`? I suppose we can merge into a single one, which means estimating the pod usage based on its request, regardless of whether it has a pod metric or not.

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -355,6 +358,43 @@ func (p *Plugin) estimatedAssignedPodUsed(nodeName string, nodeMetric *slov1alph
 	return estimatedUsed, estimatedPods
 }
 
+func (p *Plugin) shouldForceEstimatePod(info *podAssignInfo, now time.Time) bool {
```
saintube: To reduce the calculation overhead during the Filter phase, we can pre-calculate the estimated result in the assignInfo.

apis/extension/load_aware.go
```diff
@@ -27,6 +28,12 @@ const (
 	// AnnotationCustomUsageThresholds represents the user-defined resource utilization threshold.
 	// For specific value definitions, see CustomUsageThresholds
 	AnnotationCustomUsageThresholds = SchedulingDomainPrefix + "/usage-thresholds"
+	// AnnotationCustomForceEstimationSecondsAfterPodScheduled represents the user-defined
+	// force estimation seconds after pod scheduled.
+	AnnotationCustomForceEstimationSecondsAfterPodScheduled = SchedulingDomainPrefix + "/force-load-estimation-seconds-after-pod-scheduled"
```
saintube: Apart from the estimating window size, perhaps we can allow the pod to customize its scaling factor, WDYT?

pkg/scheduler/apis/config/v1beta3/types.go
zheng-weihao: They use different pod conditions. For some workload which has long init container, use forceEstimationSecondsAfterInitialized is more appropriate. But we don't want to keep estimation if init containers keep failing and restarting / kubelet pod creation fails (Initialized condition does not exists or status is False). For this case we will also set forceEstimationSecondsAfterPodScheduled as the max force estimation duration before main containers started.

apis/extension/load_aware.go
zheng-weihao: good idea, maybe we should change `allowForceEstimationFromMetadata` to `allowCustomizeEstimationFromMetadata`

pkg/scheduler/plugins/loadaware/load_aware.go
zheng-weihao: do we need another PR or I just change this PR's title?

saintube: Since it is not a functional change, implementing it in this PR is preferred.

pkg/scheduler/apis/config/v1beta3/types.go
saintube: OK.

apis/extension/load_aware.go
zheng-weihao: updated

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -92,13 +97,27 @@ func (p *podAssignCache) assign(nodeName string, pod *corev1.Pod) {
 		m = make(map[types.UID]*podAssignInfo)
 		p.podInfoItems[nodeName] = m
 	}
-
+	estimated, err := p.estimator.EstimatePod(pod)
```
saintube: calculate outside the Lock

zheng-weihao: done

saintube (commented):
/lgtm

PR #2537, opened 2025-07-18
scheduler: only requeue gang when gangWorth and by activate

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -477,6 +477,21 @@ func (gang *Gang) removeWaitingGang() {
 	gang.GangGroupInfo.RemoveWaitingGang(gang.Name)
 }
 
+func (gang *Gang) isGangWorthRequeue() bool {
+	gang.lock.Lock()
```
saintube: RLock?
same below

PR #2539, opened 2025-07-18
scheduler: add DisableFilterWhenHighLoad to LoadAwareScheduling plugin

googs1025:

- Add new config field `DisableFilterWhenHighLoad` to `LoadAwareSchedulingArgs`
When enabled, the plugin skips node resource usage filtering during scheduling
This is useful in high-load scenarios where strict filtering may block scheduling

fixes: https://github.com/koordinator-sh/koordinator/issues/2538

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -31,6 +31,9 @@ import (
 type LoadAwareSchedulingArgs struct {
 	metav1.TypeMeta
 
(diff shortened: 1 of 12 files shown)


ZiMengSheng:
Why not just turn off the Filter extension point?

googs1025:
we don't need this arg. we can turn off Filter point in scheduler config.

PR #2555, opened 2025-07-28
scheduler: loadaware fix cache for pod conditions changed

zheng-weihao:

bug fix for pod cache in load aware plugin, since items are not updated when conditions changed

### V. Checklist


pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -157,7 +192,8 @@ func (p *podAssignCache) OnUpdate(oldObj, newObj interface{}) {
 		if util.IsPodTerminated(pod) { // pod become terminated
 			p.unAssign(pod.Spec.NodeName, pod)
 		}
```
googs1025: should we just `return` here? 🤔

googs1025:
```go
if util.IsPodTerminated(pod) { // pod become terminated
  p.unAssign(pod.Spec.NodeName, pod)
  return
}
```

zheng-weihao: assign func already have IsPodTerminated check that operation will skip if pod is termitated.

I got your point. Current code is kind of confusing for reading. I'll refine that.

saintube (approved):
/lgtm

PR #2557, opened 2025-07-30
manager/koordlet: add pod-level CPU burst strategies for fine-grained control

### LINKED ISSUE #2553: [proposal] Support dynamic and fine-grained enabling of cpuBurst via QOSGreyCtrlPlugin + ConfigMap
**What is your proposal**:

I propose introducing a more flexible and dynamic control mechanism for the `cpuBurst` feature by leveraging the existing `QOSGreyCtrlPlugin` infrastructure. My idea is to enable or disable `cpuBurst` on a per-pod basis using a user-defined ConfigMap, without relying on pod-level annotations or requiring pod restarts.

This would allow me—and others operating large-scale Kubernetes clusters—to safely roll out `cpuBurst` in a fine-grained and non-intrusive way.

---

**Why is this needed**:

The current mechanisms for enabling `cpuBurst` pose two major operational challenges:

1. **Pod-level annotations**: These require pod restarts or rollouts, which are disruptive and impractical at scale.
2. **Node-level SLO settings**: These apply uniformly to all pods on the node, lacking the flexibility needed for controlled rollouts.

Both approaches introduce significant blast radius concerns during changes and make it hard to experiment with or gradually enable the feature.

---

Yes. I suggest adding a new implementation of `QOSGreyCtrlPlugin` (e.g., `CpuBurstGreyCtrlPlugin`) that does the following:

* Reads a central ConfigMap (e.g., `cpu-burst-grey-control`) from the `koordinator-system` namespace;
* Allows matching rules based on pod name, namespace, or label selectors;
* Dynamically injects the appropriate `cpuBurst` configuration at runtime when no pod-level annotation is found.

This would:

* Eliminate the need to restart pods;
* Allow targeted and safe rollout of `cpuBurst`;
* Improve manageability and control for large-scale production clusters.

I’d love to hear your thoughts on this direction. If this makes sense to you, I’m happy to provide a prototype or contribute a PR.

Thanks!

ktalg:

Adds **Pod-level grey control** for the CPU Burst feature without breaking the existing
`ConfigMap ➜ koord-manager ➜ NodeSLO ➜ koordlet` pipeline.

* **API extension** – `CPUBurstCfg` now contains an optional `podStrategies` field that lets users target Pods by **namespace and/or label selector** and optionally override per-Pod annotations.
* **koord-manager** – when converting `slo-controller-config` to `NodeSLO`, it copies the full `podStrategies` list together with the node/cluster strategies.
* **koordlet** – `cpuburst` plugin merges `podStrategies` at runtime and applies the selected strategy during the normal pod scan; no extra grey-control plugin or additional watch channel is introduced.
* Keeps full backward compatibility (old configs work unchanged).

close #2553

1. Edit the `slo-controller-config` ConfigMap and add a `podStrategies` entry
eg.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: slo-controller-config
  namespace: koordinator-system
data:
  cpu-burst-config: |
    {
      "clusterStrategy": {
        "policy": "none"
      },
      "nodeStrategies": [
        {
          "nodeSelector": {
            "matchLabels": {
              "kubernetes.io/hostname": "ci-testing-worker"
            }
          },
          "cpuBurstStrategy": {
            "policy": "auto",
            "cpuBurstPercent": 200,
            "cfsQuotaBurstPercent": 300,
            "cfsQuotaBurstPeriodSeconds": 6
          }
        }
      ],
      "podStrategies": [
        {
          "namespace": "frontend",
          "podSelector": {
            "matchLabels": {
              "app": "payment"
            }
          },
          "overridePod": true,
          "cpuBurstConfig": {
            "policy": "auto",
            "cpuBurstPercent": 100,
            "cfsQuotaBurstPercent": 600,
            "cfsQuotaBurstPeriodSeconds": 10
          }
        },
        {
          "namespace": "batch-jobs",
          "podSelector": {
            "matchLabels": {
              "workload-type": "low-priority"
            }
          },
          "overridePod": false,
          "cpuBurstConfig": {
            "policy": "cpuBurstOnly",
            "cpuBurstPercent": 50
          }
        }
      ]
    }

```
2. Ensure koord-manager reconciles the change into each node’s `NodeSLO`.
3. Launch a matching Pod and observe on the node:

   * `cpu.cfs_burst_us` / `cpu.max.burst` set as expected.
   * Logs show `genPodBurstConfig` picked the pod-specific strategy.
4. Non-matching Pods and nodes continue to follow node / cluster strategy.

### V. Checklist

* [ ] I have written necessary docs and comments
* [x] I have added necessary unit tests and integration tests
* [x] All checks passed in `make test`


ktalg:
Hi @ZiMengSheng @saintube 👋 Just a gentle ping on this PR 

The feature has been fully verified in our test clusters — including dynamic pod-level burst control without restarts, and full backward compatibility with existing configurations.

Let me know if anything needs to be adjusted. Thanks a lot for your time and review.

saintube:
@ktalg Thanks for your contributions!
For your use case, I'm wondering if we could use the pod annotation API `koordinator.sh/cpuBurst` to control the pod-level configuration. A controller may be added to patch the annotations to target pods, or we can reuse the ColocationProfile to fulfill it.

ktalg:
@saintube 

Thanks for the thoughtful suggestions! I've carefully considered both of the proposed alternatives, and would love to share some thoughts:

1. If we go with the new controller approach, it would require introducing a new configuration source — either a new CRD or reuse of the existing `slo-controller-config`. A new CRD might be too heavy just for this purpose, while sharing `slo-controller-config` between multiple controllers feels a bit unclean and risks tight coupling. Additionally, we'd also need to introduce a new webhook to inject annotations for newly created Pods.

2. Handling cases where Pods already have manually specified burst annotations (and how to safely override or restore them) adds extra complexity around state reconciliation and ownership tracking.

3. As for using `ColocationProfile`, it currently only affects newly created Pods via admission webhook. This doesn't align well with the goal of dynamically updating burst configurations on already-running Pods.

Compared to those options, the current implementation — which extends `NodeSLO` with a `podStrategies` field and delegates the merging to `koordlet` — offers a lightweight, non-intrusive, and backward-compatible way to dynamically control burst behavior without touching the Pods themselves or requiring additional components.

Happy to continue the discussion — open to adjustments if there's a direction the community prefers!

apis/slo/v1alpha1/nodeslo_types.go
```diff
@@ -434,6 +445,8 @@ type NodeSLOSpec struct {
 	ResourceQOSStrategy *ResourceQOSStrategy `json:"resourceQOSStrategy,omitempty"`
 	// CPU Burst Strategy
 	CPUBurstStrategy *CPUBurstStrategy `json:"cpuBurstStrategy,omitempty"`
+	// Pod-level CPU burst configuration strategies, allows fine-grained control over
+	PodCPUBurstStrategies []PodCPUBurstStrategy `json:"podCPUBurstStrategies,omitempty"`
```
saintube, line 449: Why not store in the CPUBurstStrategy?

saintube (commented):
I'm convinced. Any comments @zwzhang0107 @songtao98

apis/slo/v1alpha1/nodeslo_types.go
ktalg: I kept it separate mainly to avoid mixing node-level and pod-level logic in one struct. Putting everything in CPUBurstStrategy might complicate the override order and make future extensions trickier. This way keeps things cleaner and more maintainable.

ktalg: I considered that, but felt it's cleaner to keep pod-level configs separate from `CPUBurstStrategy`. Mixing node and pod strategies in one struct might blur scope boundaries, complicate override logic, and make future extensions harder. Keeping them parallel fields keeps things simple and explicit.

saintube: OK.

zwzhang0107:
@ktalg  hi, thank for the explanation.
I'm wondering pod strategies on NodeSLO might be hard to maintenance. Workloads may spread on different nodes, which lead the CR inflating.
We used to implement a [ConfigMap](https://www.alibabacloud.com/help/en/ack/ack-managed-and-ack-dedicated/user-guide/cpu-burst?spm=a2c63.p38356.0.i2#700fef8712m21) to enable CPU Burst on namespace level before open-sourced(predecessors of Koordinator). 

The former design may not following the format of Koordinator, so I suggest we can implement the podStrategies with an new independent ConfigMap for grey control, following the effective order of `pod-level annotation -> pod-level ConfigMap -> global-level ConfigMap`.

ktalg:
@zwzhang0107 
Thanks — I thought about your suggestion carefully and I’m on board: this keeps changes small and avoids disturbing existing resources and workflows.

Here’s the change I’m planning:

* add a small controller whose job is to apply/adjust the existing `koordinator.sh/cpuBurst` annotation on targeted Pods (new and running ones) according to gray rules;
* introduce a dedicated ConfigMap (e.g. `cpu-burst-grey-config` in `koordinator-system`) to hold those gray rollout rules and policies.

In the gray config each rule can specify how to handle existing or future Pod-level `cpuBurst` annotations — i.e. whether the controller should **never** overwrite, **merge** missing fields, or **always** replace the Pod annotation. My intent is to default to non-destructive behavior (do not overwrite) and make more aggressive options explicitly opt-in.

Do you think this approach is reasonable? Any suggestions on the ConfigMap placement/naming, default override policy, rate limits/rollback strategy, or anything else I should consider before I implement it?

songtao98:
@zwzhang0107 PTAL

zwzhang0107:
> In the gray config each rule can specify how to handle existing or future Pod-level `cpuBurst` annotations — i.e. whether the controller should **never** overwrite, **merge** missing fields, or **always** replace the Pod annotation. My intent is to default to non-destructive behavior (do not overwrite) and make more aggressive options explicitly opt-in.
> 
> Do you think this approach is reasonable? Any suggestions on the ConfigMap placement/naming, default override policy, rate limits/rollback strategy, or anything else I should consider before I implement it?

@ktalg 
Love the controller idea! It keeps things contained and follows origin patterns nicely.

A few quick thoughts:

- For gray rollout, maybe add support for multiple rule groups (e.g., by namespace or pod labels) to make targeting more flexible
- The overwrite behavior could be a simple config option (merge/replace) with "preserve" as default
- Naming-wise, cpu-burst-config in koordinator-system feels clean enough

This should cover the basics while leaving room for future policy types. Happy to chat more about any of these points! 😊

saintube:
@ktalg any update

ZiMengSheng:
This change has had no review progress in 14 months and the CPU burst API has since evolved. Closing to keep the backlog focused — if pod-level CPU burst strategies are still relevant, please open a proposal describing the desired API shape on top of current main, and a fresh PR can follow from there. Thanks for the contribution!

PR #2582, opened 2025-08-18
scheduler: improve load aware perf by resources cache and vectorization

zheng-weihao:
#### Performance Improvement

Improve load aware filtering and scoring performance by storing all intermediate results in cache to avoid duplicated evaluation on assigned pods and converting resources list to vector to avoid value retrieval from many small map.

Vectorization is useful in load aware because we can should only evaluate on real limited resources that we can collect or estimate.

#### Add `prodUsageIncludeSys` load aware plugin args

`prodUsageIncludeSys` indicates whether to include system usage (not used by pods) when summing up current node usage for prod pods.

#### Add `supportedResources` load aware plugin args

`supportedResources` is the list of extra resource names that can be used in load-aware scheduling since we use vectorization to convert resource list map to vector. Unknown resource names will be ignored by plugin in calculation.

`cpu`, `memory` and all other resources that show up in args are supported by default.

If more resource are added in koordlet collection, don't show up as filter thresholds or score weights in plugin args and only set up in custom node annotations, we should pass these resource names in `supportedResources` explicitly.

#### Changes

`estimatedSecondsAfterPodScheduled` and `estimatedSecondsAfterInitialized` use node metric's update time instead of time.Now to compare and judge if an existing pod should try to use estimation usage.

NONE

### V. Checklist


pkg/scheduler/apis/config/v1/types.go
```diff
@@ -65,6 +68,14 @@ type LoadAwareSchedulingArgs struct {
...
 	// Aggregated supports resource utilization filtering and scoring based on percentile statistics
 	Aggregated *LoadAwareSchedulingAggregatedArgs `json:"aggregated,omitempty"`
+	// SupportedResourceNames is the list of extra resource names that can be used in load-aware scheduling.
+	// cpu, memory and all other resources that show up in args are supported by default.
+	//
+	// If more resource are added in collection, don't show up as
+	// filter thresholds or score weights in plugin args
+	// and only set up in custom node annotations,
+	// we should pass these resource names in plugin args explicitly.
+	SupportedResources []corev1.ResourceName
```
saintube: miss a json tag

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -92,20 +97,38 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	if err != nil {
 		return nil, err
 	}
-	assignCache := newPodAssignCache(estimator, pluginArgs)
+	// cpu and memory are added by default for custom usage thresholds compatibility
+	resourceNames := sets.New(corev1.ResourceCPU, corev1.ResourceMemory)
```
saintube: recommended to wrap the logic as a function

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -66,11 +69,13 @@ var (
...
 
 type Plugin struct {
-	handle           framework.Handle
-	args             *config.LoadAwareSchedulingArgs
-	nodeMetricLister slolisters.NodeMetricLister
-	estimator        estimator.Estimator
-	podAssignCache   *podAssignCache
+	handle         framework.Handle
+	args           *config.LoadAwareSchedulingArgs
+	vectorizer     ResourceVectorizer
```
saintube, line 75: why not use the `resource.Quantity` and the `corev1.ResourceList`

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -38,25 +42,29 @@ var (
 
 // podAssignCache stores the Pod information that has been successfully scheduled or is about to be bound
 type podAssignCache struct {
-	lock sync.RWMutex
+	lock  sync.RWMutex
+	items map[string]*nodeMetric
```
saintube: Recommended to set a specific name, like `nodeMetricItems`

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -211,3 +230,264 @@ func (p *podAssignCache) OnDelete(obj interface{}) {
...
+	}
+	return m
+}
+
+func (p *podAssignCache) initPods(m *nodeMetric, pods map[types.UID]*podAssignInfo) {
+	m.nodeDelta = p.vectorizer.EmptyVec()
+	m.prodDelta = p.vectorizer.EmptyVec()
+	m.nodeEstimated = p.vectorizer.EmptyVec()
+
+	if klog.V(6).Enabled() {
```
saintube: Why initialize with klog level

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -226,146 +271,116 @@ func (p *Plugin) Score(ctx context.Context, state *framework.CycleState, pod *co
...
-			}
+	if podEstimated == nil {
+		list, err := p.estimator.EstimatePod(pod)
+		if err != nil {
+			// use len=0 but not empty vector to indicate error occurred
+			cycleState.Write(incomingPodEstimatedStateKey, ResourceVector{})
+			return err
 		}
+		podEstimated = p.vectorizer.ToFactorVec(list)
+		cycleState.Write(incomingPodEstimatedStateKey, podEstimated)
```
saintube, line 307: We can preprocess the scheduling pod during the PreFilter phase.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -211,3 +230,264 @@ func (p *podAssignCache) OnDelete(obj interface{}) {
...
+
+	e := pod.estimated
+	if e == nil {
+		return
+	}
+	// 1. when usage is not collected
+	// 2. when pod miss lastest metrics update
+	// 3. when pod metrics is still in the report interval
+	// 4. when pod is configured in estimation
+	should := u == nil ||
```
saintube, line 627: The load-aware plugin requires real-time metrics during the scheduling cycle. But this judgment only triggers calculation upon new events, which breaks the timeliness of the estimation.

pkg/scheduler/plugins/loadaware/load_aware.go
zheng-weihao: it reduces the time for map retrival and quantity checking for add or sub.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: These pods details are only used for debug log print. So they are collected only when klog is enabled (V level 6 is the same as klog level in Filter and Score)

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: Actually, in the real time calculation code only estimatedDeadline check is depend on current timestamp, which is newly added and okay to replace with metrics updateTime.

Other data are all from cache: existing pods are fetched from this pod assign cache, node metrics are fetched from informer cache.

pkg/scheduler/plugins/loadaware/load_aware.go
zheng-weihao: done

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: done

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -154,37 +209,19 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
-				nodeMetric,
-				filterProfile.AggregatedUsage.UsageAggregatedDuration,
-				filterProfile.AggregatedUsage.UsageAggregationType,
-			)
-			usageThresholds = filterProfile.AggregatedUsage.UsageThresholds
-		} else {
-			nodeUsage = &nodeMetric.Status.NodeMetric.NodeUsage
-			usageThresholds = filterProfile.UsageThresholds
-		}
+	if ret := p.filterNodeUsage(node.Name, pod, usageThresholds, estimated, allocatable, isAgg); ret != nil {
```
saintube: Duplicates L224?

zheng-weihao: not really, this check is useful if addEstimatedOfIncoming returns error

saintube: AddEstimatedOfIncoming returning an error is a small probability event, while filtering twice is the common case.

zheng-weihao: done. keep filterNodeUsage checking with node usage only when addEstimatedOfIncoming returns with error

pkg/scheduler/plugins/loadaware/load_aware.go
zheng-weihao: done

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: done

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -22,111 +22,246 @@ import (
...
-			pod:               pod,
-			estimated:         estimated,
-			estimatedDeadline: estimatedDeadline,
+	newPod := &podAssignInfo{
+		timestamp:         timestamp,
+		pod:               pod,
+		estimated:         estimated,
+		estimatedDeadline: estimatedDeadline,
+	}
+	for {
```
saintube: What does this spin lock for?

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -211,3 +343,287 @@ func (p *podAssignCache) OnDelete(obj interface{}) {
 	}
 	p.unAssign(pod.Spec.NodeName, pod)
 }
+
+func (n *nodeInfo) AddOrUpdatePod(pod *podAssignInfo, locked bool) bool {
+	if locked {
```
saintube: Recommended to simplify this implementation. We accept some redundant code to improve readability.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: p.getOrCreateNodeInfo might get a deleted nodeInfo, we should not update on that cache which will miss some update event. So when we detect a nodeInfo is deleted in update, it just return not ok and caller should retry to fetch a new nodeInfo.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: When we create new nodeInfo, it should be locked for writing data at begin, which is a must that prevent reading data from a new and empty nodeInfo. The state is passing by param because sync.RWMutex don't provide IsLocked checking method.

I will add some comments for it.

zheng-weihao: done

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
saintube: If the node is deleted, why not just abort the pod assign and return?

zheng-weihao: the pod adding or updating will be missed if we just abort

saintube: What is the case that a node is deleted while a pod is actually assigned? I suppose it is a consequence of out-of-order execution between event handlers of Pod and Node that we should ignore.

zheng-weihao: this cache is not related to node deletion. It tries to delete when node metrics is not found and pod clean up on node. When node metrics object is deleted and all pod finished on one node, we might missing some incoming assign pod when deletion is triggered.

saintube: Only the deletion of the nodemetric can trigger a deletion of the nodeInfo map. We don't need to remove the nodeInfo when assigned pods are deleted.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -211,3 +349,294 @@ func (p *podAssignCache) OnDelete(obj interface{}) {
 	}
 	p.unAssign(pod.Spec.NodeName, pod)
 }
+
+// AddOrUpdatePod add or update pod to nodeInfo.
+// It returns false is nodeInfo is already deleted and caller should get a new nodeInfo and retry.
+// Unlock is called whether nodeInfo is locked before calling or not.
+func (n *nodeInfo) AddOrUpdatePod(pod *podAssignInfo, locked bool) bool {
+	if n.deleted {
```
saintube, line 408: If the node metric is deleted, the filtering of the node should either reject the scheduling pod or skip the check (determined by the plugin args), neither of which needs the assigned pod. So I suppose we can abort the pod event, and let the nodeInfo be missing until the node metric comes back.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: This implementation will leak nodeInfo without node metrics (koordlet not installed or something else cause node metric is missing)

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: For long run pod, we should not expect the status of pod to change to trigger a new event.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
```diff
@@ -211,3 +349,294 @@ func (p *podAssignCache) OnDelete(obj interface{}) {
...
+			default:
+				return
+			}
+			p.DeleteNodeMetric(m.Name)
+		},
+	}
+}
+
+func (p *podAssignCache) AddOrUpdateNodeMetric(metric *slov1alpha1.NodeMetric) {
+	for {
```
saintube: As discussed, please expand the loop to avoid indeterministic retries.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
saintube: As discussed, please expand the loop.

zheng-weihao: done, add or update can only try for 5 times at most now.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: done, add or update can only try for 5 times at most now.

zheng-weihao: only retry once now

zheng-weihao: AddOrUpdate's retry only fails when another event handler create new object and clean it up in very very short term, which creates a new deleted nodeInfo that retry might get. This almost impossible to happen considering the informer list and watch mechanism.

pkg/scheduler/plugins/loadaware/pod_assign_cache.go
zheng-weihao: only retry once now

saintube (commented):
/lgtm

pkg/scheduler/apis/config/types.go
```diff
@@ -49,6 +49,9 @@ type LoadAwareSchedulingArgs struct {
 	// ProdUsageThresholds indicates the resource utilization threshold of Prod Pods compared to the whole machine.
 	// Not enabled by default
 	ProdUsageThresholds map[corev1.ResourceName]int64
+	// ProdUsageIncludeSys indicates whether to include system usage (not used by pods)
```
ZiMengSheng, line 52: 1. Can you add user documentation here: https://koordinator.sh/docs/user-manuals/load-aware-scheduling
2. Can you provide an example of a conceptual model diagram of resource usage on a node, such as Prod usage, Sys usage, etc., and allow users to check it themselves through NodeMetric?

pkg/scheduler/plugins/loadaware/service.go
```diff
@@ -30,6 +32,14 @@ var _ services.APIServiceProvider = &Plugin{}
 
 type NodeAssignInfoData struct {
 	Pods []PodAssignInfoData `json:"pods,omitempty"`
+
```
ZiMengSheng, line 35: Same as above, hope to provide user documentation here: https://koordinator.sh/docs/user-manuals/load-aware-scheduling

pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -121,6 +136,17 @@ func (p *Plugin) EventsToRegister() []framework.ClusterEventWithHint {
 	}
 }
 
+func (p *Plugin) PreFilter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod) (*framework.PreFilterResult, *framework.Status) {
+	// add estimated resources for incoming pod to cache
+	_ = p.addEstimatedOfIncoming(p.vectorizer.EmptyVec(), state, pod)
```
ZiMengSheng, line 141: OK, so here PreFilter is counted once, and then Filter & Score are both used.

saintube:
@ZiMengSheng @zheng-weihao Let's update the documentation later in v1.7.
https://github.com/koordinator-sh/website/issues/226

zheng-weihao:
> @ZiMengSheng @zheng-weihao Let's update the documentation later in v1.7. [koordinator-sh/website#226](https://github.com/koordinator-sh/website/issues/226)

+1, I also have a new feature issue on load aware #2596. I can help updating these docs later.

PR #2586, opened 2025-08-23
webhook: fix elasticquota validation error for min > max

zheng-weihao:

Currently, webhook reject elasticquotas' mutation with go struct %v printf when min resource value exceeds max. It is hard to read.

```
resourceKey memory of quota e2e-test min :map[cpu:{{1 0} {<nil>} 1 DecimalSI} memory:{{4294967296 0} {<nil>} 4Gi BinarySI}] > max,map[cpu:{{1 0} {<nil>} 1 DecimalSI} memory:{{4 0} {<nil>} 4 DecimalSI}]
```

### V. Checklist


saintube (approved):
/lgtm

PR #2595, opened 2025-08-28
webhook: fix pod validating error that can't forbid illegal comibinations of QoS and priorityClass

IULeen:
Fix the pod validating webhook error that couldn't correctly forbid illegal combinations of QoS and priorityClass.

None
For example, the pod validating webhook should forbid the combination of **`BE`** QoS and **`koord-prod`**,
and return reason message that _`Error from server: error when creating "STDIN": admission webhook "vpod.koordinator.sh" denied the request: Pod: Forbidden: koordinator.sh/qosClass=BE and priorityClass=koord-prod cannot be used in combination`_
```shell
#!/bin/bash
priority=("koord-prod" "koord-mid" "koord-batch" "koord-free" "system-cluster-critical")
qos=("SYSTEM" "LSE" "LSR" "LS" "BE" "NONE")

for p in "${priority[@]}"; do
  for q in "${qos[@]}"; do
  q_lower=$([ -n "$q" ] && echo "$q" | tr A-Z a-z || echo "")
  cat <<EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  name: t1-${p}-${q_lower}
  namespace: test-koordinator
  labels:
    koordinator.sh/qosClass: ${q}
spec:
  priorityClassName: ${p}
  tolerations:
  - operator: Exists
  containers:
  - name: nginx
    image: nginx:latest
    imagePullPolicy: IfNotPresent
    resources:
        limits:
          cpu: "1000m"
          memory: "500Mi"
        requests:
          cpu: "1000m"
          memory: "500Mi"
EOF
done
done
```

However, the pod validating webhook didn't effective and I successfully created these illegal pods.
pod-webhook-error

In the original wrong logic, the `clusterColocationProfileValidatingPod` function will only return `nil` err, even if errors do occur.
By this PR modify, it could return non-nil err and correctly forbid illegal combinations.
image

### V. Checklist

### DIFF
--- a/pkg/webhook/pod/validating/cluster_colocation_profile.go
+++ b/pkg/webhook/pod/validating/cluster_colocation_profile.go
@@ -65,7 +65,7 @@ func (h *PodValidatingHandler) clusterColocationProfileValidatingPod(ctx context
 		allowed = false
 		reason = err.Error()
 	}
-	return allowed, reason, nil
+	return allowed, reason, err
 }
 
 func validateRequiredQoSClass(pod *corev1.Pod) field.ErrorList {
--- a/pkg/webhook/pod/validating/cluster_colocation_profile_test.go
+++ b/pkg/webhook/pod/validating/cluster_colocation_profile_test.go
@@ -87,6 +87,7 @@ func TestClusterColocationProfileValidatingPod(t *testing.T) {
 			},
 			wantAllowed: false,
 			wantReason:  `labels.koordinator.sh/qosClass: Invalid value: "LS": field is immutable`,
+			wantErr:     true,
 		},
 		{
 			name:      "validate remove QoS",
@@ -103,6 +104,7 @@ func TestClusterColocationProfileValidatingPod(t *testing.T) {
 			},
 			wantAllowed: false,
 			wantReason:  `labels.koordinator.sh/qosClass: Invalid value: "": field is immutable`,
+			wantErr:     true,
 		},
 		{
 			name:      "validate defined QoS",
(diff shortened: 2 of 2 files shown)


pkg/webhook/pod/validating/cluster_colocation_profile.go
```diff
@@ -65,7 +65,7 @@ func (h *PodValidatingHandler) clusterColocationProfileValidatingPod(ctx context
 		allowed = false
 		reason = err.Error()
 	}
-	return allowed, reason, nil
+	return allowed, reason, err
```
ZiMengSheng, line 68: Do we need to return err? reason is not enough?

IULeen: Yes, we need. Because the caller of this function needs the `err` value to determine whether to reject the creation of the pod. However, this function always return `nil` err, so it will never return `false`  and  actual `reason`. This means that the below code snippet (L94-L96) is actually unreachable.
https://github.com/koordinator-sh/koordinator/blob/7b41f4f71ee335df5bd34ef06ab9a41799be4bc2/pkg/webhook/pod/validating/validating_handler.go#L91-L96

In contrast, this following function call will return actual `err` value to refuse illegal pods' creation.
https://github.com/koordinator-sh/koordinator/blob/7b41f4f71ee335df5bd34ef06ab9a41799be4bc2/pkg/webhook/pod/validating/validating_handler.go#L83-L88
https://github.com/koordinator-sh/koordinator/blob/7b41f4f71ee335df5bd34ef06ab9a41799be4bc2/pkg/webhook/pod/validating/verify_annotations.go#L53-L57

IULeen:
@ZiMengSheng  I rebased this branch, and all checks have passed now. When this branch can be merged? I might need to rebase it again.

IULeen:
@ZiMengSheng  @saintube Could this pull-request be merged into main branch? Thanks for your reply!

saintube (approved):
/lgtm

PR #2601, opened 2025-09-02
scheduler: add questionedObjectKet and topologyKeyToExplain

ZiMengSheng:
### V. Checklist


apis/extension/unified/schedule_explanation.go
```diff
@@ -0,0 +1,23 @@
+package unified
```
saintube: invalid pkg

PR #2603, opened 2025-09-03
scheduler: loadaware support dominantResourceWeight

### LINKED ISSUE #2596: [proposal] Support dominant resource weight for scoring in load aware scheduling.
**What is your proposal**:

We can add new scoring weight arg for dominant resource in load aware scheduling.

The node's dominant resource is defined as the most used resource on node:

> dominant resource score = min(allocatable - usedOfExistingPods - estimatedOfIncomingPod) / allocatable)

This idea comes from the concept of Dominant Resource Fairness, known as DRF. Paper is here: [www.cs.berkeley.edu/~matei/papers/2011/nsdi_drf.pdf](https://people.eecs.berkeley.edu/~matei/papers/2011/nsdi_drf.pdf)

**Why is this needed**:

Currently, We can only set resource weight base on resource name. The weight is hard to set when resources usage are already unbalanced in a multi-resource environment because of multiple node type, long run pods' affinity specs or other reasons.

For example, if node A uses cpu heavily and memory are available, node B uses memory heavily and cpu are available. We should prefer to schedule cpu-heavy pods on B and memory-heavy pods on A.

Provided in proposal.

zheng-weihao:

Add a new argument `dominantResourceWeight` in load aware scheduling plugin, which is a integer >= 0 and <= 100. When dominantResourceWeight is not zero, it is used as the weight for dominant resource's score on node.

The node's dominant resource is defined as the most used resource on node:

> dominant resource score = min(allocatable - usedOfExistingPods - estimatedOfIncomingPod) / allocatable)

When dominantResourceWeight is set non zero, plugin will not use default resource weight list.

fixes #2596

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/loadaware/load_aware.go
+++ b/pkg/scheduler/plugins/loadaware/load_aware.go
@@ -110,7 +110,7 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	nodeMetricInformer := koordInformers.Slo().V1alpha1().NodeMetrics()
 	frameworkexthelper.ForceSyncFromInformer(context.TODO().Done(), koordInformers, nodeMetricInformer.Informer(), assignCache.NodeMetricHandler())
 	scoreWeights := vectorizer.ToFactorVec(pluginArgs.ResourceWeights)
-	if scoreWeights.Empty() {
+	if pluginArgs.DominantResourceWeight == 0 && scoreWeights.Empty() {
 		scoreWeights = nil
 	}
 	return &Plugin{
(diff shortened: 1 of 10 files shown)


pkg/scheduler/plugins/loadaware/load_aware.go
```diff
@@ -343,12 +343,20 @@ func (p *Plugin) filterNodeUsage(nodeName string, pod *corev1.Pod, usageThreshol
...
 	for i, weight := range resToWeightMap {
-		nodeScore += leastUsedScore(used[i], allocatable[i]) * weight
+		score := leastUsedScore(used[i], allocatable[i])
+		nodeScore += score * weight
 		weightSum += weight
+		if dominantScore > score {
+			dominantScore = score
+		}
 	}
+	nodeScore += dominantScore * dominantWeight
```
saintube, line 359: Need to ensure the final score does not exceed the framework.MaxNodeScore.

zheng-weihao: dominantScore won't exceed MaxNodeScore.

And final score is the weighted average of resource scores and dominant score, so it won't exceed MaxNodeScore.

PR #2612, opened 2025-09-08
scheduler: only reject sibling pods when preemption failed

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -331,7 +331,7 @@ func (pgMgr *PodGroupManager) PostFilter(ctx context.Context, state *framework.C
 		gangSchedulingContext.failedMessage = message
 	}
 
-	if gang.getGangMode() == extension.GangModeStrict {
+	if gang.getGangMode() == extension.GangModeStrict || !postFilterStatus.IsSuccess() {
```
saintube: 加下注释说明下吧

saintube (approved):
/lgtm

PR #2613, opened 2025-09-11
scheduler: takeover nominatingInfo when waitingPod rejected

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/coscheduling/core/nominating_info.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"errors"
+	"time"
+
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+var (
+	JobPreemptionRejectPlugin = Name + ".PreemptionRejectPlugin"
```
saintube: please add comments to clarify these

PR #2624, opened 2025-09-18
scheduler: fine-grained device scheduling support Cambricon dynamic sMLU

zqzten:

This PR makes fine-grained device scheduling support Cambricon dynamic sMLU. The core scheduling logic is the same as NVIDIA so that this work only covers DP adaption.

Note that Cambricon DP does not support external MLU (full card) scheduling so we can only support dynamic sMLU (GPU share) for now.

How to use:
```yaml
koordinator.sh/gpu.shared: "1"
koordinator.sh/gpu-core: "5"
koordinator.sh/gpu-memory: 1Gi
# below are only used to trigger Cambricon DP
cambricon.com/mlu.smlu.vcore: "1"
cambricon.com/mlu.smlu.vmemory: "1"
```

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
+++ b/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
@@ -17,17 +17,22 @@ limitations under the License.
 package deviceshare
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
```diff
@@ -141,6 +179,74 @@ func (a *huaweiGPUDevicePluginAdapter) Adapt(object metav1.Object, allocation []
...
+	}
+	vcore := core.Value()
+
+	memory := allocation[0].Resources[apiext.ResourceGPUMemory]
+	if memory.Value() < cambriconVMemoryUnit {
+		return fmt.Errorf("gpu memory must not be less than %d bytes", cambriconVMemoryUnit)
+	}
+	vmemory := memory.Value() / cambriconVMemoryUnit
+
+	if err := a.lockNode(ctx, ctx.client, ctx.node); err != nil {
```
ZiMengSheng, line 203: If err != nil, does it mean rescheduling?

zqzten: Yes, it will lead to a prebind error and trigger rescheduling.

PR #2636, opened 2025-09-25
koordlet: fix partition in buildXPUDeviceAnnotations()

ZhuZhezz:
Partition annotations should only exist if the device information includes p2p links.

None

### V. Checklist


pkg/koordlet/statesinformer/impl/states_device_linux.go
```diff
@@ -568,7 +568,7 @@ func (s *statesInformer) buildXPUDeviceLabels(xpuDevices koordletuti.XPUDevices)
 
 func (s *statesInformer) buildXPUDeviceAnnotations(xpuDevices koordletuti.XPUDevices) map[string]string {
 	partitionTable := getPartitionTableFromXPUDevices(xpuDevices)
-	if partitionTable == nil {
+	if partitionTable == nil || len(partitionTable) == 0 {
```
zqzten:
```suggestion
	if len(partitionTable) == 0 {
```

PR #2647, opened 2025-10-14
scheduler: fix nominated reservation during preemption

saintube:

koord-scheduler: fix the nominated reservation is deleted even when preempting success.

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
@@ -71,7 +71,13 @@ func MakeReservationErrorHandler(
 		// If the Pod failed to schedule or no post-filter plugins, should remove exist NominatedReservation of the Pod.
 		if extendedHandle, ok := fwk.(frameworkext.ExtendedHandle); ok {
 			if reservationNominator := extendedHandle.GetReservationNominator(); reservationNominator != nil {
-				reservationNominator.DeleteNominatedReservePodOrReservation(pod)
+				// If the pod preempting successfully, we should keep the nomination of the pod to reservation, and other pods can not allocate
+				// the preempted reserved resources in the next cycle.
+				if nominatingInfo.NominatingMode == framework.ModeOverride && nominatingInfo.NominatedNodeName == "" || nominatingInfo.NominatingMode == framework.ModeNoop && pod.Status.NominatedNodeName == "" {
+					reservationNominator.DeleteNominatedReservePodOrReservation(pod)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
```diff
@@ -71,7 +71,13 @@ func MakeReservationErrorHandler(
 		// If the Pod failed to schedule or no post-filter plugins, should remove exist NominatedReservation of the Pod.
 		if extendedHandle, ok := fwk.(frameworkext.ExtendedHandle); ok {
 			if reservationNominator := extendedHandle.GetReservationNominator(); reservationNominator != nil {
-				reservationNominator.DeleteNominatedReservePodOrReservation(pod)
+				// If the normal pod preempting successfully, we should keep the nomination of the pod to reservation.
```
ZiMengSheng: The above is the community implementation. We recommend aligning it to support the `<nil, Unschedulable>` case. The suggested changes are as follows:
```
if (nominatingInfo.Mode() == framework.ModeOverride && nominatingInfo.NominatedNodeName == "") || (nominatingInfo.Mode() == framework.ModeNoop && pod.Status.NominatedNodeName == "") {
  reservationNominator.DeleteNominatedReservePodOrReservation(po
}
```
Possible preemption return scenarios:
```
// Preempt returns a PostFilterResult carrying the suggested nominatedNodeName, along with a Status.
// The semantics of returned <PostFilterResult, Status> vary depending on different scenarios:
//
// - <nil, Error>. This denotes a transient/rare error that may be self-healed in future cycles.
//
// - <nil, Unschedulable>. This status is mostly as expected, as the preemptor is waiting for the
// victims to be fully terminated.
//
// - In both cases above, a nil PostFilterResult is returned to keep the pod's nominatedNodeName unchanged.
//
// - <non-nil PostFilterResult, Unschedulable>. It indicates the pod cannot be scheduled even with preemption.
// In this case, a non-nil PostFilterResult is returned and result.NominatingMode instructs how to deal with
// the nominatedNodeName.
//
// - <non-nil PostFilterResult, Success>. It's the regular happy path
// and the non-empty nominatedNodeName will be applied to the preemptor pod.
func (ev *Evaluator) Preempt(ctx context.Context, pod *v1.Pod, m framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status)
```

saintube: OK

PR #2649, opened 2025-10-16
koordlet: add metric for cpuset sharepool cpus.

tan90github:

- We expect to be able to monitor and observe the CPUSet share pool information, similar to `koordlet_be_suppress_cpu_cores`. 
- The observability of cpuset is insufficient currently. In some cases, we need to know the number of share pool CPU on the node, so as to guide us whether the node has allocated too many LSR/LSE Pods and how many cores the normal pods can actually use (in scenarios with resource oversale). Currently, this can only be achieved by adjusting the log level.

### V. Checklist

### DIFF
--- a/pkg/koordlet/metrics/cpu_cpuset.go
+++ b/pkg/koordlet/metrics/cpu_cpuset.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	CPUSetSharePoolCPUS = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "cpuset_share_pool_cpu_cores",
+		Help:      "Number of share pool cores",
+	}, []string{NodeKey})
+
+	CPUSetBESharePoolCPUS = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "cpuset_be_share_pool_cpu_cores",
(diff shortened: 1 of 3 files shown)


pkg/koordlet/metrics/cpu_cpuset.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	CPUSetSharePoolCPUS = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "cpuset_share_pool_cpu_cores",
```
saintube, line 24: Perhaps we also need the metric about the BE sharepool?

pkg/koordlet/metrics/cpu_cpuset.go
```diff
@@ -0,0 +1,53 @@
...
+	}, []string{NodeKey})
+
+	CPUSetBESharePoolCPUS = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "cpuset_be_share_pool_cpu_cores",
+		Help:      "Number of be share pool cores",
+	}, []string{NodeKey})
+
+	CPUSetCollector = []prometheus.Collector{
+		CPUSetSharePoolCPUS,
```
saintube, line 35: need to register all the new metrics

pkg/koordlet/metrics/cpu_cpuset.go
tan90github: lgtm, added.

pkg/koordlet/metrics/cpu_cpuset.go
tan90github: done

pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
```diff
@@ -112,10 +113,29 @@ func (r *cpusetRule) getContainerCPUSet(containerReq *protocol.ContainerRequest)
 	}
 
 	allSharePoolCPUs := make([]string, 0, len(r.sharePools))
+	var allShareCPUSetCount, beShareCPUSetCount int
 	for _, nodeSharePool := range r.sharePools {
 		allSharePoolCPUs = append(allSharePoolCPUs, nodeSharePool.CPUSet)
+		nodeSharePoolCPUSet, err := cpuset.Parse(nodeSharePool.CPUSet)
```
saintube: How about recording the sharepool in the statesinformer?
This method is invoked heavily at the container-level instead of the sharepool updating. I'm afraid that adding the cpuset parsing here introduces non-negligible overhead.

tan90github: good advice

PR #2650, opened 2025-10-16
koordlet: exclude device NUMANode

ZiMengSheng:
### V. Checklist

### DIFF
--- a/pkg/koordlet/util/meminfo.go
+++ b/pkg/koordlet/util/meminfo.go
@@ -255,6 +255,8 @@ func GetNodeNUMAInfo() (*NodeNUMAInfo, error) {
 		return nil, fmt.Errorf("failed to read NUMA dir, err: %w", err)
 	}
 
+	giNUMANodes := system.GetNUMANodesHasGI()
+
 	result := &NodeNUMAInfo{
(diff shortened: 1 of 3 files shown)


pkg/koordlet/util/generic_initiator.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
```
saintube: why not place in the system pkg

pkg/koordlet/util/meminfo.go
```diff
@@ -273,6 +275,11 @@ func GetNodeNUMAInfo() (*NodeNUMAInfo, error) {
 		}
 		nodeID := int32(nodeIDRaw)
 
+		if giNUMANodes.Has(nodeID) {
+			klog.V(4).Infof("node %d has GI, skip", nodeID)
```
saintube: Suggested to set as V(5), otherwise it would be printed every second.

PR #2653, opened 2025-10-20
scheduler: support allocated by desinated resource

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -180,6 +184,11 @@ func (p *Plugin) PreFilter(ctx context.Context, cycleState *framework.CycleState
 	if !status.IsSuccess() {
 		return nil, status
 	}
+	if schedulingHint := hinter.GetSchedulingHintState(cycleState); schedulingHint != nil {
+		if _, ok := schedulingHint.Extensions[Name]; !ok {
+			state.designatedAllocation = nil
```
saintube: We should allocate from the designated devices only when the scheduling hint is specified?

ZiMengSheng: Is that the case now?

PR #2660, opened 2025-10-23
scheduler: coscheduling tweak and fix

zqzten:

* improve gang param parse error observability
* fix some network topology aware scheduling bug

### V. Checklist

### DIFF
--- a/apis/extension/coscheduling.go
+++ b/apis/extension/coscheduling.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"sort"
 	"strconv"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/coscheduling/coscheduling.go
```diff
@@ -119,6 +119,10 @@ func (cs *Coscheduling) Name() string {
 func (cs *Coscheduling) PreEnqueue(ctx context.Context, pod *v1.Pod) *framework.Status {
 	if err := cs.pgMgr.PreEnqueue(ctx, pod); err != nil {
 		klog.ErrorS(err, "PreEnqueue failed", "pod", klog.KObj(pod))
+		// Currently PreEnqueue error would result in no events, so we manually event here.
```
ZiMengSheng: 1. PreEnqueue is now used not only to detect PodGroup creation and MinMember, but also to ensure that a group group in the queue has only one Pod. When used for the latter, events should not be emitted. 
2. Furthermore, if PreEnqueue is called too often, it will be difficult to determine the number of coroutines involved. 

Therefore, if events need to be emitted, I recommend placing them in the controller.

PR #2661, opened 2025-10-23
scheduler: improve ut for network-topology aware scheduling

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/coscheduling/core/network_topology_workflow_test.go
```diff
@@ -132,6 +132,7 @@ func TestPodGroupManager_NetworkTopology(t *testing.T) {
 		existingNominatedPods  []*corev1.Pod
 		filterPlugin           schedulertesting.RegisterPluginFunc
 		wantScheduleStatus     *framework.Status
+		wantPlanedPodToNode    map[string]string
```
zqzten:
```suggestion
		wantPlannedPodToNode    map[string]string
```

PR #2688, opened 2025-11-11
scheduler: fix coscheduling prebind with illegal labels

saintube:

koord-scheduler: Fix the issue that pods with gang scheduling may fail the pre-bind due to the Coscheduling plugin patches the pod with invalid labels.

It is introduced in #2655. The value `GangGroupId` may be "ns-xxx/job-yyy" or "job-a,job-b,job-c" which are not the legal format of the label values. In this patch, we use annotations instead of labels to denote the gang binding info.

### V. Checklist

### DIFF
--- a/apis/extension/coscheduling.go
+++ b/apis/extension/coscheduling.go
@@ -187,8 +187,8 @@ func SortPodsByIndex(pods []*corev1.Pod) {
 }
 
 const (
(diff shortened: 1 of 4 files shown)


ZiMengSheng (commented):
/lgtm
/approve

PR #2692, opened 2025-11-12
scheduler: support reservation scale changed by spec

lijunxin559:

It is necessary to provide changes to the requested resources for reservations that are currently available. Since the source of resource requests is templates, it is hoped that changes can also be made through the same fields. However, currently only the status field can be provided to reverse correct the scheduling ledger, which results in template resource request changes being forcibly corrected even if they are modified.

Add features ReservationEnableUpdateSpec to enable reservation scale changed by spec template, and ignore status allocatable changes when ReservationEnabling Update Spec is set

### V. Checklist


ZiMengSheng:
@saintube PTAL

PR #2703, opened 2025-11-20
koordlet: set pod priority value to default evict priority

lijunxin559:
### V. Checklist


pkg/koordlet/qosmanager/plugins/util/evict_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/koordlet/util/testutil"
 )
 
+const DefaultEvictPriority = 5999
```
saintube: use a private variable

PR #2707, opened 2025-11-25
chore(deps): use exported kubernetes qos package

hunshcn:
The use of k/k should be minimized.

### DIFF
--- a/apis/extension/qos_utils.go
+++ b/apis/extension/qos_utils.go
@@ -18,7 +18,7 @@ package extension
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	"k8s.io/kubectl/pkg/util/qos"
 )
 
 // NOTE: functions in this file can be overwritten for extension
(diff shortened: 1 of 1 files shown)


apis/extension/qos_utils.go
```diff
@@ -18,7 +18,7 @@ package extension
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	"k8s.io/kubectl/pkg/util/qos"
```
ZiMengSheng, line 21: why we reference kubectl

hunshcn: k/k should not be imported.

We should use this exported package to instead.

hunshcn: This GetPodQOS func in 
`k8s.io/kubernetes/pkg/apis/core/v1/helper/qos` and
`k8s.io/kubectl/pkg/util/qos` are same.

hunshcn: And this will lead to abnormal import of k/k for the use of clientset.

ZiMengSheng (approved):

Issue #2581, opened 2025-08-18
[proposal] Introducing QueueingHint into the elasticquota plugin


qinfustu:
**What is your proposal**:

The Kubernetes scheduling framework provides a QueueingHint mechanism. If this mechanism could be integrated into elastic quota management, it would help mitigate the issue where multiple resource quotas exist in a cluster. Specifically, when some quotas have a large number of high-priority pods stuck in a pending state, they may monopolize scheduling cycles and prevent pods in other quotas—even those with sufficient resources—from being scheduled.

​​Introduction to QueueingHint​​：
1. https://kubernetes.io/blog/2024/12/12/scheduler-queueinghint/
2. https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/4247-queueinghint/README.md
3. https://cloud.tencent.com/developer/article/2393595

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

ZiMengSheng:
Good Suggestion! Maybe moving some quota check to preEnqueue is enough!

qinfustu:
> Good Suggestion! Maybe moving some quota check to preEnqueue is enough!

I understand that there are two aspects to reducing wasted scheduling cycles here:

1. **QueueingHint** enables precise control over which cluster events can trigger a pod to move from the **unschedulableQ** to the **activeQ**. If **QueueingHint** is properly implemented, it can significantly reduce ineffective scheduling. For example, in the elasticquota plugin, pod deletion events where no quota is configured can be ignored.
2. **preEnqueue** is used to determine which pods are eligible to enter the scheduling logic. Pods rejected during the **preEnqueue** phase will also rely on **ClusterEventWithHint** to trigger their re-entry into the **preEnqueue** phase.

ZiMengSheng:
> > Good Suggestion! Maybe moving some quota check to preEnqueue is enough!
> 
> I understand that there are two aspects to reducing wasted scheduling cycles here:
> 
> 1. **QueueingHint** enables precise control over which cluster events can trigger a pod to move from the **unschedulableQ** to the **activeQ**. If **QueueingHint** is properly implemented, it can significantly reduce ineffective scheduling. For example, in the elasticquota plugin, pod deletion events where no quota is configured can be ignored.
[...]

PR #2717, opened 2025-12-02
scheduler: Introducing QueueingHint into the elasticquota plugin

BUPT-wxq:
proposal: https://github.com/koordinator-sh/koordinator/issues/2581

This PR implements QueueHint optimization for ElasticQuota plugin to filter unnecessary pod re-queuing events, significantly reducing invalid scheduling retries and improving scheduler performance. QueueHint only performs coarse-grained filtering, not detailed resource size comparisons.

**New Feature: `EnableQueueHint` Parameter**

A new `EnableQueueHint` parameter has been added to `ElasticQuotaArgs` to control the QueueHint functionality:

- **When `EnableQueueHint` is `false` (default):** The plugin behavior remains unchanged - registers Pod Delete and ElasticQuota All events without QueueingHintFn, maintaining backward compatibility.

- **When `EnableQueueHint` is `true`:** The plugin registers Pod Delete and ElasticQuota Update events with QueueingHintFn for intelligent filtering.

**ElasticQuota Update Event Filtering (`isSchedulableAfterQuotaChanged`):**

The QueueHint mechanism performs coarse-grained filtering to determine if a quota update may affect pod schedulability:

1. **Basic Filtering:**
   - Filters re-queuing when unschedulable pod has no quota association
   - Filters re-queuing when modified quota and pod are in different quota trees

2. **Change Detection:**
   - Uses `IsQuotaChange` to detect changes in quota spec (Max, Min, SharedWeight, ParentName)
   - Uses `IsQuotaUpdated` (from hook plugins) to detect changes from hook plugin perspective
   - Filters re-queuing when quota has not changed (both `IsQuotaChange` and `IsQuotaUpdated` return false)

3. **Path-based Filtering:**
   - If quota has changed, checks if the modified quota is in the pod's path to root quota
   - Allows re-queuing if the modified quota is the pod's direct quota
   - When `EnableCheckParentQuota` is enabled, allows re-queuing if the modified quota is an ancestor quota (in the path from pod's quota to root)
   - Filters re-queuing when the modified quota is not in the pod's path to root

**Pod Delete Event Filtering (`isSchedulableAfterPodDeletion`):**

When `EnableQueueHint` is enabled, pod delete events are also filtered using QueueHint logic:

1. **Basic Filtering:**
   - Filters re-queuing when deleted pod has no quota association
   - Filters re-queuing when unschedulable pod has no quota association
   - Filters re-queuing when deleted pod and unschedulable pod are in different quota trees
   - Filters re-queuing when deleted pod  has not been assigned resource

2. **Path-based Filtering:**
   - Allows re-queuing if the deleted pod is in the same quota as the unschedulable pod
   - When `EnableCheckParentQuota` is enabled, allows re-queuing if the deleted pod's quota is in the unschedulable pod's path to root
   - Filters re-queuing when the deleted pod's quota is not in the unschedulable pod's path to root

**Why only check quota/pod events in the pod's path to root:**

1. **These quotas are the direct factors that affect whether a pod can be scheduled** - The quotas in the pod's path from its direct quota to the root quota directly determine the pod's resource allocation and scheduling eligibility.

2. **Add and delete quota events will also be reflected in this path** - When a quota is added or deleted, it will trigger update events on related quotas in the hierarchy.  Separate add/delete event handlers are unnecessary.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/plugin.go
+++ b/pkg/scheduler/plugins/elasticquota/plugin.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
+	k8sfeature "k8s.io/apiserver/pkg/util/feature"
 	v1 "k8s.io/client-go/listers/core/v1"
 	policylisters "k8s.io/client-go/listers/policy/v1"
 	"k8s.io/client-go/tools/cache"
@@ -34,8 +35,6 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/framework/preemption"
 	"k8s.io/kubernetes/pkg/scheduler/metrics"
 	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
-
-	k8sfeature "k8s.io/apiserver/pkg/util/feature"
 
 	"github.com/koordinator-sh/koordinator/apis/extension"
 	"github.com/koordinator-sh/koordinator/apis/thirdparty/scheduler-plugins/pkg/apis/scheduling"


pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -215,8 +216,9 @@ func (g *Plugin) EventsToRegister() []framework.ClusterEventWithHint {
 	// https://git.k8s.io/kubernetes/pkg/scheduler/eventhandlers.go#L403-L410
 	eqGVK := fmt.Sprintf("elasticquotas.v1alpha1.%v", scheduling.GroupName)
 	return []framework.ClusterEventWithHint{
-		{Event: framework.ClusterEvent{Resource: framework.Pod, ActionType: framework.Delete}},
-		{Event: framework.ClusterEvent{Resource: framework.GVK(eqGVK), ActionType: framework.All}},
+		{Event: framework.ClusterEvent{Resource: framework.Pod, ActionType: framework.Delete}, QueueingHintFn: g.isSchedulableAfterPodDeletion},
+		{Event: framework.ClusterEvent{Resource: framework.GVK(eqGVK), ActionType: framework.Add | framework.Delete}},
+		{Event: framework.ClusterEvent{Resource: framework.GVK(eqGVK), ActionType: framework.Update}, QueueingHintFn: g.isSchedulableAfterQuotaChanged},
```
TaoYang526: Is it better to consider all quota events (not only Update events) with queueing hint so that we can skip events of non-related quota?

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +280,134 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+// QueueAfterBackoff is default queueingHintFn behavior.
+func (g *Plugin) isSchedulableAfterPodDeletion(logger klog.Logger, pod *corev1.Pod, oldObj, newObj interface{}) framework.QueueingHint {
+	deletedPod, _, err := schedutil.As[*corev1.Pod](oldObj, nil)
+	if err != nil {
+		logger.Error(err, "Failed to convert oldObj to Pod", "oldObj", oldObj)
+		return framework.QueueAfterBackoff
+	}
+
+	// Get quota information for deleted pod and unschedulable pod
+	deletedPodQuotaName, deletedPodTreeID := g.getPodAssociateQuotaNameAndTreeID(deletedPod)
```
TaoYang526: If the deleted pod is still in Pending phase, the result could be QueueSkip, right?

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +280,134 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+	unschedulablePodQuotaName, unschedulablePodTreeID := g.getPodAssociateQuotaNameAndTreeID(pod)
+
+	// If unschedulable pod has no quota association, skip
+	// If deleted pod has no quota or different tree, skip
+	if unschedulablePodQuotaName == "" || deletedPodQuotaName == "" || deletedPodTreeID != unschedulablePodTreeID {
```
TaoYang526: The deleted pod in another quota may release resources for unschedulable pod, for example, they has the same ancestor quota which reached the limit but now has enough resources for the unschedulable pod.

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +280,134 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+		return framework.QueueSkip
+	}
+
+	return framework.QueueAfterBackoff
```
TaoYang526: I think we could rethink whether pod event is necessary to be considered in queueing hint, if the pod (with event) has take effect on a specified quota, a related quota UPDATE event should be generated. Is it enough to just watch the quota events?

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +280,121 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+	if runtimeNotIncreased {
+		// Runtime not increased, skip
+		return framework.QueueSkip
+	}
+
+	return framework.QueueAfterBackoff
+}
+
+// isAncestorQuota checks if ancestorQuotaName is an ancestor of quotaInfo
+func isAncestorQuota(mgr *core.GroupQuotaManager, quotaInfo *core.QuotaInfo, ancestorQuotaName string) bool {
```
TaoYang526: 2nd parameter(ancestorQuotaName) is still need to be checked, it's not a proper name.

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +280,121 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+
+	newRuntime, err := extension.GetRuntime(modifiedQuota)
+	if err != nil {
+		return framework.QueueAfterBackoff
+	}
+
+	runtimeNotIncreased, _ := quotav1.LessThanOrEqual(newRuntime, oldRuntime)
+	if runtimeNotIncreased {
+		// Runtime not increased, skip
+		return framework.QueueSkip
```
TaoYang526: hook-plugins may also make the pod unschedulable, we can't compare the resource directly, but can check if quota updated from hook plugins' perspective via GroupQuotaManager#IsQuotaUpdated.

pkg/scheduler/plugins/elasticquota/plugin.go
TaoYang526: We should keep this if the used resource won't be detected in quota events.

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -215,8 +216,9 @@ func (g *Plugin) EventsToRegister() []framework.ClusterEventWithHint {
 	// https://git.k8s.io/kubernetes/pkg/scheduler/eventhandlers.go#L403-L410
 	eqGVK := fmt.Sprintf("elasticquotas.v1alpha1.%v", scheduling.GroupName)
 	return []framework.ClusterEventWithHint{
-		{Event: framework.ClusterEvent{Resource: framework.Pod, ActionType: framework.Delete}},
-		{Event: framework.ClusterEvent{Resource: framework.GVK(eqGVK), ActionType: framework.All}},
+		{Event: framework.ClusterEvent{Resource: framework.Pod, ActionType: framework.Delete}, QueueingHintFn: g.isSchedulableAfterPodDeletion},
```
TaoYang526: Can we make the queueing hint configurable and compatible, for example, add EnableQueueingHint field and set false by default in ElasticQuotaArgs?

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: Quota add or delete events will eventually trigger update events in the pod's quota path.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: I can directly use CheckPodIsAssigned to verify whether the pod has been assigned resources. If not assigned, skip it directly.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: Your are right!!!

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: Uses the existing `GetCurToAllParentGroupQuotaInfo` method to check if the modified quota is in the pod's path to root, removing `isAncestorQuota` func.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: You're right, thanks.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: Thanks for the suggestion! I've implemented the `EnableQueueHint` parameter in `ElasticQuotaArgs`

ZiMengSheng:
PTAL  /cc @qinfustu

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +289,124 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+// QueueAfterBackoff is default queueingHintFn behavior.
+func (g *Plugin) isSchedulableAfterQuotaChanged(logger klog.Logger, pod *corev1.Pod, oldObj, newObj interface{}) framework.QueueingHint {
+	originalQuota, modifiedQuota, err := schedutil.As[*apiv1alpha1.ElasticQuota](oldObj, newObj)
+	if err != nil {
+		logger.Error(err, "Failed to convert oldObj or newObj to ElasticQuota", "oldObj", oldObj, "newObj", newObj)
+		return framework.QueueAfterBackoff
+	}
+
+	if originalQuota == nil || modifiedQuota == nil {
+		return framework.QueueAfterBackoff
```
saintube, line 370: Why do backoff when the quota is nil

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +289,124 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+	}
+
+	// Quota has changed, check if modified quota is in the pod's path to root
+	if modifiedQuota.Name == podQuotaName {
+		// Modified quota is the pod's quota, allow queueing
+		return framework.QueueAfterBackoff
+	}
+
+	// Modified quota is not the pod's quota, check if it's in the path to root
+	if g.pluginArgs.EnableCheckParentQuota {
```
saintube, line 406: Please be careful of the overhead of the queueing hint function. It will be invoked inside the SchedulingQueue's lock for each unschedulable pod.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: QueueAfterBackoff is the previous behavior; if an error occurs or parsing fails, follow the previous behavior.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: You are right. change to no lock.

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: isSchedulableAfterPodDeletion will process it

pkg/scheduler/plugins/elasticquota/plugin.go
BUPT-wxq: I will keep it.

pkg/scheduler/plugins/elasticquota/plugin.go
saintube: Is it safe to read without a lock?

saintube: For a queueing hint, we can relax the validation to improve performance. For example, use a snapshot to check instead of the latest quota info.

BUPT-wxq: No lock is unsafe. use snapshot to check.

pkg/scheduler/apis/config/types.go
```diff
@@ -251,6 +251,10 @@ type ElasticQuotaArgs struct {
 	// EnableQueueHint if true, enables QueueHint optimization to filter unnecessary pod re-queuing events.
 	EnableQueueHint bool
 
+	// QuotaSnapshotUpdateInterval is the interval to update quota snapshot for QueueHint optimization.
+	// Defaults to 5 seconds if unspecified.
```
saintube: The defaultQuotaSnapshotUpdateInterval is set to 120s.

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -278,6 +312,161 @@ func (g *Plugin) PreFilterExtensions() framework.PreFilterExtensions {
...
+		g.quotaSnapshotLock.RLock()
+		snapshot, exists := g.quotaSnapshot[podTreeID]
```
saintube: try to wrap as a function

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -375,3 +564,41 @@ func (g *Plugin) Unreserve(ctx context.Context, state *framework.CycleState, p *
 func (g *Plugin) GetQuotaInformer() cache.SharedIndexInformer { // expose for extensions
 	return g.quotaInformer
 }
+
+// updateQuotaSnapshot periodically updates quota snapshot for all quota trees
+// This runs in background and doesn't block the main scheduling path
+func (g *Plugin) updateQuotaSnapshot() {
+	// Copy quotaToTreeMap
+	g.quotaToTreeMapLock.RLock()
+	quotaToTreeMapCopy := make(map[string]string, len(g.quotaToTreeMap))
```
saintube: try to wrap as a function

saintube:
PTAL /cc @shaloulcy

pkg/scheduler/plugins/elasticquota/plugin.go
```diff
@@ -33,13 +33,17 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/preemption"
 	"k8s.io/kubernetes/pkg/scheduler/metrics"
+	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
+
+	k8sfeature "k8s.io/apiserver/pkg/util/feature"
```
saintube: need goimports

saintube:
PTAL /cc @shaloulcy @ZiMengSheng

PR #2733, opened 2025-12-12
scheduler: run PreBindReservation plugins orderly

saintube:

Currently, the plugin extension PreBindReservation runs out of order due to the registration `updatePlugins`. However, we expect the PreBindReservation to serve as a drop-in replacement for PreBind plugins against the *Reservation* objects, maintaining the same execution order as PreBind rather than in a random sequence. For example, in some use cases, we may want a plugin implementing PreBindReservation that relies on preceding PreBindReservation plugins.

Therefore, with this change, we ensure the execution order of PreBindReservation plugins aligns with the registration order of PreBind plugins.
Please note that it requires every PreBindReservation plugin to implement the fundamental interface PreBindPlugin. Otherwise, the PreBindReservation plugin will not be invoked by the FrameworkExtender.

### V. Checklist


ZiMengSheng (approved):
/lgtm

PR #2738, opened 2025-12-15
scheduler: fix pods fail to properly use a reservation yet still consume its resources

qinfustu:
When a reservation holds resources, pods cannot correctly use the reservation, even though they are using the reserved resources.

fix: https://github.com/koordinator-sh/koordinator/issues/2729

### V. Checklist


pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -439,6 +435,14 @@ func (pl *Plugin) filterWithReservations(ctx context.Context, cycleState *framew
 		}
 	}
 
+	if len(allInsufficientResourceReasonsByReservation) > 0 {
```
saintube: If a pod without the reservation-affinity does not fit any matched reservation on that node, but fits the unallocated and unreserved resources of the node, it should pass the filter.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -861,7 +865,7 @@ func (pl *Plugin) FilterNominateReservation(ctx context.Context, cycleState *fra
 		return pl.filterWithPreAllocatablePods(ctx, cycleState, rInfo, nodeInfo, []*corev1.Pod{pod}, true)
 	}
 
-	return pl.filterWithReservations(ctx, cycleState, pod, nodeInfo, []*frameworkext.ReservationInfo{rInfo}, true)
+	return pl.filterWithReservations(ctx, cycleState, pod, nodeInfo, []*frameworkext.ReservationInfo{rInfo})
```
saintube: Just so you know, the FilterNominateReservation requires at least one reservation.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -439,6 +435,14 @@ func (pl *Plugin) filterWithReservations(ctx context.Context, cycleState *framew
 		}
 	}
 
+	if len(allInsufficientResourceReasonsByReservation) > 0 {
+		preemptibleResource := framework.NewResource(state.preemptible[node.Name])
+		insufficientResourcesByNode := fitsNode(state.podRequestsResources, nodeInfo.Allocatable, nodeRState.podRequested, nodeRState.rAllocated, nil, len(nodeRState.matchedOrIgnored), len(nodeInfo.Pods), preemptibleResource)
```
ZYecho11: The logic of fitsNode seems to have already been called in fitsNodeAndReservation, is there some duplication here?

qinfustu: > The logic of fitsNode seems to have already been called in fitsNodeAndReservation, is there some duplication here?

I understand that fitsNodeAndReservationsuses isFitsNodeSkippedto decide whether to skip fitsNode, and it mainly helps check if the Reservation requirements are met. The outer fitsNodecheck verifies whether the node can still accommodate the Pod after deducting the resources reserved by reservations.

pkg/scheduler/plugins/reservation/plugin.go
saintube: Originally, the rInfo fails the FilterNominateReservation when the pod without the reservation-affinity does not pass the fitsReservation.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -439,6 +435,17 @@ func (pl *Plugin) filterWithReservations(ctx context.Context, cycleState *framew
 		}
 	}
 
+	if len(allInsufficientResourceReasonsByReservation) > 0 {
```
ZiMengSheng: 这里看起来是想证明是由于 Reservation 导致的失败，所以来看 FitsNode。但是假设 len(insufficientResourcesByNode) > 0 和  len(insufficientResourceReasonsByReservation) > 0 的时候，其实只要 len(allInsufficientResourcesByNode) > 0  就可以不再看 FitsNode 了。大致结构类似于

```go
  func (pl *Plugin) filterWithReservations(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodeInfo *framework.NodeInfo, matchedReservations []*frameworkext.ReservationInfo, requiredFromReservation bool) *framework.Status {
    for _, rInfo := range matchedReservations {
      ...
    }

  	if requiredFromReservation {
		// originalCode ...
	} else {
		var failureReasons []string
		if len(allInsufficientResourcesByNode) > 0 {
			// If the combination of reservation and node cannot satisfy the pod, then the node alone cannot satisfy it either.
			failureReasons = make([]string, 0, len(allInsufficientResourcesByNode))
			for insufficientResourceByNode := range allInsufficientResourcesByNode {
				failureReasons = append(failureReasons, fmt.Sprintf("Insufficient %s by node", insufficientResourceByNode))
			}
		} else {
			// try to allocate from node alone
			preemptibleResource := framework.NewResource(state.preemptible[node.Name])
			insufficientResourcesByNode := fitsNode(state.podRequestsResources, nodeInfo.Allocatable, nodeRState.podRequested, nodeRState.rAllocated, nil, len(nodeRState.matchedOrIgnored), len(nodeInfo.Pods), preemptibleResource)
			failureReasons = make([]string, 0, len(insufficientResourcesByNode))
			for _, insufficientResourceByNode := range insufficientResourcesByNode {
				failureReasons = append(failureReasons, fmt.Sprintf("Insufficient %s by node", insufficientResourceByNode))
			}
		}
		if len(failureReasons) > 0 {
			return framework.NewStatus(framework.Unschedulable, failureReasons...)
		}
	}
	return nil
}
```

pkg/scheduler/apis/config/v1beta3/types.go
```diff
@@ -244,13 +244,6 @@ type ElasticQuotaArgs struct {
 	// DisableDefaultQuotaPreemption if true, will not preempt pods in default quota.
 	DisableDefaultQuotaPreemption *bool `json:"disableDefaultQuotaPreemption,omitempty"`
 
-	// EnableQueueHint if true, enables QueueHint optimization to filter unnecessary pod re-queuing events.
```
ZiMengSheng, line 247: git rebase upstream/main

PR #2750, opened 2025-12-24
scheduler: fix filter issue for non-required pre-allocation reservation

TaoYang526:

This PR fixes an issue in the `filterWithPreAllocatablePods` function where the Filter phase could pass even when node resources were insufficient, leading to subsequent Reserve phase failures.

UT

**Problem:**
When `isPreAllocationRequired` is `false`, the previous implementation would directly return success without validating whether the node has sufficient resources. This caused a mismatch between Filter and Reserve phases - Filter passed but Reserve failed due to insufficient resources.

**Solution:**
Now when `isPreAllocationRequired` is `false`, we first check if the reservation can be placed using node-unallocated resources:
- If sufficient: return success immediately (no need to use pre-allocatable pods)
- If insufficient: continue to check pre-allocatable pods to find one that can be released to satisfy the reservation requirements

This ensures consistency between Filter and Reserve phases.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -548,7 +548,7 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
 		return framework.NewStatus(framework.Unschedulable, failureReasons...)
 	} else {
 		if len(allInsufficientResourcesByNode) > 0 {
-			// If the combination of reservation and node cannot satisfy the pod, then the node alone cannot satisfy it either.
+			// If the combination of pre-allocatable and node cannot satisfy the reservation, then the node alone cannot satisfy it either.
 			failureReasons = buildNodeFailureReasons(allInsufficientResourcesByNode.List())
 		} else {
 			// Check if the reserve pod can be placed with node-unallocated resource when pre-allocation is not required.


TaoYang526:
@saintube Could you please help review this PR? Thanks!

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -492,6 +486,18 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
 	preemptible := framework.NewResource(state.preemptible[node.Name])
 	state.preemptLock.RUnlock()
 
+	var insufficientResourcesByNodeUnallocated []string
```
ZiMengSheng: In the current implementation, this code will execute regardless. However, if there are Pods that can fulfill the PreAllocation requirement, this code will not need to execute.

TaoYang526: @ZiMengSheng Please let me know if there is any more concerns. Thanks!

TaoYang526:
Last update: separate failure reason handling by isPreAllocationRequired.
@saintube @ZiMengSheng  Please take a look, thanks!

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -492,6 +486,18 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
...
+	if !isPreAllocationRequired {
+		// Check if the reserve pod can be placed with node-unallocated resource when pre-allocation is not required.
+		// For reserve pod, matchedOrIgnored should be 0, both rAllocated and rRemained should be nil.
+		insufficientResourcesByNodeUnallocated = fitsNode(state.podRequestsResources, nodeInfo.Allocatable,
```
saintube: To keep consistency with the `filterWithReservations`, this check is recommended to move after the loop in L503~L544.

TaoYang526: @saintube The current order is intentional for performance reasons:
* `fitsNode` is a lightweight check - just resource comparison
* The loop (L503-544) is more expensive - it iterates through pre-allocatable pods and runs `fitsNodeAndReservation` + `RunReservationFilterPlugins` for each

So checking node-unallocated resources first acts as a fast-path: if sufficient, we skip the expensive loop entirely. This is especially beneficial when pre-allocatable pods list is large.

For `filterWithReservations`, the loop checks multiple reservations which is the core logic. But here, the node resource check is independent and can short-circuit early.

Thoughts?

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -537,19 +531,39 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
...
+	if isPreAllocationRequired {
+		failureReasons = append(failureReasons, buildNodeFailureReasons(allInsufficientResourcesByNode.List())...)
+		failureReasons = append(failureReasons, allInsufficientResourceReasonsByReservation...)
+		if len(failureReasons) == 0 {
+			failureReasons = append(failureReasons, ErrReasonNoPodsMeetPreAllocationRequirements)
+		}
+		return framework.NewStatus(framework.Unschedulable, failureReasons...)
+	} else {
+		if len(allInsufficientResourcesByNode) > 0 {
+			// If the combination of reservation and node cannot satisfy the pod, then the node alone cannot satisfy it either.
```
ZiMengSheng: the combination of pre-allocatable pods?

saintube:
```suggestion
			// If the combination of pre-allocatable and node cannot satisfy the reservation, then the node alone cannot satisfy it either.
```

PR #2757, opened 2025-12-29
scheduler: add ReservationNominate method

### LINKED ISSUE #2753: [BUG] Missing nominated reservation for DeviceShare Reserve when the Filter phase pass only one node
**What happened**:

Currently, the DeviceShare and NodeNUMAResource plugins, which support allocating from reservations, require the Reservation plugin to pick a nominated reservation on the node before their Reserve phase. And the Reservation plugin does nomination at the phase PreScore and Reserve.

However, with the introduced extended phase ResizePod, the reserved resources allocation of DeviceShare is moved forward before the Reservation's Reserve. Meanwhile, the PreScore phase would be skipped when the Filter phase only selects one node. In this case, the reservation nomination is skipped, making the DeviceShare plugin not allocate reserved resources.

Since the reservation nomination is a special phase in Koordinator, which should not be skipped or reordered by the internal framework, we propose adding a dedicated phase in the FrameworkExtender for it.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

saintube:

#2753

Fix the bug where the Reservation nomination and the PreAllocation nomination could be missing when there is only one node passing the Filter phase.

fixes #2753

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/fake.go
+++ b/pkg/scheduler/frameworkext/fake.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 
+	"github.com/koordinator-sh/koordinator/apis/extension"
 	reservationutil "github.com/koordinator-sh/koordinator/pkg/util/reservation"
 )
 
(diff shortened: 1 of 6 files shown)


ZiMengSheng (approved):
/lgtm
/approve

PR #2765, opened 2026-01-07
scheduler: fix incorrect resource check for pre-allocation reservation when no pre-allocatable pods exist on node

TaoYang526:

When there are no pre-allocatable pods on a node, nodeReservationStates doesn't contain an entry for that node. The code creates an empty nodeReservationState with nil podRequested, causing incorrect resource availability calculation in filterWithPreAllocatablePods.

This PR fixes the issue by initializing podRequested from nodeInfo.Requested.Clone() when no pre-allocatable pods exist on the target node.

UT

### V. Checklist


TaoYang526:
@saintube @ZiMengSheng PTAL, thanks!

PR #2768, opened 2026-01-08
scheduler: support cluster mode and multiple pre-allocated pods for pre-allocation

TaoYang526:

This PR enhances the Reservation PreAllocation feature with two major capabilities:
1. Cluster Mode for PreAllocation
Introduces a new cluster-wide mode for selecting pre-allocatable pods. Instead of using only the reservation's owner matchers, this mode uses cluster-wide label/annotation selectors to identify and prioritize pre-allocatable pod candidates.
2. Multiple Allocatable Pods Support
Extends the pre-allocation mechanism to support accumulating resources from multiple pods to satisfy a reservation's resource requirements, rather than being limited to a single pod.

To support new user stories in [reservation proposal](https://github.com/TaoYang526/koordinator/blob/reservation_preallocation_improve/docs/proposals/scheduling/20220609-resource-reservation.md): 

* Story 6
As a cluster administrator, I want to use **Cluster Mode** to select pre-allocatable pods for reservation. In a multi-tenant cluster, pre-allocatable pods may belong to different owners. Using Reservation's Owner matchers is not flexible enough because pre-allocatable pods are often managed centrally and should be decoupled from reservation ownership. Additionally, clusters typically run various workload types, where batch jobs with predictable running time are ideal candidates for pre-allocation since they can complete earlier and release their reserved resources. With Cluster Mode, the scheduler identifies pre-allocatable pods using cluster-wide label selectors (e.g., `scheduling.koordinator.sh/is-pre-allocatable=true`) and prioritizes them via priority annotations, enabling centralized management of pre-allocatable pods across the entire cluster.

* Story 7
As an application administrator, I want to pre-allocate resources from **multiple running pods** to satisfy a reservation. This is useful when no single pod can provide all the required resources due to resource fragmentation. With Multiple Pre-Allocated Pods enabled, the scheduler can accumulate resources from several pre-allocatable pods until the reservation requirements are met.

UT

#### API Changes
- Added `PreAllocationPolicy` struct to `ReservationSpec` with:
  - `Mode`: Supports `Default` (use owner matchers) and `Cluster` (use cluster-wide selectors)
  - `EnableMultiple`: Allows multiple pods to be pre-allocated for a single reservation

- Added new constants in `extension/constants.go`:
  - `LabelPodPreAllocatable`: Label key (`pod.koordinator.sh/is-pre-allocatable`) to identify pre-allocatable pods
  - `AnnotationPodPreAllocatableScore`: Annotation key (`pod.koordinator.sh/pre-allocatable-score`) for scoring pods

#### Scheduler Configuration
- Added `PreAllocationConfig` to `ReservationArgs`:
  - `EnableClusterMode`: Enable cluster-wide pre-allocation mode
  - `ClusterLabelKey`: Customizable label key for identifying candidates
  - `ClusterScoreAnnotationKey`: Customizable annotation key for scoring
  - `PreferNoPreAllocatedPods`: Prefer placing reservations without using pre-allocatable pods when possible

#### Implementation Changes
- **Cache Layer** (`cache.go`):
  - Added btree-based cache for pre-allocatable candidates per node
  - Supports efficient insertion, deletion, and sorted retrieval by score

- **Pod Event Handler** (`pod_eventhandler.go`):
  - Added handlers to maintain pre-allocatable candidates cache on pod add/update/delete events
  - Incremental updates for score and node changes

- **Nominator** (`nominator.go`):
  - Updated to support nominating multiple pre-allocatable pods
  - Added `NominateMultiplePreAllocation` method

#### Plugin Scheduling Phases (`plugin.go`, `scoring.go`)

**Filter Phase:**
- Added `filterWithMultiplePreAllocatablePods` to accumulate resources from multiple pods until reservation requirements are satisfied
- Added `filterWithPreAllocatablePod` (refactored from original logic) for single-pod pre-allocation
- Added fast-path: when `PreferNoPreAllocatedPods` is enabled and pre-allocation is not required, skip pre-allocatable pod selection if node has sufficient resources

**Score Phase:**
- Simplified `PreScore`: removed separate `preScoreForPreAllocation` path, unified with normal pod flow
- Updated `scoreForPreAllocation` scoring logic for multiple pods:
  - Nodes without pre-allocatable pods get highest score (`MaxNodeScore`)
  - Score decreases as more pre-allocatable pods are selected: `MaxNodeScore / (1 + selectedPodsCount)`
- Added `sortPreAllocatablePodsForDefaultMode` to prioritize pods by score before selection

**Reserve Phase:**
- Updated to handle multiple nominated pods (`nominatedPods` slice instead of single `nominatedPod`)
- Call `NominatePreAllocations` for multiple pod nomination
- Updated `assumePods` to assume multiple pods in cache

**Unreserve Phase:**
- Added `unreservePod` helper to handle individual pod unreservation
- Updated to unreserve multiple nominated pods on failure

**PreBind Phase:**
- Updated `PreBindReservation` to handle multiple pre-allocated pods in status update

### V. Checklist


apis/extension/constants.go
```diff
@@ -51,6 +51,14 @@ const (
...
 	// LabelPodSkipEnhancedValidation is the pod label key used to opt out a pod from enhanced validation.
 	LabelPodSkipEnhancedValidation = PodDomainPrefix + "/skip-enhanced-validation"
+
+	// LabelPodPreAllocatable is the label key used to identify pre-allocatable pods in cluster mode.
+	// When set to "true", the pod can be selected as a pre-allocatable candidate.
+	LabelPodPreAllocatable = PodDomainPrefix + "/is-pre-allocatable"
+
+	// AnnotationPodPreAllocatableScore is the annotation key used to score pre-allocatable pods in cluster mode.
+	// The value should be a numeric string. Higher scores indicate higher priority for pre-allocation.
+	AnnotationPodPreAllocatableScore = PodDomainPrefix + "/pre-allocatable-score"
```
saintube: Perhaps pre-allocatable-priority?

saintube:
@TaoYang526 Please add some use cases about the cluster mode pre-allocation. We can update the [reservation proposal](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220609-resource-reservation.md) if needed.

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -540,7 +543,24 @@ func (pl *Plugin) prepareMatchReservationStateForReservePod(ctx context.Context,
...
 }
 
-func listPreAllocatableCandidates(podLister listercorev1.PodLister, rInfo *frameworkext.ReservationInfo) (map[string][]*corev1.Pod, error) {
+func listPreAllocatableCandidates(podLister listercorev1.PodLister, cache *reservationCache, rInfo *frameworkext.ReservationInfo, args *config.ReservationArgs) (map[string][]*corev1.Pod, error) {
+	preAllocationMode := reservationutil.GetPreAllocationMode(rInfo.Reservation)
+	// Cluster mode: retrieve from cache which already has sorted candidates
+	if preAllocationMode == schedulingv1alpha1.PreAllocationModeCluster {
+		// Cluster mode: retrieve from cache which already has sorted candidates
+		// Check if cluster mode is enabled in plugin args
+		if args == nil || args.PreAllocationConfig == nil || !args.PreAllocationConfig.EnableClusterMode {
```
saintube: Recommend setting this during the plugin initialization

TaoYang526:
> @TaoYang526 Please add some use cases about the cluster mode pre-allocation. We can update the [reservation proposal](https://github.com/koordinator-sh/koordinator/blob/main/docs/proposals/scheduling/20220609-resource-reservation.md) if needed.

@saintube Thanks for the review!
Updates in last commit:
(1) Rename label key from xxx/pre-allocatable-score to xxx/pre-allocatable-priority.
(2) Add enablePreAllocationClusterMode field in Plugin to simplify the validation.
(3) Update the reservation proposal: [diff](https://github.com/koordinator-sh/koordinator/pull/2768/files#diff-9b8ab3c1065460c16fdb74eca4c387949600a9673b8cca4f9cfcc97568bddb42), including new stories, API changes and new use cases.
(4) Add more UT to test pre-allocation reservations with different configuration(mode and enableMultiple).

pkg/scheduler/plugins/reservation/transformer.go
TaoYang526: Add enablePreAllocationClusterMode field in Plugin to simplify the validation.

apis/extension/constants.go
TaoYang526: Make sense. Renamed label key from xxx/pre-allocatable-score to xxx/pre-allocatable-priority.

docs/proposals/scheduling/20220609-resource-reservation.md
```diff
@@ -328,6 +371,63 @@ If a node has multiple Reservations with Aligned or Restricted policies, and the
 
 The default policy (currently is none) cannot coexist with the policies Aligned and Restricted.
 
+#### Pre-Allocation Policy
```
saintube, line 374: /cc @ZiMengSheng PTAL

pkg/scheduler/plugins/reservation/cache.go
```diff
@@ -452,3 +508,125 @@ func (cache *reservationCache) ListAvailableReservationInfosOnNode(nodeName stri
 	}
 	return result
 }
+
+func (cache *reservationCache) listAllNodes() []string {
```
saintube: why not use `ListAllNodes`

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -486,6 +501,33 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
...
+	checkNodeUnallocatedOnceFn := func() []string {
+		if !checkNodeUnallocatedDone {
+			// use nodeInfo.Requested if nodeRState.podRequested is nil (when there is no pre-allocatable pods).
+			podRequested := nodeRState.podRequested
+			if podRequested == nil {
+				podRequested = nodeInfo.Requested.Clone()
+			}
+			// Check if the reserve pod can be placed with node-unallocated resource when pre-allocation is not required.
+			// For reserve pod, matchedOrIgnored should be 0, both rAllocated and rRemained should be nil.
+			insufficientResourcesByNodeUnallocated = fitsNode(state.podRequestsResources, nodeInfo.Allocatable,
```
saintube, line 520: As discussed, we prefer to make the pattern of the error message the same as the fitsWithReservation when using the default mode.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -1262,3 +1528,21 @@ func (pl *Plugin) DeleteReservation(r *schedulingv1alpha1.Reservation) *framewor
...
+	}
+	return false
+}
+
+// isMultiplePAPodsEnabled checks if multiple pre-allocated pods are enabled for the reservation.
+func isMultiplePAPodsEnabled(rInfo *frameworkext.ReservationInfo) bool {
+	if rInfo == nil {
+		return false
+	}
+	return reservationutil.IsMultiplePAPodsEnabled(rInfo.Reservation)
```
saintube: recommended to wrap as a method of the ReservationInfo

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -527,6 +569,7 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
 		}
 
 		if len(insufficientResourcesByNode) <= 0 && len(insufficientResourceReasonsByReservation) <= 0 {
+			nodeRState.selectedPreAllocatablePods = preAllocatablePods
```
saintube: Does it need a lock?

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -625,10 +639,8 @@ func checkReservationMatchedOrIgnored(pod *corev1.Pod, rInfo *frameworkext.Reser
 }
 
 func checkPreAllocatableMatched(rInfo *frameworkext.ReservationInfo, candidatePod *corev1.Pod, diagnosisState *nodeDiagnosisState, node *corev1.Node) (bool, error) {
-	// check if candidate pod matches the reservation
-	if !rInfo.MatchOwners(candidatePod) {
-		return false, nil
-	}
+	// NOTE: No need to check OwnerMatchers here because pods are already filtered by
+	// listPreAllocatableCandidates based on the PreAllocation mode (Default or Cluster)
```
saintube: We still need to verify it since the listPreAllocatableCandidates does not guarantee all the matchers apart from the selector matches.

pkg/scheduler/plugins/reservation/plugin.go
TaoYang526: I think there's no change on the pattern of the error message when using the default mode: 
it will be called in 
```
		} else if len(preAllocatablePods) > 0 {
			// If no insufficient resources by node with pre-allocatable pods, use the node-unallocated reasons,
			// avoid passing in Filter phase but later fail in Reserve phase.
			// When no pre-allocatable pods exist, NodeResourcesFit performs the node-unallocated check.
			failureReasons = buildNodeFailureReasons(checkNodeUnallocatedOnceFn())
		}
```
same as the original logic:
```
		} else {
			// Check if the reserve pod can be placed with node-unallocated resource when pre-allocation is not required.
			// For reserve pod, matchedOrIgnored should be 0, both rAllocated and rRemained should be nil.
			insufficientResourcesByNodeUnallocated := fitsNode(state.podRequestsResources, nodeInfo.Allocatable,
				nodeRState.podRequested, nil, nil, 0, len(nodeInfo.Pods), preemptible)
			// If no insufficient resources by node with pre-allocatable pods, use the node-unallocated reasons.
			failureReasons = buildNodeFailureReasons(insufficientResourcesByNodeUnallocated)
		}
```
New added logic is that when we configured PreAllocationConfig#PreferNoPreAllocatedPods to be true, the node unallocated resources will be checked at first, and the check result can be reused later in bellow logic:
```
		} else if checkNodeUnallocatedDone {
			// If node-unallocated resources have already been checked, reuse the reasons directly.
			failureReasons = buildNodeFailureReasons(insufficientResourcesByNodeUnallocated)
		}
```

pkg/scheduler/plugins/reservation/cache.go
TaoYang526: listAllNodes() is unused now, will be removed in next commit.

pkg/scheduler/plugins/reservation/plugin.go
TaoYang526: Good suggestion!

pkg/scheduler/plugins/reservation/plugin.go
saintube: OK. I missed the arg checkNodeUnallocatedDone.

pkg/scheduler/plugins/reservation/plugin.go
TaoYang526: The state of a specified node could only be processed once in the Filter phase, right? Please correct me if I lost something. Thanks.

pkg/scheduler/plugins/reservation/transformer.go
TaoYang526: Yes, I missed the ObjectRef and ReservationControllerReference validations in ReservationOwnerMatcher#Match. To make it right when multiple pre-allocated pods enabled, we need to add these validations in listPreAllocatableCandidates, right?

TaoYang526:
Addressed comments and rebased from main. 
@saintube PTAL. Thanks!

pkg/scheduler/plugins/reservation/cache.go
```diff
@@ -18,34 +18,73 @@ package reservation
...
+	if p.priority != other.priority {
+		return p.priority > other.priority
+	}
+	// If priorities are equal, order by UID for stability
+	return p.pod.UID < other.pod.UID
+}
+
+// preAllocatablePodCache manages sorted pre-allocatable pods for a node using btree
+type preAllocatablePodCache struct {
+	tree  *btree.BTree                         // Sorted storage by priority
```
saintube, line 57: Could you please add some benchmark tests for this optimization?

pkg/scheduler/plugins/reservation/transformer.go
saintube: I think the checkPreAllocatableMatched and listPreAllocatableCandidates are coupled for the pre-allocatable pods selection.

pkg/scheduler/plugins/reservation/plugin.go
saintube: Yes, it is safe currently. Perhaps we need some comments on the definitions to avoid unsafe misuse.

TaoYang526: Done. Added comments on nodeReservationState definition to clarify thread-safety constraints and prevent unsafe misuse.
```
// nodeReservationState stores the reservation state for a specific node during the scheduling cycle.
// NOTE: This struct is NOT thread-safe. It is designed to be accessed sequentially within a single
// scheduling cycle. In the Filter phase, each node's state is processed exactly once, so concurrent
// access to the same nodeReservationState instance should not occur. Do not access or modify the same
// nodeReservationState from multiple goroutines.
type nodeReservationState struct {
   ...
}
```

pkg/scheduler/plugins/reservation/transformer.go
TaoYang526: Yes, these two functions are coupled for pre-allocatable pod selection. I've added the comment for checkPreAllocatableMatched function to clarify the relationship:
```
// NOTE: This function is coupled with listPreAllocatableCandidates for pre-allocatable pod selection:
// - Default mode: OwnerMatchers (ObjectRef, Controller, Labels) are validated in listPreAllocatableCandidates
//   via ownerMatcher.Match(), so no need to re-check here.
// - Cluster mode: Pods are retrieved from a global cache without reservation-specific OwnerMatchers validation.
//   The cache contains all pods matching the pre-allocatable label, and the reservation's owners field
//   is not used for filtering in cluster mode by design.
```

pkg/scheduler/plugins/reservation/cache.go
TaoYang526: Done. Added a multi-node benchmark test BenchmarkPreAllocatablePodCache that covers all operations of preAllocatablePodCache.

Benchmark Results:
```
BenchmarkPreAllocatablePodCache
BenchmarkPreAllocatablePodCache/Add
BenchmarkPreAllocatablePodCache/Add/nodes_10_podsPerNode_100
BenchmarkPreAllocatablePodCache/Add/nodes_10_podsPerNode_100-12         	    3823	    303271 ns/op
BenchmarkPreAllocatablePodCache/Add/nodes_100_podsPerNode_100
BenchmarkPreAllocatablePodCache/Add/nodes_100_podsPerNode_100-12        	     345	   3388741 ns/op
BenchmarkPreAllocatablePodCache/Add/nodes_1000_podsPerNode_100
BenchmarkPreAllocatablePodCache/Add/nodes_1000_podsPerNode_100-12       	      27	  39754080 ns/op
BenchmarkPreAllocatablePodCache/Delete
BenchmarkPreAllocatablePodCache/Delete/nodes_10_podsPerNode_100
BenchmarkPreAllocatablePodCache/Delete/nodes_10_podsPerNode_100-12      	    5223	    222793 ns/op
BenchmarkPreAllocatablePodCache/Delete/nodes_100_podsPerNode_100
BenchmarkPreAllocatablePodCache/Delete/nodes_100_podsPerNode_100-12     	     548	   2240961 ns/op
BenchmarkPreAllocatablePodCache/Delete/nodes_1000_podsPerNode_100
BenchmarkPreAllocatablePodCache/Delete/nodes_1000_podsPerNode_100-12    	      51	  22494264 ns/op
BenchmarkPreAllocatablePodCache/UpdatePriority
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_10_podsPerNode_100
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_10_podsPerNode_100-12         	    2060	    630488 ns/op
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_100_podsPerNode_100
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_100_podsPerNode_100-12        	     187	   6815537 ns/op
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_1000_podsPerNode_100
BenchmarkPreAllocatablePodCache/UpdatePriority/nodes_1000_podsPerNode_100-12       	      18	  69665998 ns/op
BenchmarkPreAllocatablePodCache/GetAll
BenchmarkPreAllocatablePodCache/GetAll/nodes_10_podsPerNode_100
BenchmarkPreAllocatablePodCache/GetAll/nodes_10_podsPerNode_100-12                 	  145009	      8475 ns/op
BenchmarkPreAllocatablePodCache/GetAll/nodes_100_podsPerNode_100
BenchmarkPreAllocatablePodCache/GetAll/nodes_100_podsPerNode_100-12                	   13887	     82404 ns/op
BenchmarkPreAllocatablePodCache/GetAll/nodes_1000_podsPerNode_100
BenchmarkPreAllocatablePodCache/GetAll/nodes_1000_podsPerNode_100-12               	    1306	    943600 ns/op
```

The btree-based cache shows good performance - GetAll remains fast (0.94ms for 100K pods) since it only iterates over node-level entries, while write operations(Add/Delete/UpdatePriority) scale linearly with pod count (39.75ms, 22.49ms, 69.67ms for 100K pods).

pkg/scheduler/plugins/reservation/transformer.go
```diff
@@ -624,12 +641,13 @@ func checkReservationMatchedOrIgnored(pod *corev1.Pod, rInfo *frameworkext.Reser
...
 
+// NOTE: This function is coupled with listPreAllocatableCandidates for pre-allocatable pod selection:
+//   - Default mode: OwnerMatchers (ObjectRef, Controller, Labels) are validated in listPreAllocatableCandidates
+//     via ownerMatcher.Match(), so no need to re-check here.
+//   - Cluster mode: Pods are retrieved from a global cache without reservation-specific OwnerMatchers validation.
+//     The cache contains all pods matching the pre-allocatable label, and the reservation's owners field
+//     is not used for filtering in cluster mode by design.
 func checkPreAllocatableMatched(rInfo *frameworkext.ReservationInfo, candidatePod *corev1.Pod, diagnosisState *nodeDiagnosisState, node *corev1.Node) (bool, error) {
-	// check if candidate pod matches the reservation
-	if !rInfo.MatchOwners(candidatePod) {
```
saintube, line 629: This match can be heavy, so we invoked it inside the Parallelizer. We do not need to change this behavior.

pkg/scheduler/plugins/reservation/cache.go
```diff
@@ -270,6 +309,23 @@ func (cache *reservationCache) assumePod(reservationUID types.UID, pod *corev1.P
...
+	}
+	if rInfo.IsTerminating() {
+		return fmt.Errorf("target reservation is terminating")
+	}
+	for _, pod := range pods {
+		rInfo.AddAssignedPod(pod)
+	}
+	return nil
+}
+
```
saintube, line 335: Why not a forgetPods

TaoYang526: As we discussed, assumed pods are already removed when calling reservationCache#forgetReservation, so forgetPod only applies to non-preallocation scenarios. I'd like to rename it to forgetPods and updating the second parameter to accept allocatedPods instead of pod (reserved pod) for better clarity.

pkg/scheduler/plugins/reservation/transformer.go
TaoYang526: Make sense. We can keep this match for the default mode.

TaoYang526:
New Updates:
* keep MatchOwners check in checkPreAllocatableMatched for default mode.
* rename forgetPod to forgetPods and updating the second parameter to accept allocatedPods.
* make other cache operations more explicit and easier to maintain: add assumePods/addPods/deletePods as basic operations, keep assumePod/addPod/deletePod for testing.

pkg/scheduler/plugins/reservation/plugin.go
saintube: One last thing is that the Filter plugin can be invoked multiple times in one scheduling cycle.
e.g. evaluateNominatedNodeName, preemption, and maybe for some special uses
So please ensure this assignment does not pollute the calculation.

saintube:
Apart from https://github.com/koordinator-sh/koordinator/pull/2768#discussion_r2768855225, I will leave the approval to @ZiMengSheng for a double check.

ZiMengSheng (approved):

pkg/scheduler/plugins/reservation/plugin.go
TaoYang526: Sorry for the late response. I'll carefully review all the related invocations.

TaoYang526: @saintube Thanks for the review! This is a valid concern. I have updated the PR to address this concern: changed `filterWithPreAllocatablePods` to return selected pods instead of modifying shared state. Now `nodeRState.selectedPreAllocatablePods` is only updated in the Filter phase.

Another update is to add support for multiple pre-allocatable pods in `Plugin#ReservationNominate`, I missed the new process logic introduced by #2757 before.

**Note:** Since `deviceshare` and `nodenumaresource` plugins currently only support single pre-allocatable pod, `Plugin#NominatePreAllocations` (called when multiple pre-allocatable pods is enabled) does not call `RunNominateReservationFilterPlugins`. I plan to add this support in a follow-up PR. Does this make sense?

saintube: @TaoYang526 I mean, the preemption evaluator or any other plugin can call the `framework.FunFilterWithNominatedPods` in a scheduling cycle. So we should pay attention to the correctness and the idempotence if we want to update the cycle state in a Filter plugin.

TaoYang526: @saintube Thank you for the clarification!
The check condition before updating selectedPreAllocatablePods has been removed, to make sure selectedPreAllocatablePods is always kept up-to-date with the current nodeInfo state on each Filter invocation.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -527,7 +572,8 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
 		}
 
 		if len(insufficientResourcesByNode) <= 0 && len(insufficientResourceReasonsByReservation) <= 0 {
-			return nil
+			selectedPreAllocatablePods = preAllocatablePods
```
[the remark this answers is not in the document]
TaoYang526: The current implementation is intentional. In single-pod mode, passing the full preAllocatablePods list is required like before.
1. Two-phase filtering with different concerns: The Filter phase (filterWithPreAllocatablePod) performs resource feasibility checks via fitsNodeAndReservation, while the Nominate phase (NominatePreAllocation → RunNominateReservationFilterPlugins) runs additional plugin checks via FilterNominateReservation (e.g., device share, NUMA resource allocation).
2. Nominate phase validates each candidate: FilterNominateReservation is called for each pre-allocatable pod during nomination, performing device/NUMA-specific checks that the Filter phase doesn't cover. The best-fitting pod is selected after scoring.

pkg/scheduler/plugins/reservation/pod_eventhandler.go
```diff
@@ -147,4 +151,77 @@ func (h *podEventHandler) deletePod(pod *corev1.Pod) {
...
+	}
+	return priority
+}
+
+// updatePreAllocatableCandidatesCache updates the cached pre-allocatable candidates when pod changes
+// Uses incremental updates with btree for efficiency
+func (h *podEventHandler) updatePreAllocatableCandidatesCache(oldPod, newPod *corev1.Pod) {
+	if newPod == nil || newPod.Spec.NodeName == "" {
+		return
+	}
```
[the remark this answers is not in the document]
TaoYang526: In normal Kubernetes operations, spec.nodeName is immutable after a Pod is bound, so the transition from non-empty to empty nodeName should not happen. However, we handle this edge case defensively to ensure cache consistency.

pkg/scheduler/plugins/reservation/scoring.go
```diff
@@ -41,17 +41,9 @@ const (
 
 func (pl *Plugin) PreScore(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod, nodes []*corev1.Node) *framework.Status {
 	if reservationutil.IsReservePod(pod) {
-		if reservationutil.IsReservePodPreAllocation(pod) {
-			return pl.preScoreForPreAllocation(ctx, cycleState, pod, nodes)
-		}
-
 		return framework.NewStatus(framework.Skip)
```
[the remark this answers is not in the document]
TaoYang526: This appears to be a merge conflict issue where the recent changes were reverted. The changes will be restored.

pkg/scheduler/plugins/reservation/cache.go
```diff
@@ -452,3 +527,112 @@ func (cache *reservationCache) ListAvailableReservationInfosOnNode(nodeName stri
...
+	}
+	// Add or update the pod
+	priority := cache.getPreAllocatablePriorityFromPod(pod)
+	item := &preAllocatablePodItem{
+		pod:      pod,
+		priority: priority,
+	}
+	podCache.tree.ReplaceOrInsert(item)
+	podCache.index[pod.UID] = item
+}
```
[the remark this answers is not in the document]
TaoYang526: The current design explicitly separates the two responsibilities:
* addPreAllocatableCandidateOnNode is only called when a pod first becomes a candidate (Case 1: non-candidate → candidate), so the UID is guaranteed to not exist in the cache at that point. ReplaceOrInsert is therefore safe here — if a duplicate Add event fires with the same priority, the btree treats (priority, UID) as the same key and replaces in place.
* Priority changes are always routed through updatePreAllocatableCandidatePriority (Case 3 in updatePreAllocatableCandidatesCache), which already does delete-then-insert to keep the tree and index consistent.

So the scenario where the same UID is added with a different priority via addPreAllocatableCandidateOnNode cannot occur under the current call structure. The two methods have a clear, intentional split: add for first-time insertion, update for priority changes.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -527,7 +573,8 @@ func (pl *Plugin) filterWithPreAllocatablePods(ctx context.Context, cycleState *
...
```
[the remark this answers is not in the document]
TaoYang526: Same as above: https://github.com/koordinator-sh/koordinator/pull/2768#discussion_r2938010552

saintube:
@TaoYang526 Please resolve the conflicts.

TaoYang526:
> @TaoYang526 Please resolve the conflicts.

@saintube Done. Thanks for reminding that.

saintube:
@TaoYang526 Please resolve all comments to merge :)

TaoYang526:
Thanks @saintube @ZiMengSheng for the review and commit !

PR #2773, opened 2026-01-12
scheduler: fix representative pod stuck in unschedulablePods

ZiMengSheng:
### V. Checklist


pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -212,6 +212,12 @@ func (pgMgr *PodGroupManager) PreEnqueue(ctx context.Context, pod *corev1.Pod) (
 	}
 	err = pgMgr.basicGangRequirementsCheck(gang, pod)
 	if err != nil {
+		// use IsPodRepresentative to avoid write lock collisions.
+		if gang.IsPodRepresentative(pod) {
```
saintube, line 216: Recommended to add a UT for this case if it is reproducible

PR #2775, opened 2026-01-13
koordlet: fix bug kubelet restart failed when enable cpu qos feature

### LINKED ISSUE #2732: [BUG] kubelet restart failed when  enable cpu qos feature
**What happened**:

While testing the CPU QoS feature using this community document
（https://koordinator.sh/docs/v1.6/user-manuals/cpu-qos#use-cpu-qos-based-on-core-scheduling）

, we encountered startup failures with the kubelet component.

kubelet log 
```
kubelet[2879019]: E1211 20:21:29.889413 2879019 kubelet.go:1389] "Failed to start ContainerManager" err="failed to initialize top level QOS containers: failed to update top level BestEffort QOS cgroup : failed to write \"2\": write /sys/fs/cgroup/cpu,cpuacct/kubepods.slice/kubepods-besteffort.slice/cpu.shares: invalid argument"
Dec 11 20:21:29 nodeXXX  systemd[1]: kubelet.service: Main process exited, code=exited, status=1/FAILURE
```
i also can't update /sys/fs/cgroup/cpu,cpuacct/kubepods.slice/kubepods-besteffort.slice/cpu.shares file 
```
echo 2 > /sys/fs/cgroup/cpuacct/kubepods.slice/kubepods-besteffort.slice/cpu.shares
-bash: echo: write error: Invalid argument
```
When I update the value of file /sys/fs/cgroup/cpuacct/kubepods.slice/kubepods-besteffort.slice/cpu.idle to 0 , kubelet successfuly started 

please check if this is related to the kernel code 
https://github.com/torvalds/linux/blob/v6.18/kernel/sched/fair.c#L13526

**Environment**:
- k8s 1.21
- koordinator 1.6.0
- os ubuntu 24.04
- kernel 6.6.88-2048

hahahaheihei:

This pr fix the bug ，where kubelet restart failed due to enable cpu qos feature .
Use pod level directory cpu.idle rather qosclass root directory cpu.idle 
eg:
use 
```
/sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-podxxxxx.slice/cpu.idle
```
not use
```
/sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.idle
```

fix #2732 

- disable SetKubeQOSCPUIdle
- enable SetPodQOSCPUIdle

### V. Checklist

- [✔] I have written necessary docs and comments
- [✔] I have added necessary unit tests and integration tests
- [✔] All checks passed in `make test`

### DIFF
--- a/pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
+++ b/pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
@@ -24,6 +24,7 @@ import (
 	gocache "github.com/patrickmn/go-cache"
 	"go.uber.org/atomic"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 	"k8s.io/utils/ptr"
 
 	"github.com/koordinator-sh/koordinator/apis/extension"
@@ -116,9 +117,9 @@ func (p *Plugin) Register(op hooks.Options) {
 	reconciler.RegisterCgroupReconciler(reconciler.SandboxLevel, sysutil.VirtualCoreSchedCookie,
 		"set core sched cookie to process groups of sandbox container specified",
 		p.SetContainerCookie, reconciler.PodQOSFilter(), podQOSConditions...)
-	// TODO: support host application
-	reconciler.RegisterCgroupReconciler(reconciler.KubeQOSLevel, sysutil.CPUIdle, "reconcile QoS level cpu idle",
-		p.SetKubeQOSCPUIdle, reconciler.NoneFilter())
+	reconciler.RegisterCgroupReconciler(reconciler.PodLevel, sysutil.CPUIdle,
+		"set pod level cpu.idle ", p.SetPodQOSCPUIdle, reconciler.NoneFilter())
+	p.Setup(op)
 	p.Setup(op)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
```diff
@@ -24,6 +24,7 @@ import (
 	gocache "github.com/patrickmn/go-cache"
 	"go.uber.org/atomic"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
```
saintube, line 27: please use `k8s.io/utils/ptr` instead

pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
```diff
@@ -116,9 +117,9 @@ func (p *Plugin) Register(op hooks.Options) {
 	reconciler.RegisterCgroupReconciler(reconciler.SandboxLevel, sysutil.VirtualCoreSchedCookie,
 		"set core sched cookie to process groups of sandbox container specified",
 		p.SetContainerCookie, reconciler.PodQOSFilter(), podQOSConditions...)
-	// TODO: support host application
-	reconciler.RegisterCgroupReconciler(reconciler.KubeQOSLevel, sysutil.CPUIdle, "reconcile QoS level cpu idle",
-		p.SetKubeQOSCPUIdle, reconciler.NoneFilter())
+	reconciler.RegisterCgroupReconciler(reconciler.PodLevel, sysutil.CPUIdle,
+		"set pod level cpu.idle ", p.SetPodQOSCPUIdle, reconciler.NoneFilter())
+	p.Setup(op)
```
saintube, line 122: why duplicate Setup

pkg/koordlet/runtimehooks/hooks/coresched/rule.go
```diff
@@ -213,19 +222,6 @@ func (p *Plugin) ruleUpdateCb(target *statesinformer.CallbackTarget) error {
 }
 
 func (p *Plugin) refreshForAllPods(podMetas []*statesinformer.PodMeta) error {
-	for _, kubeQOS := range []corev1.PodQOSClass{
```
saintube, line 216: I prefer to make the QoS-level reconciliation optional for the future migration. And currently, we can disable it by default.
IIUC, the pod-level cpu.idle will disable its cpu.shares, which breaks the fairness between besteffort pods. So it is not a optimal solution.

hahahaheihei: thanks for your reply , i test this code again the pod-level cpu.idle will disable koordinator batch resource. maybe we need a new solution.

saintube: OK. Temporarily disabling the pod-level cpu.shares is a workaround if it is acceptable in your use case. You can still enable the container-level cpu.shares to retain the fairness. As it is not a final solution for the cpu.idle, it is up to you whether to implement it in this PR.

PR #2776, opened 2026-01-13
scheduler: nodenumaresource and deviceshare support unassigned event

saintube:

In multi-scheduler scenarios, we allow pods to transition from the assigned state to an unassigned state for retry purposes. In this case, the event handlers of plugin caches should adapt the transition to release resources.

### V. Checklist


ZiMengSheng (approved):
/lgtm
/approve

PR #2787, opened 2026-01-22
scheduler: support prefer node hint

saintube:

Make the koord-scheduler support the scheduling hint of the preferred nodes.

Introduce the `PreferNodesPlugin` in the FrameworkExtender, allowing the plugin to provide a node list that the scheduling pipeline first tries to filter before entering the regular `findNodesThatPassFilters`. It introduces the overhead of once RunFilterPluginsWithNominatedPlugins, which should be acceptable.

In contrast to the `.status.nominatedNodeName`, the hint uses a list of nodes, and the hinted pods do not account for the nodes before assuming.

**Compatibilty**:

When the user also enables the network-topology-aware scheduling and generates the PreFilterResult by the FindOneNode plugin according to the best plan, the preference hint is ignored.

The SchedulingHint plugin also considers the intersection of the original PreFilterResult. When the preferred nodes are not included in the PreFilterResult, the scheduler does not break the constraint of the existing PreFilter, but aborts the hint instead.

In the scenario of job queueing, the queue component may have some suggestions about which nodes the dequeued pods should try to schedule first. The patch adds the api for the queue to hint the scheduler for the preferred node when it does not want the pod to be counted on the nominated node before the scheduler predicates. We hint at a node list instead of a single node, which serves as a backup in case the most preferred one fails for some reason.

1. Set the scheduling hint annotations for the Pod:

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    internal.scheduling.koordinator.sh/scheduling-hint: '{"preferredNodeNames": ["most-prefer-node-name", "second-prefer-node-name"]}'
```

2. Then check the scheduler's log if the PreferOneNodePlugin runs and returns a result.

### V. Checklist


ZiMengSheng:
A very useful building block. However, I'd like to understand the queue component's requirements for us:
1. When preemption is needed, should it also prioritize preemption on certain nodes?
2. When calculating network topology, should it also prioritize trying on certain nodes first?

apis/extension/multi_scheduler.go
```diff
@@ -35,7 +35,12 @@ func GetSchedulerName(pod *corev1.Pod) string {
 }
 
 type SchedulingHint struct {
```
hormes, line 37: In my view, schedulinghint should be an internal scheduling protocol, rather than something exposed to clients for use.

saintube:
> A very useful building block. However, I'd like to understand the queue component's requirements for us:
> 
> 1. When preemption is needed, should it also prioritize preemption on certain nodes?
> 2. When calculating network topology, should it also prioritize trying on certain nodes first?

1. It is a good point. Like the nominatedNodeName in the vanilla kube-scheduler, the preference to filter is not the same as the preference for preempt. However, I believe it can be helpful if we give a chance to try some candidate nodes first when the pods are dequeued for an intended preemption. The problem is that there is no unified interface for the DryRunPreemption, so it requires us to implement this ability inside preemption plugins.
2. For the network-topology-aware scheduling, I suppose the current offering is enough. If the queue wants to nominate job pods with nominated nodes according to the topology constraints, it is OK to use this api or the nominatedNodeName. If the calculation of the NTW is to be done by the scheduler, we do not need to set any preferred node.

saintube:
PTAL /cc @ZiMengSheng @hormes

apis/extension/coscheduling.go
```diff
@@ -114,8 +114,16 @@ func GetGangWaitTime(pod *corev1.Pod) (time.Duration, error) {
...
 	// AnnotationBindGangGroupId is the annotation key for the pod's gang group id to bind.
-	AnnotationBindGangGroupId = SchedulingDomainPrefix + "/bind-gang-group-id"
+	AnnotationBindGangGroupId = InternalSchedulingDomainPrefix + "/bind-gang-group-id"
+	// DEPRECATED: This api is marked as internal and will be removed next version.
+	// Please use the domain `internal.scheduling.koordinator.sh/` instead.
+	// DeprecatedAnnotationBindGangMemberCount is the annotation key for the count of the pod's gang group member pods to bind.
+	DeprecatedAnnotationBindGangMemberCount = SchedulingDomainPrefix + "/bind-gang-member-count"
 	// AnnotationBindGangMemberCount is the annotation key for the count of the pod's gang group member pods to bind.
-	AnnotationBindGangMemberCount = SchedulingDomainPrefix + "/bind-gang-member-count"
+	AnnotationBindGangMemberCount = InternalSchedulingDomainPrefix + "/bind-gang-member-count"
```
ZiMengSheng, line 128: There shouldn't be any break change, right?

pkg/scheduler/plugins/schedulinghint/scheduling_hint.go
```diff
@@ -29,13 +29,20 @@ import (
...
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/hinter"
+	"github.com/koordinator-sh/koordinator/pkg/util"
 )
 
-const Name = "SchedulingHint"
+const (
+	Name = "SchedulingHint"
+
+	maxHintNodes = 100 // max count of hinted preferred nodes
```
ZiMengSheng: 搞成插件配置之类的吧

apis/extension/coscheduling.go
saintube: We also patch DeprecatedAnnotationBindGangMemberCount in the current version.

pkg/scheduler/plugins/schedulinghint/scheduling_hint.go
saintube: done

PR #2800, opened 2026-02-04
koordlet: add cpuset sharepool cpu info to metrics.

tan90github:
This PR adds the `koordlet_cpuset_share_pool_info` metric, which exports the specific CPU IDs currently assigned to the Share Pool. The same applies to the BE Share Pool.

For CPU pinning scenarios (e.g., LSE/LSR pods), we should monitor the Share Pool's usage. If too many exclusive cpus are allocated, non-pinned pods may be crowded into a small Share Pool, causing severe CPU contention and node hotspots.

With the `koordlet_cpuset_share_pool_info` metric, we can easily monitor the CPU utilization of the sharepool through the following PromQL metric, and prevent hot spot issues in advance.

```
avg(
  irate(node_cpu_seconds_total{instance=~"kind-worker1", mode=~"user"}[5m])
  and on(node, cpu)
  koordlet_cpuset_share_pool_info
)
```

no

Go to prometheus, and query follow promql
- koordlet_cpuset_share_pool_info{}
- koordlet_cpuset_be_share_pool_info{}
### V. Checklist

### DIFF
--- a/pkg/features/koordlet_features.go
+++ b/pkg/features/koordlet_features.go
@@ -182,13 +182,13 @@ const (
 	// owner: @tan90github
 	// alpha: v1.8
 	//
-	// CPUSetMetricEnhance enables recording per-cpu info metrics of share pool and BE share pool.
+	// PerCPUMetric enables recording per-cpu info metrics of share pool and BE share pool.
 	// NOTE: Enabling this feature will generate high-cardinality time series at node × share pool cpu numbers level:
 	// - koordlet_cpuset_share_pool_info{node, cpu}
 	// - koordlet_cpuset_be_share_pool_info{node, cpu}
 	// which may significantly increase the pressure on Prometheus/TSDB/remote-write.
 	// It is recommended to enable it only during small clusters or troubleshooting/capacity assessment.
-	CPUSetMetricEnhance featuregate.Feature = "CPUSetMetricEnhance"
+	PerCPUMetric featuregate.Feature = "PerCPUMetric"
 )
 
 func init() {
@@ -224,7 +224,7 @@ var (
 		HugePageReport:         {Default: false, PreRelease: featuregate.Alpha},
 		PodResourcesProxy:      {Default: false, PreRelease: featuregate.Alpha},
 		HamiCoreVGPUMonitor:    {Default: false, PreRelease: featuregate.Alpha},
-		CPUSetMetricEnhance:    {Default: false, PreRelease: featuregate.Alpha},
+		PerCPUMetric:           {Default: false, PreRelease: featuregate.Alpha},
 	}
 )
 
--- a/pkg/features/koordlet_features.go
+++ b/pkg/features/koordlet_features.go
@@ -178,6 +178,17 @@ const (
 	//
 	// HamiCoreVGPUMonitor enables the vGPU monitoring for HAMi-core.
 	HamiCoreVGPUMonitor featuregate.Feature = "HamiCoreVGPUMonitor"
+
+	// owner: @tan90github
+	// alpha: v1.8
+	//
+	// CPUSetMetricEnhance enables recording per-cpu info metrics of share pool and BE share pool.
+	// NOTE: Enabling this feature will generate high-cardinality time series at node × share pool cpu numbers level:
+	// - koordlet_cpuset_share_pool_info{node, cpu}
+	// - koordlet_cpuset_be_share_pool_info{node, cpu}
+	// which may significantly increase the pressure on Prometheus/TSDB/remote-write.
+	// It is recommended to enable it only during small clusters or troubleshooting/capacity assessment.
+	CPUSetMetricEnhance featuregate.Feature = "CPUSetMetricEnhance"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


pkg/koordlet/metrics/cpu_cpuset.go
```diff
@@ -31,9 +35,23 @@ var (
 		Help:      "Number of be share pool cores",
 	}, []string{NodeKey})
 
+	CPUSetSharePoolInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: KoordletSubsystem,
+		Name:      "cpuset_share_pool_info",
+		Help:      "CPUSet share pool info, means this cpu id is in cpuset share pool",
+	}, []string{NodeKey, CPUIDKey})
```
saintube, line 42: This might produce too many metrics. How about recording the cpuset string as a label key, and the cpus count as the value?

tan90github: > This might produce too many metrics. How about recording the cpuset string as a label key, and the cpus count as the value?

1. You're right, this will indeed generate more metrics, but I think this is as expected, just like `node_cpu_seconds_total` in node_exporter.
2. I believe that in most scenarios, the cpu sharePool id information is more effective than a string because it can be aggregated with other metrics, such as: `node_cpu_seconds_total`.

saintube: The per-cpu metric is heavy in a large cluster. So I recommend introducing a feature-gate or cmdline arg to enable it only when needed.

tan90github: Good suggestion, it has been revised.

pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
```diff
@@ -192,6 +192,8 @@ func (p *cpusetPlugin) parseRule(nodeTopoIf interface{}) (bool, error) {
 		}
 	}
 
+	// record sharePool, beSharePool cpu numbers and cpu ID info.
+	recordPerCPUInfo := features.DefaultKoordletFeatureGate.Enabled(features.CPUSetMetricEnhance)
```
saintube: It can be set once the plugin is initialized.

pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
```diff
@@ -200,6 +202,12 @@ func (p *cpusetPlugin) parseRule(nodeTopoIf interface{}) (bool, error) {
 			continue
 		}
 		shareCPUSetCount += nodeSharePoolCPUSet.Size()
+
+		if recordPerCPUInfo {
+			for _, e := range nodeSharePoolCPUSet.ToSlice() {
+				metrics.RecordCPUSetSharePoolInfo(e)
```
saintube, line 211: We should also refresh the metrics about the former share pool.
e.g. 
t0: sharepool = {2, 3, 4, 5}, metrics = {{"cpu"="2", "cpu"="3", "cpu"="4", "cpu"="5"}}
t1: sharepool = {4, 5}, metrics = {{"cpu"="4", "cpu"="5"}}; the metrics for cpu 2, 3 should be cleaned.

pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
tan90github: fixed

pkg/koordlet/runtimehooks/hooks/cpuset/rule.go
tan90github: fixed

tan90github: I tested it. I created two 4-core CPU-bound Pods respectively. The sharePool info metrics were deleted as expected, and they are consistent with the actual bound CPU core IDs.

image
image

pkg/features/koordlet_features.go
```diff
@@ -178,6 +178,17 @@ const (
...
+	// owner: @tan90github
+	// alpha: v1.8
+	//
+	// CPUSetMetricEnhance enables recording per-cpu info metrics of share pool and BE share pool.
+	// NOTE: Enabling this feature will generate high-cardinality time series at node × share pool cpu numbers level:
+	// - koordlet_cpuset_share_pool_info{node, cpu}
+	// - koordlet_cpuset_be_share_pool_info{node, cpu}
+	// which may significantly increase the pressure on Prometheus/TSDB/remote-write.
+	// It is recommended to enable it only during small clusters or troubleshooting/capacity assessment.
+	CPUSetMetricEnhance featuregate.Feature = "CPUSetMetricEnhance"
```
saintube: how about `PerCPUMetric`

tan90github: ok, done.

tan90github:
@saintube PTAL, I've addressed the Copilot review suggestions. Thanks for taking a look when you get a chance.

tan90github:
@saintube Are there any improvements needed for this PR?

saintube:
@tan90github Sorry for the late response.
/lgtm

Issue #2326, opened 2025-01-30
[BUG] CPU Burst not working


abriko:
Hi koordinator team,

**What happened**:

I tried to reproduce the CPU burst feature based on the [document](https://koordinator.sh/docs/user-manuals/cpu-burst/), but I didn't see any difference between enabling and disabling it.

I did a quick investigation and found the following logs in koordlet:

```
...
I0130 08:41:39.268583   68132 cpu_burst.go:612] update container default/apache-demo/apache cpu burst failed, dir kubepods.slice/kubepods-podf993afb8_4ac9_4b63_b262_a9644fb56022.slice/cri-containerd-6b0e2e6a55b754b59455855793e2e91c33bee3f91d98c88b07b9db945d1bc7b3.scope, updated false, err write /host-cgroup/kubepods.slice/kubepods-podf993afb8_4ac9_4b63_b262_a9644fb56022.slice/cri-containerd-6b0e2e6a55b754b59455855793e2e91c33bee3f91d98c88b07b9db945d1bc7b3.scope/cpu.max.burst: invalid argument                                                
I0130 08:41:39.268745   68132 cpu_burst.go:635] update pod default/apache-demo cpu burst failed, dir kubepods.slice/kubepods-podf993afb8_4ac9_4b63_b262_a9644fb56022.slice, updated false, err write /host-cgroup/kubepods.slice/kubepods-podf993afb8_4ac9_4b63_b262_a9644fb56022.slice/cpu.max.burst: invalid argument                                       
...
```

full logs: [logs.zip](https://github.com/user-attachments/files/18599689/logs.zip)

**What you expected to happen**:

After CPU Burst is enabled, the latency reduced.  (behavior is similar to what's described in the document.)

**How to reproduce it (as minimally and precisely as possible)**:

1. Create GKE cluster with standard mode, 1 node pool and 3 nodes (1 per zone), machine type e2-standard-4.
2. Install koordinator via helm.
3. Follow [document](https://koordinator.sh/docs/user-manuals/cpu-burst/) enable CPU Burst for koordlet.
4. Follow [document](https://koordinator.sh/docs/user-manuals/cpu-burst/) deploy apache-demo pod.
5. Trigger http test. 

**Anything else we need to know?**:

**Environment**:
- App version: 
1.5.0

- Kubernetes version (use `kubectl version`): 
Client Version: v1.30.8-dispatcher
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
[...]

PR #2801, opened 2026-02-04
fix(koordlet):Resolve CPUBurst triggering cfsScaleDown on CgroupV2 node

AutuSnow:
Resolve CPUBurst triggering cfsScaleDown on CgroupV2 node

fixes:
https://github.com/koordinator-sh/koordinator/issues/2778
https://github.com/koordinator-sh/koordinator/issues/2326

### V. Checklist


pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
```diff
@@ -544,23 +610,43 @@ func (b *cpuBurst) applyContainerCFSQuota(podMeta *statesinformer.PodMeta, conta
...
-	for _, update := range sortOfUpdateQuota {
-		if err := update(); err != nil {
-			return err
+	if deltaContainerCFS < 0 {
+		// Scale-down: burst first, then quota
+		// Order: container burst -> container quota -> pod burst -> pod quota
+		sortOfUpdate := []func() error{
+			func() error { return updateContainerBurst(targetContainerCFS) },
+			updateContainerCFSQuota,
+			updatePodBurst,
```
saintube: Here, the applyCFSQuotaBurst sets the cfs_burst_us again after the applyCPUBurst. Naturally, they should be handled in a single function.

pkg/util/sloconfig/nodeslo_config.go
```diff
@@ -417,7 +417,7 @@ func DefaultCPUBurstStrategy() *slov1alpha1.CPUBurstStrategy {
 func DefaultCPUBurstConfig() slov1alpha1.CPUBurstConfig {
 	return slov1alpha1.CPUBurstConfig{
 		Policy:                     slov1alpha1.CPUBurstNone,
-		CPUBurstPercent:            ptr.To[int64](1000),
+		CPUBurstPercent:            ptr.To[int64](100),
```
saintube, line 420: please also update the slo-controller-config [example](https://github.com/koordinator-sh/koordinator/blob/main/apis/configuration/slo_controller_config.go#L397)

AutuSnow:
@saintube The modification of this code may be a bit significant, increasing the cost of review. Let me briefly explain:
1. The applyCPUBurst method has been removed because it was originally set separately for cpu.cfs_burst_us,Based on the discussion at the weekly meeting, I have removed this function
2. The applyCFSQuota Burst has merged the burst logic, and now both quota and burst are processed in one loop
3. Added applyContainerCPUBurst auxiliary method, extracting the logic of container level burst settings
4. Add the calcSafeCPUBurstVal function to ensure that the burst value does not exceed quota and meets the cgroup v2 constraint
5. Write order control is to reduce burst first and then quota when scaling down; When scaling up, first increase quota and then increase burst

saintube (commented):
/lgtm

PR #2805, opened 2026-02-05
scheduler: support configure defaultGangMatchPolicy

ZiMengSheng:
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/gang.go
+++ b/pkg/scheduler/plugins/coscheduling/core/gang.go
@@ -140,13 +140,13 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
 
 	matchPolicy := extension.GetGangMatchPolicy(pod)
 	if matchPolicy == "" {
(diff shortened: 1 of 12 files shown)


pkg/scheduler/plugins/coscheduling/core/gang.go
```diff
@@ -139,14 +139,11 @@ func (gang *Gang) tryInitByPodConfig(pod *v1.Pod, args *config.CoschedulingArgs)
 	gang.Mode = mode
 
 	matchPolicy := extension.GetGangMatchPolicy(pod)
-	if matchPolicy == "" {
-		matchPolicy = extension.GangMatchPolicyOnceSatisfied
-	}
 	if matchPolicy != extension.GangMatchPolicyOnlyWaiting && matchPolicy != extension.GangMatchPolicyWaitingAndRunning &&
 		matchPolicy != extension.GangMatchPolicyOnceSatisfied {
 		klog.Errorf("pod's annotation AnnotationGangMatchPolicy illegal, gangName: %v, value: %v",
```
saintube, line 147: If we consider this situation as the default, shall we change the log verbosity

PR #2806, opened 2026-02-05
koordlet: Fix invalid argument error in cfs_quota_us and cfs_burst_us validation

CFH2436:
1. **Apply CPU Burst Limit**: Added a logic to cap `cfs_burst_us` based on `cfs_quota_us` in `applyCPUBurst` to avoid `invalid argument` errors during updates.
2. **Synchronous Quota Adjustment**: Ensured `cfs_burst_us` is adjusted alongside `cfs_quota_us` in `applyCFSQuota`.
3. **Refactoring**: Extracted common logic into `applyContainerCPUBurst` and `applyPodCPUBurst` for better maintainability.
4. **CgroupReader Enhancement**: Added support for reading current `cfs_burst_us` values.

https://github.com/koordinator-sh/koordinator/issues/2778

After the fix, cfs_burst_us is now more closely coupled with cfs_quota_us and will be adjusted synchronously with changes to the quota. While this resolves the invalid argument error, I would like to bring up a concern: does this coupling deviate from the original design intent of CPU Burst at the Linux kernel level? I would appreciate the community's insight and evaluation on this approach.
### V. Checklist

### DIFF
--- a/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
+++ b/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
@@ -245,8 +245,8 @@ func (b *cpuBurst) start() {
 		klog.V(5).Infof("get pod %v/%v cpu burst config: %v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
 		// set cpu.cfs_burst_us for pod and containers
 		b.applyCPUBurst(cpuBurstCfg, podMeta)
-		// scale cpu.cfs_quota_us for pod and containers
-		b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)
+		// scale cpu.cfs_quota_us and cpu.cfs_burst_us for pod and containers
+		b.applyCFSQuotaWithBurst(cpuBurstCfg, podMeta, nodeState)
 	}
 	b.Recycle()
 }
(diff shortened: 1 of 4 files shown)


pkg/koordlet/resourceexecutor/reader.go
```diff
@@ -470,6 +479,16 @@ func (r *CgroupV2Reader) ReadNetClsId(parentDir string) (uint32, error) {
 	return readCgroupAndParseUint32(parentDir, resource)
 }
 
+func (r *CgroupV2Reader) ReadCPUBurst(parentDir string) (int64, error) {
+	resource, ok := sysutil.DefaultRegistry.Get(sysutil.CgroupVersionV2, sysutil.CPUBurstName)
+	if !ok {
+		return -1, ErrResourceNotRegistered
+	}
+
+	// since cpu.max.burst is a simgle value in cgroupV2 file, just convert the value to int64 and return
```
saintube, line 488: typo:

```suggestion
	// since cpu.max.burst is a single value in cgroupV2 file, just convert the value to int64 and return
```

And is there any difference between v1 and v2 to highlight?

pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
```diff
@@ -245,8 +245,8 @@ func (b *cpuBurst) start() {
 		klog.V(5).Infof("get pod %v/%v cpu burst config: %v", podMeta.Pod.Namespace, podMeta.Pod.Name, cpuBurstCfg)
 		// set cpu.cfs_burst_us for pod and containers
 		b.applyCPUBurst(cpuBurstCfg, podMeta)
-		// scale cpu.cfs_quota_us for pod and containers
-		b.applyCFSQuotaBurst(cpuBurstCfg, podMeta, nodeState)
+		// scale cpu.cfs_quota_us and cpu.cfs_burst_us for pod and containers
+		b.applyCFSQuotaWithBurst(cpuBurstCfg, podMeta, nodeState)
```
saintube, line 249: If we decide to scale both the cfs_quota_us and the cfs_burst_us, we should handle them in a single function for readability.

CFH2436: The initial intention of retaining `applyCFSQuotaBurst` was to account for scenarios where `policy=cpuBurstOnly`. For `genOperationByContainer` method checks `cfsQuotaBurstEnabled`, and it will return `cfsReset` in this case.

For readability, this part of the logic can be refactored.

pkg/koordlet/resourceexecutor/reader.go
CFH2436: In this context, there is no difference between v1 and v2.

CFH2436:
Since #2801 is already on the top of this fix, I'll close this PR to keep our focus in one place.
Thanks for your time.

PR #2817, opened 2026-02-27
koord-descheduler: use raw-allocatable for threshold calculation in descheduler lowNodeLoad plugin.

tan90github:

Problem:

In the cpu-normalization or resource amplification scenario, the node's Status.Allocatable is amplified (eg: 16 cores → 32 cores). However, the descheduler lowNodeLoad plugin calculates thresholds using this amplified allocatable:
`threshold = allocatable * thresholdPercent  (eg: 32 * 70% = 22.4)`
This causes the threshold to be much higher than expected. Even when actual node usage reaches 100% (16 cores), it still cannot trigger the hotspot threshold (22.4 cores), making the descheduler ineffective.

Fix:

Use the raw-allocatable stored in node annotation (node.koordinator.sh/raw-allocatable) for threshold calculations instead of the amplified Status.Allocatable. This ensures the usage percentage is calculated correctly based on the actual physical resources.

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/loadaware/low_node_load_test.go
+++ b/pkg/descheduler/framework/plugins/loadaware/low_node_load_test.go
@@ -35,6 +35,7 @@ import (
 	coretesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/events"
 
+	"github.com/koordinator-sh/koordinator/apis/extension"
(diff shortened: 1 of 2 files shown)


tan90github:
It seems that the test failed due to the e2e itself？

image

tan90github:
@songtao98 @ZiMengSheng PTAL, thanks!

songtao98 (approved):
/lgtm

PR #2823, opened 2026-03-04
fix(koordlet): add GPU device info to OCI spec in NRI injection to prevent permission loss

### LINKED ISSUE #2809: [question] Perhaps there is some flaw when using NRI to inject into the GPU
**What happened:**

When using NRI to inject into the GPU, there might be a problem of losing device cgroup permissions. 
In our production, we use the NRI injection similar to `koordinator` for implementation. We set the `NVIDIA_VISIBLE_DEVICES` environment variable and then let the `nvidia-container-runtime` handle it.

However, a very strange thing occurred. In the device cgroup of the injected container, the device numbers of these GPU cards would suddenly disappear from the `devices.list`. After investigation, we found that it was triggered by `systemd reload` command. Once the node executed `systemctl daemon-reload`, the device cgroup would be reset, and then the GPU device permissions would be lost directly.

Further research revealed that the corresponding container's permission configuration was not saved in `/run/systemd/transient/`. The specific location is `/run/systemd/transient/*/50-DeviceAllow.conf`. This should be written by [runc](https://github.com/opencontainers/runc/blob/f047c6b0f88f3299e71d6be7508dfdaa824f2117/vendor/github.com/opencontainers/cgroups/systemd/common.go#L138) after reading the OCI spec. In the implementation of the `device-plugin`, [kubelet](https://github.com/kubernetes/kubernetes/blob/5adfc48e19d5fbc4af5b0d31aeb9f0c13c01cf5d/pkg/kubelet/kuberuntime/kuberuntime_container.go#L440) writes the OCI spec so that runc can write to `/run/systemd/transient`. This logic is not implemented in the NRI mechanism.

This issue only occurs when the `nvidia-container-runtime` is using the `legacy` mode; the `cdi` mode can be used to avoid it.

I'm not sure if koordlet has a similar issue. It's not very convenient for me to simulate an environment. Judging from the code, it seems that there is no logic to supplement the OCI spec. It's very likely that it can be reproduced.

Here, it might be necessary to consider calling `adjustment.AddDevice` in the NRI injection logic to replace the kubelet's device-plugin mechanism to complete the device information in the OCI spec. Otherwise, `systemctl daemon-reload` will cause all GPU permissions on the entire node to be lost.

**Environment:**

Could someone reproduce this scenario conveniently? You could check whether the GPU device permissions of the pod have been persisted in systemd by viewing `/run/systemd/transient/`. The prerequisite is to set the `nvidia-container-runtime` to the legacy mode, that is, to use the method of `NVIDIA Container Runtime Hook`.

AutuSnow:

/kind bug

When using NRI to inject GPU access into containers, the GPU hook only sets the NVIDIA_VISIBLE_DEVICES environment variable. It does not add GPU device information (/dev/nvidia*) to the OCI spec via AddContainerDevices.
This causes a problem: runc never writes GPU device cgroup rules to /run/systemd/transient/*/50-DeviceAllow.conf. When systemctl daemon-reload is triggered on the node, systemd resets the cgroup configuration, and all GPU device permissions for NRI-injected containers are lost.
﻿This fix adds GPU device entries (per-GPU devices + control devices like nvidiactl, nvidia-uvm, nvidia-uvm-tools) to AddContainerDevices, following the same pattern as the existing RDMA device injection. The devices are collected atomically to avoid partial injection on failure.
﻿
Additionally, the original monolithic InjectContainerGPUEnv function is refactored into three focused helpers for better readability:
injectGPUEnv 
injectGPUDevices
injectHAMiResources

Fixes https://github.com/koordinator-sh/koordinator/issues/2809

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/hooks/gpu/gpu.go
+++ b/pkg/koordlet/runtimehooks/hooks/gpu/gpu.go
@@ -33,7 +33,21 @@ import (
 	rmconfig "github.com/koordinator-sh/koordinator/pkg/runtimeproxy/config"
 )
 
-const GpuAllocEnv = "NVIDIA_VISIBLE_DEVICES"
+const (
+	GpuAllocEnv = "NVIDIA_VISIBLE_DEVICES"
+
+	nvidiaDevicePrefix   = "/dev/nvidia"
+	nvidiaCtlDevice      = "/dev/nvidiactl"
+	nvidiaUVMDevice      = "/dev/nvidia-uvm"
+	nvidiaUVMToolsDevice = "/dev/nvidia-uvm-tools"
+)
+
+// nvidiaControlDevices are common NVIDIA control devices shared by all GPU containers.
+var nvidiaControlDevices = []string{nvidiaCtlDevice, nvidiaUVMDevice, nvidiaUVMToolsDevice}
+
+// getDeviceNumbers is a function variable for getting device major/minor numbers,
+// allowing substitution in tests.
+var getDeviceNumbers = system.GetDeviceNumbers
 
 type gpuPlugin struct{}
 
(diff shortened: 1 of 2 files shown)


AutuSnow:
I noticed that this code was written by @ZiMengSheng because I split the original large function to increase the readability of the code. Therefore, I still hope you can take a look at this issue first

ZiMengSheng:
A test report is needed here.

Issue #2837, opened 2026-04-01
[proposal]Lambda-G Scoring Function for Resource Imbalance Detection and Balanced Rescheduling (#2332)


0x-auth:
**What is your proposal**:

Integrate Lambda-G V3 scoring algorithm as a Score plugin for koord-scheduler to address the node resource imbalance problem described in #2332.

Lambda-G V3 scores pod placement using a hybrid of **variance minimization** (how balanced the node becomes after placement) and **cosine alignment** (does the pod's resource shape match the node's free capacity). This combination beats all existing Kubernetes scoring strategies across 5 heterogeneous cluster scenarios.

Design doc PR: #2839
Open-source auditor: https://github.com/0x-auth/lambda-g-auditor
Scheduler implementation: https://github.com/0x-auth/lambda-g-scheduler

**Why is this needed**:

Current scoring strategies don't consider the *shape* of resource usage across dimensions:

- **LeastAllocated**: Node at 90% CPU / 10% RAM scores ~50. Node at 50% CPU / 50% RAM also scores ~50. The first node has stranded RAM.
- **BalancedAllocation**: Minimizes post-placement variance, but when multiple nodes have similar variance, it picks arbitrarily. It doesn't ask "does this pod's shape *match* what this node needs?"
- **DominantResource**: Only looks at the single most-consumed dimension, missing cross-dimensional interactions.

At scale (50+ nodes), cross-dimensional fragmentation wastes 10-20% of compute budget. In GPU clusters running mixed inference workloads, GPU-Memory can be 95% consumed while GPU-Compute sits at 20% — stranding expensive silicon.

**Score Function (V3 — Hybrid Variance-Alignment)**:

```
score = 0.6 × variance_score
      + 0.2 × alignment_score
      + 0.1 × headroom_score
      − pressure_penalty
      − strand_penalty
```

**Components:**

1. **Variance Score (60%)** — Same signal as BalancedAllocation. How evenly utilized will this node be across all dimensions after placing this pod?
```
after_frac[i] = (used[i] + pod.req[i]) / capacity[i]
variance = mean((after_frac[i] - mean)²)
variance_score = max(0, (1.0 - variance × 4) × 100)
```

2. **Cosine Alignment (20%)** — Does the pod's resource shape point in the same direction as the node's free capacity? A CPU-heavy pod scores higher on a node with lots of free CPU.
```
alignment_score = cosine_similarity(node_free_frac, pod_req_frac) × 100
```

[...]

Issue #2332, opened 2025-02-10
[proposal] 重调度解决单机维度不同类型资源不均衡问题


JBinin:
**What is your proposal**:
集群中的节点，经过一段时间，可能会出现某些节点不同类型资源不均衡的现象，比如某个节点的 CPU 分配率 90%，memory 分配率 50%，导致该节点的 memory 资源成为碎片资源而不能得到充分利用
目前重调度的策略中还没有针对这类情况进行均衡的策略

**Why is this needed**:
希望能够减少由于单机内不同类型资源不均衡导致的碎片，从而无法充分利用集群资源的问题

**Is there a suggested solution, if so, please add it**:
1. 新增一个重调度插件，对集群中节点的不同类型资源不均衡程度超过一定阈值（可配置）的节点进行重调度，配合调度器的 NodeResourcesBalancedAllocation 插件，对新 Pod 进行调度，来使得新的调度结果，机器内不同类型资源更加均衡
2. 可以使用标准差 std 来描述节点不同类型资源不均衡程度，与默认调度器的 NodeResourcesBalancedAllocation 插件一致（https://github.com/kubernetes/kubernetes/blob/b7c55c2ed2f8a5522e53e839c615168550c0966e/pkg/scheduler/framework/plugins/noderesources/balanced_allocation.go#L153C2-L172）
3. 节点上，驱逐之后使得 std 更小的 Pod 具有更高的驱逐优先级
4. 依靠目前koordinator已有的资源预留驱逐模式，来保证驱逐的稳定性

zwzhang0107:
@songtao98 PTAL

songtao98:
![Image](https://github.com/user-attachments/assets/40e218e2-d98e-4eed-8a74-0cb2e9ffdd3f) @JBinin FIY, just an example if I understand it right.

songtao98:
@JBinin Another issue I'm thinking about be like: Suppose if koord-descheduler can evict a pod B from node 1, the ideally scheduling action is to swap it with Pod F on Node 3. This can make the whole cluster balanced. 

But as far as I know, the current **NodeResourcesBalancedAllocation** plugin in scheduler might place Pod B on node 2 because node 2 is more balanced and it has a higher score when choosing nodes. We may consider how to optimize this.

LY-today:
[...]

PR #2839, opened 2026-04-02
proposal: Lambda-G Symmetric Exhaustion Score Plugin for 6D resource …

0x-auth:
Ⅰ. Describe what this PR does
Proposal for a Score plugin (Lambda-G V3) that optimizes pod placement for cross-dimensional resource balance across 6 dimensions (CPU, Memory, GPU-Core, GPU-Memory, IOPS, Network).

V3 uses a hybrid approach: variance minimization (60%) as the stability base + cosine alignment (20%) as a directional tiebreaker + headroom (10%) for schedulability, minus pressure and strand penalties. This combination beats all existing K8s scoring strategies across 5 heterogeneous cluster scenarios.

Addresses the resource imbalance problem in #2332. Detailed proposal in issue #2837.

Ⅱ. Does this pull request fix one issue?
Refs #2837

Ⅲ. Describe how to verify it
Review the proposal document at `docs/proposals/scheduling/20260402-lambda-g-symmetric-exhaustion.md`.

Key benchmark results (Balance Score — higher is better):

| Scenario | BalancedAlloc | DominantRes | **Lambda-G V3** |
|---|---|---|---|
| Mixed GPU — AI Workload | 78.7 | 79.4 | **81.8** |
| GPU — Inference Heavy | 81.3 | 76.6 | **81.9** |
| GPU — Training Heavy | 79.4 | 78.9 | **82.8** |
| CPU + Few GPUs | 72.3 | 68.8 | **74.7** |
| Scale (60n×300p) | 74.1 | 73.8 | **76.7** |

V3 wins 5/5 scenarios. 23% fewer stranded nodes, 53% less waste vs BalancedAllocation.

Benchmark reproduction: https://github.com/0x-auth/lambda-g-auditor

Ⅳ. Special notes for reviews
This is a proposal-only PR — no implementation code. Implementation PR with plugin code and tests will follow after design review.

NVIDIA's KAI-Scheduler has also expressed interest: kai-scheduler/KAI-Scheduler#1373

Ⅴ. Checklist
- [ ] I have added necessary unit tests and integration tests (implementation PR)
- [x] All checks passed in `make test` (proposal only, no code changes)

### DIFF
--- a/docs/proposals/scheduling/20260402-lambda-g-symmetric-exhaustion.md
+++ b/docs/proposals/scheduling/20260402-lambda-g-symmetric-exhaustion.md
@@ -1,0 +1,423 @@
+---
+title: Lambda-G Symmetric Exhaustion Score Plugin
+authors:
+  - "@0x-auth"
+reviewers:
+  - TBD
+creation-date: 2026-04-02
+last-updated: 2026-04-22
+status: provisional
+see-also:
+  - "/docs/proposals/scheduling/20220510-load-aware-scheduling.md"
+---
+
+# Lambda-G Symmetric Exhaustion Score Plugin
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals/Future Work](#non-goalsfuture-work)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Implementation Details](#implementation-detailsnotesconstraints)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Alternatives](#alternatives)
+- [Upgrade Strategy](#upgrade-strategy)
+- [Additional Details](#additional-details)
+  - [Test Plan](#test-plan-optional)
+- [Implementation History](#implementation-history)
+
+## Glossary
+
+- **Symmetric Exhaustion**: A scheduling objective where all resource dimensions on a node drain evenly toward zero. The opposite of "stranded resources."
+- **Stranded Resource**: A resource dimension (e.g., RAM) that is mostly free but unusable because another dimension (e.g., CPU) is fully consumed on the same node.
+- **Resource Vector**: A normalized vector `[CPU, Memory, GPU-Core, GPU-Memory, ...]` representing either free capacity or pod request as fractions in `[0, 1]`.
+
+## Summary
+
+This proposal introduces a Score plugin for koord-scheduler that scores nodes based on how much *more balanced* they become after placing a pod. The plugin combines post-placement variance minimization with cosine alignment between the pod's resource request vector and the node's free capacity vector. Unlike `LeastAllocated` which minimizes total usage, or `BalancedAllocation` which only considers variance, Lambda-G V3 uses directional alignment as a tiebreaker to steer pods toward nodes where their resource shape fills the gap.
+
+The plugin natively handles Koordinator's `koordinator.sh/gpu-core` and `koordinator.sh/gpu-memory` extended resources, addressing GPU-specific imbalance patterns (VRAM full but compute idle, or compute maxed but VRAM unused) that are common in AI inference workloads.
+
+## Motivation
+
+Default Kubernetes `LeastAllocated` scoring treats these two nodes equivalently:
+
+- Node A: 90% CPU used, 10% RAM used -> score ~50
+- Node B: 50% CPU used, 50% RAM used -> score ~50
+
+Node A has 90% of its RAM stranded -- paid for but unusable because no CPU-light/RAM-heavy pod can be scheduled there. At scale (50+ nodes), this wastes 10-20% of compute budget.
+
+The problem is worse with GPUs. In AI inference clusters, VRAM is often fully consumed while GPU compute sits idle (common with large model serving), or GPU compute is maxed while VRAM is unused (common with many small models). Existing scoring plugins do not consider cross-dimensional balance.
+
+Issue [#2332](https://github.com/koordinator-sh/koordinator/issues/2332) describes this resource imbalance problem. LoadAwareScheduling helps with utilization thresholds but doesn't optimize for balance *across* dimensions -- it can tell you a node is 80% used, but not that CPU is at 95% while RAM is at 20%.
+
+### The Problem in Detail
+
+Koordinator currently supports two node-ordering strategies relevant to balance:
+
+- **Bin-packing (MostAllocated):** Prefers nodes with the highest overall utilization. A node at 95% GPU-Memory / 15% CPU looks "mostly full" -- bin-packing will avoid it, stranding CPU capacity.
+- **Spread (LeastAllocated):** Prefers nodes with the most free resources. The same node looks "partially used" -- spread will send more pods to it, but those pods may need GPU-Memory that isn't there.
+
+Neither strategy considers the *shape* of resource usage across dimensions. Both treat a node at 50% CPU / 50% RAM identically to one at 90% CPU / 10% RAM. The second node has stranded RAM -- you're paying for memory nobody can use.
+
+The existing **BalancedAllocation** strategy (minimize per-node variance between dimensions) is better, but it only asks "will this node be more even after placement?" It does not ask "does this pod's resource shape *match* what this node needs?" When multiple nodes have similar variance, BalancedAllocation picks arbitrarily. Lambda-G uses directional alignment to pick the node where the pod actually fills the gap.
+
+### Real-World Impact
+
+In GPU clusters running mixed inference workloads, cross-dimensional fragmentation is common:
+
+| Workload Type | GPU-Mem | GPU-Compute | CPU | RAM | Stranded Resource |
+|---|---|---|---|---|---|
+| LLaMA 70B serving | ~95% | ~20% | ~30% | ~60% | GPU-Compute |
+| Batch small-model inference | ~30% | ~85% | ~70% | ~20% | GPU-Memory, RAM |
+| CPU preprocessing pipeline | ~0% | ~0% | ~90% | ~25% | RAM, all GPU |
+
+### Goals
+
+1. Provide a Score plugin that steers pod placement toward symmetric resource exhaustion across multiple dimensions.
+2. Reduce stranded resources compared to `LeastAllocated` and `BalancedAllocation` in heterogeneous clusters.
+3. Natively handle `koordinator.sh/gpu-core` and `koordinator.sh/gpu-memory` extended resources without additional configuration.
+4. Maintain sub-millisecond scoring latency (pure arithmetic, no API calls).
+
+### Non-Goals/Future Work
+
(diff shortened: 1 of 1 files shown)


ZiMengSheng:
Thanks for the detailed writeup. Before we go further on this proposal I'd like to see an explicit discussion of the functional boundary between this plugin and two already-existing pieces in the stack. Right now the design doc briefly touches on `LoadAwareScheduling` under Non-Goals but doesn't really answer the harder question.

**1. vs. `NodeResourcesBalancedAllocation` (upstream K8s)**

The 60% variance component of V3 is, by the design doc's own admission, "the same signal as Kubernetes BalancedAllocation." So 60% of the scoring weight is duplicating an already-available plugin. The genuinely new pieces are:

- 20% cosine alignment (pod request shape vs node free capacity)
- 10% headroom (a diluted `LeastAllocated`)
- pressure penalty (utilization gate)
- strand penalty (80/20 pair threshold)

Could you explain why these should be packaged as a **new top-level Score plugin** rather than shipped as **arguments/extensions to `NodeResourcesBalancedAllocation`** (or a Koordinator-flavored variant of it)? Concretely, something like:

```yaml
plugins:
  score:
    enabled:
      - name: NodeResourcesBalancedAllocation
        args:
          resources: [cpu, memory, koordinator.sh/gpu-core, koordinator.sh/gpu-memory]
          alignmentWeight: 0.2       # cosine alignment tiebreaker
          headroomWeight: 0.1
          pressureThreshold: 0.85
          strandPair: {high: 0.80, low: 0.20, penalty: 15}
```

seems to cover the same behavior without introducing a brand-new plugin name, a new configuration surface, and a new set of tests. It would also let operators mix in the new signals without giving up the existing variance-only behavior. What's the argument for a separate plugin?

Related: cosine alignment specifically only differs from variance when two nodes are near-tied on variance. Have you considered gating alignment as a **conditional tiebreaker** (only apply when `|variance_A - variance_B| < ε`) rather than an unconditional 20% linear term? That's closer to how it's described in the prose ("directional tiebreaker") than how it's actually computed in the formula.

**2. vs. `LoadAwareScheduling`**

The design doc says LoadAware "caps utilization" and Lambda-G "balances within those caps," and lists them as complementary. But the pressure penalty in V3

```
if used_after > 0.92: penalty += (used_after - 0.92) * 500
elif used_after > 0.85: penalty += (used_after - 0.85) * 50
```

is itself a utilization-threshold mechanism, just expressed as a smooth penalty instead of a hard cap. That's functionally overlapping with LoadAware's utilization thresholds, and when both are enabled it's not obvious what the composite behavior is. A few concrete questions:

- If an operator has `LoadAwareScheduling` configured with `usageThresholds.cpu = 80`, and V3 starts penalizing at 85%, which one wins? Are they additive? Does one shadow the other?
- LoadAware uses **real-time metrics** (NodeMetric CR), whereas V3's pressure penalty uses **static allocatable** (`node.used[i] + pod.req[i]) / node.capacity[i]`). These two definitions of "utilization" don't agree in practice — a node can be 90% allocated but 30% used, or vice versa. The proposal should state which one V3 is targeting and why, and how the two plugins should be composed in a real deployment.
- Would it make sense for V3 to **reuse LoadAware's utilization view** (via the same estimator/NodeMetric path) instead of computing its own from requests? Otherwise clusters using both will see two different "utilization" numbers driving two different scoring signals.

**3. What I'd like to see in the doc**

A new subsection under Implementation Details called something like "Relationship to existing plugins," covering:

- explicit column-by-column comparison table: `LeastAllocated` / `MostAllocated` / `BalancedAllocation` / `LoadAwareScheduling` / Lambda-G, showing which signal each one uses (allocatable vs real usage, per-dimension vs aggregate, variance vs alignment vs threshold), and where they conflict when co-enabled;
- a clear justification for why the new signals cannot be added as args on `BalancedAllocation`, or why doing so is worse than a new plugin;
- a recommended `KoordinatorSchedulerConfiguration` that shows Lambda-G + LoadAware co-existing, with the intended precedence and any weight guidance.

Without this the proposal is hard to evaluate against the existing plugin set, and there's a real risk of shipping three plugins that overlap on 60-70% of their signal. Thanks!

0x-auth:
Thanks @ZiMengSheng, this is exactly the boundary discussion the proposal was missing, and you are right that it is the crux. Let me answer each point directly, concede what should be conceded, and lay out concrete doc + benchmark changes.

## 1. vs `NodeResourcesBalancedAllocation`

You are correct that the 60% variance term is the same signal as `BalancedAllocation`, and I should not have buried that. The honest framing is: the variance term is not the contribution, it is the stable base that the two genuinely new signals ride on. The new signals are the cosine alignment and the per-pair strand penalty, and the question is whether those fit inside `BalancedAllocation`'s arg surface or need their own plugin.

My argument for a separate plugin, stated concretely so it can be argued against:

- `BalancedAllocation` scores a node from a single scalar (post-placement variance across a resource list). The strand penalty is not a function of that scalar, it is a function of the **pairwise** structure of the after-placement vector (`after[i] > 0.80 AND after[j] < 0.20`). Expressing that as args on `BalancedAllocation` means teaching it to iterate resource pairs and carry a penalty table, which changes its scoring model from "variance of a set" to "variance plus pairwise gating." That is a different plugin wearing the old name.
- Cosine alignment needs the pod request vector and the node free vector as first-class inputs. `BalancedAllocation` today only needs the after-placement usage vector. Adding alignment means threading the request shape through as a second vector.

That said, I do not want to hand-wave this. The right way to settle it is to actually build the `args-on-BalancedAllocation` variant you sketched and compare it head to head with the standalone plugin on the same 5 scenarios, on: lines of new code, size of the new config surface, and whether the pairwise strand behavior survives the translation. I will add that comparison to the doc and post the numbers here. If the arg variant covers the behavior without materially growing `BalancedAllocation`'s semantics, then you are right and it should not be a new plugin.

### On cosine alignment as a conditional tiebreaker

This is the best catch in the review. You are right that the prose ("directional tiebreaker") and the formula (unconditional 0.2 linear term) do not describe the same thing, and the formula is what runs. Two ways to reconcile, and I would like your read on which is preferable for upstream:

- **(a) Make the formula match the prose:** gate alignment as `if |variance_A - variance_B| < epsilon then apply alignment`, i.e. a true tiebreaker that is zero unless variance is near-tied. This matches the intent and removes the "alignment over-steers and creates new imbalances" failure mode I mention under weights (which is exactly what an unconditional term risks).
- **(b) Make the prose match the formula:** stop calling it a tiebreaker and defend the unconditional term on its own merits.

I lean (a), because your framing is closer to correct than mine was. I will benchmark the epsilon-gated version against the current linear term on the 5 scenarios and report whether the gated version keeps the stranded-resource reduction. If it does, (a) becomes the design and the 0.2 weight goes away in favor of an epsilon.

## 2. vs `LoadAwareScheduling`

The allocatable-vs-realusage point is the sharpest one and the proposal was wrong to list the two as cleanly complementary without addressing it. The precise distinction:

- Lambda-G's pressure penalty operates on **static allocatable**: `(node.used[i] + pod.req[i]) / node.capacity[i]`, i.e. request-based, known at schedule time, deterministic, no metrics dependency. It targets **allocation-time balance**: do not schedule a pod that would strand a dimension based on what is reserved.
- `LoadAwareScheduling` operates on **real-time NodeMetric**: it targets **runtime utilization**: do not schedule onto a node that is actually hot right now.

You are right that these disagree in practice (90% allocated / 30% used, and the reverse), and that when both are enabled the composite behavior is undefined in the current doc. Answering your three questions:

- **Which wins at the 80/85 boundary:** as written, neither shadows the other, they are additive, and that is a bug not a design. Two independent utilization gates on two different definitions of utilization is exactly the "three overlapping plugins" risk you flagged.
- **Real-time vs static:** they are targeting different quantities on purpose (schedulable headroom vs live load), but the proposal never says so, which is the actual defect.
- **Should V3 reuse LoadAware's utilization view:** I think the pressure penalty should not compute its own utilization when LoadAware is present. The cleaner composition is: `LoadAwareScheduling` owns the runtime-utilization gate (it already has the NodeMetric/estimator path), and Lambda-G owns cross-dimensional balance and stranding on allocatable. Concretely I would make the pressure penalty default-off and only engage when LoadAware is not enabled, so there is a single utilization authority in any given config. That removes the overlap you are pointing at while keeping the strand/alignment signal, which LoadAware does not have.

## 3. What I will add to the doc

Agreed on all three, I will add a "Relationship to existing plugins" subsection under Implementation Details containing:

- the column-by-column table across `LeastAllocated` / `MostAllocated` / `BalancedAllocation` / `LoadAwareScheduling` / Lambda-G, with columns for signal source (allocatable vs real usage), granularity (per-dimension vs aggregate), and scoring model (variance vs alignment vs threshold), and an explicit "conflicts when co-enabled" row;
- the `args-on-BalancedAllocation` prototype comparison with the numbers, as the justification for (or against) a separate plugin;
- a recommended `KoordinatorSchedulerConfiguration` showing Lambda-G + LoadAware co-existing with the pressure penalty deferred to LoadAware and the intended precedence.

I will hold the plugin registration until the doc lands the comparison table and the epsilon-gated benchmark, so we are not shipping overlap. Will follow up here with the updated doc and the benchmark numbers. Thanks again for the depth here.

0x-auth:
@ZiMengSheng ran both comparisons you asked for. Script is committed at `benchmarks/bench_variants.py` in the scheduler repo, same scenarios and seed as the original numbers.

**1. Epsilon gated alignment**

I said I leaned toward gating it as a true tiebreaker. The data says that is wrong, so I am reporting against my own preference.

```
BalancedAllocation (baseline)      77.2 bal    44 stranded   $144,588
V3 unconditional 0.2 alignment     79.6        34            $ 67,975
V3 eps-gated, eps=0.5              77.7        47            $120,284
V3 eps-gated, eps=1.0              77.0        49            $123,335
V3 eps-gated, eps=2.0              77.3        47            $138,709
V3 eps-gated, eps=5.0              77.0        49            $ 96,469
V3 eps-gated, eps=10.0             76.6        44            $108,929
```

Every epsilon loses. Stranded goes from 34 back to 44 to 49, level with plain BalancedAllocation. The unconditional term is doing real work, not over steering.

So the fix is the other direction from what I proposed: keep the formula, correct the prose. It is not a tiebreaker and the doc should stop calling it one.

One implementation note that came out of building this. A true tiebreaker cannot be expressed as a Score plugin at all. The scorer signature is per node, the comparison is between nodes. It needs two pass selection, which is a bigger change than an arg.

**2. args on BalancedAllocation**

```
BalancedAllocation (plain)         77.2 bal    44 stranded   $144,588
+ strand + pressure as args        79.2        44            $ 96,471
full V3 (separate plugin)          79.6        34            $ 67,975
```

You are right on balance. 79.2 against 79.6 is a wash, and that does not justify a new plugin.

Stranding is where they separate: 44 against 34. The alignment term accounts for the entire stranding reduction, which is the thing the plugin exists to do. So the argument for a separate plugin is narrower than my PR claimed, but it is not nothing.

If the stranding reduction is not worth a separate plugin to you, then args on BalancedAllocation is the better outcome and I will rework it that way. Your call, I do not have a strong position now that the numbers are in.

Still owed from my earlier comment: the comparison table and the LoadAware co-existence config. Those are doc work, not blocked on anything.

Issue #2851, opened 2026-04-14
[Umbrella] Align Koordinator scheduling with Kubernetes 1.35 capabilities


saintube:
**What is your proposal**:

Following the dependency upgrade to k8s 1.35.2 in #2822, several advanced scheduling features (such as GangScheduling, OpportunisticBatching, and Workload-Aware Preemption) are currently disabled by default. This umbrella issue tracks the effort to fully restore and enhance compatibility with the latest upstream Kubernetes scheduling semantics and features.

To ensure the Koordinator leverages the latest upstream improvements to maximize scheduling efficiency and workload awareness, the following upstream features need to be adopted, integrated, or validated within Koordinator:

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

allexistence:
A possible way to address this is to first ensure API-level compatibility with DRA GA (`resource.k8s.io/v1`) and avoid conflicts by disabling/bypassing DeviceShare when DRA is enabled, then incrementally add node-side support (koordlet + koord-device-daemon) for `ResourceClaim` handling.

PR #2863, opened 2026-04-14
scheduler: implement SignPlugin across core plugins (#2854)

### LINKED ISSUE #2854: Opportunistic Batching
Implement SignPlugin Interface for Opportunistic Batching

Kubernetes 1.35 introduces Opportunistic Batching to improve scheduling throughput by batching pods with the same scheduling constraints. Plugins need to implement the SignPlugin interface to declare their "signature" so the scheduler can group pods correctly.

What needs to be done:
1. [ ] Implement SignPlugin for Reservation plugin — covering reservation affinity, owner references, resource requirements
2. [ ] Implement SignPlugin for NodeNUMAResource plugin — covering NUMA topology policy, CPU bind policy, resource requests
3. [ ] Implement SignPlugin for DeviceShare plugin — covering device resource requests (GPU/RDMA/FPGA type and amount)
4. [ ] Implement SignPlugin for other koordinator plugins (e.g., ElasticQuota, LoadAware, Coscheduling) if they affect scheduling decisions
5. [ ] Add unit tests for all Sign() implementations to verify correct grouping behavior
6. [ ] Verify that unimplemented plugins do not silently disable global batching

thc1006:

Kubernetes 1.35 builds a per-pod signature from every Filter and Score plugin's `SignPod` and reuses scheduling results across pods whose signatures match (KEP-5598). This adds `SignPod` to six Koordinator plugins.

Signatures are still disabled in the default deployment, so nothing here takes effect there yet. `SchedulingHint` leads the `preFilter` list in `config/manager/scheduler-config.yaml` and has no `SignPlugin`, and one such plugin sets `enableSignatures = false` for the whole profile.

It cannot simply be given a signer. Its `PreFilter` narrows nodes from `hinter.GetSchedulingHintState`, and that state has two writers: the plugin's own `BeforePreFilter`, from a pod annotation, and `pkg/scheduler/batch/engine.go`, per node and per cycle. `SignPod` is handed only the pod, so it can reach the first and not the second. Two pods with equal signatures would not get equal PreFilter results while it is active, so the framework declining to batch is correct. What that leaves is a direction question, raised on this PR: whether upstream batching replaces `pkg/scheduler/batch`, coexists with it, or is something it migrates toward.

Each fragment mirrors an input that plugin's own Filter or Score reads. Inputs an in-tree signer already covers are left alone, since the signature is the union of every plugin's fragments: aggregate requests go to `noderesources/fit`, node affinity and nodeSelector to `nodeaffinity`.

| Plugin | Fragments |
| --- | --- |
| `Reservation` | reserve-pod target and its pre-allocation annotation; reservation-affinity and exact-match (strict parse); reservation-ignored; pre-allocation-required; canonicalized owner-matching inputs |
| `DeviceShare` | canonicalized device-type/resource/amount triples; `DeviceAllocated`, and for pods with device requests `DeviceAllocateHint` / `DeviceJointAllocate`; `GPUPartitionSpec` for a non-zero GPU request; `GPUIsolationProvider` for a shared GPU request |
| `NodeNUMAResource` | `numa-topology-spec` and `resource-spec`; and for non-zero-request pods, reservation-affinity presence, pre-allocation-required, and the effective `AllowUseCPUSet` |
| `ElasticQuota` | resolved `quotaName\|treeID` via the same resolver PreFilter uses |
| `LoadAware` | priority class; DaemonSet ownership; aggregate limits; custom scaling factors |
| `Coscheduling` | gang id; network-topology-aware pods are refused instead of signed |

Coscheduling refuses because `sortNodesByTopology` ranks nodes by where the rest of the gang already sits, and `PodGroupManager.PreFilter` pins each later gang pod to its own planned node, which `findNodesThatFitPod` applies only after trying the batch hint.

Part of #2854. See the scope note below for what is still missing before it can close.

```
go test -race ./pkg/scheduler/plugins/{reservation,deviceshare,nodenumaresource,elasticquota,loadaware,coscheduling} ./pkg/scheduler/frameworkext -count=1
```

`signature_realplugins_test.go` builds frameworks from the production factories, registered before construction because that is when `computeBatchablePlugins` collects them, and asserts on the framework's own `SignPod`. One case reproduces the shipped `preFilter` list and asserts the signature is empty; the rest keep `Reservation` out, since it signs pod UID and name and would separate every pair regardless.

Each fragment this PR adds was deleted in turn to confirm the matching case fails without it.

Four plugins have no signer. Only the first ships enabled.

| Plugin | Extension points |
| --- | --- |
| `SchedulingHint` | PreFilter |
| `ScheduleAdmission` | PreFilter |
| `NodeResourcesFitPlus` | PreScore, Score |
| `ScarceResourceAvoidance` | PreScore, Score |

They are left out rather than given an empty `SignPod`, which would assert they have no pod input worth distinguishing. Establishing that is per-plugin work I have not done here.

`DeviceShare`, `NodeNUMAResource` and `Reservation` also branch on that hint state in their own `PreFilter` without signing it, for the same reason `SchedulingHint` cannot be signed. Their doc comments say so. It is masked while signatures are off.

A signature being non-empty proves nothing on its own, since `frameworkImpl.SignPod` seeds the map with the scheduler name. Batch reuse is a further step again, needing consecutive cycles and the previous node no longer fitting, and is not covered here.

### Ⅴ. Checklist


thc1006:
Copilot review follow-up (after commit 13c246ea):

**Addressed**

- r3081722972 (ElasticQuota correctness): now uses `getPodAssociateQuotaNameAndTreeID` so the signature tracks the resolved quota identity, not raw labels (commit 44deae44).
- r3081723020 (ElasticQuota tree-only): fragment is only emitted when the resolved quota name is non-empty (commit 44deae44).
- r3081723038, r3081723061 (JSON canonicalisation in NodeNUMAResource and Reservation): added `canonicalJSON` so semantically equal JSON annotations produce the same fragment value regardless of whitespace or key order. Covered by new subtests (commit 44deae44).
- r3083410754 (Coscheduling subtest name): renamed "pod without gang annotations" to "pod without PodGroup label" to match the actual input (commit 13c246ea).

**Respectfully declined (same note applies to r3083410711, r3083410741, r3083410768, r3083410775, r3083410788, r3083410797)**

These comments claim `status.IsSuccess()` will panic when `SignPod` returns `(*fwktype.Status)(nil)`. It will not. `k8s.io/kube-scheduler/framework` defines `Status` methods that are nil-safe by design:

```go
// interface.go:129-134
func (s *Status) Code() Code {
    if s == nil {
        return Success
    }
    return s.code
}

// interface.go:174-177
// IsSuccess returns true if and only if "Status" is nil or Code is "Success".
func (s *Status) IsSuccess() bool {
    return s.Code() == Success
}
```

The contract is explicit: a nil `*Status` **is** a Success status. Every `SignPod` test in this PR runs green locally and in CI with this assertion. Rewriting it to `status == nil || status.IsSuccess()` would duplicate a contract the framework already guarantees and, more importantly, mislead future readers into thinking the framework's nil handling is unreliable. Keeping the assertions as-is.

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -182,6 +184,218 @@ func (pl *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWit
...
+//   - aggregate pod resource requests — matched against reservation capacity;
+//   - required node affinity / nodeSelector — matched against the reservation's
+//     target node;
+//   - pod affinity / topologySpreadConstraints and required pod-affinity terms —
+//     they flip isPodAllNodesPreRestoreRequired and switch the preRestore path.
+func (pl *Plugin) SignPod(_ context.Context, pod *corev1.Pod) ([]fwktype.SignFragment, *fwktype.Status) {
+	if reservationutil.IsReservePod(pod) {
+		return []fwktype.SignFragment{{
+			Key:   "koord.Reservation.reservePodFor",
+			Value: reservationutil.GetReservationNameFromReservePod(pod),
```
saintube, line 233: Why pick the reservation name as the sign?

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -182,6 +184,218 @@ func (pl *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWit
...
+		fragments = append(fragments, fwktype.SignFragment{
+			Key:   "koord.Reservation.requests",
+			Value: canonicalResourceList(requests),
+		})
+	}
+	// Required node affinity / nodeSelector constrain which reservations (or
+	// which reservation's target node) a pod can land on. Include the raw
+	// NodeAffinity and NodeSelector so two pods that would filter onto
+	// different nodes cannot reuse a cached scheduling decision.
+	if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil &&
```
saintube: Does this duplicate the sign of the NodeAffinity plugin?

thc1006: Good catch, thank you — yes, it was a duplicate. The overall KEP-5598 pod signature is the union of all in-profile plugins' fragments, so inputs already signed by an upstream in-tree plugin do not need to be re-signed here.

Dropped four redundant fragments in 8850276a:

- `koord.Reservation.requests` — upstream `noderesources/fit.SignPod` signs pod resource requests via `computePodResourceRequest`.
- `koord.Reservation.requiredNodeAffinity` and `koord.Reservation.nodeSelector` — upstream `nodeaffinity.SignPod` signs both.
- `koord.Reservation.allNodesPreRestoreRequired` — unreachable in practice: upstream `interpodaffinity.SignPod` and `podtopologyspread.SignPod` return `Unschedulable` for pods carrying PodAffinity/PodAntiAffinity or TopologySpreadConstraints, which already disables batching for those pods.

SignPod's docstring now documents the boundary with upstream signing.

pkg/scheduler/plugins/reservation/plugin.go
thc1006: Reserve pods are 1:1 with their Reservation object, and the reserve-pod PreFilter path (plugin.go:408) fetches that Reservation by name via `pl.rLister.Get(rName)`; every downstream decision — `ValidateReservation`, `state.rInfo`, the target node, allocatable capacity, `state.isPreAllocationRequired` — derives from that one Reservation. So the reservation name is the minimal unique identity of the reserve-pod scheduling path: two reserve pods pointing at the same Reservation would resolve to the same object and produce identical outcomes through this plugin.

In practice only one reserve pod exists per Reservation, so batching rarely kicks in on this short-circuit anyway. The reason to still return a non-empty fragment is that `SignPlugin` must return a signature (not `nil`) for the framework to keep opportunistic batching enabled for the overall profile; otherwise the whole scheduler profile falls back to disabled-for-all-pods.

Happy to change the value (e.g. use `pod.UID` or the reserve pod's namespaced name) if you'd prefer a pod-centric identity; they're all equivalent for uniqueness since the reserve pod's Name is itself derived from the reservation key.

ZiMengSheng:
@thc1006 Please resolve the conflicts

thc1006:
> @thc1006 Please resolve the conflicts

OK later I will fix

thc1006:
Done in 3b0d30bc. Conflict was just the var assertion blocks in coscheduling.go and elasticquota/plugin.go. Both SignPlugin and the new InformerFactoryProvider from #2875 are kept. CI green now.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -186,6 +188,157 @@ func (p *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWith
...
+	if err != nil {
+		// PreFilter (preparePod) returns UnschedulableAndUnresolvable for the
+		// same parse failure; mirror that exactly so the pod is handled
+		// identically whether opportunistic batching is on or off.
+		return nil, fwktype.NewStatus(fwktype.UnschedulableAndUnresolvable, err.Error())
+	}
+	// Parse reservation affinity only to mirror PreFilter's failure mode on
+	// malformed input; the presence bit is not emitted as a fragment because
+	// the Reservation plugin's SignPod already signs the affinity annotation.
+	if _, err := reservationutil.GetRequiredReservationAffinity(pod); err != nil {
```
saintube: This seems to be useless.

Same below.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -186,6 +188,157 @@ func (p *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWith
...
+		b, mErr := json.Marshal(spec)
+		if mErr != nil {
+			return nil, fwktype.NewStatus(fwktype.UnschedulableAndUnresolvable, mErr.Error())
+		}
+		fragments = append(fragments, fwktype.SignFragment{
+			Key:   "koord.DeviceShare.annotation:" + apiext.AnnotationGPUPartitionSpec,
+			Value: string(b),
+		})
+	}
+	if apiext.IsPreAllocationRequired(pod.Labels) {
```
saintube: Does this duplicate the reservation's Sign?

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -182,6 +184,158 @@ func (pl *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWit
...
+// For a normal pod the plugin-unique inputs are:
+//
+//   - reservation-affinity annotation (drives matching);
+//   - reservation-ignored label (skips reservation matching entirely);
+//   - exact-match-reservation annotation (extra strict matching);
+//   - owner-matching inputs — labels, pod.{UID,Name,Namespace,APIVersion}
+//     and every ownerRef field — consumed by the reservation cache's
+//     IsMatchable / MatchOwners / MatchObjectRef / MatchReservationControllerReference.
+func (pl *Plugin) SignPod(_ context.Context, pod *corev1.Pod) ([]fwktype.SignFragment, *fwktype.Status) {
+	if reservationutil.IsReservePod(pod) {
```
saintube, line 230: I suppose the Spec.PreAllocation should be a part of the sign, including whether the pre-allocation is required.

ZiMengSheng:
Pre-reviewed — this is the highest-quality PR in the current queue. The SignPod fragment selection shows real understanding of KEP-5598: per-plugin fragments dedupe across the profile union (reservation-affinity and pre-allocation label left to the Reservation plugin's SignPod), parsed-struct marshaling collapses semantically equal annotations, and the gating mirrors preparePod's `state.skip` branches so zero-request pods with malformed annotations batch identically to their PreFilter behavior. Six plugins covered with dedicated test files each. No blocking findings — ready for final review; the main thing to double-check is that each plugin's fragment list is complete against what its Filter/Score actually reads.

ZiMengSheng:
Also cc @saintube for a final look — this is the KEP-5598 core piece of the #2851 umbrella (SignPod across six plugins). @thc1006 @saintube PTAL at your convenience; happy to merge once we have a second pair of eyes on the fragment completeness (the one substantive check left: each plugin's fragment list vs. what its Filter/Score actually reads).

pkg/scheduler/plugins/deviceshare/plugin.go
thc1006: You're right, both were. The reservation-affinity parse produced no fragment and only mirrored a status PreFilter reaches on its own, and the pre-allocation fragment duplicated one the Reservation plugin already emits. Both are gone in 111425b2, with the pre-allocation input moved to the plugin that owns it.

pkg/scheduler/plugins/deviceshare/plugin.go
thc1006: It did. The keys differed, so the framework wouldn't have deduplicated them even though the value was the same function of the pod. Removed in 111425b2. DeviceShare reads that label only through `state.isReservationRequired`, so the shared signature still covers it.

pkg/scheduler/plugins/reservation/plugin.go
thc1006: Agreed, and it is now. Reserve pods get `koord.Reservation.reservePodPreAllocation` from `IsReservePodPreAllocation`, which mirrors `Spec.PreAllocation`, and normal pods get `koord.Reservation.preAllocationRequired` from the label. Both in 111425b2.

thc1006:
Thanks both. I've replied to the three open inline threads above; all three were fixed in 111425b2 and I should have said so then instead of leaving them open.

@ZiMengSheng, on the completeness check you left as the last substantive item: going through each plugin's fragments against what its Filter and Score actually read turned up three gaps. The fixes are written and tested locally but not pushed yet, since landing them will drop the /lgtm and I'd rather flag that first.

Coscheduling is the one I'd most like a second opinion on. `SignPod` returns only the gang id, but `PreScore` ranks nodes through `sortNodesByTopology`, which orders them by how many matching pods already sit under each topology node. That changes as soon as the previous pod of a batch is placed, so one pod's ranking isn't the next pod's ranking. `PodGroupManager.PreFilter` also pins every gang pod after the first to a single planned node, and `findNodesThatFitPod` tries the batch hint before it applies `PreFilterResult`, with no Coscheduling Filter to re-check it. No pod-level fragment repairs either of those, so I've made those pods refuse to sign, which is the opt-out KEP-5598 describes for plugins that depend on global placement. Ordinary gangs keep their fragment, since `AwareNetworkTopology` defaults to false.

NodeNUMAResource wasn't signing `AllowUseCPUSet`, which gates the entire CPU-binding block of its `PreFilter`. The default profile covers it by coincidence: InterPodAffinity signs `v1.Pod.Labels` and LoadAware signs the same effective priority class. Neither is guaranteed, though. InterPodAffinity emits nothing when `IgnorePreferredTermsOfExistingPods` is set, LoadAware may not be enabled, and priority can come from `pod.Spec.Priority` rather than a label, so two pods with identical labels and different `spec.Priority` share a signature today and get different CPU-binding treatment. Signing the derived boolean removes the dependency without splitting batches any finer than the label fragment already does.

DeviceShare parsed the GPU partition annotation whenever the pod requested any device. `parseGPURequirements` is that annotation's only production reader and it returns before touching it when GPU requests are zero, so RDMA-only and FPGA-only pods were being taken out of batching, and rejected on malformed JSON, over an input that never reaches their scheduling decision. The fix gates it on a non-zero GPU request.

The other three signers are unchanged. Each fix has a test that fails without it, and I'll post the input-to-fragment coverage table once these land, so the completeness question is reviewable rather than something you have to take my word for.

thc1006:
The `/lgtm` dropped when I pushed, as expected. Before you re-read this, one thing is not visible in the diff: I narrowed what the PR claims.

It no longer says it covers all six plugins that participate in Filter/Score. `computeBatchablePlugins` collects the PreFilter, Filter, PreScore and Score plugins, and `SchedulingHint`, `NodeResourcesFitPlus` and `ScarceResourceAvoidance` still turn signatures off for any profile that configures them. The body lists all three, and `fixes #2854` is now `Part of #2854`, so merging this will not close the umbrella issue.

@saintube, the three threads you opened in May are answered above. All three were addressed back in `111425b2`, and I should have said so at the time instead of leaving them open.

One more input turned up while I was checking the rest. DeviceShare's Filter compares `LabelGPUIsolationProvider` against the node's label for shared GPU requests, and no signer covered it, so two pods differing only in that label could share a signature. It is signed now, under the same gate Filter uses.

`signature_integration_test.go` asserts on the framework's own `SignPod` instead of calling plugins directly, including that a single unsigned plugin at PreFilter, PreScore or Score disables the profile, and it requires a non-empty signature before comparing two. It uses small fake plugins. Putting the six real ones into one profile still needs their args and informers, and I have not done that.

Production is 600 of the 2300 added lines.

ZiMengSheng:
Re-reviewed at `2f9c98eb`. The earlier `lgtm` was dropped by the bot on 09-06 15:51 after the two force-pushes, so this is a fresh pass over the whole PR plus the three new commits (`78b302a4`, `1903bb7d`, `2f9c98eb`, +932/-3).

## What holds up well

- **Fragment keys are fully namespaced and collision-free.** I extracted every key produced by the six plugins; all are `koord.<Plugin>.<field>`, no two plugins can shadow each other. That matters, because `frameworkImpl.SignPod` merges fragments into one `map[string]any` and the contract explicitly says "Only one fragment with a given key will be included" — a collision would be silently dropped, not reported.
- **Determinism is handled correctly.** `canonicalDeviceRequests`, `canonicalResourceList` and `canonicalOwnerInputs` each follow every `for k := range m` with a `sort.Strings`. Fragment *ordering* between plugins is already irrelevant (the framework marshals a map, so keys are sorted by `encoding/json`), but each fragment's own `Value` has to be stable, and it is.
- Each fragment is justified against the corresponding gate in `preparePod`, with line references. That made the review of "is anything that affects PreFilter/Filter/Score left unsigned within these six plugins" tractable.
- `coscheduling.SignPod` opting out with `Unschedulable` for network-topology-aware gangs is the right call and correctly uses `Unschedulable` (not `Error`) per the interface doc.

## Blocking: signatures are disabled in every configuration we ship

`SignPlugin` is all-or-nothing per profile — `interface.go`:

> If an enabled plugin that does Scoring, Prescoring, Filtering or Prefiltering does not implement this interface we will turn off batching for all pods.

And `framework/runtime/framework.go:787-802` implements exactly that: one unsupported candidate plugin sets `f.enableSignatures = false` for the profile, after which `SignPod` returns `nil` at line 821 for every pod.

Counting by extension-point method on `main`, ten plugin packages are candidates. This PR covers six:

| plugin | ext points | SignPod |
|---|---|---|
| coscheduling | PreFilter, Filter, PreScore, Score | yes |
| deviceshare | PreFilter, Filter, PreScore, Score | yes |
| elasticquota | PreFilter | yes |
| loadaware | PreFilter, Filter, Score | yes |
| nodenumaresource | PreFilter, Filter, PreScore, Score | yes |
| reservation | PreFilter, Filter, PreScore, Score | yes |
| **schedulinghint** | **PreFilter** | **no** |
| scheduleadmission | PreFilter | no |
| noderesourcefitplus | PreScore, Score | no |
| scarceresourceavoidance | PreScore, Score | no |

`SchedulingHint` is the blocking one, because it is **enabled by default**: `config/manager/scheduler-config.yaml` lists it as the *first* `preFilter.enabled` entry, and the released charts do the same (`versions/v1.7.0` and `versions/v1.8.0` of `koord-scheduler-config.yaml`). The `OpportunisticBatching` gate is `{Version: 1.35, Default: true, PreRelease: Beta}`, so the feature is on and it is our own config that switches signatures back off.

I also checked the other side: every in-tree upstream plugin registered at these four extension points in 1.35.2 does implement `SignPod`, and `frameworkext.PluginFactoryProxy` returns the plugin unwrapped, so the type assertion survives. `SchedulingHint` really is the only thing standing in the way.

Probe against the shipped plugin set, reusing the helpers from `signature_realplugins_test.go`:

```
I0907 framework.go:800] "Disabling signatures for profile because plugins do not support it."
    profile="koord-scheduler" plugins=["SchedulingHint"]

WITHOUT SchedulingHint: 105 bytes -> {"koord.Reservation.ownerInputs":"...","v1.Pod.Spec.SchedulerName":"koord-scheduler"}
WITH    SchedulingHint:   0 bytes -> ""
```

So as it stands, +1747 lines of `SignPod` are inert in the default deployment. You clearly know part of this — `TestProfileSignature_UnsupportedPluginDisablesTheProfile` asserts precisely this behaviour. My objection isn't that you missed it; it's that the PR description and #2854 present this as landing Opportunistic Batching, while the shipped effect is zero and the remaining four plugins aren't tracked anywhere.

## Why "just add SignPod to SchedulingHint" is not the fix

`schedulinghint.PreFilter` narrows nodes from `hinter.GetSchedulingHintState(state)` — a `CycleState` value, not anything derivable from the Pod. `SignPod(ctx, pod)` structurally cannot observe it. And the only writer of that state is our own `pkg/scheduler/batch/engine.go:181`.

So the framework disabling batching here is *correct*, not a gap: two pods with identical signatures would not get identical PreFilter results while `SchedulingHint` is active, and reusing a cached result across them would be wrong. What this PR actually surfaces is that **upstream KEP-5598 batching and our in-house `pkg/scheduler/batch` engine are mutually exclusive as currently wired**, and that is a design call I don't want to make inside a code review.

@saintube — could you settle the direction: does upstream batching *replace* `pkg/scheduler/batch`, coexist with it (in which case `SchedulingHint` has to leave the default profile for signatures to ever engage), or is `pkg/scheduler/batch` a migration step toward it?

## Related, lower severity

`deviceshare/plugin.go:195`, `nodenumaresource/plugin.go:294` and `reservation/scheduling_hint.go:38` also read the hint state inside `PreFilter` and change behaviour on it — `deviceshare` clears `designatedAllocation`/`designatedVF` when no hint is present; `nodenumaresource` builds a fixed allocation from `GetResourceStatus` when one is. None of the three `SignPod` implementations reference that state (grep count: 0).

Today this is masked: `SchedulingHint` being enabled kills signatures anyway. But the engine writes the hint state unconditionally, independent of whether the plugin is enabled. Disable `SchedulingHint` while keeping the batch engine and signatures switch on while an unsigned input still steers PreFilter. Worth a comment at minimum, so the coupling is deliberate rather than accidental.

## To unblock this

Either of these works for me:

1. **Scope it honestly.** Update the PR description to state that signatures stay disabled in the default profile until the batching/`pkg/scheduler/batch` question is resolved; open a follow-up issue enumerating `SchedulingHint`, `ScheduleAdmission`, `NodeResourcesFitPlus`, `ScarceResourceAvoidance`; note on #2854 that Opportunistic Batching is not yet effective. Then this is a reviewable, self-consistent step and I'm happy to `/lgtm`.
2. Or land the direction decision first and adjust the default profile accordingly in this PR.

On merge order, unchanged from my earlier note and now verified: I merged `upstream/main` (77 commits ahead of your merge-base) into `2f9c98eb` with no conflicts, `go build ./pkg/scheduler/... ./cmd/koord-scheduler/...` is clean and all 18 affected packages pass. This PR touches `reservation/plugin.go` in only three shallow places, and five other PRs queue behind those same files, so it should still go in first — please rebase onto `main` when you push the next revision.

thc1006:
You are right, and I reproduced it before changing anything. `TestProfileSignature_ShippedProfileDisablesSignatures` builds the shipped `preFilter` list out of the production factories, asserts the signature is empty, then asserts the same five signers produce one once `SchedulingHint` is dropped. Your probe is a test now, so it stays true.

I agree the framework is right here rather than gapped. `schedulinghint.PreFilter` narrows nodes from a `CycleState` value that `SignPod` never sees, so signing it would be a lie, and an empty `SignPod` would be a worse one.

Taking option 1. The description now opens with signatures being disabled in the default deployment, lists all four unsigned plugins and which of them ship enabled, and says the batching question is not settled here. I had missed `ScheduleAdmission` because my merge base predated it. It is in the table now.

Your lower-severity point holds as well. DeviceShare, NodeNUMAResource and Reservation branch on the same hint state in their own PreFilter and none of them sign it, and I could not find a way to sign it either, for the same reason you gave. Their doc comments record the coupling and that it is masked only while signatures are off for the profile.

Rebased onto main, `behind_by` is 0. The conflicts were the `var _` assertion blocks in elasticquota and coscheduling plus one import block in nodenumaresource.

@saintube, the direction call is yours: does upstream batching replace `pkg/scheduler/batch`, coexist with it, or is `pkg/scheduler/batch` a step toward it? I will follow whichever way you decide, and open the follow-up issue for the remaining plugins once it is clear what they should be tracked against.

ZiMengSheng:
Verified `cdbf1bf6`. Option 1 is delivered in every part I asked for, and the two profile-level tests are real rather than vacuous. `/lgtm`.

## What I checked

**Rebase and merge order.** `git rev-list --count refs/review/2863..upstream/main` is `0`, so this is on top of `81e01347`. `go build ./pkg/scheduler/... ./cmd/koord-scheduler/...` is clean, and `go test -count=1` passes for all 24 packages under `pkg/scheduler/frameworkext/...` and the six plugin trees. CI is green including `Check auto-generated codes` and `Check manifests`. Merge-order position is unchanged: this still goes in first.

**The profile tests are mutation-verified.** An assertion of the form `assert.Empty(t, ...SignPod(...))` passes for the wrong reason as easily as the right one, so I dropped a `SignPod` onto `schedulinghint.Plugin` returning one fragment and re-ran:

```
--- FAIL: TestProfileSignature_UnsupportedPluginDisablesTheProfile
    signature_realplugins_test.go:223: Should be empty, but was {...}
--- FAIL: TestProfileSignature_ShippedProfileDisablesSignatures
    signature_realplugins_test.go:431: Should be empty, but was {...}
```

Both fail, and only those two. So they really are driving `computeBatchablePlugins` through the production factories, and the `require.NotEmpty` guard on the without-hint half is doing its job. `TestProfileSignatureCoverage` covering PreFilter, PreScore and Score separately is the right call too — collecting only Filter and Score would have been an easy thing to get wrong and impossible to notice.

**Doc comments.** The hint-state coupling is recorded at all three sites (`deviceshare/plugin.go:215`, `nodenumaresource/plugin.go:285`, `reservation/plugin.go:215`), each saying both that `SignPod` cannot see the state and that it is masked while signatures are off. That is what I wanted: the coupling is now deliberate.

**Description and #2854.** The scope note leads with signatures being disabled, the four unsigned plugins are tabled with which one ships enabled, and the note on #2854 says Opportunistic Batching is not yet effective. `ScheduleAdmission` is in the table. Leaving the four out rather than giving them an empty `SignPod` is the right instinct — an empty signer is an assertion, not an abstention.

**One correction to my own earlier note.** I wrote that ten plugin packages are candidates. `NodeResourcesFitPlus` lives in `pkg/scheduler/plugins/noderesourcefitplus`, not `noderesourcesfitplus`; the count and the table were right, the path in my message was not.

## For the follow-up issue: owner inputs make reuse structurally zero, not just config-zero

This does not block, and the description already says batch reuse "is a further step again". But the follow-up issue should carry it, because it is the next wall and it is not the `SchedulingHint` wall.

`canonicalOwnerInputs` includes `name=` and `uid=` unconditionally. Probe on the shipped `preFilter` list *minus* `SchedulingHint` — two pods of the same ReplicaSet, identical spec, differing only in name and UID:

```
A = {"koord.ElasticQuota.quota":"koordinator-default-quota|",
     "koord.Reservation.ownerInputs":"ns=default|name=nginx-abc-11111|uid=nginx-abc-11111|label:app=nginx|owner:apiVersion=apps/v1/kind=ReplicaSet/name=nginx-abc/uid=rs-uid/controller=",
     "v1.Pod.Spec.SchedulerName":"koord-scheduler"}
B = {... name=nginx-abc-22222|uid=nginx-abc-22222 ...}
DIFFERENT
```

So once the direction question is settled and signatures do engage, every ordinary pod still gets a unique signature in any profile that enables `Reservation` — which is every profile we ship, at PreFilter, Filter, PreScore and Score. Hit rate stays zero. `TestProfileSignature_ReservationSignsPodIdentity` records this as a trade-off, and I agree the fragment is *sound*: `MatchObjectRef` really does read `pod.UID`, `pod.Name`, `pod.Namespace` and `pod.APIVersion` (`pkg/util/reservation/reservation.go:471-477`), so a signature that omitted them could let a cached PreFilter result cross an owner-match boundary. Conservative is correct here.

What I'd want the follow-up to consider is narrowing rather than dropping. `Plugin` already holds `rLister`, so `SignPod` can ask which `ObjectReference` fields any live reservation's `spec.owners[*].Object` actually sets, and include only those pod fields. Owners in practice are `LabelSelector` or `Controller` shaped, and `Owners == nil` matches nothing at all — in both cases `name` and `uid` fall out and same-workload pods collapse to one signature, with no loss of soundness because no matcher can distinguish them.

The trade-off to decide deliberately, not in passing: that makes a fragment depend on cluster state rather than on the pod alone, so the signature's meaning shifts when a reservation is created or deleted. The staleness window is bounded by `maxBatchAge` (500ms), which is probably fine, but "probably fine" is the kind of thing that should be written down before it is relied on. Same bucket as the `pkg/scheduler/batch` question, so please fold it into the same follow-up.

/lgtm

@saintube — two things now wait on the direction call: whether upstream batching replaces `pkg/scheduler/batch`, coexists with it (in which case `SchedulingHint` has to leave the default profile), or is a migration target; and, downstream of that, whether owner-input granularity is worth the state dependency above. This PR is self-consistent without either answer.

thc1006:
Thank you for the mutation run on the two profile tests. Dropping a `SignPod` onto `schedulinghint.Plugin` and checking that exactly those two fail is a sharper test of them than I had, and I would rather have that on record than my own word for it.

**TL;DR:** folded into #3198, reproduced your probe, and I agree the fragment wants narrowing rather than dropping.

I rebuilt the shipped `preFilter` set without `SchedulingHint` from the production factories and got your two signatures byte for byte, so #3198 now carries the second wall with the soundness argument, the narrowing idea and the state-dependency trade-off, next to the `pkg/scheduler/batch` question.

One thing I added there for whoever picks it up: working out which `ObjectReference` fields are in use means reading every live reservation's `spec.owners`, and `SignPod` runs once per pod. That points at a value maintained by the reservation event handler and read as a snapshot, rather than a scan on the signing path. It also moves the staleness question, since the value would then change on informer events instead of on lister reads.

Noted your correction on the package path as well. `noderesourcefitplus` is what the table uses.

Nothing further from my side on this one. Both open questions are the direction call.

PR #2872, opened 2026-04-17
scheduler: support workload auditor

ZiMengSheng:
### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
+++ b/pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
@@ -1,0 +1,430 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workloadauditor
+
+import (
+	"fmt"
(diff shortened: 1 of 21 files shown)


pkg/scheduler/plugins/coscheduling/core/gang_cache.go
```diff
@@ -142,26 +146,38 @@ func (gangCache *GangCache) onPodAddInternal(obj interface{}, action string) {
 
 		gangGroup := gang.getGangGroup()
 		gangGroupId := util.GetGangGroupId(gangGroup)
-		gangGroupInfo := gangCache.getGangGroupInfo(gangGroupId, gangGroup, true)
+		gangGroupInfo, created := gangCache.getGangGroupInfo(gangGroupId, gangGroup, true)
 		gang.SetGangGroupInfo(gangGroupInfo)
+		if created && pod.Spec.NodeName == "" && gangCache.isResponsibleForPod(pod) && gangCache.workloadAuditor != nil {
+			gangCache.workloadAuditor.AddGangGroup(gang.GangGroupId)
```
saintube: gang.GangGroupId or gangGroupId, which one is better?

pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
```diff
@@ -0,0 +1,430 @@
...
+
+func (w *workloadAuditorImpl) RecordGangGroup(gangGroupID string, pod *corev1.Pod, recordType RecordType, message string) {
+	if !w.Config.Enabled {
+		return
+	}
+	start := time.Now()
+	defer func() {
+		RecordMethodDurationSeconds.WithLabelValues("RecordGangGroup").Observe(time.Since(start).Seconds())
+	}()
+	record := w.getRecord(gangGroupID)
```
saintube, line 122: There might be a race case with the AddPod handler between L122-L126.
Imagine a pod event triggers a store to the new gangGroupID and has not locked the record, while another pod loads it concurrently, making the records out-of-order.

PR #2873, opened 2026-04-21
scheduler: provide schedule suggestion for job/pod scheduling failure

ZiMengSheng:

This PR introduces a Suggestion mechanism in the scheduler's diagnosis system. After a job/pod scheduling failure, the scheduler can now provide actionable advice to external components (e.g., descheduler, autoscaler, or platform controllers) via a typed ScheduleSuggestion attached to the Diagnosis.

Example Pod condition message:
```
message: '0/72 nodes are available: ..., Suggestion: {type: EvictWorkloadSelf, message: "XXX"}.'
```

Diagnosis now embeds sync.Mutex (suggestionMu), so it must not be copied by value. Existing test assertions were updated to use pointer comparison (*Diagnosis → compare via pointer).
The SetSuggestion method enforces set-once semantics: the first plugin to set a suggestion wins, subsequent calls return false.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -345,6 +345,9 @@ func (pgMgr *PodGroupManager) BeforePreFilter(ctx context.Context, cycleState fw
 	}
 	if gangSchedulingContext.failedMessage != "" {
 		diagnosis.IsRootCausePod = false
+		if gangSchedulingContext.suggestion != nil {
+			diagnosis.SetSuggestion(gangSchedulingContext.suggestion)
+		}
 		return fmt.Errorf("%s", gangSchedulingContext.failedMessage)
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/scheduler/plugins/coscheduling/core/core.go
```diff
@@ -384,6 +387,13 @@ func (pgMgr *PodGroupManager) AfterPostFilter(ctx context.Context, state fwktype
 	}
 
 	message := pgMgr.summaryAndRecordFailedMessage(state, pod, filteredNodeStatusMap)
+	diagnosis := frameworkext.GetDiagnosis(state)
+	if diagnosis != nil && diagnosis.Suggestion != nil {
```
saintube: Why not wrap with RWMutex?

ZiMengSheng: somewhat not necessary, so omit it, add it is ok

PR #2875, opened 2026-04-23
scheduler: refactor ForceSyncFromInformer to align with vanilla kube-scheduler behavior

saintube:

koord-scheduler: This change revises the ForceSyncFromInformer mechanism to eliminate the following side effects:

- Inconsistent Startup Order: Previously, ForceSyncFromInformer started the informer factory before the Scheduler’s Run() method. This caused inconsistency between koord-scheduler and the vanilla kube-scheduler.
- High Startup Latency: To ensure event handlers were invoked synchronously, the previous implementation serially processed all stored events for each registered handler. This introduced non-negligible overhead to scheduler startup latency.

We now leverage the upstream Kubernetes solution (https://github.com/kubernetes/kubernetes/issues/113763, https://github.com/kubernetes/kubernetes/pull/116729) to replace this workaround.

With this patch, plugin event handlers are registered earlier in the lifecycle. While this aligns koord-scheduler with the standard boot procedure used by most scheduler plugins, existing koord plugins that depend on a specific startup order between event handlers and the scheduler's `Run()` method may require adaptation.

### V. Checklist

### DIFF
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -68,6 +68,7 @@ import (
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/defaultprofile"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/eventhandlers"
+	frameworkexthelper "github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/helper"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/informer"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/networktopology"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/services"
@@ -170,6 +171,12 @@ func runCommand(cmd *cobra.Command, opts *options.Options, registryOptions ...Op
 
 // Run executes the scheduler based on the given configuration. It only returns on error or when context is done.
 func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *scheduler.Scheduler, extenderFactory *frameworkext.FrameworkExtenderFactory, customWorkflow CustomWorkflow) error {
+	// Wrap the incoming ctx so that Run itself owns a cancel function; this lets
+	// leader-election callbacks trigger a graceful shutdown (e.g. when plugin
+	// initialization fails) by canceling the scheduler context instead of
+	// abruptly terminating via klog.Fatalf.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	logger := klog.FromContext(ctx)
 	// To help debugging, immediately log version
 	logger.Info("Starting Koordinator Scheduler version", "version", version.Get())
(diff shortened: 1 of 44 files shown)


pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -592,7 +593,14 @@ func (n *nodeDeviceCache) getAllNodeDeviceSummary() map[string]*NodeDeviceSummar
 
 func (n *nodeDeviceCache) gcNodeDevice(ctx context.Context, informerFactory informers.SharedInformerFactory, period time.Duration) {
 	nodeInformer := informerFactory.Core().V1().Nodes().Informer()
+	// Register a no-op handler to collect the registration and ensure the
+	// node informer is included in the framework's handler-sync tracking.
 	frameworkexthelper.ForceSyncFromInformer(ctx.Done(), informerFactory, nodeInformer, nil)
```
ZiMengSheng: Is this still needed? Can we put it to controller, to assure it will start after acquire leader?

pkg/scheduler/frameworkext/interface.go
```diff
@@ -295,3 +296,25 @@ type NextPodPlugin interface {
...
 type ForgetPodHandler func(pod *corev1.Pod)
+
+// SharedInformerFactory is a minimal interface for informer factories
+// that can be started and cache-synced.
+type SharedInformerFactory interface {
+	Start(stopCh <-chan struct{})
+	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
+}
+
+// InformerFactoryProvider is implemented by plugins that create their own
```
ZiMengSheng, line 307: Another approach might be to have a framework initialization process, similar to noderesourcetopology, and then allow plugins to access this information via a handler. This would allow for the configuration of the transformer.

pkg/scheduler/plugins/deviceshare/device_cache.go
saintube: I suppose it is not a bug, so we can also use frameworkext.WaitForHandlerSync to keep the behavior consistent.

pkg/scheduler/frameworkext/interface.go
saintube: Yes, it is an option that depends on whether the informer factory is common for plugins. With this patch, the plugins are free to configure transform hooks inside their `New()`.

cmd/koord-scheduler/app/server.go
```diff
@@ -233,7 +234,46 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
...
+		// other events and does not require this cross-factory ordering.
+
+		// Step 1: start plugin informer factories registered via InformerFactoryProvider.
+		for _, f := range extenderFactory.GetPluginInformerFactories() {
+			f.Start(ctx.Done())
+		}
+		for _, f := range extenderFactory.GetPluginInformerFactories() {
+			f.WaitForCacheSync(ctx.Done())
+		}
+		// Step 2: wait for replaceHandlers (e.g. ElasticQuota's ReplaceQuotas) to
```
ZiMengSheng: a little not graceful

saintube: refactor as InformersSynced hooks

ZiMengSheng (approved):
/lgtm
/approve

PR #2876, opened 2026-04-23
koord-scheduler: fix integer rounding fragmentation in ElasticQuota runtime redistribution

qinfustu:

Fix an integer rounding bug in `ElasticQuota` runtime redistribution.

In `iterationForRedistribution`, each child's share of the lendable pool is
rounded independently with `int64(float64(...) + 0.5)`. For integer-only
resources like `nvidia.com/gpu`, the rounded shares can sum to less than the
parent's total, leaving capacity stranded — `parent.runtime > Σ(child.runtime)`.

This PR replaces the per-node rounding with the Largest-Remainder (Hamilton)
method (new helper `computeHamiltonDeltas`):

- base = `floor(weight * totalRes / totalSharedWeight)`
- distribute the leftover units to the nodes with the largest fractional parts

This guarantees `Σ(delta) == totalRes` exactly.

Fixes a GPU fragmentation issue observed in production:
parent quota with `max=81` GPUs, children's runtime sum only reached 78,
so 3 GPUs were permanently stranded and pending Pods could not be scheduled.
No linked issue.

A regression test is added:

```bash
go test ./pkg/scheduler/plugins/elasticquota/core/... \
    -run TestRuntimeQuotaCalculator_IterationAdjustQuota_IntegerDistribution -v

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
+++ b/pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
@@ -17,6 +17,8 @@ limitations under the License.
 package core
 
 import (
+	"math/bits"
+	"sort"
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go [hunk 1]
```diff
@@ -172,6 +180,71 @@ func (qt *quotaTree) iterationForRedistribution(totalRes, totalSharedWeight int6
...
+		remainders = append(remainders, remainderEntry{
+			index:      i,
+			fractional: share - float64(base),
+			name:       node.quotaName,
+		})
+	}
+
+	residual := totalRes - distributed
+	if residual <= 0 || len(remainders) == 0 {
+		return deltas
```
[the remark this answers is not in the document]
qinfustu: use 128-bit integer arithmetic to avoid float64 precision loss

pkg/scheduler/plugins/elasticquota/core/runtime_quota_calculator.go
```diff
@@ -172,6 +180,70 @@ func (qt *quotaTree) iterationForRedistribution(totalRes, totalSharedWeight int6
 	}
 }
 
+// computeHamiltonDeltas splits totalRes into per-node integer deltas proportional
+// to node.sharedWeight using the largest-remainder method:
```
ZiMengSheng, line 184: how about the time complexity

qinfustu: The time complexity changes from O(N) to O(N log N) due to an added sort step. This adds a few microseconds per scheduling cycle, which is negligible compared to the millisecond-level scheduling latency.

ZiMengSheng (approved):
/lgtm

PR #2883, opened 2026-04-29
scheduler: fix tryAllocateFromNode

saintube:

koord-scheduler: Fix a bug when the pod without the reservation-affinity fails to allocate NUMA CPU resources from the node unreserved part. Revise the DeviceShare logic to avoid the same kind of error.

### V. Checklist


ZiMengSheng (commented):
/lgtm
/approve

PR #2885, opened 2026-04-30
scheduler: integrate NominatedNodeNameForExpectation with reservation preemption 

### LINKED ISSUE #2858: NominatedNodeName for expectation
NominatedNodeName Expectation Mechanism

Kubernetes 1.35 introduces a new expectation mechanism tied to NominatedNodeName to avoid race conditions between preemption and pod scheduling.

What needs to be done:

- [ ] Audit how the Reservation preemption sets and clears NominatedNodeName
- [ ] Integrate with the new k8s expectation mechanism in pkg/scheduler/plugins/reservation/preemption.go
- [ ] Verify that preemption logic for clearing nominatedNodeName is compatible with the new mechanism
- [ ] Add regression tests for race condition scenarios

adity1raut:

Kubernetes 1.35 introduces `NominatedNodeNameForExpectation` (Beta, default on), which sets
`NominatedNodeName` (NNN) during the **binding cycle** for pods awaiting `WaitOnPermit` or
`PreBind` — not just during preemption. This creates two race conditions in Koordinator's
reservation preemption:

**Race 1 — stale ReservationNominator on NNN clearing:**
When k8s preemption's `clearNominatedNodeName` patches a lower-priority pod's NNN to `""`,
the informer fires a pod update event. But `pod_eventhandler.updatePod` returns early for
unassigned pods (`Spec.NodeName == ""`), leaving `ReservationNominator` with stale entries
that can block subsequent scheduling decisions.

**Race 2 — async informer lag after preemption:**
After a successful preemption cycle, the NNN-clearing informer event may not propagate
before the next scheduling cycle reads `ReservationNominator`, causing stale reserve pod
nominations to interfere.

**Fix 1 — `pod_eventhandler.go`:**
When an unassigned pod's NNN transitions from non-empty → empty, call
`DeleteNominatedReservePodOrReservation` to immediately sync the nominator instead of
returning early.

**Fix 2 — `preemption.go`:**
After `pe.Preempt` succeeds, synchronously clear `ReservationNominator` entries for
lower-priority reserve pods on the nominated node via a new
`clearLowerPriorityReservePodNominations` helper — providing immediate consistency without
waiting for informer propagation.

fixes #2858

1. Run the new unit tests
2. Run the full reservation plugin test suite to confirm no regressions

### DIFF
--- a/pkg/scheduler/plugins/reservation/pod_eventhandler.go
+++ b/pkg/scheduler/plugins/reservation/pod_eventhandler.go
@@ -93,6 +93,15 @@ func (h *podEventHandler) updatePod(oldPod, newPod *corev1.Pod) {
 	}
 
 	if !assignedPod(newPod) {
+		// When NominatedNodeName is cleared for an unassigned pod (e.g., by the k8s
+		// preemption's clearNominatedNodeName), sync the ReservationNominator to prevent
+		// stale nominations. With the NominatedNodeNameForExpectation feature gate (k8s 1.35+,
+		// Beta by default), NNN is also set during the binding cycle for pods awaiting
+		// WaitOnPermit or PreBind, so more pods will now have NNN set and may be subject
+		// to clearing by a concurrent preemption cycle.
+		if oldPod != nil && oldPod.Status.NominatedNodeName != "" && newPod.Status.NominatedNodeName == "" {
+			h.nominator.DeleteNominatedReservePodOrReservation(newPod)
+		}
 		return
 	}
 
--- a/pkg/scheduler/plugins/reservation/pod_eventhandler_test.go
+++ b/pkg/scheduler/plugins/reservation/pod_eventhandler_test.go
@@ -425,6 +425,176 @@ func TestPodEventHandler_PreAllocatableCacheSync(t *testing.T) {
 	}
 }
 
+// TestPodEventHandler_NominatedNodeNameClearing tests that when an unassigned pod's
+// NominatedNodeName is cleared (e.g., by preemption's clearNominatedNodeName, or after a
+// binding-cycle failure with the NominatedNodeNameForExpectation feature), the
+// ReservationNominator is updated to prevent stale nominations from causing scheduling conflicts.
+func TestPodEventHandler_NominatedNodeNameClearing(t *testing.T) {
+	tests := []struct {
+		name        string
+		setupFunc   func(handler *podEventHandler, pod *corev1.Pod, podInfo *framework.PodInfo)
+		oldPod      func(pod *corev1.Pod) *corev1.Pod
+		newPod      func(pod *corev1.Pod) *corev1.Pod
+		verifyFunc  func(t *testing.T, handler *podEventHandler, pod *corev1.Pod, podInfo *framework.PodInfo)
+		isReserve   bool
+	}{
+		{
+			name: "reserve pod NNN cleared while unassigned removes reservation nomination",
+			setupFunc: func(handler *podEventHandler, pod *corev1.Pod, podInfo *framework.PodInfo) {
+				handler.nominator.AddNominatedReservePod(podInfo, "test-node-1")
+			},
+			oldPod: func(pod *corev1.Pod) *corev1.Pod {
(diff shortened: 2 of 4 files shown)


saintube:
@adity1raut Thanks for the suggestion! However, this change seems to depend on the Reservation API supporting NominatedNodeName. Could you please help implement that prerequisite as well?

adity1raut:
> @adity1raut Thanks for the suggestion! However, this change seems to depend on the Reservation API supporting NominatedNodeName. Could you please help implement that prerequisite as well?

yeah Sure

saintube:
@adity1raut hi, is there any update about the Reservation NominatedNodeName API?

adity1raut:
> @adity1raut hi, is there any update about the Reservation NominatedNodeName API?

Working on it

ZiMengSheng:
Pre-review, done as part of a cross-check over the four PRs that touch the reservation nominator (#3084, #3114, #3156, this one). Cluster analysis: https://github.com/koordinator-sh/koordinator/pull/3084#issuecomment-5564939241

**This one is genuinely independent of the other three and the reasoning holds up.** I went in expecting overlap and found none: the others react to `Reservation` events, this reacts to `Pod` events and to the preemption cycle itself.

Things I checked rather than took on trust:

- **Your premise about the feature gate is accurate.** `NominatedNodeNameForExpectation` is `{Version: 1.35, Default: true, PreRelease: Beta}`, so NNN really is set during the binding cycle by default now and the population of pods carrying an NNN is larger than before. The PR description isn't overstating it.
- **`clearLowerPriorityReservePodNominations` mirrors upstream exactly, and that is the right call.** I initially read "clear every lower-priority nominee on the node, not just the victims" as over-broad — it isn't. Upstream `getLowerPriorityNominatedPods` does precisely that: `NominatedPodsForNode(nodeName)` filtered on `PodPriority(pi) < PodPriority(pod)`, no victim check. You're applying the same rule to the reserve-pod nominator that upstream applies to the pod nominator.
- **It fills a real gap rather than duplicating upstream.** Upstream's clearing runs over `PodNominator.NominatedPodsForNode`, which never sees koord's reserve pods — they live in a separate store reached through `NominatedReservePodForNode`. So nothing was cleaning these up.
- **The iteration is safe.** `NominatedReservePodForNode` deep-copies before returning ("Make a copy of the nominated Pods so the caller can mutate safely"), so deleting inside the loop can't corrupt the slice you're walking.
- **The `pod_eventhandler.go` guard is correctly placed.** Sitting inside `if !assignedPod(newPod)` means the NNN-cleared-on-successful-bind transition doesn't trigger it, which is exactly the case that would otherwise misfire now that NNN is set during binding. And `DeleteNominatedReservePodOrReservation` on a regular pod routes to `RemoveNominatedReservation`, which is the right store.

Two points to address:

1. **Blocking, mechanical: this is `CONFLICTING` against `main`.** Your merge-base is 85 commits behind and `pkg/scheduler/plugins/reservation/pod_eventhandler.go` conflicts. Please rebase — I can't run a real pass on a tree that doesn't merge. Everything below is from reading the diff against your merge-base, so re-confirm it after the rebase.

2. **Interaction with #3084 that you should know about.** That PR persists the nomination into `Reservation.status.NominatedNodeName` and re-syncs the in-memory nominator from it on every `OnUpdate`. Your clearing here is in-memory only and doesn't touch `Reservation.status`. Once #3084 lands, a nomination you drop in `clearLowerPriorityReservePodNominations` gets re-added from the persisted status on the next reservation update — the "stale nomination" you're racing to remove comes back. Whichever lands second needs to close that loop, either by clearing the persisted field too or by having the restore path validate against the current spec. I've raised the same class of problem on #3084 and #3156; worth reading that thread before you rebase, since it may change what you want this function to do.

One smaller note, not a blocker: neither path clears `nominatedPreAllocatable`, so a pre-allocation nomination can outlive the pod nomination it belonged to. That's pre-existing and deliberately out of scope for #3156 too — tracked in #3158 — so I'm not asking you to fix it here, just flagging that this PR doesn't close it either.

PR #2888, opened 2026-05-02
test: add unit tests for reservation controller reservation_eventhandler.go

adity1raut:

- Add `reservation_eventhandler_test.go` for
  `pkg/scheduler/plugins/reservation/controller/reservation_eventhandler.go`
- File had zero test coverage — all 6 functions now at 100%

### V. Checklist


ZiMengSheng:

/lgtm

saintube (approved):
/lgtm

PR #2896, opened 2026-05-07
test: add 100% unit test coverage for GangGroupInfo in coscheduling core

### LINKED ISSUE #2895: [Test]: add unit tests for pkg/scheduler/plugins/coscheduling/core/ganggroup.go
**What happened**:
- `pkg/scheduler/plugins/coscheduling/core/ganggroup.go` contains the `GangGroupInfo` struct
  and all its methods that manage the lifecycle of a gang group during co-scheduling. This file
  has **zero unit test coverage** despite containing non-trivial concurrent state logic with
  mutex-protected reads/writes and Prometheus metric side effects.

- A corresponding `ganggroup_test.go` file should exist in the same package covering all
  exported and key unexported methods of `GangGroupInfo`:
  - `NewGangGroupInfo` — correct field initialization
  - `SetInitialized` / `IsInitialized` — state transition, concurrent-safe
  - `setResourceSatisfied` / `isGangOnceResourceSatisfied` — irreversibility once set true
  - `AddWaitingGang` — inserts all GangGroup IDs; metric `WaitingGangGroupNumber` incremented only on first call
  - `RemoveWaitingGang` — removes single gang ID; metric decremented when set becomes empty; `BindingMemberPods` cleared
  - `ClearWaitingGang` — clears all; metric decremented; `BindingMemberPods` cleared
  - `RecordIfNoRepresentatives` — sets key when empty, returns existing key when already set
  - `DeleteIfRepresentative` — clears key only if the pod matches
  - `IsRepresentative` — returns true only for the representative pod
  - `ClearCurrentRepresentative` — always clears
  - `SetBindingMembers` / `GetBindingMembers` — round-trip correctness

adity1raut:
Adds ganggroup_test.go covering all methods of GangGroupInfo including concurrent-safety cases, Prometheus metric side-effects on WaitingGangGroupNumber (increment/decrement/no-op paths), irreversibility of OnceResourceSatisfied, BindingMemberPods clearing semantics, and a full add→bind→clear lifecycle test.

## Summary

- `ganggroup.go` had zero test coverage despite containing non-trivial concurrent state logic, mutex-protected reads/writes, and Prometheus metric side-effects
- Adds `ganggroup_test.go` with 35 tests achieving **100% statement coverage** on every function in the file
- Fixed #2895 

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/ganggroup_test.go
+++ b/pkg/scheduler/plugins/coscheduling/core/ganggroup_test.go
@@ -1,0 +1,382 @@
+/*
+Copyright 2026 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package core
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


ZiMengSheng (approved):
/lgtm
/approve

PR #2900, opened 2026-05-09
feat(scheduler): add EnableAsyncPreemption to Reservation and Coscheduling

### LINKED ISSUE #2857: Asynchronous Preemption
Add EnableAsyncPreemption Parameter to Reservation Plugin

Kubernetes 1.35 enables Asynchronous Preemption by default, where preemption victim deletion is handled asynchronously. The Reservation plugin and the Coscheduling plugin need a feature gate to control this behavior.

vibhor-5:

This PR implements **Asynchronous Preemption** support for the `Reservation` and `Coscheduling` plugins, aligning with Kubernetes 1.35 behavior.

**Changes:**
- **Configuration:** Added `EnableAsyncPreemption` (boolean) to `ReservationArgs` and `CoschedulingArgs` (internal and `v1`).
- **Defaults:** `EnableAsyncPreemption` defaults to `true` to match K8s 1.35.
- **Reservation Plugin:** Passes the new flag to the standard Kubernetes preemption evaluator.
- **Coscheduling Plugin:** 
    - Updated `NewPreemptionEvaluator` to respect the flag.
    - Modified logic to allow preemption even if terminating pods exist on nominated nodes when enabled.
    - **Bug Fix:** Fixed a critical bug in `placeToSchedulePods` where assumed `NodeInfo` and `CycleState` were not being persisted across iterations during dry-run preemption, which could lead to incorrect resource estimations for multi-pod groups.
- **Testing:** Added a dedicated unit test suite for async preemption logic and updated existing tests to maintain compatibility.

fixes #2857

1.  **Unit Tests:**
    ```bash
    go test -v ./pkg/scheduler/plugins/coscheduling/core/ -run Test_podEligibleToPreemptOthers_Async
    ```
2.  **Full Test Suite:**
    ```bash
    make test
    ```

The bug fix in `pkg/scheduler/plugins/coscheduling/core/preemption.go` (`placeToSchedulePods`) is essential for correct multi-pod preemption simulation, as without it, each pod in a gang was scheduled against the original node state instead of an incrementally updated one.

### V. Checklist


ZiMengSheng:
@vibhor-5 Please fix the fmt Issue

pkg/scheduler/plugins/coscheduling/core/preemption.go
```diff
@@ -392,8 +394,10 @@ func (ev *preemptionEvaluatorImpl) podEligibleToPreemptOthers(ctx context.Contex
 				if ev.isPreemptionAllowed(nodeInfo, p, pod) && podTerminatingByPreemption(p.GetPod()) {
 					terminatingPodKey := framework.GetNamespacedName(p.GetPod().Namespace, p.GetPod().Name)
 					jobPreemptionState.TerminatingPodOnNominatedNode[terminatingPodKey] = nomNodeName
-					// There is a terminating pod on the nominated node.
-					return false, ReasonTerminatingVictimOnNominatedNode
+					if !ev.enableAsyncPreemption {
```
ZiMengSheng, line 399: why remove it

pkg/scheduler/plugins/coscheduling/core/preemption.go
```diff
@@ -680,10 +684,12 @@ func (ev *preemptionEvaluatorImpl) placeToSchedulePods(
 				// 2. another to determine the victim on the candidate node.
 				// Here we choose clone to avoid the modification of cycleState affecting the subsequent determination of Victim
 				assumedCycleState = cycleStates[selectedNode.Node().Name].Clone()
+				assumedCycleStates[selectedNode.Node().Name] = assumedCycleState
```
ZiMengSheng, line 688: Please also add UT for this

pkg/scheduler/plugins/coscheduling/core/preemption.go
vibhor-5: I removed that specific line and its comment to streamline the code during the implementation of the EnableAsyncPreemption feature gate. My intent was to make the logic conditional on the feature gate (if !ev.enableAsyncPreemption) rather than an unconditional early return.

 Since the original logic was:

   if ... {
       // There is a terminating pod on the nominated node.
       return false, ReasonTerminatingVictimOnNominatedNode
   }

 By wrapping it in the feature gate check, the logic becomes:

   if ... {
       // ... logic for terminating pod key ...
       if !ev.enableAsyncPreemption {
           return false, ReasonTerminatingVictimOnNominatedNode
       }
   }

  The comment was essentially describing the block I was modifying, and I deemed it redundant given the new structure. However, I understand the desire for clarity. Would you like me to restore the comment for better documentation of this specific branch?

pkg/scheduler/plugins/coscheduling/core/preemption.go
vibhor-5: added in `preemption_async_test.go`

vibhor-5:
@ZiMengSheng fixed the ci issues, added back the comment and added unit test for the change 

PTAL

vibhor-5:
@ZiMengSheng ci is all green lemme know if any further changes are needed

ZiMengSheng:
Pre-reviewed — config plumbing follows the standard pattern (internal + v1 types, conversion, deepcopy, defaults), and passing `EnableAsyncPreemption` into the standard K8s preemption evaluator is exactly the right integration point rather than reimplementing async victim handling. Two things for final review: (1) `defaults.go` flips the default to `true` to match K8s 1.35 — confirm the reservation plugin's async path handles the gang-victim case (coscheduling) that the sync path special-cases, since async victim deletion reorders gang teardown; (2) the `network_topology_workflow_test.go` one-liner change looks like a mechanical rename fallout — worth confirming it doesn't weaken that test.

ZiMengSheng:
Final review. I rebuilt this on top of current `upstream/main` (merge-base `5bd00b43` is 77 commits behind; `git merge upstream/main` is clean, `go build ./pkg/scheduler/... ./cmd/koord-scheduler/...` and `go test ./pkg/scheduler/plugins/coscheduling/... ./pkg/scheduler/plugins/reservation/... ./pkg/scheduler/apis/config/...` are green, and CI including `Check auto-generated codes` is green). So nothing here is a mechanical problem.

The problem is semantic. Production code is 26 lines out of +553; those 26 lines make three decisions that don't match what `EnableAsyncPreemption` means upstream, plus they carry one unrelated bug fix.

### 1. In coscheduling, this flag does not make anything asynchronous — it only removes a safety check (blocking)

In Kubernetes 1.35, `defaultpreemption.PodEligibleToPreemptOthers` never consults `pl.fts`; the "terminating victim on the nominated node" check is unconditional. `EnableAsyncPreemption` gates exactly three things there: passing the flag to `preemption.NewEvaluator` (`default_preemption.go:106`), the `PreEnqueue` in-flight gate (`:138`), and the `{Pod/Delete, Skip}` hint in `EventsToRegister` (`:147`). Async preemption is "do the victim deletions in a goroutine and refuse to re-enter for the same preemptor"; it is not "relax the terminating-victim guard".

And in this PR's coscheduling half there is no async machinery at all:

```
git grep -n 'go func\|prepareCandidateAsync\|preempting' -- pkg/scheduler/plugins/coscheduling/    # nothing
git grep -n 'enableAsyncPreemption' -- pkg/scheduler/plugins/coscheduling/                          # 6 hits: field, ctor, assignment, :373, :398 log, :399
```

No goroutine, no `preempting` set, no `PreEnqueue`. So with `enableAsyncPreemption=true` the only observable effect in coscheduling is: a gang may preempt again while victims from its previous round are still terminating. Defaulted to `true` (see 5), that ships double-preemption of gangs to everyone.

This is the real answer to my 2026-05-11 question at `:399`. It wasn't just about the deleted comment — restructuring into `if !ev.enableAsyncPreemption` changed behavior. Please either drop both hunks and keep the guard unconditional (matching upstream), or take the behavior change out of this PR and justify it separately.

### 2. `:373` is dead code, and the test that appears to cover it cannot fail (blocking)

`podEligibleToPreemptOthers` has exactly one `return false`, and it is already behind `!ev.enableAsyncPreemption`. So `anyTerminatingPodOnNominatedNode` can only become true when the flag is off, which makes `&& !ev.enableAsyncPreemption` at `:373` an unreachable difference.

Mutation proof — I deleted `&& !ev.enableAsyncPreemption` from `:373` and re-ran the tree:

```
--- if anyTerminatingPodOnNominatedNode && !ev.enableAsyncPreemption {
+++ if anyTerminatingPodOnNominatedNode {
ok  .../coscheduling/core   2.181s      # all green, including Test_jobEligibleToPreemptOthers_Async
```

`Test_jobEligibleToPreemptOthers_Async` drives the real `podEligibleToPreemptOthers` rather than a stub, so all three of its cases are produced by `:399` alone; it re-tests `:399` one level up. For contrast, mutating `:399` (`if !ev.enableAsyncPreemption` → `if true`) does fail `Test_podEligibleToPreemptOthers_Async` — that site is genuinely covered.

### 3. On the reservation side the async contract is wired to nothing (blocking)

`preemption.go:111` builds the evaluator inside `PostFilter`:

```go
pe := preemption.NewEvaluator(Name, pm.fh, pm, pm.enableAsyncPreemption)
```

`ev.preempting` exists for exactly one purpose — the upstream comment says "In order to prevent requesting preemption of the same pod multiple times for the same preemptor" — and its only reader is `IsPodRunningPreemption`, which only `PreEnqueue` calls. A per-call evaluator allocates that set and throws it away, so:

- the dedup set never survives a scheduling cycle;
- upstream keeps `pl.Evaluator` as a field and reads it from `PreEnqueue`, so even the promoted `DefaultPreemption.PreEnqueue` would consult `pm.DefaultPreemption.Evaluator` — a *different* evaluator that never runs `Preempt`, so the gate could never fire;
- and it is unreachable regardless: `git grep 'PreEnqueue\|IsPodRunningPreemption' -- pkg/scheduler/plugins/reservation/` (non-test) returns nothing, and `Plugin` holds `preemptionMgr *PreemptionMgr` as a plain field rather than embedding it, so promoted methods are invisible to the framework.

Before the flag this was harmless: `NewEvaluator(..., false)` takes the synchronous `prepareCandidate` path and never touches instance state. This PR turns a harmless per-call construction into a load-bearing one.

To be fair about severity: `PodEligibleToPreemptOthers` is *not* overridden on `PreemptionMgr`, so it is still promoted from `*DefaultPreemption` and still unconditional. Once `nominatedNodeName` is persisted, repeat preemption is blocked there. The residual exposure is the window before the nomination is visible, plus the extra scheduling cycles you'd otherwise have saved with the in-flight gate. Wrong, not catastrophic — but it means the flag currently buys nothing on this side while adding risk.

Fix: construct the evaluator once in `newPreemptionMgr` and store it (the `pe.PodLister = newDelegatingPodLister(...)` assignment has to move accordingly — note it is per-pod today, so this needs care), then implement `PreEnqueue` on `Plugin` delegating to `IsPodRunningPreemption`. Current reservation-side coverage is two lines asserting the field got plumbed; behavior is untested.

### 4. The `assumedCycleStates` / `assumedNodeInfos` fix is unrelated, ungated, and half-covered (scope)

On `upstream/main` both maps are created (`:652-653`) and read (`:675`, `:684`, and again in `findFeasibleNodes`) but never written — so when several gang members land on the same node, each iteration re-clones and loses the accumulated `addPod` effect. That is a real bug and the fix looks right. It is also independent of `EnableAsyncPreemption`, ungated, and changes behavior for every user.

Re my 2026-05-11 "please also add UT for this" at `:688` — you answered "added in `preemption_async_test.go`". Per-line mutation says otherwise:

| mutation | result |
|---|---|
| delete `assumedCycleStates[selectedNode.Node().Name] = assumedCycleState` (`:688`) | **all green — uncovered** |
| delete `assumedNodeInfos[selectedNode.Node().Name] = assumedNodeInfo` (`:693`) | `Test_..._placeToSchedulePods_StatePersistence` FAILs |

So the exact line I asked about is still unguarded; only the `NodeInfo` half is. This is Copilot's `preemption_state_test.go:105` comment, confirmed by experiment rather than by reading. Please assert that the *same* `CycleState` pointer is reused for the same node across iterations.

### 5. Default is `true` while the doc comment says `false`

`defaults.go:53` sets `defaultEnableAsyncPreemption = ptr.To[bool](true)`, `defaults_test.go` asserts `true`, and both `apis/config/v1/types.go:354` and `apis/config/types.go:361` say `// default is false`. Copilot already flagged the comment; the deeper issue is the value. My ask: default `false` for one release, consistent with `EnablePreemption=false` and `EnableQueueHint=false` in the same file — and mandatory as long as 1 and 3 stand, since today "on" means "guard removed, nothing gained". Then make the comments match.

Nit: `SetDefaults_ReservationArgs` / `SetDefaults_CoschedulingArgs` alias the shared package-level pointer. #2927 touches this same file and introduces `ptr.To(*defaultEnableQueueHint)` with a comment explaining why; please follow that after rebase.

### Not a finding

Copilot's panic warning about `framework.NewNodeToStatus(nil, nil)` is wrong. `Status.Code()` is nil-safe — `kube-scheduler@v0.35.2 framework/interface.go:129-133` returns `Success` for a nil receiver. `Get` returning nil for an absent node yields `Success`, which is not `UnschedulableAndUnresolvable`, so the tests take the intended path. No change needed; don't spend time on it.

### On splitting by domain

I owed this PR an explicit answer. Production diff is 26 lines, so splitting reservation from coscheduling costs more than it saves — that is not the split worth doing. What must come out is 4. Suggested shape:

1. the `assumed{CycleState,NodeInfo}` bug fix, on its own, with an issue and a test that covers `:688`;
2. the reservation flag, with the evaluator hoisted to `newPreemptionMgr` and `PreEnqueue` implemented on `Plugin`;
3. coscheduling — only after we agree what the flag should mean there, since right now it means the opposite of upstream.

1 can go in quickly and independently.

### Merge ordering, measured

Correcting my earlier assumption that the generated files would collide: merging `refs/review/2900` into #2927-plus-main conflicts in exactly one file, `pkg/scheduler/apis/config/v1/defaults_test.go`, and it is an add/add — both PRs create it and it does not exist on `main`. `zz_generated.conversion.go` and `zz_generated.deepcopy.go` merge cleanly even though both PRs add fields to `ReservationArgs`. So #2927 → #2900 is a convenience, not a hard constraint; whoever lands second resolves a few lines of test file and re-runs codegen to be safe.

Holding so this doesn't merge on the strength of a green CI while 1, 3 and 5 are open. Happy to remove the hold as soon as the default flips and 4 is split out; 1 and 3 need a decision from us, not just from you.

/hold

vibhor-5:
Thanks for the detailed review @ZiMengSheng!

I have rebased the branch on latest `upstream/main` and updated the PR addressing all review comments:

1. **Defaulting (`EnableAsyncPreemption` set to `false`)**:
   - Updated `defaultEnableAsyncPreemption` to `false` for both `ReservationArgs` and `CoschedulingArgs` in `pkg/scheduler/apis/config/v1/defaults.go`.
   - Followed the #2927 pattern using `ptr.To(*defaultEnableAsyncPreemption)` to avoid pointer aliasing.
   - Updated doc comments in `apis/config/types.go` and `v1/types.go` and updated `defaults_test.go`.

2. **Reservation Evaluator & `PreEnqueue` Implementation**:
   - Hoisted `preemption.NewEvaluator` creation to `newPreemptionMgr` and stored `evaluator` on `PreemptionMgr` so in-flight preemption state (`ev.preempting`) persists across scheduling cycles.
   - Implemented `PreEnqueue` on `Plugin` (`_ fwktype.PreEnqueuePlugin = &Plugin{}`) and `PreemptionMgr`, delegating to `evaluator.IsPodRunningPreemption(p.UID)`.
   - Simplified `delegatingPodLister` to remove per-pod instance caching (`cachedPod`) and look up reserve pods directly in `Get`.
   - Added `TestPreEnqueue` unit test coverage in `preemption_test.go`.

3. **Coscheduling Preemption Safety Checks**:
   - Restored the unconditional terminating victim check (`ReasonTerminatingVictimOnNominatedNode`) in `podEligibleToPreemptOthers` and `jobEligibleToPreemptOthers` to prevent gang double-preemption.

4. **Splitting Unrelated Scope**:
   - Removed the `assumedCycleStates`/`assumedNodeInfos` dry-run state persistence changes and associated tests from this PR so it can be submitted in a separate targeted PR/issue.

PR #2905, opened 2026-05-09
fix/dead logging branch in setChild map existence check

### LINKED ISSUE #2904: [BUG]: Dead logging branch in setChild due to map existence check after assignment
**What happened**:
The logging logic inside setChild in gang.go checks whether a child exists in the map after inserting it into the map.
Current Implementation-
```
gang.Children[podId] = pod
if _, ok := gang.Children[podId]; !ok {
    klog.V(6).Infof("SetChild, ...")
} else {
    klog.V(6).Infof("UpdateChild, ...")
}
```
Since the entry is already inserted before the existence check, ok is always true.

As a result:

"SetChild" log branch is never executed
"UpdateChild" log is always emitted, even for newly added children

The same issue exists for PendingChildren.
The log output should correctly distinguish between:

newly inserted child entries (SetChild)
updated existing child entries (UpdateChild)

The same behavior should apply to PendingChildren.
Create a new Gang
Call setChild() with a pod that does not already exist in gang.Children
Observe logs

**Environment**:
- App version: current main
- Kubernetes version (use `kubectl version`): v1.36.0
  - OS version:24.02
  - Kernal version:6.x
  - Cgroup driver: systemd
- Others:

IITI-tushar:

This PR fixes a dead logging branch in setChild within pkg/scheduler/plugins/coscheduling/core/gang.go.
Since the key had already been inserted, the condition was always false. As a result:

* SetChild logs were never emitted
* UpdateChild logs were always emitted, even for newly added children

This PR changes the logic to check key existence before insertion so logs correctly distinguish between new child creation and child updates.

The same fix is also applied to PendingChildren.
fixes #2904 

Create a new gang and add a child pod that does not already exist in gang.Children
Observe logs from setChild
Verify that:
newly inserted children emit SetChild
existing children emit UpdateChild

Example verification:

first insertion → SetChild
subsequent insertion/update → UpdateChild

This PR only changes logging branch behavior and does not modify scheduling logic or gang state management behavior.
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/gang.go
+++ b/pkg/scheduler/plugins/coscheduling/core/gang.go
@@ -398,15 +398,18 @@ func (gang *Gang) setChild(pod *v1.Pod) {
 	defer gang.lock.Unlock()
 
 	podId := util.GetId(pod.Namespace, pod.Name)
+	_, existed := gang.Children[podId]
 	gang.Children[podId] = pod
-	if _, ok := gang.Children[podId]; !ok {
+
+	if !existed {
 		klog.V(6).Infof("SetChild, gangName: %v, childName: %v", gang.Name, podId)
 	} else {
 		klog.V(6).Infof("UpdateChild, gangName: %v, childName: %v", gang.Name, podId)
 	}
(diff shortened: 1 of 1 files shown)


ZiMengSheng (commented):
/lgtm
/approve

Issue #2803, opened 2026-02-04
[proposal] Provide a kind of JobNomination mechanism


ZiMengSheng:
**What is your proposal**:

Sometimes, when a job fails, we want to retain the job's resources in the scheduler and only restart some or all Pods. However, we don't want to explicitly create reservations, especially when a node has only one Pod. 

We hope Koordinator can provide a mechanism to reserve resources for jobs, prioritizing these resources when the Pod to which the job belongs is scheduled, and preventing other jobs from using those resources. JobNomination CR can be applied by the user as an enhancement to the job scheduling functionality.

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

Nesar976:
HI @ZiMengSheng Can i work on this issue

ZiMengSheng:
Filling in the sections I left empty when I opened this in February. @Nesar976 implemented this in #2907 and then waited over three months for a direction I never gave — that was my fault, and this comment is the answer he should have had in May. I'm not editing the original body so the history stays honest; treat this comment as the proposal.

## Motivation

When a pod of a distributed/batch Job fails, its resources become immediately allocatable to anyone. The replacement pod the Job controller creates then has to compete for them like a new arrival. Two things go wrong:

- **Locality loss.** A large training job that loses one worker can find that worker's node taken, and topology/affinity constraints then push the replacement somewhere that degrades the whole job — or make it unschedulable entirely while the other N-1 workers sit idle holding their own resources.
- **Partial-recovery inefficiency.** The job is alive and holding most of its allocation, but the one hole it needs to fill is the one thing it can't guarantee.

Users can already solve this by authoring a `Reservation` by hand before the pod fails, which is impractical — you'd have to predict the failure. So the gap is not the reservation *mechanism*, it's that nobody creates the reservation at the right moment.

I want to correct an ambiguity in my own original text. I wrote "we don't want to explicitly create reservations". I meant *users shouldn't have to author them*. It was read — reasonably — as "don't use the Reservation mechanism", and #2907 accordingly built a parallel in-memory ledger inside koord-scheduler. See my review there for why that costs us correctness (GPU/scalar resources unprotected, cross-namespace key collisions, double-subtraction against the scheduler cache, state lost on restart).

## Goals

1. On qualifying pod failure, automatically retain that pod's resources **on its node** so a replacement pod from the same Job can reclaim them.
2. Opt-in per Job. Never a cluster-wide default.
[...]

PR #2907, opened 2026-05-09
feat: add job nomination retention scheduler plugin

Nesar976:
### I. Describe what this PR does

This PR adds a lightweight JobNomination retention plugin for Koordinator scheduler.

It temporarily retains scheduling information for failed/restarting Job Pods and prioritizes replacement Pods on previously associated nodes when possible, while safely falling back to normal scheduling behavior.

### II. Does this pull request fix one issue?

Related to #2803

### III. Describe how to verify it

- Run unit tests
- Verify nomination retention and expiration behavior
- Verify fallback scheduling works correctly

### IV. Special notes for reviews

The implementation is intentionally minimal and avoids scheduler architecture changes.

### V. Checklist

- [x] Added tests
- [x] Added documentation
- [x] Relevant tests passed

### DIFF
--- a/cmd/koord-scheduler/main.go
+++ b/cmd/koord-scheduler/main.go
@@ -34,6 +34,7 @@ import (
 	noderesourcesfitplus "github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/noderesourcefitplus"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/reservation"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/scarceresourceavoidance"
+	"github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/jobnomination"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/schedulinghint"
 
 	// Ensure metric package is initialized
@@ -53,6 +54,7 @@ var koordinatorPlugins = map[string]frameworkruntime.PluginFactory{
 	noderesourcesfitplus.Name:    noderesourcesfitplus.New,
 	scarceresourceavoidance.Name: scarceresourceavoidance.New,
 	schedulinghint.Name:          schedulinghint.New,
+	jobnomination.Name:           jobnomination.New,
 }
 
 func flatten(plugins map[string]frameworkruntime.PluginFactory) []app.Option {
--- a/docs/jobnomination.md
+++ b/docs/jobnomination.md
@@ -1,0 +1,25 @@
+# JobNomination Resource Retention Mechanism
+
+## The Problem
+Currently, when a Pod from a distributed/batch Job fails, its resources immediately become available to other workloads. Because replacement Pods from the same Job may experience scheduling delays, they can lose their previously assigned nodes. This loss of stability means large distributed Jobs may lose locality, and partial Job recovery becomes inefficient.
+
+## The Solution: JobNomination Plugin
+The `JobNomination` scheduler plugin introduces a lightweight resource retention mechanism that allows failed Pods belonging to the same Job to regain scheduling priority on previously used resources, without introducing a full reservation framework.
+
+### Retention Flow
+1. **Detection:** The plugin watches Pod events. When a Pod belonging to a Job transitions to a `Failed` phase, the plugin records its requested resources and the node it was running on.
+2. **In-Memory Tracking:** The requested resources are tracked in an in-memory cache against the Job's UID and the Node's name. A bounded retention timeout (default: 2 minutes) is set for the nomination.
+3. **Protection (Filter phase):** During the `Filter` phase, the plugin calculates the total resources retained by *other* jobs on the node and deducts them from the node's available resources. This prevents unrelated jobs from immediately consuming those retained resources during the retention window.
+4. **Preference (Score phase):** During the `Score` phase, if a replacement Pod from the *same* Job is being scheduled, it receives a maximum score (100) for the node where its resources are retained.
+5. **Consumption (Reserve phase):** When a replacement Pod is successfully reserved on the nominated node, its corresponding nomination is removed from the cache, preventing the Job from double-dipping retained resources.
+
+### Fallback Behavior
+- **Retention Expiration:** The tracking cache uses an expiration mechanism. After the 2-minute window, expired nominations are lazily dropped or cleaned up by a background goroutine. The node's resources then become fully available to all workloads.
+- **Fairness:** The cluster will not be hard-locked globally because the retention window is strictly bounded.
+- **Conflict Handling:** If a retained resource cannot be fulfilled because of other factors, the normal scheduler behaviors take over.
+
+### Limitations of First Iteration
+1. **In-Memory Only:** The state is purely in-memory within the koord-scheduler process. If the scheduler restarts, the tracking state is lost.
+2. **Best-Effort Reservation:** The plugin currently implements retention by hiding resources from unrelated workloads. It does not enforce a rigid guarantee if the node's actual state drifts or if other pods were force-scheduled.
+3. **Simple Timeouts:** Expirations are globally defined at plugin initialization and are not yet configurable per Job.
+4. **Scope Control:** The current iteration does not involve multi-scheduler coordination or CRD-heavy persistent arbitration systems.
--- a/pkg/scheduler/plugins/jobnomination/plugin.go
+++ b/pkg/scheduler/plugins/jobnomination/plugin.go
@@ -1,0 +1,329 @@
+package jobnomination
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	resourceapi "k8s.io/component-helpers/resource"
+	fwktype "k8s.io/kube-scheduler/framework"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+
+	"github.com/koordinator-sh/koordinator/apis/extension"
+)
+
+const (
+	Name = "JobNomination"
+	// defaultRetentionDuration is the default time retained resources are kept
+	defaultRetentionDuration = 2 * time.Minute
+)
+
+// Nomination tracks retained resources for a failed pod on a specific node.
+type Nomination struct {
+	Resources  *framework.Resource
+	Expiration time.Time
+	PodUID     types.UID
+}
+
+// JobNomination is a plugin that retains resources for failed Job pods.
+type JobNomination struct {
+	handle            fwktype.Handle
+	retentionDuration time.Duration
+
+	// lock protects the cache
+	lock sync.RWMutex
+	// cache maps JobID -> NodeName -> list of Nominations (since multiple pods of same job can fail on same node)
+	cache map[string]map[string][]*Nomination
+}
+
+var _ fwktype.PreFilterPlugin = &JobNomination{}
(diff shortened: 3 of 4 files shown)


ZiMengSheng:
@Nesar976  Please fix the UT

cmd/koord-scheduler/main.go
```diff
@@ -53,6 +54,7 @@ var koordinatorPlugins = map[string]frameworkruntime.PluginFactory{
 	noderesourcesfitplus.Name:    noderesourcesfitplus.New,
 	scarceresourceavoidance.Name: scarceresourceavoidance.New,
 	schedulinghint.Name:          schedulinghint.New,
+	jobnomination.Name:           jobnomination.New,
```
ZiMengSheng, line 57: please fmt the doc

docs/jobnomination.md
```diff
@@ -0,0 +1,25 @@
+# JobNomination Resource Retention Mechanism
+
+## The Problem
+Currently, when a Pod from a distributed/batch Job fails, its resources immediately become available to other workloads. Because replacement Pods from the same Job may experience scheduling delays, they can lose their previously assigned nodes. This loss of stability means large distributed Jobs may lose locality, and partial Job recovery becomes inefficient.
+
+## The Solution: JobNomination Plugin
+The `JobNomination` scheduler plugin introduces a lightweight resource retention mechanism that allows failed Pods belonging to the same Job to regain scheduling priority on previously used resources, without introducing a full reservation framework.
+
+### Retention Flow
```
ZiMengSheng, line 9: Not every Job needs this — there should be a toggle to enable or disable reservation explicitly.

saintube (commented):
@Nesar976 Please an user manual to the [website](https://github.com/koordinator-sh/website) after this PR merged.
See https://koordinator.sh/docs/user-manuals/gang-scheduling

ZiMengSheng:
Thanks for the patience here, @Nesar976 — and sorry. You picked this up in May, I left two one-line comments on 2026-05-26 and then went quiet for over three months. Worse, the proposal I assigned you (#2803) is three sentences long with the "Why is this needed" and "Is there a suggested solution" sections left empty. You had to invent the entire design yourself, and the direction question below is one you had no way to answer without me. That's on me, not on you.

Let me pay that debt now: first the direction, then the concrete defects I verified.

## 1. Direction: this should be built on `Reservation`, not on a new in-memory ledger

When I wrote #2803 I said "we don't want to explicitly create reservations". I meant *the user shouldn't have to author them by hand* — I did not mean we should avoid the `Reservation` mechanism itself. That was ambiguous and this PR reasonably read it the other way.

The problem is that `Reservation` already provides every single thing this plugin hand-rolls, and provides it better:

| what `plugin.go` does by hand | `ReservationSpec` field | why it's better |
|---|---|---|
| `cache[jobID][nodeName]` | `Template.Spec.NodeName` | pins to the original node, persisted in etcd |
| `getJobID()` (bare gang name / owner UID) | `Owners[].Controller` (`ReservationControllerReference` — carries a `Namespace` field) or `Owners[].LabelSelector` | namespace-scoped, so no cross-tenant collision |
| hardcoded `defaultRetentionDuration = 2m` | `TTL` / `Expires` | per-job configurable — your doc lists this as limitation #3 |
| `Reserve()` consuming one nomination | `AllocateOnce` (defaults true) | already correct, already tested |
| `Nomination.Resources` (MilliCPU/Memory only) | `Template.Spec.Containers[].Resources` | full resource vector including GPU/RDMA — see defect 3 |
| `Filter()` subtracting `retainedByOthers` | the reservation plugin's existing accounting | consistent with NodeInfo, preemption, and quota — see defect 4 |
| — | etcd-backed | your doc's limitation #1 (state lost on scheduler restart) disappears |

And we already have a working precedent for exactly this shape in-tree: `PodMigrationJob` reserves a failed/evicted pod's resources by *automatically creating a Reservation on the controller's behalf*. See `pkg/descheduler/controllers/migration/reservation/util.go:42-103` — it copies the pod's ObjectMeta+Spec into `Template`, calls `GenerateReserveResourceOwners(pod)`, forces `AllocateOnce=true`, and sets `TTL`. The only line that differs for our case is that migration deliberately clears `NodeName` (it wants a *different* node); JobNomination wants to keep it.

So my proposed shape is:

- an opt-in marker on the Job/PodGroup (this is also the answer to my own 2026-05-26 comment "there should be a toggle to enable or disable reservation explicitly" — it must be opt-in, never cluster-wide default),
- a controller that, on qualifying pod failure, creates a `Reservation` with `Template.Spec.NodeName` = the failed pod's node, `Owners[].Controller` = the pod's controller ownerRef *with namespace*, `TTL` = the configured window, `AllocateOnce=true`,
- **zero** new scheduler-side filter/score logic — the existing `Reservation` plugin already prefers and protects reserved resources.

That deletes roughly all 329 lines of `plugin.go` and makes the feature strictly stronger. I'd like to settle this on #2803 before you write more code — I'll post the filled-in design there and tag you.

## 2. Defects I verified in the current implementation

I ran these as probes against your branch (`ae3f83d5`) to be sure I wasn't guessing. Recording them because several are instructive regardless of which direction we take.

**2.1 (blocking) `deletePod` retains resources for pods that never failed.** `deletePod` → `trackPodRetention` unconditionally (`plugin.go:115`); `isPodFailed` is only consulted on the *update* path. So a Succeeded pod cleaned up after a normal Job completion, a Running pod removed by a node drain, and a Pending pod removed by a scale-down all create nominations. Probe:

```
P1 phases tracked: Succeeded=1 Running=1 Pending=1 (total jobs in cache=3)
P1 drained node-1 (16 cpu, 12 retained by 3 healthy deletions): unrelated 8-cpu pod -> Unschedulable "Node resources retained by other jobs"
```

Draining a node therefore makes that node partially unschedulable to *every* unrelated workload for 2 minutes. `docs/jobnomination.md` says "When a Pod belonging to a Job transitions to a `Failed` phase" — the doc describes the intended behaviour, the code does something much broader.

**2.2 (blocking) `getJobID` collides across namespaces.** `extension.GetGangName` (`apis/extension/coscheduling.go:98-108`) returns the *bare* gang name with no namespace — coscheduling itself always wraps it in `util.GetId(ns, name)`. Probe:

```
P2 getJobID(team-a/trainer-0)="trainer"  getJobID(team-b/trainer-0)="trainer"
P2 cache key "trainer" holds 2 nomination(s) on node-1
P2 team-b replacement Score on node-1 = 100 (reuses team-a's retention)
```

Two tenants that both name their PodGroup `trainer` share one retention pool, and team-b's pods get score 100 on nodes retained for team-a. Separately, the two branches of `getJobID` return different kinds of key (a name vs. an owner UID) into the same map, so their key spaces aren't comparable.

**2.3 (blocking) scalar resources are recorded but never protected.** `Filter` only accumulates and compares `MilliCPU`/`Memory`/`EphemeralStorage`/`AllowedPodNumber` (`plugin.go:222-243`) and never touches `Resource.ScalarResources`. Probe:

```
P3 nomination recorded: MilliCPU=1000 Memory=0 ScalarResources=map[nvidia.com/gpu:8]
P3 unrelated 8-GPU pod against the 8 retained GPUs -> Success ""
```

The GPUs are in the nomination and are simply ignored, so an unrelated job takes all 8. Since #2803 is motivated by distributed training, this means the feature does not work for its primary use case.

**2.4 (blocking) double subtraction against the scheduler cache.** `Filter` computes `allocatable - requested - retainedByOthers`. But a Failed-and-not-yet-deleted pod is still in the scheduler cache: `assignedPod` is just `len(pod.Spec.NodeName) != 0` (`k8s.io/kubernetes/pkg/scheduler/eventhandlers.go:469-471`, i.e. in the vendored upstream scheduler, not this repo) with no phase filter, and `NodeInfo.AddPod` doesn't filter by phase either. So during the whole window between "pod failed" and "pod deleted" — which for training operators that keep failed pods around for log inspection is the *normal* state, not an edge case — the same resources are subtracted twice. Your unit test constructs `nodeInfo` with `requested = 0`, which is exactly why it doesn't catch this.

**2.5 (blocking) `Reserve` consumes, `Unreserve` is empty.** `Reserve` deletes the nomination (`plugin.go:307-319`) but `Unreserve` is a no-op (`plugin.go:328`). If any later stage (PreBind/Bind) fails, the retention is gone permanently and the replacement pod loses its preference on retry — i.e. it fails open precisely on the path where it's most needed. Also, consumption is independent of the pod's actual request: a 1m-CPU pod can consume a nomination holding 8 GPUs.

**2.6 (blocking) `Filter` is O(all jobs) per node per cycle.** The loop at `plugin.go:212` walks the entire cache for every candidate node while holding `RLock`. On a 5k-node cluster with a few thousand tracked jobs that's a full cache scan 5k times per scheduling cycle. The cache wants `nodeName` as the primary key, or a snapshot taken once in `PreFilter`.

**2.7 (blocking) registering an empty `PreFilter` will silently disable batching.** `PreFilter` returns Success and does nothing, `PreFilterExtensions` returns nil (`plugin.go:194-200`) — pure overhead. It's also actively harmful: with KEP-5598 OpportunisticBatching now Beta/default-on in 1.35, `computeBatchablePlugins` (`pkg/scheduler/framework/runtime/framework.go:759-802`) sets `enableSignatures = false` for the **whole profile** if *any* preFilter/filter/preScore/score plugin doesn't implement `fwk.SignPlugin`. This plugin implements PreFilter, Filter *and* Score without `SignPlugin`, so enabling it in a profile turns off signature caching for every other plugin too. See #2863, which is adding `SignPod` to six plugins. Since `PreFilter` does nothing, just drop both methods and the `PreFilterPlugin` assertion — but `Filter` and `Score` still need `SignPod` (or a conscious decision to accept the loss).

**2.8 (blocking) `New` discards its context.** `New(_ context.Context, ...)` ignores the ctx and starts `go jn.cleanupRoutine(context.Background())` (`plugin.go:62`), so the goroutine outlives the scheduler and leaks once per plugin instantiation (multi-profile setups instantiate more than once). Use the ctx you were handed.

## 3. Still open from May

- `cmd/koord-scheduler/main.go:37` — my "please fmt the doc" comment: `jobnomination` is inserted after `scarceresourceavoidance`, breaking the import block's sort order (`j` sorts before `n`). Not yet fixed.
- My "there should be a toggle" comment — addressed by the opt-in marker in section 1.
- @saintube asked for a website user manual. Please hold that until the direction is settled, otherwise you'll write it twice.

## 4. Non-blocking

- `docs/jobnomination.md` is in the wrong place. Designs go to `docs/proposals/scheduling/YYYYMMDD-<name>.md` and follow `docs/proposals/YYYYMMDD-template.md` (Motivation / Goals / Non-Goals / API / Risks / Alternatives / Test Plan). The content you wrote is good — the "Limitations of First Iteration" section is honest and useful — it just needs to live in the right file with the right sections.
- Use `Resource.Add()` instead of hand-summing four fields; and accumulating `AllowedPodNumber` from pod requests is always 0, so it's dead arithmetic.
- The branch is 81 commits behind main and `mergeStateStatus` is `DIRTY` — it needs a rebase before CI is meaningful.
- The plugin isn't added to `config/manager/scheduler-config.yaml`, so it's off by default. That's the right call for now; please say so explicitly in the doc.
- `isPodFailed` only matches `PodFailed`. Worth stating in the doc whether container-level restarts and operator-initiated deletions are in scope — right now 2.1 catches them by accident rather than by design.

## 5. Sequencing

1. Settle the direction on #2803 (my action — I'll post the filled-in design and tag you).
2. Land the design doc under `docs/proposals/scheduling/`.
3. Then implement, most likely as a controller rather than a scheduler plugin.

Holding this so it can't merge while the direction is unsettled — this is not a reflection on your work. If we land on "keep the in-memory plugin" after all, 2.1–2.8 above are the concrete fix list and I'll review promptly. Thanks again for your patience, and sorry for the delay.

/hold

saintube:
@ZiMengSheng

PR #2911, opened 2026-05-10
fix: stale gang cache by handling informer tombstones in delete handlers

### LINKED ISSUE #2910: [BUG]: Handle DeletedFinalStateUnknown tombstones in coscheduling gang cache delete handlers
**What happened**:

When the Kubernetes informer's watch connection is interrupted and later re-synced, it may deliver missed delete events wrapped in a `cache.DeletedFinalStateUnknown` tombstone instead of the raw object. Both `onPodDelete` and `onPodGroupDelete` in the coscheduling gang cache perform a bare type-assertion `(obj.(*v1.Pod)` / `obj.(*v1alpha1.PodGroup))` that fails on tombstones, causing them to silently return without cleaning up the gang cache. As a result, pods and pod groups that were deleted during a watch gap remain in the cache indefinitely, leading to a stale gang cache.

Gang cache entries for deleted Pods and PodGroups should be cleaned up correctly even when the informer delivers a tombstone instead of a direct object (e.g., after a watch re-sync following a temporary API server disruption).

* Create a Gang (via annotation or PodGroup CRD) with at least one Pod.
* While the scheduler is running, briefly disrupt connectivity to the API server.
* Delete the Pod (or PodGroup) during the disruption window.
* After the scheduler reconnects, the gang cache still holds the deleted Pod/PodGroup — onPodDelete / onPodGroupDelete was called with a DeletedFinalStateUnknown tombstone, the type assertion failed silently, and no cleanup occurred.

Every other delete event handler in the same repo `(elasticquota, nodenumaresource, deviceshare, reservation) `already handles cache.`DeletedFinalStateUnknown` correctly. The fix follows the same pattern used throughout the codebase.
**Environment**:
- App version:main
  - OS version:24.02
  - Kernal version:
  - Cgroup driver: systemd
- Others:

IITI-tushar:
Fixes missing `cache.DeletedFinalStateUnknown` tombstone handling in `onPodDelete` and `onPodGroupDelete` inside the coscheduling gang cache (`gang_cache.go`)

When the Kubernetes informer misses a delete event (e.g., due to a watch timeout or API-server blip), it delivers a `DeletedFinalStateUnknown` tombstone on `resync`. The previous bare type-assertions silently returned on mismatch, leaving deleted Pods and PodGroups forever in the gang cache. This PR adds the standard tombstone-unwrap pattern already used by `elasticquota, nodenumaresource, deviceshare, and reservation delete handlers.`

fixes #2910 

* Run go test ./pkg/scheduler/plugins/coscheduling/... — all existing tests pass.
* Six new unit tests cover the three cases for each handler:
    *  Valid tombstone → cleanup executes correctly.
    * Tombstone wrapping a wrong type → event is ignored, no panic.
    * Completely unknown object type → event is ignored, no panic.
* Only `gang_cache.go and gang_cache_test.go ` are changed.
* `"k8s.io/client-go/tools/cache"` is the only new import — it is already a transitive dependency of the module.
* The fix mirrors the pattern in `elasticquota/pod_handler.go`exactly.
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/gang_cache.go
+++ b/pkg/scheduler/plugins/coscheduling/core/gang_cache.go
@@ -22,6 +22,7 @@ import (
 	"github.com/go-logr/logr"
 	v1 "k8s.io/api/core/v1"
 	listerv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	fwktype "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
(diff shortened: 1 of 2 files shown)


ZiMengSheng (approved):
/lgtm
/approve

PR #2912, opened 2026-05-10
descheduler: adjust DetectorCacheTimeout when less than DeschedulingInterval

### LINKED ISSUE #2626: [BUG] koord-descheduler evict failed if DeschedulingInterval and DetectorCacheTimeout mismatch
**What happened**:
DeschedulingInterval is allowed to configured, but if it is set larger than DetectorCacheTimeout, nodeAnomalyDetectors and prodAnomalyDetectors will expire, then no nodes will be abnormal. 

We might need a default DetectorCacheTimeout to be larger than DeschedulingInterval. Or some way to check and limit it. 

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

madmecodes:
## Summary

- When `DeschedulingInterval` exceeds `DetectorCacheTimeout`, the anomaly detector cache entries (in `nodeAnomalyDetectors` and `prodAnomalyDetectors`) expire between descheduling cycles. This means the detectors never accumulate enough consecutive abnormalities to trigger eviction, so evictions silently stop working.
- This fix adds a check during descheduler setup that auto-adjusts `DetectorCacheTimeout` to match `DeschedulingInterval` when the timeout is smaller, and logs a message so operators know it happened.

Fixes #2626

## Test plan

- [x] Existing unit tests pass (`go test ./pkg/descheduler/...`)
- [x] Build passes (`go build ./cmd/koord-descheduler/...`)
- [ ] Manual test: set `DeschedulingInterval` to 10m and `DetectorCacheTimeout` to 5m, confirm log message appears and evictions still work

### DIFF
--- a/cmd/koord-descheduler/app/server.go
+++ b/cmd/koord-descheduler/app/server.go
@@ -316,6 +316,28 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions 
 
 	completedProfiles := make([]deschedulerconfig.DeschedulerProfile, 0)
 
+	// Ensure DetectorCacheTimeout >= DeschedulingInterval for LowNodeLoad plugins.
+	// If DetectorCacheTimeout is smaller than DeschedulingInterval, the anomaly
+	// detector cache entries expire between descheduling cycles, causing evictions
+	// to silently stop working because the detector never accumulates enough
+	// consecutive abnormalities.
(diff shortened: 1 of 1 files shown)


ZiMengSheng:
@songtao98 PTAL

songtao98:
good job!
/lgtm

PR #2913, opened 2026-05-10
refactor(scheduler): make reservation Bind non-blocking

### LINKED ISSUE #2853: Non-blocking API Calls
Non-blocking API Calls Adaptation (Informer / Client)

k8s 1.35 promotes non-blocking API calls in scheduler plugins (Bind, Reserve, etc.) to avoid blocking the scheduling goroutine. We should identify and refactor the scheduler plugins in Koordinator that perform synchronous API calls in Bind/Reserve/PostBind phases.

xenonnn4w:
# PR Description

## I. Describe what this PR does

- Makes reservation `Bind()` non-blocking by moving Reservation status updates to an async worker, avoiding scheduler goroutine blocking.
- Adjusts reservation bind tests to tolerate async status updates.

## II. Does this pull request fix one issue?

* fixes #2853

## III. Describe how to verify it

```bash
go test ./pkg/scheduler/plugins/reservation -run TestBind -count=1
```

## IV. Special notes for reviews

- Reservation status updates are now async with retry and timeout; `Bind()` no longer blocks on API calls.

## V. Checklist

- [x] All checks passed in make test

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -1573,39 +1574,53 @@ func (pl *Plugin) Bind(ctx context.Context, cycleState fwktype.CycleState, pod *
...
-		if err = reservationutil.SetReservationAvailable(reservation, nodeName); err != nil {
-			return err
-		}
-		_, err = pl.client.Reservations().UpdateStatus(context.TODO(), reservation, metav1.UpdateOptions{})
-		if err != nil {
-			klog.V(4).ErrorS(err, "failed to update reservation", "reservation", klog.KObj(reservation))
-		}
-		return err
-	})
+	reservation, err := pl.rLister.Get(rName)
```
ZiMengSheng, line 1577: We may don't need that cause BindCycle is already in a separate goroutine.

xenonnn4w: oh i see, so we have two options now:
1. Drop this PR; sync is fine since the bind cycle is already concurrent across pods, and revisit when we adopt the upstream AsyncAPICall API.
2. Keep this PR open as a placeholder and switch the implementation to AsyncAPICall once koord-scheduler exposes it.

PR #2914, opened 2026-05-10
fix: rename PageCacheCollector to UsageWithCacheCollector

### LINKED ISSUE #2508: [BUG] pageCacheCollector 模块采集 PageCache 错误
**What happened**:

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

// MemWithPageCacheUsageBytes returns the usage of mem with page cache bytes.
func (i *MemInfo) MemUsageWithPageCache() uint64 {
	// total - free
	return (i.MemTotal - i.MemFree) * 1024
}

func (m *MemoryStatRaw) UsageWithPageCache() int64 {
	// memory.stat usage: total_inactive_anon + total_active_anon + total_unevictable + total_activefile + total_inactivefile
	return m.InactiveAnon + m.ActiveAnon + m.Unevictable + m.ActiveFile + m.InactiveFile
}

page cache 不应该是这个， 这里面包含了。匿名页 + 文件页 + slab 页 等等，

madmecodes:
## Summary

- Renamed `PageCacheCollector` to `UsageWithCacheCollector` since it actually collects memory usage *including* page cache, not page cache itself
- Renamed config field `EnablePageCacheCollector` -> `EnableUsageWithCacheCollector` and CLI flag `--enable-pagecache-collector` -> `--enable-usage-with-cache-collector`
- Updated internal struct/method names and log messages to match the new naming
- Fixed misleading `MemUsageWithPageCache` doc comment in `meminfo.go`

The collector was named `PageCacheCollector` but it collects `total - free` at the node level and `inactive_anon + active_anon + unevictable + active_file + inactive_file` at the cgroup level, which is memory usage with cache included -- not page cache alone. The maintainer confirmed the rename in the issue discussion.

## Test plan

- [x] `go build ./pkg/koordlet/...` passes
- [x] `go test ./pkg/koordlet/metricsadvisor/collectors/pagecache/...` passes
- [x] `go test ./pkg/koordlet/metricsadvisor/framework/...` passes

Fixes #2508

### DIFF
--- a/pkg/koordlet/metricsadvisor/collectors/pagecache/page_cache_collector.go
+++ b/pkg/koordlet/metricsadvisor/collectors/pagecache/page_cache_collector.go
@@ -34,22 +34,22 @@ import (
 )
 
 const (
-	CollectorName = "PageCacheCollector"
+	CollectorName = "UsageWithCacheCollector"
 )
 
 var (
 	timeNow = time.Now
 )
 
-type pageCacheCollector struct {
-	collectInterval       time.Duration
-	started               *atomic.Bool
-	appendableDB          metriccache.Appendable
-	metricDB              metriccache.MetricCache
-	statesInformer        statesinformer.StatesInformer
-	cgroupReader          resourceexecutor.CgroupReader
-	podFilter             framework.PodFilter
-	coldPageCollectorGate bool
+type usageWithCacheCollector struct {
+	collectInterval             time.Duration
+	started                     *atomic.Bool
+	appendableDB                metriccache.Appendable
+	metricDB                    metriccache.MetricCache
+	statesInformer              statesinformer.StatesInformer
+	cgroupReader                resourceexecutor.CgroupReader
+	podFilter                   framework.PodFilter
+	usageWithCacheCollectorGate bool
 }
 
 func New(opt *framework.Options) framework.Collector {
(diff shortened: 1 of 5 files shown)


ZiMengSheng:
### The rename is correct, but the flag change will break existing deployments

I agree with the premise. `PageCacheCollector` genuinely is misnamed — it collects memory usage *including* page cache, not page cache itself. @pingxin403's #3132 independently confirms this reading of the semantics (it deals with `NodeMemoryCollectPolicy` being `UsageWithPageCache` / `UsageWithHotPageCache`). So the direction is right and worth doing.

**Blocking problem:** you replace the CLI flag outright rather than deprecating it.

```go
- fs.BoolVar(&c.EnablePageCacheCollector, "enable-pagecache-collector", ...)
+ fs.BoolVar(&c.EnableUsageWithCacheCollector, "enable-usage-with-cache-collector", ...)
```

koordlet runs as a DaemonSet. Any existing deployment that passes `--enable-pagecache-collector` in its args will fail to start with `unknown flag` after upgrading — **koordlet CrashLoopBackOff across every node in the cluster**, with no way to recover except editing the DaemonSet. That's not an acceptable upgrade experience for a cosmetic rename.

Please keep the old flag working for at least one release:

```go
fs.BoolVar(&c.EnableUsageWithCacheCollector, "enable-usage-with-cache-collector", c.EnableUsageWithCacheCollector, "...")
fs.BoolVar(&c.EnableUsageWithCacheCollector, "enable-pagecache-collector", c.EnableUsageWithCacheCollector, "...")
fs.MarkDeprecated("enable-pagecache-collector", "use --enable-usage-with-cache-collector instead")
```

Both flags then write the same field, old deployments keep working, and users get a deprecation warning in the logs telling them what to change.

The Go-level renames (`PageCacheCollector` → `UsageWithCacheCollector`, the config field, internal strings) are fine as-is — those are internal and `git grep` shows `EnablePageCacheCollector` is only referenced within `metricsadvisor`, so the blast radius is contained.

Two smaller notes:

- Consider whether the package directory `collectors/pagecache/` should be renamed too, or deliberately left alone. Either is defensible; just be intentional, since a half-renamed concept is more confusing than either extreme.
- I checked `main` for other references to the old flag string outside Go code (helm chart, docs) and found none, so the deprecation shim is the only compatibility work needed.

**On overlap:** I cross-checked this against #3132 since both touch page-cache-related metrics. They do **not** conflict — you're in `collectors/pagecache/` + `framework/config.go`, he's in `collectors/sysresource/`. No file overlap, and the two can proceed in parallel. If anything they're complementary: his PR is the concrete evidence that the old name was misleading.

PR #2917, opened 2026-05-11
feat: support DRA in DeviceShare and update informers

### LINKED ISSUE #2855: DRA
Kubernetest 1.35 graduates DRA to GA, with a significantly redesigned API (resource.k8s.io/v1). The key objects are ResourceClaim, ResourceSlice, and DeviceClass.

The Koordinator should be compatible if the DRA plugin and related feature gates are enabled. However, there is some overlap between the DRA and the Koordinator DeviceShare to assess, and the node-side of the DRA is currently unsupported by the koordlet and the koord-device-daemon.

vibhor-5:
Kubernetes 1.35 graduates DRA to GA with the resource.k8s.io/v1 API.
- Update descheduler informers to use resource.k8s.io/v1 for DRA resources
- Skip DeviceShare plugin logic if Pod uses DRA (ResourceClaims) to prevent conflicts
- Add RBAC permissions for resource.k8s.io/v1 to manager and koordlet
- Add unit tests for DeviceShare DRA skip logic

Closes #2855

### DIFF
--- a/config/rbac/koordlet.yaml
+++ b/config/rbac/koordlet.yaml
@@ -54,6 +54,16 @@ rules:
     - events
   verbs:
     - '*'
+- apiGroups:
+    - resource.k8s.io
+  resources:
+    - resourceclaims
+    - resourceslices
+    - deviceclasses
+  verbs:
+    - get
+    - list
+    - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -165,3 +165,26 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - resource.k8s.io
+  resources:
+  - resourceclaims
+  - resourceclaimtemplates
+  - resourceslices
+  - deviceclasses
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - delete
+  - patch
+  - update
+- apiGroups:
+  - resource.k8s.io
+  resources:
+  - resourceclaims/status
+  verbs:
+  - get
+  - patch
+  - update
--- a/pkg/descheduler/informers/generic.go
+++ b/pkg/descheduler/informers/generic.go
@@ -52,6 +52,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	rbacv1alpha1 "k8s.io/api/rbac/v1alpha1"
 	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
+	resourcev1 "k8s.io/api/resource/v1"
 	resourcev1beta1 "k8s.io/api/resource/v1beta1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	schedulingv1alpha1 "k8s.io/api/scheduling/v1alpha1"
@@ -319,6 +320,16 @@ func (f *sharedInformerFactory) ForResource(resource schema.GroupVersionResource
 		return &genericInformer{resource: resource.GroupResource(), informer: f.Rbac().V1beta1().Roles().Informer()}, nil
 	case rbacv1beta1.SchemeGroupVersion.WithResource("rolebindings"):
 		return &genericInformer{resource: resource.GroupResource(), informer: f.Rbac().V1beta1().RoleBindings().Informer()}, nil
+
+		// Group=resource.k8s.io, Version=v1
+	case resourcev1.SchemeGroupVersion.WithResource("deviceclasses"):
+		return &genericInformer{resource: resource.GroupResource(), informer: f.Resource().V1().DeviceClasses().Informer()}, nil
+	case resourcev1.SchemeGroupVersion.WithResource("resourceclaims"):
+		return &genericInformer{resource: resource.GroupResource(), informer: f.Resource().V1().ResourceClaims().Informer()}, nil
+	case resourcev1.SchemeGroupVersion.WithResource("resourceclaimtemplates"):
+		return &genericInformer{resource: resource.GroupResource(), informer: f.Resource().V1().ResourceClaimTemplates().Informer()}, nil
+	case resourcev1.SchemeGroupVersion.WithResource("resourceslices"):
+		return &genericInformer{resource: resource.GroupResource(), informer: f.Resource().V1().ResourceSlices().Informer()}, nil
 
 		// Group=resource.k8s.io, Version=v1beta1
 	case resourcev1beta1.SchemeGroupVersion.WithResource("deviceclasses"):
--- a/pkg/scheduler/plugins/deviceshare/utils.go
+++ b/pkg/scheduler/plugins/deviceshare/utils.go
@@ -360,6 +360,10 @@ func preparePod(pod *corev1.Pod, gpuSharedResourceTemplatesCache *gpuSharedResou
 		preemptibleInRRs:   map[string]map[types.UID]map[schedulingv1alpha1.DeviceType]deviceResources{},
 	}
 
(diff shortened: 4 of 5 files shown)


pkg/scheduler/plugins/coscheduling/core/preemption.go
```diff
@@ -76,9 +76,10 @@ type preemptionEvaluatorImpl struct {
 	gangCache             *GangCache
 	gangContextHolder     *GangSchedulingContextHolder
 	networkTopologySolver NetworkTopologySolver
+	enableAsyncPreemption bool
```
ZiMengSheng: don't put two things in one PR

vibhor-5: My bad that was a mistake.
only related changes now in the PR.

ZiMengSheng:
Pre-review found a **blocking bug** in the DRA skip path:

```go
// utils.go preparePod():
info := &podInfo{...}          // info constructed
if len(pod.Spec.ResourceClaims) > 0 {
    return                     // ← returns (nil, nil)
}
```

`preparePod` returns `(state *preFilterState, status *fwktype.Status)`. For a DRA pod this returns `(nil, nil)`. In the Kubernetes scheduling framework a **nil Status is treated as Success**, so the caller in `PreFilter` (plugin.go:196) passes the `!status.IsSuccess()` check and then dereferences the nil state at plugin.go:206 (`state.designatedAllocation = nil` when no scheduling hint) — **nil pointer panic, takes down koord-scheduler on the first DRA pod**.

The added test only calls `preparePod` directly (asserting nil/nil), which is exactly the blind spot — it never exercises the PreFilter path.

Suggested fix — return a success status with an empty state so the plugin proceeds as a no-op for DRA pods:

```go
if len(pod.Spec.ResourceClaims) > 0 {
    return &preFilterState{skip: true, ...}, nil   // or framework.NewStatus(framework.Skip)
}
```

…matching how the existing `state.skip` zero-request path flows through PreFilter/Filter. Also worth moving the check to the top of the function before constructing `info`, and adding a PreFilter-level test with a ResourceClaim pod to lock the contract.

The informer/RBAC parts look fine. Pinging @vibhor-5 for the fix.

ZiMengSheng:
@vibhor-5 I need to retract my previous comment. **The blocking bug I reported yesterday does not exist** — I misread your code, and my "suggested fix" was already exactly what you had written. Sorry for the noise; let me correct it precisely and then give you the review you should have gotten.

## Retraction

I claimed `preparePod` returns `(nil, nil)` for a DRA pod and that `PreFilter` would then panic. Three things in that comment were wrong:

1. I quoted the code as `info := &podInfo{...}`. There is no `info` variable and no `podInfo` type in that function — I fabricated it. The actual code assigns to the **named return value**: the signature is `func preparePod(...) (state *preFilterState, status *fwktype.Status)` and the body does `state = &preFilterState{skip: true, ...}` (`utils.go:356-360`).
2. Because `state` is the named return, the bare `return` at `utils.go:364` returns `(&preFilterState{skip: true, ...}, nil)` — **not** `(nil, nil)`. There is nothing to dereference nil.
3. I said your test "asserts nil/nil". It asserts `state.skip` — which would itself panic if `state` were nil. The test passing was already evidence against my claim, and I read past it.

Verified rather than re-reasoned, this time. I ran a probe through the full `PreFilter` path on your head `9147e8a9`:

```
PROBE preparePod: state==nil? false  status==nil? true  state.skip=true  status.IsSuccess()=true
PROBE PreFilter returned: result=<nil> code=Skip (Skip=true) -> no panic
```

So a DRA pod flows: `preparePod` → `skip=true` → `PreFilter` sets `state.designatedAllocation = nil` on a **valid** state → `cycleState.Write` → `return nil, NewStatus(Skip)`. DeviceShare cleanly no-ops for DRA pods, which is the correct behaviour, reached via the existing `state.skip` path. That is precisely the design I "suggested" you adopt. You'd already done it. My apologies.

## The real problem: `Check manifests` is failing, and I said RBAC "looks fine"

I also wrote "The informer/RBAC parts look fine." That was wrong in the opposite direction — `unit-tests(Check manifests)` is **red** right now, and RBAC is why.

`config/rbac/role.yaml` is a **generated** file. `make manifests` runs `controller-gen rbac:roleName=manager-role crd webhook paths="./..."` (`Makefile:67-68`), and `role.yaml` is the `manager-role` it emits. There is no `// +kubebuilder:rbac:groups=resource.k8s.io,...` marker anywhere in the tree (I grepped: zero hits), so controller-gen doesn't know about your rules and strips them. Reproduced locally:

```
$ make manifests && diff role.yaml.before config/rbac/role.yaml
168,190d167
< - apiGroups:
<   - resource.k8s.io
<   resources:
<   - resourceclaims
...   (all 23 lines you added, deleted)
```

That's the CI failure. But fixing it by adding a marker would be wrong too, because there are two deeper issues underneath:

**The rules are on the wrong component.** `role.yaml` is `manager-role` — that's **koord-manager**, which doesn't touch DRA at all. The components that do have their own hand-written (non-generated) roles: `config/rbac/scheduler_role.yaml` and `config/rbac/descheduler_role.yaml`.

**And for the scheduler they're already there.** `scheduler_role.yaml:66-86` already grants exactly what you added — `resourceclaims`, `resourceclaimtemplates`, `resourceslices`, `deviceclasses` (get/list/watch) plus `resourceclaims/status`. So the DeviceShare half of this PR needs no RBAC change whatsoever.

**What may actually be missing is the descheduler.** `descheduler_role.yaml` has zero `resource.k8s.io` rules, and your `pkg/descheduler/informers/generic.go` change adds the `resource.k8s.io/v1` cases to the descheduler's generic informer factory. Worth checking before you add anything, though: `ForResource` only constructs an informer when some plugin actually asks for that GVR. If nothing requests these GVRs today, the informer never starts and no permission is needed yet — in which case the right move is to add nothing now and let whoever wires up the first consumer add the rule. Please check whether anything requests them; I don't want to send you adding speculative permissions.

**`koordlet.yaml`.** This one is hand-written (its ClusterRole is `koordlet-role`, not `manager-role`), so it survives `make manifests` — but the PR contains no koordlet code at all. Nothing in koordlet reads DRA objects, so those 10 lines grant permissions with no consumer. Please drop them unless you can point at the code that needs them.

## Suggested resolution

1. Revert both `config/rbac/role.yaml` and `config/rbac/koordlet.yaml` to main. That alone should turn `Check manifests` green.
2. Confirm whether any descheduler plugin requests the `resource.k8s.io` GVRs. If yes, add the rules to `config/rbac/descheduler_role.yaml` (hand-written, no marker needed). If no, add nothing.
3. Rebase — the branch is 69 commits behind main.

Everything else looks good to me: the `utils.go` skip is correct and minimally invasive, the `generic.go` v1 cases mirror the existing v1beta1 block exactly, and the test is reasonable. One optional nit: moving the `len(pod.Spec.ResourceClaims) > 0` check above the `state = &preFilterState{...}` construction would read slightly better, but as written it's correct *because* of the named return — so if you move it, you must return an explicit `&preFilterState{skip: true, ...}` rather than a bare `return`. Given that subtlety is what tripped me up, leaving it exactly as-is is the safer choice.

Sorry again for the bogus blocking report. Thanks also for splitting out the unrelated coscheduling change back in May when I asked — that was clean.

vibhor-5:
@ZiMengSheng Thanks for the review and detailed follow-up!

I have updated the PR accordingly:
1. **Rebased** the branch onto `upstream/main`.
2. **RBAC Cleanup**: Reverted `config/rbac/role.yaml` and `config/rbac/koordlet.yaml` back to `main`. `make manifests` now runs cleanly in CI.
3. Verified unit tests for `deviceshare` pass cleanly (`go test ./pkg/scheduler/plugins/deviceshare/...`).

PR #2920, opened 2026-05-11
test(scheduler): add unit tests for elasticquota plugin preempt.go

### LINKED ISSUE #2919: [Test] Add unit tests for `pkg/scheduler/plugins/elasticquota/preempt.go`
**What happened**:

`pkg/scheduler/plugins/elasticquota/preempt.go` implements preemption behavior for the ElasticQuota scheduler plugin (canPreempt, PodEligibleToPreemptOthers, SelectVictimsOnNode, PDB-aware victim grouping, PDB lister wiring, and small PostFilter/preemption helper hooks). 

It has no dedicated unit tests, so behavior around default-quota preemption flags, nominated-node eligibility, victim selection vs. quota limits, and PDB violation grouping was easy to regress.

Add `preempt_test.go` covering `canPreempt`, `PodEligibleToPreemptOthers`, `SelectVictimsOnNode`,
`filterPodsWithPDBViolation`, `getPDBLister`.

Apolo151:

Adds `pkg/scheduler/plugins/elasticquota/preempt_test.go` to give direct unit coverage for preempt.go, which previously had none.

The new tests exercise:

- `canPreempt`: check when preemption is blocked (can’t preempt that pod) and when it’s allowed, including the case where default-quota preemption is turned off by default.
- `PodEligibleToPreemptOthers`: We check “can this pod try to preempt others?” for things like never-preempt policy, nominated node status, and pods that are already shutting down on that node.
- `SelectVictimsOnNode`: We check picking who to evict: nobody to evict, everyone can stay if there’s enough quota room, extra victim when quota is too tight, and skipping pods that must not be preempted. 
    - Tests use the same test setup as real scheduling (including pod IDs) so the scheduler code doesn’t blow up.
- `filterPodsWithPDBViolation`: We check how pods are split into “would break a PDB” vs “safe” for a few real-world-ish cases (wrong namespace, already disrupted, bad selector, etc.).
- `getPDBLister`: We check that the PDB lister is only returned when the feature is on and the API says PDBs exist.

No production code changes.

Fixes #2919

Run the tests:

```bash
go test -count=1 -cover ./pkg/scheduler/plugins/elasticquota/
```

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/preempt_test.go
+++ b/pkg/scheduler/plugins/elasticquota/preempt_test.go
@@ -1,0 +1,654 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package elasticquota
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


ZiMengSheng (approved):
/lgtm
/approve

PR #2921, opened 2026-05-11
scheduler: fix memleak when reservation is deleted before binding completed

saintube:

koord-scheduler: fix a memory leak bug when scheduling a deleting reservation.

1. For the Reservation plugin,

When a reservation is created and immediately deleted during the binding cycle, the reservation plugin's Unreserve may fail to get the reservation from the lister and skip the cache cleanup. This leads to memory leaks in the reservationCache.

This patch constructs a temporary reservation object using the reserve pod's info when the lister returns an error, ensuring the cache is always cleaned up during Unreserve.

2. For the LoadAwareScheduling plugin,

The LoadAware plugin stores the reserve pod during the Reserve phase, but never removes it.
To fix the problem, we skip the load-aware Reserve for the reserve pods.

### V. Checklist


ZiMengSheng (commented):
/lgtm
/approve

ZiMengSheng (commented):
/lgtm
/approve

ZiMengSheng (approved):

PR #2922, opened 2026-05-11
scheduler: account for reservation CPUs in amplified CPU filter

### LINKED ISSUE #2158: [BUG] NodeNUMAResource should support allocating reserved cpus when CPU amplification enabled
**What happened**:

The fine-grained cpu orchestration has already supported allocating reserved CPUs for the owner pods when the node does not enable the cpu amplification. But when the cpu amplification is enabled, the owner pod can fail to allocate the reserved CPUs because of insufficient CPUs.

Both when the cpu amplification is disabled and it is enabled, the owner pods can allocate the reserved CPUs correctly.

**Environment**:
- App version: v1.5.0
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

madmecodes:
## Summary

When CPU amplification is enabled, `filterAmplifiedCPUs` was rejecting owner pods that should be able to reuse reservation CPUs. The function counts reservation CPUSet CPUs as consumed (amplified) capacity but didn't know those CPUs could be reused by the pod being scheduled.

The fix reads the reservation restore state from the cycle state and subtracts the amplified contribution of reusable reservation CPUs from the effective requested amount, so the capacity check reflects that the owner pod replaces the reservation's usage rather than adding to it.

Also added tests covering the reservation + amplification scenario (cpuset owner pod, cpushare owner pod, and fully-occupied node cases).

Fixes #2158

## Test plan

- [x] existing `TestFilterWithAmplifiedCPUs` tests pass (no regressions)
- [x] new `TestFilterAmplifiedCPUsWithReservation` tests pass
- [x] full `./pkg/scheduler/...` test suite passes (2106 tests)

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/plugin.go
+++ b/pkg/scheduler/plugins/nodenumaresource/plugin.go
@@ -383,7 +383,7 @@ func (p *Plugin) Filter(ctx context.Context, cycleState fwktype.CycleState, pod 
 	if !status.IsSuccess() {
 		return status
 	}
-	if status := p.filterAmplifiedCPUs(state.requests.Cpu().MilliValue(), nodeInfo, requestCPUBind); !status.IsSuccess() {
+	if status := p.filterAmplifiedCPUs(cycleState, state.requests.Cpu().MilliValue(), nodeInfo, requestCPUBind); !status.IsSuccess() {
 		return status
 	}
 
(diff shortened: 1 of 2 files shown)


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 9 with no activity since. The reservation-CPU accounting idea is still valid; if you'd like to continue, feel free to reopen or open a fresh PR referencing this one — happy to review.

Issue #2852, opened 2026-04-14
SchedulerQueueingHints


saintube:
Extend EventsToRegister / QueueingHints for Core Plugins

Kubernetes 1.35 enables SchedulerQueueingHints by default. Plugins that return incorrect or missing QueueingHintFn in EventsToRegister() cause the scheduler to fall back to re-queuing all pods on every event, hurting performance.

What needs to be done:

- [ ] Audit all Koordinator plugins that implement EventsToRegister() — verify QueueingHintFn is correctly set (currently Reservation plugin registers events without QueueingHintFn)
- [ ] Implement proper QueueingHintFn for Reservation plugin events (Pod Delete, Reservation Add/Update/Delete)
- [ ] Audit Coscheduling, NodeNUMAResource, DeviceShare, ElasticQuota plugins for missing hints
- [ ] Add unit tests for each QueueingHintFn to verify Queue/QueueSkip return values

PR #2862, opened 2026-04-14
scheduler: extend EventsToRegister / QueueingHints for core plugins (#2852)

thc1006:
### Ⅰ. Describe what this PR does

Kubernetes 1.35 enables `SchedulerQueueingHints` by default. A plugin's `EventsToRegister()` entry that omits `QueueingHintFn` causes the scheduler to fall back to re-queuing every unschedulable pod on every matching cluster event, losing the benefit of the feature gate.

This PR audits every Koordinator scheduler plugin that implements `EventsToRegister` and closes the gaps:

| Plugin | Before | After |
| --- | --- | --- |
| `ElasticQuota` | Already hinted (opt-in `EnableQueueHint`) | unchanged |
| `Reservation` | Events, no hints | Hinted opt-in via `EnableQueueHint` |
| `NodeNUMAResource` | Events, no hints | Hinted opt-in via `EnableQueueHint` |
| `DeviceShare` | Events, no hints | Hinted opt-in via `EnableQueueHint` |
| `LoadAware` | Events, no hints | Hinted opt-in via `EnableQueueHint` |
| `Coscheduling` | Returns no events (gang gates in `PreEnqueue`) | Comment + invariant test |

The flag matches the existing `ElasticQuota` pattern: off by default, so this PR is byte-for-byte equivalent to the previous behaviour unless an operator opts in.

Per-plugin hint semantics:

- **Reservation** — Pod Delete is `Queue` only when the waiting pod uses reservations and the deleted pod was either a reserve pod or was bound to a reservation. Reservation Add is `Queue` only when the new reservation is already active. Reservation Update is `Queue` on the inactive→active transition. Reservation Delete is `Queue`.
- **NodeNUMAResource** — Pod Delete is `Queue` only when both the waiting pod and the deleted pod use NUMA-aware allocation. NRT Add/Update are `Queue`; NRT Delete is `QueueSkip`.
- **DeviceShare** — Pod Delete is `Queue` only when both pods request device resources (GPU/RDMA/FPGA). Device Add/Update are `Queue`; Device Delete is `QueueSkip`.
- **LoadAware** — Pod Delete is `Queue` only when the deleted pod was bound to a node. NodeMetric Add/Update are `Queue`; NodeMetric Delete is `QueueSkip`.
- **Coscheduling** — unchanged behaviour. The plugin deliberately returns no events; gang gating lives in `PreEnqueue`. An expanded comment and an invariant test pin this decision.

### Ⅱ. Does this pull request fix one issue?

[...]

PR #2927, opened 2026-05-12
scheduler: reservation support SchedulerQueueingHints (#2852)

thc1006:

Wires the Reservation plugin into SchedulerQueueingHints (KEP-4247), behind `ReservationArgs.EnableQueueHint`, default false. The plugin already registers two events; this attaches a `QueueingHintFn` to each so a pod it rejected is only requeued by events that can change that verdict.

The framework side is already live regardless of that flag. `SchedulerQueueingHints` has been GA and `LockToDefault: true` since 1.34, and `buildQueueingHintMap` substitutes `defaultQueueingHintFn`, which always returns `Queue`, wherever a plugin leaves `QueueingHintFn` nil. Both registered events resolve to that today, so `EnableQueueHint=false` is bit-for-bit current behaviour and this change is strictly additive.

Most of the value is in the Reservation Update path. This scheduler writes an Unschedulable condition after every failed attempt, and those writes currently requeue every waiter. The hint skips them when nothing a Filter reads has moved: same UID, phase, `status.nodeName`, `status.allocatable`, `status.allocated`, `status.currentOwners`, deletion state, generation, labels and annotations.

The other two paths requeue unconditionally, and it took three revisions to find out why.

Reservation events arrive as `*unstructured.Unstructured`, not the typed object. The scheduler has no typed informer for a custom resource, so `addAllEventHandlers` falls back to `dynInformerFactory.ForResource(gvr)`. The first revision type-asserted, which never matched, so every event took the error path and answered `Queue`. Fail-open, so nothing was lost, but none of the filtering above ran. #3128 fixed the same thing for ElasticQuota and the decoder here follows it, except that it returns the decode error where `toElasticQuota` collapses to `nil`, so an undecodable old object cannot be read as an Add. Porting that back is a follow-up.

Pod deletions cannot be skipped on the event object. `deletePod` unwraps a `DeletedFinalStateUnknown` before notifying the queue and passes the carried pod on, calling it possibly stale in its own comment, while `cache.RemovePod` removes `currState.pod` and treats an empty `Spec.NodeName` as a missed delete rather than corruption. So a pre-binding copy of a pod the cache just released reaches the hint with empty placement fields.

An Add is the first sight of the object by that observer, not its creation. `processDeltas` emits `OnAdd` whenever the object is absent from the local store, so a relist can deliver a reservation whose earlier life this observer missed, after the typed listener has already dropped its nomination. `BeforeFilter` counts a nominated reserve pod against every pod evaluated for that node, so that release matters and nothing in the object shows it happened.

Part of #2852.

```
go test -race ./pkg/scheduler/plugins/reservation/... ./pkg/scheduler/apis/config/... -count=1
go test ./pkg/scheduler/plugins/reservation -run '^$' -bench BenchmarkIsSchedulableAfterReservationChange -benchmem
```

Every hint case runs twice, once with the typed object and once with the same object converted to unstructured, and both must agree. The `QueueSkip` rows are the ones that pin the decoder; a table of `Queue`-only expectations passes against the old type-error fallback too.

Two tests assert the premises rather than the hint: that `restoreUnmatchedReservations` subtracts more from the node's requested total as the allocation grows, and that the typed listener frees the node before the Add arrives.

The scheduling queue runs the hint once per pod this plugin rejected, with the same two event objects each time, so the cost that matters is one event against many pods. The hint compares with `maps.Equal`, `slices.Equal` and `quotav1.Equals` rather than `apiequality.Semantic.DeepEqual`, which reflects over every `Quantity` and `ObjectReference`, and it keeps the last decoded pair so `FromUnstructured` runs once per event rather than once per pod. Benchmarks use distinct old and new objects with 10 labels, 5 annotations, 10 owners and four resource names, on this machine:

```
GenerationBump                        407 ns/op      0 B/op    0 allocs/op
StatusHeartbeat                       834 ns/op      0 B/op    0 allocs/op   typed compare only
UnstructuredHeartbeatDecode         53201 ns/op  16223 B/op  340 allocs/op   first call for an event
UnstructuredHeartbeatCached           937 ns/op      0 B/op    0 allocs/op   every later call
UnstructuredAdd                     48831 ns/op  20116 B/op  375 allocs/op
OneEvent1000Pods                  1021062 ns/op  16209 B/op  340 allocs/op   one heartbeat, 1000 rejected pods
```

Before the typed compares and the decode cache, the same 1000-pod event took 69.5 ms and 557,071 allocations on the same harness. Through a real `queue.PriorityQueue`, which adds the queue's own per-pod work, the same event costs 2.0 ms with the cache and 53.7 ms with it disabled (`BenchmarkQueueingHintThroughSchedulingQueue_OneEvent1000Pods`). A previous revision of this description quoted heartbeat numbers taken with the same pointer for old and new, which short-circuits the map comparison; those were not representative.

The Pod/Delete registration keeps an explicit callback that always returns `Queue`, so the reasoning stays next to the code. It is the same function as the framework's `defaultQueueingHintFn`, which is what a nil `QueueingHintFn` gets, so the two are interchangeable. Naming that equivalence in the doc comment is queued as a follow-up, since the merged ElasticQuota hint does narrow its Pod/Delete path and a reader will look for the same filter here.

Out of scope and tracked elsewhere: unregistered wake-up sources in #3145, the missing happens-before between the plugin cache and the queue in #3148, and two correctness bugs found while writing this in #3154 and #3155. `EnableQueueHint` stays default-off until #3145 lands.

#3156 narrows the stale-nomination window once the typed informer store has observed the invalidating change. It does not order that against the dynamic informer that emits these events, so the two are independent and can merge in either order.

### Ⅴ. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/plugin.go
+++ b/pkg/scheduler/plugins/reservation/plugin.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,10 +35,12 @@ import (
 	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	k8sfeature "k8s.io/apiserver/pkg/util/feature"
 	listercorev1 "k8s.io/client-go/listers/core/v1"
+	clientcache "k8s.io/client-go/tools/cache"
 	resourceapi "k8s.io/component-helpers/resource"
 	"k8s.io/klog/v2"
 	fwktype "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
+	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
 
 	apiext "github.com/koordinator-sh/koordinator/apis/extension"
 	schedulingv1alpha1 "github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1"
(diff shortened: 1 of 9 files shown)


ZiMengSheng:
Pre-reviewed as part of the #2852 umbrella triage. The split from #2862 into per-plugin PRs matches the maintainer review request, and this Reservation slice is self-contained: config plumbing (types/defaults/conversion/deepcopy all updated consistently), two QueueingHintFn callbacks gated behind `EnableQueueHint` (default-off — good conservative rollout), and a 1011-line test file. One thing to verify in final review: whether the queueing-hint callbacks are registered on **all** EventsToRegister entries or only the two the description mentions — an event registered without a hint callback silently loses the optimization, so the coverage should be explicit.

thc1006:
Thanks for the triage.

Both entries have one. `EventsToRegister` registers exactly two: `Pod` with `Delete`, and `reservations.v1alpha1.scheduling.koordinator.sh` with `Add|Update|Delete`. With `EnableQueueHint` on they get `isSchedulableAfterPodDeletion` and `isSchedulableAfterReservationChange`; with it off both stay nil and the registrations are otherwise unchanged. There's no third entry that could quietly lose the optimization.

`TestPlugin_EventsToRegister` already asserted the count, the resources and the action masks. I've extended it to also call each registered callback with an input the other one answers differently, since a nil check alone would still pass if the two were wired the wrong way round.

Unrelated to your question, but I found a bug in my own benchmark while re-reading this. `BenchmarkIsSchedulableAfterReservationChange_OwnerOnlyWaiter` bumped `metadata.generation`, which returns `Queue` long before `reservationOwnerMatches` is reached, so it never measured the owner-matching path its name refers to. It drives the Add path instead. The two are 1961 ns/op and 236 ns/op here, so the old figure was off by about an order of magnitude. Both the test and the benchmark change are local so far; I'll push them together.

Worth separating out: every registered event has a hint, but not every event that could unblock a rejected waiter is registered. Waiter pod updates, pre-allocation candidate pods and node label updates aren't registered at all. Those are in #3145 rather than here.

thc1006:
Thanks for the triage. Two things changed since then that I would rather you hear from me than find in the diff.

The Pod/Delete hint no longer filters anything. It used to skip a deleted pod whose placement fields were both empty. That turned out not to hold: the scheduler unwraps a `DeletedFinalStateUnknown` before it notifies the queue and passes the carried pod on, calling it possibly stale in its own comment, and `cache.RemovePod` then removes the pod the cache holds while treating an empty `Spec.NodeName` as a missed delete rather than corruption. A pre-binding copy of a pod the cache just released can reach the hint with nothing in those fields. So it requeues unconditionally now, and that registration contributes no filtering at all. I left an explicit callback there to keep the reasoning next to the code, but dropping the hint from that registration would behave the same. Your call.

The Reservation hint was not running at all before `cd60523f`. That event resource has no typed informer, so the scheduler falls back to the dynamic one and the objects arrive as `*unstructured.Unstructured`. The typed assertion never matched them, so every event took the error path and answered Queue. Nothing was lost, since that path is conservative, but none of the filtering you triaged could run. #3128 fixed the same thing for ElasticQuota and I followed it.

The second commit is separate. Allocation growth and `currentOwners` changes requeue now, because the restore path subtracts a reservation's allocation from the node's requested total, and a pod slot is a different limit from the quantities. The reasoning is in the commit message.

One note on reading order, since the diff is large: 289 of the 1825 added lines are production code.

ZiMengSheng:
Final review. I verified every load-bearing claim in the description against `upstream/main` and against the vendored `k8s.io/kubernetes@v1.35.2` / `k8s.io/kube-scheduler@v0.35.6`, ran the suite on a merge with current `main`, and mutation-tested the hint. Result: **`/lgtm`**, with two non-blocking follow-ups below.

## Coverage verdict (the thing I was checking)

#2852's checklist for this plugin is the event surface the plugin actually registers: `Pod/Delete` and `Reservation/Add|Update|Delete`. Both are wired. The three unregistered sources (the unschedulable pod's own update, pre-allocation candidate Pod Add/Update, Node label updates feeding reservation affinity) are **not** regressions — the plugin registers none of them today either — and they are tracked in #3145 with `help wanted`, alongside #3148 for the cache/wake-up ordering gap. That is the right decomposition; I'm not asking for any of it here.

## Risk statement the description is missing

`SchedulerQueueingHints` is **GA and `LockToDefault: true` since 1.34** (`pkg/features/kube_features.go:1762-1766`). So the framework-side mechanism is already unconditionally live on `main`: the plugin's two registered events currently resolve to `defaultQueueingHintFn` (`pkg/scheduler/scheduler.go:462-466`, substituted at `:493-496` whenever `QueueingHintFn == nil`), i.e. always `Queue`. That makes `EnableQueueHint=false` bit-for-bit today's behavior, and the PR strictly additive. Worth one line in the description — it's a stronger safety argument than "default false".

## What behavior is actually delivered

Only `isSchedulableAfterReservationChange`, and only on Update. Add and Delete always `Queue`. The two `QueueSkip` paths are:

- both `Available`, same node: condition-only status writes;
- both non-available, same phase: status writes that move neither `generation` nor labels nor annotations.

That looks like a thin slice, but it is the right slice: the only in-tree writers on Reservations are `UpdateStatus` (`reservation_handler.go:330`, `controller/controller.go:379`) and `Delete` (`garbage_collection.go:47`). Nothing in-tree mutates a Reservation's spec, labels or annotations. So the high-frequency event — the unschedulable condition this scheduler records after every failed reserve-pod attempt — is exactly what gets skipped, while the ungated `Queue` branches for spec/label/annotation are driven only by user writes and are inherently rare. The cost/benefit holds.

## Mutation verification

Per our rule that guard tests must be shown to fail, I removed one comparison at a time from `isSchedulableAfterReservationChange` and re-ran `TestPlugin_QueueingHint*`:

| removed | result |
| --- | --- |
| `Status.CurrentOwners` DeepEqual | FAIL — `ReservationChange_UnstructuredParity/owners_released,_quantities_unchanged` |
| `Generation` check in the non-available branch | FAIL |
| `Status.Allocatable` DeepEqual | FAIL |
| `DeletionTimestamp.IsZero()` mismatch | FAIL — `ReservationChange_UnstructuredParity/entering_deletion` |

Four for four, each with a named case. The tests pin the logic rather than restating it. `go build ./pkg/scheduler/... ./cmd/koord-scheduler/...` and `go test ./pkg/scheduler/plugins/reservation/... ./pkg/scheduler/apis/config/...` are green on a clean merge with `main` (33 commits ahead of your merge-base, no conflicts).

I also checked the one structural premise the `generation` reasoning rests on: `config/crd/bases/scheduling.koordinator.sh_reservations.yaml:411-412` does enable the status subresource, so `metadata.generation` bumps on spec writes only. And `IsReservationAvailable` (`pkg/util/reservation/reservation.go:238-240`) keys off phase **and** node name, both already compared above the branch — so your comment that same-phase-and-same-node implies matching availability is exact, and the two branches are a clean dichotomy rather than an accidental one.

## Consistency with the merged precedent — good

The arg shape (internal `bool` / external `*bool`, wiring gated inside `EventsToRegister`) matches the already-merged ElasticQuota implementation exactly, and you reuse the existing `defaultEnableQueueHint` package var rather than adding a second default. `reservationForQueueingHint` is a strict improvement on `toElasticQuota`: keeping the decode error instead of collapsing to `nil` means an undecodable old object cannot be mistaken for an Add. That distinction matters — mistaking it for an Add turns a reasoned wake-up into an unconditional one and hides the bug. Please open a follow-up to port that back to `toElasticQuota`.

## Follow-up 1 (non-blocking): `isSchedulableAfterPodDeletion` is a provable no-op

Its body is `return fwktype.Queue, nil`, which is *the same function* as `defaultQueueingHintFn` — the one the framework already installs when `QueueingHintFn` is nil. Since the framework gate is locked on, the Pod/Delete branch is behaviorally identical in both states of `EnableQueueHint`, and the 150-line table below it has every case expecting `Queue`, so it cannot fail for any input.

I am not asking you to narrow it — your argument for why the event object can't support a skip is correct, and `cache.RemovePod` does treat an empty `Spec.NodeName` as a missed delete rather than as proof. The problem is discoverability: the merged ElasticQuota `isSchedulableAfterPodDeletion` (`elasticquota/plugin.go:486-540`) *does* narrow, so the next reader will assume this one does too and go looking for the filter. Either drop the function and let the default apply, or keep it and name `defaultQueueingHintFn` in the doc comment so the equivalence is explicit rather than something a reader has to derive. Your call; both are fine by me.

## Follow-up 2 (non-blocking): the defaults.go comment cuts against its neighbors

`SetDefaults_ReservationArgs` copies the pointee with a three-line comment about not aliasing the package-level default. Twenty lines down, `SetDefaults_ElasticQuotaArgs` aliases it five times (`defaults.go:218-232`) — including `EnableQueueHint` itself. The hazard is real (a caller writing through `*args.EnableQueueHint` would corrupt the default for every object defaulted afterwards), so keep your version; but as it stands the comment reads as if the neighbors were considered and accepted. Trim it to one line and file the ElasticQuota five separately, or say explicitly that they are pre-existing and out of scope.

## Merge ordering

Re-measured, and it is looser than I had it:

- **#2927 × #2863 merges clean.** #2863 (SignPlugin) only appends to `reservation/plugin.go`; no interaction with your hunks. So landing #2863 first costs you nothing.
- **#2927 × #2900** conflicts in exactly one file: `apis/config/v1/defaults_test.go`, add/add, because both PRs create it. The `zz_generated.{conversion,deepcopy}.go` files merge cleanly despite both adding a field to `ReservationArgs` — so whoever lands second resolves a test file, and CI's `unit-tests(Check auto-generated codes)` will say if codegen needs a re-run. I had this recorded as a hard ordering constraint; it is a convenience, not a constraint.

Nothing here blocks. Holding for a second approver on the `koord-scheduler-approvers` side rather than self-approving, given the size.

/lgtm

thc1006:
I'll open a PR the moment this merges that drops the Pod/Delete callback and lets `defaultQueueingHintFn` cover it, moves the reasoning to the registration site, removes the two tests that could not fail, and trims the defaults comment to one line; does that work for you?

pkg/scheduler/plugins/reservation/plugin.go
```diff
@@ -181,10 +184,219 @@ func (pl *Plugin) EventsToRegister(_ context.Context) ([]fwktype.ClusterEventWit
...
+	//   - an Available reservation's reserve pod is in the scheduler cache, and
+	//     it carries the reservation's labels, so other pods' inter-pod affinity
+	//     and topology spread are evaluated against them.
+	// Accounting changes are handled ungated above; a status write that moves
+	// neither Allocated nor CurrentOwners, such as the condition refresh this
+	// scheduler records after every attempt, is skipped.
+	if oldR.Generation != newR.Generation {
+		return fwktype.Queue, nil
+	}
+	if !apiequality.Semantic.DeepEqual(oldR.Labels, newR.Labels) ||
```
saintube: These DeepEquals are kind of heavy. Is there any benchmark to ensure we filter meaningful events and do not make the queuing worse?

thc1006: You were right, and the benchmarks I had were hiding it: they passed the same pointer as old and new, which short-circuits DeepEqual before it reaches the maps.

With distinct objects of a realistic size (10 labels, 5 annotations, 10 assigned owners, four resource names), the six `Semantic.DeepEqual` calls cost 17 µs and 217 allocations per call, and 115 µs with 100 owners. The reflection over `Quantity` and `ObjectReference` is nearly all of it: `Semantic.DeepEqual` on a four-entry `Allocatable` is 2.4 µs and 25 allocations, where `quotav1.Equals` on the same map is 136 ns and none. And since the queue runs the hint once per rejected pod with the same two event objects, `FromUnstructured` was also running once per pod.

b6d9874b fixes both. `maps.Equal`, `slices.Equal` and `quotav1.Equals` replace `Semantic.DeepEqual`, with a test pinning that they agree on the nil, empty, subset, order and Quantity-format edges. A one-entry cache keyed by object identity and resourceVersion makes an event decode once; a Delete returns before decoding at all.

Same harness, one heartbeat against 1000 rejected pods:

```
before   69.5 ms   557,071 allocs
after     1.0 ms       340 allocs
```

The description now carries these numbers instead of the old table. It is a single commit on top of what was reviewed, if that is easier to look at.

thc1006: One more number, since the question was about the queue rather than the hint alone. 3c4f132f drives the hint through a real `queue.PriorityQueue` with 1000 pods rejected by this plugin, so the queue's own per-pod work is included: one heartbeat costs 2.0 ms with the decode cache and 53.7 ms with it disabled on the same harness. The test alongside it checks that the cache ends up holding the very objects the queue passed, so the once-per-event decode is not an assumption about the queue any more.

PR #2929, opened 2026-05-12
 fix: replace custom OnceValues with sync.OnceValues from stdlib

### LINKED ISSUE #2928: [BUG] OnceValues in pkg/util/utils.go reimplements sync.OnceValues available since Go 1.21
**What happened**:
`pkg/util/utils.go:335–361` contains a hand-rolled implementation of `sync.OnceValues` with a comment `// TODO: Replace this function with the standard library after go1.21+ version`. `go.mod` declares `go 1.25.0`.

The custom implementation is removed. The one call site (`pkg/koordlet/util/system/resctrl.go:75`) uses `sync.OnceValues` directly from the standard library.

```
grep -n "OnceValues" pkg/util/utils.go
# line 336: func OnceValues(f func() ([]int, error)) func() ([]int, error)

head -3 go.mod
# go 1.25.0
```
`sync.OnceValues` has been stable since Go 1.21; the TODO precondition has been satisfied for four major releases.

- `pkg/util/utils.go:335–361` — full implementation to delete, plus the `sync` import which becomes unused.
- `pkg/koordlet/util/system/resctrl.go:75` — sole call site: `CacheIdsCacheFunc = util.OnceValues(GetCacheIds)` → `sync.OnceValues(GetCacheIds)`. The `pkg/util` import becomes unused in that file after the swap; `sync` is already imported there for `sync.Mutex`.
- `pkg/util/utils_test.go:1309–1326` — `TestOnceValues` tests the deleted function and must be removed. The stdlib's own test suite covers `sync.OnceValues` correctness.

mdryaaan:

  Removes the hand-rolled `OnceValues` function from `pkg/util/utils.go` and  replaces the single call site in `pkg/koordlet/util/system/resctrl.go` with  `sync.OnceValues` from the standard library. The custom implementation  predates Go 1.21; `go.mod` now declares `go 1.25.0`, so the stdlib version has been available for four major releases.

FIxes #2928 

  1. `go build ./pkg/util/...` — confirms the deleted function is no longer  referenced within the package.
  2. `go build ./pkg/koordlet/util/system/...` — confirms `resctrl.go` compiles with `sync.OnceValues` and without the `pkg/util` import.
  3. `make fast-test` — all existing tests pass; `TestOnceValues` was removed because the stdlib has its own tests for `sync.OnceValues`.

  `sync.OnceValues` is generic (`func OnceValues[F1, F2 any](...)`), while the deleted function had a concrete signature `func(func() ([]int, error))`. There was exactly one call site (`resctrl.go:75`); `GetCacheIds` returns `([]int, error)`, which satisfies the generic instantiatio  `sync.OnceValues[[]int, error]` without any cast.

  ### V. Checklist

### DIFF
--- a/pkg/koordlet/util/system/resctrl.go
+++ b/pkg/koordlet/util/system/resctrl.go
@@ -31,8 +31,6 @@ import (
 
 	"go.uber.org/multierr"
 	"k8s.io/klog/v2"
-
-	"github.com/koordinator-sh/koordinator/pkg/util"
 )
 
 const (
@@ -72,7 +70,7 @@ var (
 )
 
 func init() {
-	CacheIdsCacheFunc = util.OnceValues(GetCacheIds)
+	CacheIdsCacheFunc = sync.OnceValues(GetCacheIds)
 }
 
 func isCPUSupportResctrl() (bool, error) {
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
-	"sync"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


mdryaaan:
Hi @saintube @ZiMengSheng  , just tagging you for review. Could you please have a look at this PR whenever you have time? Thanks!

PR #2931, opened 2026-05-13
fix: use namespace-scoped list to resolve ElasticQuota name for pod

### LINKED ISSUE #2930: [BUG] ElasticQuota lookup uses pod.Namespace as both Namespace and Name in pod_check.go
**What happened**:
`pkg/webhook/elasticquota/pod_check.go:88` builds a `types.NamespacedName` with `pod.Namespace` as both the `Namespace` and `Name` fields:

```go
err := kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: pod.Namespace, Name: pod.Namespace}, eq)
```

This lookup only succeeds when the `ElasticQuota` name happens to equal the namespace name. Any other naming convention causes a `NotFound` error, which falls through to a list-based path at lines 95–103 that returns the first quota in the list without namespace or name filtering.

The `Get` call uses the actual quota name — derived from the pod's `scheduling.koordinator.sh/quota-name` annotation or the namespace's bound quota label — not `pod.Namespace` repeated twice.

1. Create namespace `team-a`.
2. Create an `ElasticQuota` named `team-a-quota` in namespace `team-a` (name differs from namespace name).
3. Schedule a pod into `team-a` with the koord-manager webhook enabled.
4. The webhook calls `getQuotaName`, which calls `kubeClient.Get` with `Name: "team-a"` — quota not found.
5. Falls through to `eqList` — returns `DefaultQuotaName` or the wrong quota depending on what else is listed.

- `pkg/webhook/elasticquota/pod_check.go:83–105`
- The list fallback at lines 95–103 loops `eqList.Items` and returns on the first iteration unconditionally, meaning the first quota in list order wins regardless of namespace match.
- The correct quota name should come from `pod.Annotations[extension.LabelQuotaName]` before falling back to namespace-scoped lookup.

**Environment**:
- App version: main
- Kubernetes version (use `kubectl version`): v1.35.x
- Install details (e.g. helm install args): koord-manager webhook enabled
- Others:

mdryaan:

`GetQuotaName` in `pkg/webhook/elasticquota/pod_check.go` was calling  `kubeClient.Get` with `Name: pod.Namespace`  repeating the namespace as the  quota name. This only worked when the ElasticQuota happened to share a name  with its namespace. The list fallback used a non-standard field selector  (`annotation.namespaces`) which the controller-runtime cache does not index,  so it could return quotas from any namespace and the first item was returned  unconditionally regardless of which namespace it belonged to.

 Replace the broken `Get` with a `List` scoped to the pod's own namespace, and  replace the unindexed field-selector fallback with a cluster-wide `List`  filtered by checking `GetAnnotationQuotaNamespaces` per item. Add annotation  lookup (`pod.Annotations[extension.LabelQuotaName]`) as the highest-priority  path before the existing label check.

Fixes #2930 

  1. `go test ./pkg/webhook/elasticquota/... -run TestGetQuotaName` — all six   table cases pass, including the regression case where the quota name differs from the namespace name.
  2. Create namespace `team-a`, ElasticQuota named `team-a-quota` in `team-a` (name ≠ namespace), deploy a pod into `team-a` with the webhook enabled. The pod now resolves to `team-a-quota` instead of falling through to`DefaultQuotaName`.

  The `Get` call is removed entirely — there is no convention guaranteeing a  quota's name equals its namespace name, so a namespace-scoped `List` is the  correct approach. The `fields` import is retained because `hasQuotaBoundedPods`  (same file) still uses `fields.OneTermEqualSelector`. The `DisableDefaultQuota` feature gate is preserved between the annotation check and the namespace list.

  ### V. Checklist

### DIFF
--- a/pkg/webhook/elasticquota/pod_check.go
+++ b/pkg/webhook/elasticquota/pod_check.go
@@ -21,9 +21,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -75,6 +73,10 @@ func (qt *quotaTopology) getQuotaNameFromPodNoLock(pod *corev1.Pod) string {
 }
 
 func GetQuotaName(pod *corev1.Pod, kubeClient client.Client) string {
+	if name := pod.Annotations[extension.LabelQuotaName]; name != "" {
+		return name
+	}
+
 	quotaName := extension.GetQuotaName(pod)
 	if utilfeature.DefaultFeatureGate.Enabled(features.DisableDefaultQuota) {
 		return quotaName
@@ -84,22 +86,31 @@ func GetQuotaName(pod *corev1.Pod, kubeClient client.Client) string {
 		return quotaName
 	}
 
-	eq := &v1alpha1.ElasticQuota{}
-	err := kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: pod.Namespace, Name: pod.Namespace}, eq)
-	if err == nil {
-		return eq.Name
-	} else if !errors.IsNotFound(err) {
-		klog.Errorf("Failed to Get ElasticQuota %s, err: %v", pod.Namespace, err)
-	}
-
+	// Find a quota in the pod's own namespace.
 	eqList := &v1alpha1.ElasticQuotaList{}
 	if err := kubeClient.List(context.TODO(), eqList, &client.ListOptions{
-		FieldSelector: fields.OneTermEqualSelector("annotation.namespaces", pod.Namespace),
+		Namespace: pod.Namespace,
 	}, utilclient.DisableDeepCopy); err != nil {
+		klog.Errorf("Failed to list ElasticQuota in namespace %s, err: %v", pod.Namespace, err)
+	} else {
+		for i := range eqList.Items {
+			if eqList.Items[i].Labels[extension.LabelQuotaIsParent] != "true" {
+				return eqList.Items[i].Name
+			}
+		}
+	}
+
+	// Find a quota whose AnnotationQuotaNamespaces covers this pod's namespace.
+	allEqList := &v1alpha1.ElasticQuotaList{}
+	if err := kubeClient.List(context.TODO(), allEqList, utilclient.DisableDeepCopy); err != nil {
 		return extension.DefaultQuotaName
 	}
-	for _, quota := range eqList.Items {
-		return quota.Name
+	for i := range allEqList.Items {
+		for _, ns := range extension.GetAnnotationQuotaNamespaces(&allEqList.Items[i]) {
+			if ns == pod.Namespace {
+				return allEqList.Items[i].Name
+			}
+		}
 	}
 
 	return extension.DefaultQuotaName
--- a/pkg/webhook/elasticquota/pod_check_test.go
+++ b/pkg/webhook/elasticquota/pod_check_test.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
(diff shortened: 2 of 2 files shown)


mdryaaan:
Hey @ZiMengSheng  @zwzhang0107  , Could you please have a look at this PR whenever you have time? Thanks!

ZiMengSheng:
What is the difference from user view. It seems we should respect the existing behavior.

ZiMengSheng (commented):
See detailed review comments above.

ZiMengSheng:
Thanks for tracking down #2930 and putting up a fix. The core bug is real, but I have a few concerns about the current shape of the change and would suggest tightening the scope before merging. (Superseding my previous empty review — reposting the details here for readability.)

### 1. Removing the `Get(Namespace: pod.Namespace, Name: pod.Namespace)` fast path is a semantic change, not just a bug fix

The original `Get` looks broken at first glance, but it's actually leaning on a convention that is used elsewhere in this same file — e.g. `hasQuotaBoundedPods` also does `Namespace: quotaName`, i.e. it assumes `quotaName == namespaceName`. The vast majority of tests in this package (`MakeQuota("team-a").Namespace("team-a")`) follow the same convention. So the `Get` is best read as a fast path: "if there's a same-named quota in this namespace, use it; otherwise fall back to List."

By dropping the `Get` entirely and going straight to a namespace-scoped `List` that returns the first non-parent quota, the behavior changes in a subtle way: when a namespace contains both a convention-named quota and other leaf quotas, the resolved quota name may drift depending on cache list order. Could you either

- keep the `Get` as an explicit fast path before the `List`, or
- call out in the PR description that dropping the `name == namespace` convention is intentional and explain why it's safe?

### 2. The PR description's claim about the `annotation.namespaces` indexer is inaccurate, and switching to a cluster-wide `List` is a perf regression on the webhook hot path

The indexer `annotation.namespaces` on `ElasticQuota` **is** registered in `pkg/util/fieldindex/register.go` and is also used by the scheduler plugin (`pkg/scheduler/plugins/elasticquota/plugin.go`). So the original fallback

```go
kubeClient.List(ctx, eqList, &client.ListOptions{
    FieldSelector: fields.OneTermEqualSelector("annotation.namespaces", pod.Namespace),
}, utilclient.DisableDeepCopy)
```

when the indexer is active, returns exactly the EQs whose `AnnotationQuotaNamespaces` covers `pod.Namespace`, so `return quota.Name` on the first item is not "return any random quota" — it's returning a correctly filtered result.

The new code replaces this with a cluster-wide `List` followed by a manual loop that JSON-unmarshals each quota's annotation. This is on the admission webhook hot path (every Pod admission), and scales linearly with the total number of ElasticQuotas cluster-wide. In clusters with hundreds or thousands of EQs this is a measurable regression.

Could you either:

- keep the `FieldSelector("annotation.namespaces", pod.Namespace)` path (which is the point of having the indexer), and only fix the actual defect if there is one, or
- if you have evidence that the webhook manager's cache never calls `RegisterFieldIndexes` and the indexer really is unavailable in this binary, please state that explicitly in the PR body and remove the current wording ("the controller-runtime cache does not index") which reads as a general claim about the field?

### 3. `pod.Annotations[extension.LabelQuotaName]` is scope creep and should be dropped from this PR

```go
if name := pod.Annotations[extension.LabelQuotaName]; name != "" {
    return name
}
```

A few problems here:

- `LabelQuotaName` is defined and used everywhere in the repo as a **label** key (see `extension.GetQuotaName`, `pkg/util/fieldindex/register.go`'s `label.quotaName` indexer, etc.). Reading the same key from annotations too introduces a brand-new resolution path that isn't used or documented anywhere else.
- The linked issue #2930 does not request this behavior — the bug is purely about name resolution when the pod has no quota label.
- It sits **before** the `DisableDefaultQuota` feature gate check, so it silently changes the behavior of that feature gate as well.

Suggestion: drop this block from the current PR to keep the scope tight on the actual bug fix. If there is a real use case for annotation-based quota assignment, please open a separate issue/proposal so we can decide the key name, precedence with the label, and interaction with `DisableDefaultQuota` explicitly.

### 4. Nice-to-haves (non-blocking)

- The new namespace-scoped path filters out `LabelQuotaIsParent == "true"`, but the cross-namespace path does not. Original code didn't filter parents on either path, so this isn't a regression, but for symmetry / defense-in-depth (given `ValidateAddPod` explicitly forbids binding a pod to a parent quota), it would be nicer to filter parents on both.
- When multiple non-parent quotas coexist in a single namespace, the result depends on the underlying list ordering. Worth either sorting deterministically or leaving a code comment noting that this scenario is not supported.
- Please consider adding tests for: (a) `DisableDefaultQuota` gate on, (b) namespace containing both parent and leaf quotas, (c) precedence order (annotation vs label — if you keep #3), (d) both a namespace-local quota and a cross-namespace quota being present at the same time.

### Summary

Bug is real and worth fixing. Before merging I'd like to see (1) the `name == namespace` fast path preserved or its removal explicitly justified, (2) the PR body corrected re: the `annotation.namespaces` indexer and ideally the indexer path retained instead of a cluster-wide `List`, and (3) the new `pod.Annotations[LabelQuotaName]` branch removed from this PR. The parent-filter symmetry and extra tests can be addressed here or in a follow-up. Thanks!

PR #2937, opened 2026-05-13
koordlet: select cpu burst cfs candidates by headroom

### LINKED ISSUE #2164: [proposal] cpu burst should also burst some pods when the node util nears cpuBurst.sharePoolThresholdPercent
**What is your proposal**:
when the utilization around cpuBurst.sharePoolThresholdPercent, koordlet should not reset all but try to select some pods and burst them.

**Why is this needed**:

cpu burst strategy is proposed to solve problems of cpu resource throttling for pods like java app, which needs much resource than its resource.limit during start time.

When there are multi pods start at the same time, the strategy may not handle wll. Since now if node utilization > cpuBurst.sharePoolThresholdPercent, all bursted cfs_quota will be reset, which may make the node utilizaiton << cpuBurst.sharePoolThresholdPercent. koordlet will try burst again util the cooling duration finished.

This leads to none pods could get burst in long time and stuck at starting stage.

Only reset PART of over-utilized pods when the node.util > cpuBurst.sharePoolThresholdPercent to avoid none pod could get burst.
Only burst SOME throttled pods when node.util < cpuBurst.sharePoolThresholdPercent to avoid node util too high.

officialasishkumar:

- koordlet
  - Collect CFS quota burst candidates across all burstable Pods before applying CFS quota updates.
  - Use node share-pool threshold headroom to allow only part of throttled scale-up candidates in cooling state.
  - Use node share-pool threshold overage to scale down only enough over-base candidates in overload state.
  - Keep candidate selection deterministic by stable Pod/container identity and largest excess quota first for scale-down.

Fixes #2164

- `go test ./pkg/koordlet/qosmanager/plugins/cpuburst`
- `GOOS=darwin CGO_ENABLED=0 go test -c -o /tmp/koordinator-cpuburst-darwin.test ./pkg/koordlet/qosmanager/plugins/cpuburst`
- `git diff --check`

This keeps the existing idle and unknown-state behavior, but changes cooling/overload from blanket operation gating to threshold-aware candidate selection.

### V. Checklist

### DIFF
--- a/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
+++ b/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"math/rand"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mdryaan (commented):
Hi @officialasishkumar, i have reviewed it, LGTM.
 One housekeeping note: the DCO check is failing. Please sign your commits by running:
git rebase HEAD~<number-of-commits> --signoff then force push.
 cc @saintube @ZiMengSheng

ZiMengSheng:
Closing due to inactivity (last update May 2026). The headroom-based candidate selection for CPU burst CFS quota is not implemented on main; a revived design would still be welcome if you return to it. Note the cpuburst plugin recently gained normalization handling in #3176, so a revival should rebase on top of that.

PR #2938, opened 2026-05-13
koordlet: sync qos cgroup updates to systemd

### LINKED ISSUE #1734: [BUG] The cfs_quota_us of the Cgroup BestEffort resource group is periodically set to -1
**What happened**:
The content of the file /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.cfs_quota_us is periodically set to -1 every minute.

The cfs_quota_us of the BestEffort resource group should always be greater than 0 to limit the cpu resources used by the BE Pod.

1. Set cgroup driver to systemd
2. Set cpuSuppressPolicy to 'cfsQuota' in slo-controller-config
3. Observe changes in the content of the /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.cfs_quota_us file on k8s node. koordlet will dynamically modify the file content, but every minute, the file content will be reset to -1, and gradually increase (because koordlet sets the maximum step size, the file content will slowly increase to the ideal value)

I found that when kubelet sets cpu shares of besteffort group periodically (period is 1 minute), the content of cfs_quota_us will be changed to -1 at the same time. Through debugging, I found that kubelet only set cpushare and did not modify cfs_quota_us, but it changed.

I wrote a simple Demo to simulate the operation of kubelet. This demo only set cpu shares. 

If I modify /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.cfs_quota_us and then exec the demo, the content of cpu.cfs_quota_us will be set to -1

If I change cfs_quota_us by `systemctl set-property`, for example `systemctl set-property --runtime kubepods-besteffort.slice CPUQuota=5000%`, then exec the demo, the content of cpu.cfs_quota_us has not been changed.

In addition, it seems that koordlet modifies the cfs_quota_us file through os.WriteFile.

```go
package main

import (
	"fmt"

	cgroupsystemd "github.com/opencontainers/runc/libcontainer/cgroups/systemd"
	libcontainerconfigs "github.com/opencontainers/runc/libcontainer/configs"
)

func main() {

	resources := &libcontainerconfigs.Resources{}
	resources.CpuShares = 8
	cgroups := &libcontainerconfigs.Cgroup{
		Resources: resources,
	}
	cgroups.Name = "kubepods-besteffort.slice"
	cgroups.Parent = "kubepods.slice"

	paths := map[string]string{
		"cpu": "/sys/fs/cgroup/cpu,cpuacct/kubepods.slice/kubepods-besteffort.slice",
	}

	manager, err := cgroupsystemd.NewLegacyManager(cgroups, paths)
	if err != nil {
		fmt.Printf("failed to create cgroup manager: %v", err)
		return
	}
	manager.Set(resources)
}

```

**Environment**:
- App version: v1.2.0
- Kubernetes version (use `kubectl version`): v1.22.15
  - Containerd/Docker version: containerd v1.3.9
  - OS version: Debian 10
  - Kernal version: 4.19.0-23-amd64
  - Cgroup driver: systemd
- Others:

officialasishkumar:

- koordlet
  - Set systemd unit properties when updating QoS-level Kubernetes cgroups on systemd paths.
  - Map QoS cgroup resources to systemd properties for CPU quota and memory QoS.
  - Keep direct cgroup writes for cgroupfs, unsupported resources, and pod/container-level cgroups.
  - Route merge updates through the same systemd-aware path so memory QoS reconciliation is not bypassed.

Fixes #1734

- `go mod tidy`
- `go test ./pkg/koordlet/resourceexecutor`
- `go test ./pkg/koordlet/qosmanager/plugins/cpusuppress`
- `GOOS=darwin CGO_ENABLED=0 go test -c -o /tmp/koordinator-resourceexecutor-darwin.test ./pkg/koordlet/resourceexecutor`
- `git diff --check`

The systemd sync is limited to exact QoS-level slice paths: `kubepods.slice`, `kubepods.slice/kubepods-burstable.slice`, and `kubepods.slice/kubepods-besteffort.slice`. Pod-level and container-level cgroups continue to use direct cgroup file updates only.

### V. Checklist

### DIFF
--- a/pkg/koordlet/resourceexecutor/systemd.go
+++ b/pkg/koordlet/resourceexecutor/systemd.go
@@ -1,0 +1,247 @@
+/*
+Copyright 2026 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resourceexecutor
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	systemddbus "github.com/coreos/go-systemd/v22/dbus"
+	godbus "github.com/godbus/dbus/v5"
+	"k8s.io/klog/v2"
+
+	sysutil "github.com/koordinator-sh/koordinator/pkg/koordlet/util/system"
+)
+
+const systemdSetUnitPropertiesTimeout = 5 * time.Second
+
+var errSystemdDBusUnavailable = errors.New("systemd dbus is unavailable")
+
+var kubeQOSSystemdUnits = map[string]string{
+	"kubepods.slice": "kubepods.slice",
+	"kubepods.slice/kubepods-burstable.slice":  "kubepods-burstable.slice",
+	"kubepods.slice/kubepods-besteffort.slice": "kubepods-besteffort.slice",
+}
+
+type systemdUnitPropertyWriter interface {
+	SetUnitProperties(ctx context.Context, unitName string, runtime bool, properties ...systemddbus.Property) error
+}
+
+type systemdDBusPropertyWriter struct {
+	lock sync.Mutex
+	conn *systemddbus.Conn
+}
+
+func (w *systemdDBusPropertyWriter) SetUnitProperties(ctx context.Context, unitName string, runtime bool, properties ...systemddbus.Property) error {
+	conn, err := w.getConn(ctx)
+	if err != nil {
+		return err
+	}
+	return conn.SetUnitPropertiesContext(ctx, unitName, runtime, properties...)
+}
+
+func (w *systemdDBusPropertyWriter) getConn(ctx context.Context) (*systemddbus.Conn, error) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	if w.conn != nil {
+		return w.conn, nil
+	}
+	address := systemBusAddress()
+	if address == "" {
+		return nil, errSystemdDBusUnavailable
+	}
+	conn, err := newSystemdConnection(ctx, address)
+	if err != nil {
+		return nil, err
+	}
+	w.conn = conn
+	return conn, nil
+}
+
+var systemdPropertyWriter systemdUnitPropertyWriter = &systemdDBusPropertyWriter{}
+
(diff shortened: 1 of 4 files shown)


pkg/koordlet/resourceexecutor/systemd.go
```diff
@@ -0,0 +1,247 @@
...
+	lock sync.Mutex
+	conn *systemddbus.Conn
+}
+
+func (w *systemdDBusPropertyWriter) SetUnitProperties(ctx context.Context, unitName string, runtime bool, properties ...systemddbus.Property) error {
+	conn, err := w.getConn(ctx)
+	if err != nil {
+		return err
+	}
+	return conn.SetUnitPropertiesContext(ctx, unitName, runtime, properties...)
```
mdryaaan, line 62: If SetUnitPropertiesContext returns a transport error, w.conn is left  pointing at the broken connection. Every subsequent call reuses the  stale conn, fails silently, and the systemd property is never updated 
again , the original cfs_quota_us reset bug returns for the rest of  the koordlet process lifetime.

Fix: reset w.conn = nil when SetUnitPropertiesContext fails so the 
next call re-dials.

mdryaan (commented):
Hi @officialasishkumar, one issue in systemd.go , left an inline  comment on the stale dbus connection handling.

Everything else looks correct: path detection is implicitly systemd-only (cgroupfs nodes use kubepods/ naming which won't match the map), the CPUQuotaPerSecUSec conversion is right, and MemoryMin/ MemoryLow/MemoryHigh are all covered

cc @saintube @ZiMengSheng

ZiMengSheng:
Pre-review summary (as the assigned reviewer):

**First: thanks for picking this up.** The direction matches the consensus from the #1734 discussion — syncing QoS-level cgroup updates to systemd unit properties via `SetUnitProperties` so the values survive kubelet's own set-property calls. The scope is appropriately conservative (only the three QoS slices; cgroupfs driver no-ops via the unit-name lookup).

**Items to address before this can move forward:**

1. **DCO check is failing** — the commits need `Signed-off-by` (`git commit --amend -s` and force-push).

2. **CI is red beyond DCO** (license-check, slo-controller) — likely needs a rebase onto current main; the PR has been open since May.

3. **Fail-open vs fail-closed on dbus errors**: `updateSystemdUnitPropertyForCgroup` returns the dbus error directly, which fails the whole cgroup update path (`MergeFuncUpdateCgroup` and `cgroupWriteIfDifferentWithLog`). When the systemd dbus is unavailable (e.g. transient dbus issues, or a node where the address lookup fails), koordlet would stop updating QoS cgroups entirely instead of falling back to the direct cgroupfs write that the old behavior used. Consider logging the error and falling back to the cgroupfs-only path, or at least making the failure mode explicit — this changes behavior on systemd nodes from "flaky quota" to "no updates at all" if dbus breaks.

4. **Patch coverage is 48%** — the dbus connection path, the timeout context, and the QoS unit mapping deserve test coverage given the fix targets a 3-year-old production bug.

Given the PR has been idle since May, please treat this as needing a response — if there's no update by the stale deadline (Sep 27), it will be closed automatically. If anyone else wants to take it over, the review points above are the roadmap.

PR #2940, opened 2026-05-13
ExtendedResourceNames: add mid-tier resources

### LINKED ISSUE #2939: [util] Add Mid resource (MidCPU, MidMemory) to ExtendedResourceNames
## Summary

In `pkg/util/extended_resource.go`, the `ExtendedResourceNames` variable currently only includes Batch-tier resources (`BatchCPU` and `BatchMemory`), while Mid-tier resources (`MidCPU` and `MidMemory`) are missing. This is already noted by an inline TODO in the source code.

## Current Behavior

```go
// ExtendedResourceNames todo: add mid resource
var ExtendedResourceNames = []corev1.ResourceName{
    extension.BatchCPU,
    extension.BatchMemory,
}
```

The same file already contains helper functions for Mid resources, including:

- `GetMidMilliCPUFromResourceList`
- `GetMidMemoryFromResourceList`
- `GetContainerMidMilliCPURequest`
- `GetContainerMidMemoryByteRequest`

However, `ExtendedResourceNames` does not include:

- `extension.MidCPU`
- `extension.MidMemory`

## Expected Behavior

`ExtendedResourceNames` should include both Batch and Mid extended resources for consistency with the existing utility functions in the file.

```go
var ExtendedResourceNames = []corev1.ResourceName{
    extension.BatchCPU,
    extension.BatchMemory,
    extension.MidCPU,
    extension.MidMemory,
}
```

Vatsalpatni73:

This PR adds the missing Mid-tier resources (`extension.MidCPU` and `extension.MidMemory`) to the `ExtendedResourceNames` slice in `pkg/util/extended_resource.go`. This change resolves an existing `todo` comment and ensures consistency, as the same file already contains helper functions dealing with Mid-tier resources (e.g., `GetMidMilliCPUFromResourceList`, `GetMidMemoryFromResourceList`).

fixes #2939

### V. Checklist

### DIFF
--- a/pkg/util/extended_resource.go
+++ b/pkg/util/extended_resource.go
@@ -24,10 +24,12 @@ import (
 
 // NOTE: functions in this file can be overwritten for extension
 
-// ExtendedResourceNames todo: add mid resource
+// ExtendedResourceNames
 var ExtendedResourceNames = []corev1.ResourceName{
 	extension.BatchCPU,
 	extension.BatchMemory,
+	extension.MidCPU,
+	extension.MidMemory,
 }
 
 func GetBatchMilliCPUFromResourceList(r corev1.ResourceList) int64 {


saintube:
PTAL /cc @lijunxin559

IITI-tushar:
@saintube @Vatsalpatni73 
I was looking into this change and had a few questions before adding `MidCPU` / `MidMemory` to `ExtendedResourceNames`.

Currently the codebase seems to treat Batch and Mid resources somewhat differently:

* `batchresource` and `midresource` plugins already maintain separate `ResourceNames` lists
* the webhook mutator (`extended_resource_spec.go`) only handles Batch resources
* quota evaluation already includes both Batch and Mid resources separately

A few questions:

1. Is there a concrete bug/use-case caused by Mid resources missing from `ExtendedResourceNames`, or is this mainly for API consistency with the helper functions already present in `extended_resource.go`?

2. Have all current callers of `ExtendedResourceNames` / `GetPodExtendedResources()` been audited to ensure Mid resources should also be processed there?

3. If Mid resources are intended to behave the same as Batch resources globally, should the webhook mutation path also be updated to include `MidCPU` / `MidMemory`?

4. Is there a reason the architecture currently keeps separate plugin-level `ResourceNames` lists for Batch and Mid resources instead of one unified list?

5. I also noticed helper functions for Mid resources exist but don’t appear to have dedicated tests yet. Would it make sense to add unit tests for Mid resource handling as part of this change?

Just trying to better understand whether this is a straightforward consistency fix or whether there are broader behavioral implications for resource handling paths.

pkg/util/extended_resource.go
```diff
@@ -24,10 +24,12 @@ import (
 
 // NOTE: functions in this file can be overwritten for extension
 
-// ExtendedResourceNames todo: add mid resource
+// ExtendedResourceNames
 var ExtendedResourceNames = []corev1.ResourceName{
 	extension.BatchCPU,
 	extension.BatchMemory,
+	extension.MidCPU,
```
IITI-tushar, line 31: You implemented the TODO, but I think this is not the full implementation!!!

Vatsalpatni73:
@IITI-tushar
Thanks for the feedback.

My initial intention was mainly to address the existing TODO in `extended_resource.go` and make `ExtendedResourceNames` consistent with the helper functions already present in the same file, rather than proposing a broader architectural change for Mid resource handling.

Your points about the separate plugin-level `ResourceNames` lists, the webhook mutator only handling Batch resources, and quota evaluation already treating Batch/Mid separately make sense, and I agree they should be considered before changing shared utility behavior globally.

@saintube Could you please guide whether this should remain a narrow TODO fix, or whether the related Mid resource handling paths should also be updated as part of this change?

lijunxin559:
Hello, @Vatsalpatni73 
Todo  is not meaning simply added. If there is no functional dependency, it does not need to be added for the time being. If there is a functional dependency, it needs to be fully implemented.It is recommended to start from the needs.

Vatsalpatni73:
@lijunxin559
Thank you for the clarification! Since there is no current functional dependency requiring Mid resources, I will go ahead and close this PR and the associated issue (#2939). We can revisit this in the future if a true functional need arises.
Thanks again for the guidance!

PR #2950, opened 2026-05-14
deviceshare: fix nil map panic in device plugin adapters

### LINKED ISSUE #2947: [BUG]: deviceshare/device plugin adapters panic with nil pointer when pod has no annotations or labels
**What happened**:

All five device plugin adapters (`generalDevicePluginAdapter`, `generalGPUDevicePluginAdapter`, `huaweiGPUDevicePluginAdapter`, `cambriconGPUDevicePluginAdapter`, `metaxDevicePluginAdapter`) call `object.GetAnnotations()[key] = value` and `object.GetLabels()[key] = value` directly on the map returned by the getter. In Go, `GetAnnotations()` returns `nil` when the object has no annotations, and writing to a `nil` map panics.

Example:
- test fails intermittently when pods have no annotations
- scheduler crashes during GPU/device pre-bind

1. Create a GPU pod with NO annotations and no labels set (e.g., a bare pod with only `spec.containers`).
2. Schedule it on a node with a GPU device.
3. Let the scheduler reach the `PreBind` phase.

* The annotation writes are currently safe _only_ because `SetDeviceAllocations` is called before `adaptForDevicePlugin` in `preBindObject`, leaving annotations non-nil. However the adapters have no such contract; future refactors or direct adapter calls (e.g., in tests) will panic.
* The label write at line 292 (`object.GetLabels()[apiext.LabelHAMIVGPUNodeName] = ctx.node.Name`) is conditionally reachable via a label read guard but the same unsafe pattern applies.
* All five adapters must be fixed: `generalDevicePluginAdapter`, `generalGPUDevicePluginAdapter`, `huaweiGPUDevicePluginAdapter`, `cambriconGPUDevicePluginAdapter`, `metaxDevicePluginAdapter`.

**Environment**:
- App version:
- Kubernetes version (use kubectl version):v1.36.0
  - OS version: Ubuntu 24.04
  - Kernal version:
- Others:

madmecodes:
## What type of PR is this?

/kind bug

All five device plugin adapters (`generalDevicePluginAdapter`, `generalGPUDevicePluginAdapter`, `huaweiGPUDevicePluginAdapter`, `cambriconGPUDevicePluginAdapter`, `metaxDevicePluginAdapter`) write directly to the map returned by `GetAnnotations()` / `GetLabels()`. When a pod has no pre-existing annotations or labels, these getters return `nil`, and writing to a `nil` map panics.

This PR adds `ensureAnnotations` and `ensureLabels` helpers that initialize the map via `SetAnnotations` / `SetLabels` if it is `nil`, then returns it for safe writing. All five adapters are updated to use these helpers.

Test cases for nil annotations are added for `generalDevicePluginAdapter` and `generalGPUDevicePluginAdapter`.

Fixes #2947

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
+++ b/pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go
@@ -272,12 +272,30 @@ func unlockNode(node *corev1.Node, lockKey string) {
 	delete(node.Annotations, lockKey)
 }
 
+func ensureAnnotations(object metav1.Object) map[string]string {
+	annotations := object.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+		object.SetAnnotations(annotations)
+	}
+	return annotations
+}
+
+func ensureLabels(object metav1.Object) map[string]string {
+	labels := object.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+		object.SetLabels(labels)
+	}
+	return labels
+}
+
 // generalDevicePluginAdapter annotates the bind timestamp to pod which enables users to write their own device plugins
 // that can be used with koord-scheduler.
 type generalDevicePluginAdapter struct{}
 
 func (a *generalDevicePluginAdapter) Adapt(_ *DevicePluginAdaptContext, object metav1.Object, _ []*apiext.DeviceAllocation) error {
-	object.GetAnnotations()[AnnotationBindTimestamp] = strconv.FormatInt(dpAdapterClock.Now().UnixNano(), 10)
+	ensureAnnotations(object)[AnnotationBindTimestamp] = strconv.FormatInt(dpAdapterClock.Now().UnixNano(), 10)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


ZiMengSheng:
@zqzten PTAL

zqzten (commented):
The nil cases won't happen.
For annotation: The adapt logic always happen after `apiext.SetDeviceAllocations` which has already init annotations for pod.
For label: The current logic only happen if the pod has another specific label.

ZiMengSheng:
Closing due to inactivity (last update May 2026). **Important: the bug is real and confirmed still on main** — `pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go` still has 8 direct `GetAnnotations()[key] = ...` / `GetLabels()[key] = ...` writes that panic when annotations/labels are nil. Your `ensureAnnotations`/`ensureLabels` helper approach here is the right shape for the fix. Anyone is welcome to pick this up (reopen or fresh PR referencing this one — happy to fast-track the review).

PR #2953, opened 2026-05-15
feat: support custom GPU/NPU resources in gpudeviceresource plugin

### LINKED ISSUE #2952: feat: support custom GPU/NPU resources in gpudeviceresource plugin
## Summary

The `gpudeviceresource` plugin currently reconciles only a hardcoded set of GPU/NPU resource names from devices to nodes.

File:
`pkg/slo-controller/noderesource/plugins/gpudeviceresource/utils.go`

```go
// ResourceNames are the known resources reconciled by the plugin from devices to nodes.
// TODO: support custom resources.
ResourceNames = []corev1.ResourceName{
    extension.ResourceGPU,
    extension.ResourceGPUCore,
    extension.ResourceHuaweiNPUCore,
    extension.ResourceHuaweiNPUCPU,
    extension.ResourceHuaweiNPUDVPP,
    extension.ResourceGPUMemory,
    extension.ResourceGPUMemoryRatio,
    extension.ResourceGPUShared,
}
```

This makes it difficult to support custom GPU/NPU resource types from third-party vendors or internal hardware without modifying the source code.

## Proposal

Support configurable custom resource names for reconciliation instead of relying only on the hardcoded list.

Possible approaches:
- plugin configuration
- ConfigMap/CRD-based configuration
- registration mechanism for extensions

## Acceptance Criteria

- [ ] Custom resource names can be configured
- [ ] Existing default resources continue to work
- [ ] Unit tests cover the custom resource path
- [ ] Documentation is updated

Vatsalpatni73:

This PR adds configurable registration functionality in the `gpudeviceresource` plugin to support custom GPU/NPU resources instead of relying solely on a hardcoded list.

- Added `RegisterGPUDeviceResource()` to allow custom GPU/NPU resource types to be registered at runtime.
- Added `RegisterGPUDeviceLabel()` to allow configurable device labels to be registered at runtime.
- Removed the `TODO: support custom resources.` comment since this functionality is now implemented via the new registration mechanism.
- Added unit tests for the new custom resource logic ensuring new items are appended correctly and duplicate registrations are safely ignored.

This allows other parts of the codebase, extensions, or third-party vendor integrations to easily inject their specific resource types and labels into the plugin prior to reconciliation without modifying the plugin's source directly.

fixes #2952

### V. Checklist

### DIFF
--- a/pkg/slo-controller/noderesource/plugins/gpudeviceresource/utils.go
+++ b/pkg/slo-controller/noderesource/plugins/gpudeviceresource/utils.go
@@ -24,7 +24,7 @@ import (
 
 var (
 	// ResourceNames are the known resources reconciled by the plugin from devices to nodes.
-	// TODO: support custom resources.
+	// Applications can use RegisterGPUDeviceResource to add custom resources.
 	ResourceNames = []corev1.ResourceName{
 		extension.ResourceGPU,
 		extension.ResourceGPUCore,
@@ -42,3 +42,23 @@ var (
 		extension.LabelGPUDriverVersion,
 	}
 )
+
+// RegisterGPUDeviceResource registers a custom GPU/NPU resource to be reconciled.
+func RegisterGPUDeviceResource(resourceName corev1.ResourceName) {
+	for _, name := range ResourceNames {
+		if name == resourceName {
+			return
+		}
+	}
+	ResourceNames = append(ResourceNames, resourceName)
+}
+
+// RegisterGPUDeviceLabel registers a custom GPU/NPU label to be reconciled.
+func RegisterGPUDeviceLabel(label string) {
+	for _, l := range Labels {
+		if l == label {
+			return
+		}
+	}
+	Labels = append(Labels, label)
+}
--- a/pkg/slo-controller/noderesource/plugins/gpudeviceresource/plugin.go
+++ b/pkg/slo-controller/noderesource/plugins/gpudeviceresource/plugin.go
@@ -119,14 +119,10 @@ func (p *Plugin) Prepare(_ *configuration.ColocationStrategy, node *corev1.Node,
 	// prepare node labels
 	// TBD: shall we reset labels if not exist in the NR
 	if nr.Labels != nil {
-		if _, ok := nr.Labels[extension.LabelGPUVendor]; ok {
-			node.Labels[extension.LabelGPUVendor] = nr.Labels[extension.LabelGPUVendor]
-		}
-		if _, ok := nr.Labels[extension.LabelGPUModel]; ok {
-			node.Labels[extension.LabelGPUModel] = nr.Labels[extension.LabelGPUModel]
-		}
-		if _, ok := nr.Labels[extension.LabelGPUDriverVersion]; ok {
-			node.Labels[extension.LabelGPUDriverVersion] = nr.Labels[extension.LabelGPUDriverVersion]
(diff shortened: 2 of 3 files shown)


pkg/slo-controller/noderesource/plugins/gpudeviceresource/utils.go
```diff
@@ -42,3 +42,23 @@ var (
...
 )
+
+// RegisterGPUDeviceResource registers a custom GPU/NPU resource to be reconciled.
+func RegisterGPUDeviceResource(resourceName corev1.ResourceName) {
+	for _, name := range ResourceNames {
+		if name == resourceName {
+			return
+		}
+	}
+	ResourceNames = append(ResourceNames, resourceName)
```
mdryaaan, line 53: RegisterGPUDeviceLabel adds the label to the Labels slice but it never gets propagated to the node. calculate() and Prepare() in plugin.go both hardcode LabelGPUVendor, LabelGPUModel, LabelGPUDriverVersion  they never iterate the labels slice, so a custom label registered here will affect NeedSyncMeta change detection but never actually get written to the node.

mdryaan (commented):
Hi @Vatsalpatni73,  I reviewed you PR two thing i found
- RegisterGPUDeviceLabel doesn't actually propagate custom labels  left an inline comment explaining why.
- You mention unit test added,  but you missed to add

cc @saintube @ZiMengSheng PTAL

pkg/slo-controller/noderesource/plugins/gpudeviceresource/utils.go
Vatsalpatni73: Hi @mdryaan, thanks for the review. I have just pushed the code to address the issues.

zqzten (commented):
Could you provide some practical use case of this feat?

Vatsalpatni73:
Hey @zqzten, sure! A couple of real cases are:

1. New accelerator vendors (Cambricon, Enflame, Biren, etc.) each use their own resource names, so right now supporting them means patching `utils.go` and carrying that diff. With `RegisterGPUDeviceResource()` they can just register their resource name from their own package instead.

2. Some clusters run in-house GPU-sharing setups with custom resource names (e.g. `internal.company.com/gpu-vcore`), this lets them hook into the reconciliation path without forking koordinator.

Vatsalpatni73:
Basically it's for anyone who needs to add a resource type without upstreaming a PR every time.

zqzten:
> 1. New accelerator vendors (Cambricon, Enflame, Biren, etc.) each use their own resource names, so right now supporting them means patching `utils.go` and carrying that diff. With `RegisterGPUDeviceResource()` they can just register their resource name from their own package instead.
> 2. Some clusters run in-house GPU-sharing setups with custom resource names (e.g. `internal.company.com/gpu-vcore`), this lets them hook into the reconciliation path without forking koordinator.

Actually the node resource controller doesn't need to care about the vendor specific device plugin resource names (because these would be reported by vendor dp themselves), it only reports resources that koord-scheduler uses for fine-grained device scheduling, which are defined by koordinator itself. As you can see there're no vendor resource names in the list for now, and the only special case is Huawei NPU, because it needs special scheduling logic which need additional resources (see #2496 for details). For others, we don't need to introduce additional resources to list. So in short I'm not seeing any out-of-tree usage of this for now.

PR #2956, opened 2026-05-15
fix: data race on alreadyAttemptedPods in gang scheduling context

### LINKED ISSUE #2955: [BUG]: coscheduling: data race on `alreadyAttemptedPods` in gang scheduling context causes concurrent map corruption
**What happened**:

GangSchedulingContext.alreadyAttemptedPods is a shared sets.Set[string] protected by the struct’s sync.RWMutex. However, some code paths access it without holding the lock, while NextPod updates it under a write lock.

Currently:

- NextPod checks Has() without a lock
- NextPod performs Insert() with a lock
- several network topology workflow paths read len(alreadyAttemptedPods) without any lock
- 

This can lead to races when multiple goroutines schedule pods from the same gang concurrently.

Possible effects:

- concurrent map read/write panic
- duplicate scheduling attempts for the same pod
- inconsistent scheduling state under contention

For example, two NextPod calls may both pass the unlocked Has() check before either goroutine acquires the write lock and inserts the pod into the set.

1. Enable `-race` flag in the scheduler build.
2. Configure a gang with 3+ pods and a network topology constraint.
3. Trigger scheduling of the gang pods.

Commands:

```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=5
```

Observed output/error:

```text
WARNING: DATA RACE
Read at 0x... by goroutine ...:
  runtime.mapaccess1_faststr(...)
  ...coscheduling/core/network_topology_workflow.go:58

Previous write at 0x... by goroutine ...:
  runtime.mapassign_faststr(...)
  ...coscheduling/core/core.go:185
```

* The comment on line 102 of `gang_context.go` says "secure alreadyAttemptedPods to avoid concurrent map read and write", confirming this was a known concern that was incompletely addressed.
* The `Has()` check at line 183 also creates a TOCTOU window: two goroutines can both see `Has() == false`, both acquire the lock sequentially, and both insert the same pod — causing a pod to be scheduled twice.

**Environment**:
- App version: main
  - OS version: Ubuntu 24.04
  - Kernal version:
- Others:

IITI-tushar:

`GangSchedulingContext` embeds `sync.RWMutex` to protect `alreadyAttemptedPods`, but several code paths accessed the field without holding that lock.

**In `NextPod`** — the `Has()` check raced with the subsequent `Insert()`:
```go
// BEFORE — Has() without lock, then Lock() + Insert()
if !gangSchedulingContext.alreadyAttemptedPods.Has(podKey) {   // ← unlocked read
    gangSchedulingContext.Lock()
    gangSchedulingContext.alreadyAttemptedPods.Insert(podKey)
    gangSchedulingContext.Unlock()
    return pod
}
```
Two goroutines could both observe `Has() == false`, then both insert the same pod and return it twice. Fix: hold the lock for the entire check-and-insert:
```go
// AFTER — atomic check-and-insert
gangSchedulingContext.Lock()
alreadyAttempted := gangSchedulingContext.alreadyAttemptedPods.Has(podKey)
if !alreadyAttempted {
    gangSchedulingContext.alreadyAttemptedPods.Insert(podKey)
}
gangSchedulingContext.Unlock()
if !alreadyAttempted {
    return frameworkext.CopyQueueInfoToPod(firstPod, pod)
}
```

**In `PreFilter`, `FindOneNode`, `PlanNodes`** — `len()` was called without RLock. Fix: wrap in `RLock`/`RUnlock`:
```go
gangSchedulingContext.RLock()
attempted := len(gangSchedulingContext.alreadyAttemptedPods)
gangSchedulingContext.RUnlock()
if attempted > 1 {
```

No behaviour change for single-goroutine scenarios.

fixes #2955

#### Reproduce the original issue

```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=10
```

Expected failure/error:

```text
WARNING: DATA RACE
```

#### Verify the fix

Run:

```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=10
```

Expected result:

* no data race warnings
* tests pass

* The check-and-insert atomicity fix in `NextPod` is the higher-severity portion — it can cause the same pod to be scheduled twice.
* The `len()` fixes in network topology code are lower risk in practice (scheduler cycles are sequential per pod) but still data races by Go's memory model.

---

### V. Checklist

* [x] I have written necessary docs and comments
* [x] I have added necessary unit tests and integration tests
* [x] All checks passed in `make test`

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -180,10 +180,13 @@ func (pgMgr *PodGroupManager) NextPod() *corev1.Pod {
 		for _, pod := range pods {
 			podKey := util.GetId(pod.Namespace, pod.Name)
 
-			if !gangSchedulingContext.alreadyAttemptedPods.Has(podKey) {
-				gangSchedulingContext.Lock()
+			gangSchedulingContext.Lock()
+			alreadyAttempted := gangSchedulingContext.alreadyAttemptedPods.Has(podKey)
+			if !alreadyAttempted {
 				gangSchedulingContext.alreadyAttemptedPods.Insert(podKey)
-				gangSchedulingContext.Unlock()
+			}
+			gangSchedulingContext.Unlock()
+			if !alreadyAttempted {
 				klog.Infof("NextPod: return pod %s/%s/%s, gangGroup: %+v, gangGroupStartTime: %+v", pod.Namespace, pod.Name, pod.UID, gangGroup, gangSchedulingContext.startTime)
 				// correct podInfo.Time and podInfo.Attempts
 				return frameworkext.CopyQueueInfoToPod(firstPod, pod)
--- a/pkg/scheduler/plugins/coscheduling/core/core_test.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core_test.go
@@ -18,7 +18,9 @@ package core
 
 import (
 	"context"
+	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


ZiMengSheng:
Closing due to inactivity (last update May 2026). **The data race is confirmed still on main**: in `coscheduling/core/core.go` `NextPod()`, `gangSchedulingContext.alreadyAttemptedPods.Has(podKey)` is called outside the lock while `Insert` is inside — a concurrent read/write race on the shared set. Anyone is welcome to pick this up (fresh PR or reopen); happy to fast-track the review.

PR #2959, opened 2026-05-15
webhook: enable mutation for pod update requests

### LINKED ISSUE #2954: [webhook/pod] Implement mutating logic for Pod Update in handleUpdate
## Summary

The `handleUpdate` method in `pkg/webhook/pod/mutating/mutating_handler.go` currently only contains a TODO and returns `nil` without invoking any mutating plugins.

```go
func (h *PodMutatingHandler) handleUpdate(ctx context.Context, req admission.Request, obj *corev1.Pod) error {
    // TODO: add mutating logic for pod update here
    return nil
}
```

In contrast, `handleCreate` executes multiple mutating stages for pod admission requests, including:

- `ClusterColocationProfile`
- `ExtendedResourceSpec`
- `MultiQuotaTree`
- `DeviceResourceSpec`

along with webhook duration metric recording for each stage.

As a result, pod update requests currently bypass all mutation logic, which may lead to inconsistent behavior between pod creation and update flows.

## Proposed Work

- Review which existing mutating plugins should also handle pod update events.
- Implement the required mutation flow inside `handleUpdate`, following the same structure used in `handleCreate`.
- Add webhook duration metric recording for update mutations.
- Add unit tests covering the update path.

Dev10-sys:
This fixes #2954 by enabling pod mutation logic during Update operations.

Previously, pod updates bypassed all mutation stages because `handleUpdate` was a no-op. This change reuses the existing mutation flow for both Create and Update requests while keeping plugin-specific operation checks unchanged.

The patch also:

* fixes a potential nil map panic in device resource mutation handling
* adds update path unit tests for extended resources and GPU mutations
* keeps the implementation minimal and aligned with existing webhook patterns

Tests run:

* `go test ./pkg/webhook/pod/mutating/...`
* `go test ./pkg/webhook/...`
* `go vet ./pkg/webhook/...`

### DIFF
--- a/pkg/webhook/pod/mutating/cluster_colocation_profile.go
+++ b/pkg/webhook/pod/mutating/cluster_colocation_profile.go
@@ -51,7 +51,7 @@ var (
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.koordinator.sh,resources=clustercolocationprofiles,verbs=get;list;watch
 
-func (h *PodMutatingHandler) clusterColocationProfileMutatingPod(ctx context.Context, req admission.Request, pod *corev1.Pod) error {
+func (h *PodMutatingHandler) clusterColocationProfileMutatingPod(ctx context.Context, req admission.Request, pod, oldPod *corev1.Pod) error {
 	if req.Operation != admissionv1.Create {
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


mdryaaan:
Hi @Dev10-sys , this is duplicate of https://github.com/koordinator-sh/koordinator/pull/2958

ZiMengSheng:
Closing in favor of #2958 — both implement the `TODO: add mutating logic for pod update here` in the pod mutating webhook, and #2958 is being actively carried in the current mentee review batch. Feel free to reopen if you'd like to continue driving this instead.

PR #2961, opened 2026-05-16
fix: include init containers in BE pod resource calculations

### LINKED ISSUE #2960: [BUG] BE pod resource calculation ignores init containers and pod overhead
**What happened:**
Four functions in `pkg/util/pod_resources_utils.go` sum only `pod.Spec.Containers` and skip `pod.Spec.InitContainers` and `pod.Spec.Overhead`. All four have explicit `// TODO: count init containers and pod overhead` comments. koordlet uses these values for BE CPU suppression thresholds and memory eviction decisions. Pods with heavyweight init containers are undercounted.

**How to reproduce:**
```bash
go test ./pkg/util/... -v -run TestGetPodBEMilliCPURequest
```

Before the fix, a pod with an init container requesting `4000m` CPU and a main container requesting `100m` CPU would make `GetPodBEMilliCPURequest` return `100` — the init container is ignored. A correct implementation returns `4000` (Kubernetes semantics: `max(init_max, container_sum)`).

Init container resources are included using Kubernetes scheduling semantics: `max(max(init_i), sum(containers)) + overhead`.

**What actually happened:**
`GetPodBEMilliCPURequest`, `GetPodBEMilliCPULimit`, `GetPodBEMemoryByteRequestIgnoreUnlimited`, and `GetPodBEMemoryByteLimit` all return results computed from `pod.Spec.Containers` only.

**Relevant log output:**

Image

Image

**Environment:**

- `pkg/util/pod_resources_utils.go:57–115` — all four functions have the same gap.
- `pkg/util/pod.go:47` has the same `TODO` for `GetPodTargetExtendedResources`.
- `pkg/webhook/pod/mutating/extended_resource_spec.go:45` has the same gap in webhook-side resource computation.
- Kubernetes scheduling semantics: effective pod resource = `max(max(init_i), sum(containers)) + overhead`.

mdryaaan:

  Four functions in `pkg/util/pod_resources_utils.go` calculated BE pod CPU and memory resources by summing only `pod.Spec.Containers`, ignoring  `pod.Spec.InitContainers`. This caused koordlet to undercount resource usage for pods with heavyweight init containers, leading to incorrect CPU suppression  thresholds and memory eviction decisions.

 The fix applies Kubernetes scheduling semantics — `max(max(init_i), sum(containers))` — to all four BE resource functions, matching the pattern already used by  `GetPodMilliCPULimit` in the same file. The same gap is fixed in `GetPodTargetExtendedResources` (pkg/util/pod.go) and `mutateByExtendedResources` (pkg/webhook/pod/mutating/extended_resource_spec.go), which also build per-container resource maps used by koordlet and the admission webhook.

  fixes #2960 

  ```bash
  go test ./pkg/util/... -v -run 'Test_GetPodBEMilliCPURequest|Test_GetPodBEMemoryRequest|TestGetPodTargetExtendedResources'
  go test ./pkg/webhook/pod/mutating/... -v -run TestExtendedResourceSpecMutatingPod
```
Before this fix, GetPodBEMilliCPURequest on a pod with a 4000m init container and a 100m regular container returned 100. After the fix it returns 4000. All existing tests continue to pass.

## Before fix:

Screenshot 2026-05-16 083651

Screenshot 2026-05-16 083818

## After fix

Screenshot 2026-05-16 083844

 ###  Ⅳ. Special notes for reviews

The init container loop pattern follows GetPodMilliCPULimit exactly: sum regular containers first, then MaxInt64(running_total, init_value) for each init  container. No behavior change for pods with no init containers.

 ###  V. Checklist- [x] I have written necessary docs and comments

### DIFF
--- a/pkg/util/pod.go
+++ b/pkg/util/pod.go
@@ -44,9 +44,16 @@ func GetPodTargetExtendedResources(pod *corev1.Pod, resourceNames ...corev1.Reso
 
 	extendedResources := GetEmptyPodExtendedResources()
 
-	// TODO: count init containers and pod overhead
 	for i := range pod.Spec.Containers {
 		container := &pod.Spec.Containers[i]
+		r := GetContainerTargetExtendedResources(container, resourceNames...)
+		if r == nil {
+			continue
+		}
+		extendedResources.Containers[container.Name] = *r
+	}
+	for i := range pod.Spec.InitContainers {
+		container := &pod.Spec.InitContainers[i]
 		r := GetContainerTargetExtendedResources(container, resourceNames...)
 		if r == nil {
 			continue
--- a/pkg/util/pod_resources_utils.go
+++ b/pkg/util/pod_resources_utils.go
@@ -56,7 +56,6 @@ func GetPodRequest(pod *corev1.Pod, resourceNames ...corev1.ResourceName) corev1
 
 func GetPodBEMilliCPURequest(pod *corev1.Pod) int64 {
 	podCPUMilliReq := int64(0)
-	// TODO: count init containers and pod overhead
 	for _, container := range pod.Spec.Containers {
 		containerCPUMilliReq := GetContainerBatchMilliCPURequest(&container)
 		if containerCPUMilliReq <= 0 {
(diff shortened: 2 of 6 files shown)


ZiMengSheng:
Thanks for working on this! One important update since this PR was opened: **#3020 (fix requests computation for native sidecar containers) has been merged** and already adds InitContainers handling to `GetPodBEMilliCPURequest`/`GetPodBEMilliCPULimit`/`GetPodBEMemoryByte*` in `pkg/util/pod_resources_utils.go` — but it only covers **sidecar** init containers (summed, since sidecars run concurrently).

This PR currently takes `MaxInt64` over **all** init containers, which now conflicts with #3020's semantics for sidecars (sidecars must be summed, not maxed). To move forward, could you:

1. **Rebase onto current main** — the file will have diverged.
2. **Restrict the max-semantics to non-sidecar init containers** (`!IsSidecarContainer(c)`, matching K8s pod resource semantics where plain init containers run sequentially so effective request = max), and let #3020's sidecar summing stand.
3. The `GetPodTargetExtendedResources` part in `pkg/util/pod.go` (the `TODO: count init containers and pod overhead` at line 47) is still untouched on main — that part of this PR remains valuable, but also needs the sidecar/non-sidecar distinction.

Happy to re-review once rebased.

PR #2964, opened 2026-05-16
rbac: add nodes/finalizers permission to koordlet

### LINKED ISSUE #2941: [BUG] failed to create NodeResourceTopology my-worker-0 because of missing `nodes/finalizers` permission
**What happened**:
I observed lines like the following in koordlet logs:
```
states_noderesourcetopology.go:294] failed to create NodeResourceTopology my-worker-0, err: noderesourcetopologies.topology.node.k8s.io "my-worker-0" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

No such errors.

Deploy using the v1.8.0 version of the Helm chart.

Adding `nodes/finalizers` to the `koordlet-role` ClusterRole fixes the issue, but I’m surprised this hasn’t been reported before. Am I missing sth?

**Environment**:
- App version: v1.8.0
- Kubernetes version (use `kubectl version`): v1.27

Dev10-sys:
Title:
rbac: add nodes/finalizers permission to koordlet

Description:
This fixes #2941 by adding the missing `nodes/finalizers` permission to the koordlet ClusterRole.

Koordlet creates NodeResourceTopology and Device resources with `blockOwnerDeletion=true` owner references pointing to the host Node. Kubernetes requires access to the `nodes/finalizers` subresource for this behavior, otherwise resource creation fails with forbidden errors during resource creation.

The change is intentionally minimal and only adds the required RBAC permission.

### DIFF
--- a/config/rbac/koordlet.yaml
+++ b/config/rbac/koordlet.yaml
@@ -32,6 +32,7 @@ rules:
     - nodes
     - nodes/status
     - nodes/proxy
+    - nodes/finalizers
     - pods
     - pods/status
     - persistentvolumeclaims


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 14 with no activity since. Feel free to reopen if still needed.

PR #2967, opened 2026-05-16
fix data race on Gang.HasGangInit reads

### LINKED ISSUE #2966: [BUG]: coscheduling: gang.HasGangInit field read without lock in scheduling paths
**What happened**:

`Gang.HasGangInit` is written under `gang.lock.Lock()` in `tryInitByPodConfig` and `tryInitByPodGroup`, but several scheduling paths read it directly without holding `gang.lock.RLock()`.

Current unsafe reads:

- `core.go:219` — `PreEnqueue`
- `core.go:266` — `basicGangRequirementsCheck`
- `core.go:307` — `BeforePreFilter`

In the multi-threaded scheduler, informer callbacks can initialize gangs concurrently while scheduling goroutines are executing these checks, creating a read/write data race.

This may lead to:
- race detector warnings,
- inconsistent scheduling behaviour,
- or stale `"gang has not init"` errors for already initialized gangs.

`HasGangInit` should only be accessed through a lock-protected accessor while holding `gang.lock.RLock()`.

Run the scheduler tests with the race detector enabled:

```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=20
```
Under concurrent gang initialization and scheduling, the race detector reports concurrent read/write access on `HasGangInit`.

Other code paths such as:

```
- isGangWorthRequeue()
- isGangValidForPermit()
- GetGangSummary()
```

already access `HasGangInit` safely under `gang.lock.RLock().

`The unsafe reads in core.go are inconsistent with the rest of the implementation.

The same issue also exists for the bare `gang.WaitTime` read in Permit, while a safe accessor (`getGangWaitTime())` already exists.

**Environment**:
- App version:main
  - OS version:Ubuntu 24.04
  - Kernal version:
- Others:

IITI-tushar:

`Gang.HasGangInit` was written under `gang.lock.Lock()` but read directly in multiple scheduling paths without synchronization:

- `PreEnqueue`
- `basicGangRequirementsCheck`
- `BeforePreFilter`

This creates a data race between informer callbacks initializing gangs and scheduler goroutines reading gang state.

This PR adds a lock-protected accessor:

```go
func (gang *Gang) hasGangInit() bool {
    gang.lock.RLock()
    defer gang.lock.RUnlock()
    return gang.HasGangInit
}
```
and replaces all unsafe direct field reads with the accessor.

Additionally, this PR replaces the direct `gang.WaitTime` read in Permit with the existing safe accessor:
```go
gang.getGangWaitTime()
```
No functional behaviour changes are introduced; this PR only fixes synchronization correctness.

fixes #2966 

**Reproduce the original issue**

Run:
```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=20
```
Before the fix, the race detector may report concurrent read/write access on Gang.HasGangInit.

Example:

`WARNING: DATA RACE`

**Verify the fix**

Run:
```bash
go test ./pkg/scheduler/plugins/coscheduling/... -race -count=20
```
Expected result:

- no race detector warnings
- all tests pass

- HasGangInit is a boolean field, but Go's memory model still requires synchronization for concurrent access.
- gang.WaitTime fix is included since the lock-protected accessor already existed and follows the same synchronization pattern.
- This PR only changes synchronization access patterns and does not modify scheduling logic.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -216,7 +216,7 @@ func (pgMgr *PodGroupManager) PreEnqueue(ctx context.Context, pod *corev1.Pod) (
 	}
 
 	// check if gang is initialized
-	if !gang.HasGangInit {
+	if !gang.hasGangInit() {
 		return fmt.Errorf("gang has not init, gangName: %v, podName: %v", gang.Name,
 			util.GetId(pod.Namespace, pod.Name))
 	}
@@ -263,7 +263,7 @@ func (pgMgr *PodGroupManager) basicGangRequirementsCheck(gang *Gang, pod *corev1
 			gangsOfGangIsNil = append(gangsOfGangIsNil, gangID)
 			continue
 		}
-		if !memberGang.HasGangInit {
+		if !memberGang.hasGangInit() {
 			gangsOfGangNotInit = append(gangsOfGangNotInit, gangID)
 			continue
 		}
@@ -304,7 +304,7 @@ func (pgMgr *PodGroupManager) BeforePreFilter(ctx context.Context, cycleState fw
 	}
 
 	// check if gang is initialized
-	if !gang.HasGangInit {
+	if !gang.hasGangInit() {
 		return fmt.Errorf("gang has not init, gangName: %v, podName: %v", gang.Name,
 			util.GetId(pod.Namespace, pod.Name))
 	}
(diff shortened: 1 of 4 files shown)


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 15 with no activity since. The gang data race it fixes is real; if anyone wants to pick it up, reopening this PR or submitting a fresh fix is welcome — the analysis here is a good starting point.

Issue #2856, opened 2026-04-14
Gang Scheduling


saintube:
Upstream k8s 1.35 Native Gang Scheduling (GangScheduling Plugin) Adaptation

Kubernetes 1.35 introduces a brand-new native GangScheduling plugin (feature gate: GangScheduling, alpha, disabled by default) based on the new Workload API (`scheduling.k8s.io/v1alpha1`). This is entirely separate from Koordinator's existing Coscheduling plugin that uses PodGroup CRD.

What needs to be done:

- [ ] Design the coexistence strategy for the native Gang and the Koordinator Gang.
- [ ] Integrate with the native Workload API.
- [ ] Provide the migration guide and deprecation planning.

HirazawaUi:
I’m not entirely sure how the Koordinator community manages compatibility with upstream, but I noticed that while the Workload API was proposed in v1.35, it is still in the alpha stage and undergoing rapid changes. In fact, there was a major change recently in v1.36 (ref: https://docs.google.com/document/d/1B3kLWh_U1a2g-VQ6ExokMjmb7pA8lGkF9MafSSg3JmQ/edit?tab=t.0). 

Since introducing an unstable API without compatibility guarantees would require significant maintenance effort, would it be better to wait until the Workload API reaches the beta stage before implementing this feature?

saintube:
@HirazawaUi Good point. Given the rapid changes in the Alpha stage, we've decided against implementing compatibility for now. We view upstream compatibility as a long-term target and will likely adopt it once the API matures to Beta or later.

wolf-06:
Hi @saintube 
Can I pickup the issue and contribute ?
thank you

hharshhsaini:
Hi @saintube  @ZiMengSheng , I've been looking into this issue and put together a design proposal for the coexistence strategy.

The proposal covers:
- The architectural relationship between Koordinator's `Coscheduling` plugin and the upstream native `GangScheduling` plugin
- A lightweight annotation-based opt-out mechanism (`gang.scheduling.koordinator.sh/use-native-gang-scheduling`) to prevent interference when both systems are active
- A draft migration guide and deprecation timeline for when the upstream `scheduling.k8s.io` API matures to Beta

It's aligned with the decision to defer full Workload API integration until the upstream API stabilizes.

[...]

PR #2969, opened 2026-05-16
[scheduling] Introduce GangInfo abstraction for coscheduling        

Dev10-sys:
This PR introduces a common GangInfo abstraction layer in the coscheduling plugin to reduce duplicated gang initialization logic and prepare the scheduler for future native workload based integrations.

Changes included:

Added the GangInfo interface for different gang sources
Refactored gang initialization to use shared logic through InitByGangInfo
Added implementations for:
annotation based gangs
PodGroup CRD based gangs
native Workload based gangs
Added workload event handling in GangCache
Preserved existing PodGroup and annotation scheduling behavior

The goal of this change is to make the internal gang management flow source agnostic while keeping the existing scheduling flow unchanged.

Related:

#2856
#2851

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/core.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
 	listerv1 "k8s.io/client-go/listers/core/v1"
+	listerschedulingv1alpha1 "k8s.io/client-go/listers/scheduling/v1alpha1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	fwktype "k8s.io/kube-scheduler/framework"
@@ -95,6 +96,8 @@ type PodGroupManager struct {
 	pgClient pgclientset.Interface
 	// pgLister is podgroup lister
 	pgLister pglister.PodGroupLister
+	// workloadLister is native workload lister
+	workloadLister listerschedulingv1alpha1.WorkloadLister
 	// podLister is pod lister
 	podLister listerv1.PodLister
 	// cache stores gang info
@@ -117,14 +120,35 @@ func NewPodGroupManager(
 ) *PodGroupManager {
 	pgInformer := pgSharedInformerFactory.Scheduling().V1alpha1().PodGroups()
 	podInformer := sharedInformerFactory.Core().V1().Pods()
-	gangCache := NewGangCache(args, podInformer.Lister(), pgInformer.Lister(), pgClient, handle)
+
+	workloadAPIEnabled := false
+	if handle != nil && handle.ClientSet() != nil {
+		resources, err := handle.ClientSet().Discovery().ServerResourcesForGroupVersion("scheduling.k8s.io/v1alpha1")
+		if err == nil && resources != nil {
+			for _, res := range resources.APIResources {
+				if res.Name == "workloads" {
+					workloadAPIEnabled = true
+					break
+				}
+			}
+		}
+	}
+
+	var workloadLister listerschedulingv1alpha1.WorkloadLister
+	if workloadAPIEnabled {
+		workloadInformer := sharedInformerFactory.Scheduling().V1alpha1().Workloads()
+		workloadLister = workloadInformer.Lister()
+	}
+
+	gangCache := NewGangCache(args, podInformer.Lister(), pgInformer.Lister(), workloadLister, pgClient, handle)
 	pgMgr := &PodGroupManager{
-		handle:    handle,
-		args:      args,
-		pgClient:  pgClient,
-		pgLister:  pgInformer.Lister(),
-		podLister: podInformer.Lister(),
-		cache:     gangCache,
+		handle:         handle,
+		args:           args,
+		pgClient:       pgClient,
+		pgLister:       pgInformer.Lister(),
+		workloadLister: workloadLister,
+		podLister:      podInformer.Lister(),
+		cache:          gangCache,
 	}
 	if extHandle, ok := handle.(frameworkext.ExtendedHandle); ok {
 		pgMgr.workloadAuditor = extHandle.GetWorkloadAuditor()
@@ -141,6 +165,16 @@ func NewPodGroupManager(
 		DeleteFunc: gangCache.onPodGroupDelete,
 	}
 	frameworkexthelper.ForceSyncFromInformer(context.TODO().Done(), pgSharedInformerFactory, pgInformer.Informer(), podGroupEventHandler)
+
+	if workloadAPIEnabled {
+		workloadInformer := sharedInformerFactory.Scheduling().V1alpha1().Workloads()
+		workloadEventHandler := cache.ResourceEventHandlerFuncs{
+			AddFunc:    gangCache.onWorkloadAdd,
+			UpdateFunc: gangCache.onWorkloadUpdate,
+			DeleteFunc: gangCache.onWorkloadDelete,
+		}
+		frameworkexthelper.ForceSyncFromInformer(context.TODO().Done(), sharedInformerFactory, workloadInformer.Informer(), workloadEventHandler)
+	}
 
 	podEventHandler := cache.ResourceEventHandlerFuncs{
 		AddFunc:    gangCache.onPodAdd,
--- a/pkg/scheduler/plugins/coscheduling/core/core_test.go
+++ b/pkg/scheduler/plugins/coscheduling/core/core_test.go
@@ -605,7 +605,7 @@ func TestPodGroupManager_PostFilter(t *testing.T) {
 				args: &config.CoschedulingArgs{
 					EnablePreemption: ptr.To[bool](tt.enablePreemption),
 				},
-				cache:               NewGangCache(nil, nil, nil, nil, nil),
+				cache:               NewGangCache(nil, nil, nil, nil, nil, nil),
 				preemptionEvaluator: tt.preemptionEvaluator,
 			}
 			frameworkext.InitDiagnosis(tt.args.state, tt.args.pod)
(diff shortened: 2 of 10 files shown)


ZiMengSheng:
First pass over `6d653233`. The abstraction is the right shape and the annotation/PodGroup collapse into one code path is a real simplification, so this is worth landing — but not in this state. Building and testing it at the merge base (`6dfc4d78`, 26 commits behind) is clean, and that is the problem: three of the findings below are invisible to the suite as written.

`/hold`

## Blocking

### 1. Match-policy validation is gone, silently, on the existing paths

The description says existing PodGroup and annotation behaviour is preserved. It is not. `extension.GetGangMatchPolicy` does no validation at all — `apis/extension/coscheduling.go:116-122` returns the annotation verbatim, falling back to the alias annotation. The old code validated it against the three legal constants and fell back to `args.DefaultMatchPolicy` with an error log (`gang.go:145-150` for pods, `:228-233` for PodGroups). `AnnotationGangInfo.GetMatchPolicy` and `PodGroupGangInfo.GetMatchPolicy` only test for `""`.

Same pod, `DefaultMatchPolicy: only-waiting`, annotated `gang.scheduling.koordinator.sh/match-policy: once-satisfied-TYPO`:

```
main:      E gang.go:147] pod's annotation AnnotationGangMatchPolicy illegal, gangName: ns/g1, value: once-satisfied-TYPO
           GangMatchPolicy = "only-waiting"
this PR:   (no log)
           GangMatchPolicy = "once-satisfied-TYPO"
```

The consequence is worse than the wrong value. `isGangValidForPermit` switches on it with a `default:` arm that implements OnceSatisfied semantics (`gang.go:565-571`), while `core.go:228`, `:317`, `:397` and `:592` test `getGangMatchPolicy() == extension.GangMatchPolicyOnceSatisfied` explicitly. An unrecognised policy therefore takes the **OnceSatisfied** path in permit and the **non**-OnceSatisfied path at all four other sites. Before this change both agreed, because the value was always one of the three. A typo in an annotation now splits the plugin against itself, and `default:` is the permissive arm — `|| isGangOnceResourceSatisfied()` — so it can let a gang bind that would previously have waited.

Mode, minNum, totalNum, waitTime, gangGroups and networkTopologySpec need the same audit. I checked each and the resulting *values* match `main`, including the `totalChildrenNum < minRequiredNumber` clamp. Match policy is the one that changed.

### 2. The interface cannot express failure, so every diagnostic was deleted

Eleven getters, none returning an error. Callers use `_` throughout: `util.StringToGangGroupSlice`, `extension.GetGangTotalNum`, `extension.GetGangWaitTime`, `extension.GetNetworkTopologySpec`. Every `klog.Errorf("... illegal ...")` in the old `tryInitByPodConfig` and `tryInitByPodGroup` is gone with them.

Malformed gang annotations are an operator error, and the error log was the only channel telling them so. A user who misspells `gang.scheduling.koordinator.sh/total-number` now gets a silently different gang and nothing in the log. Sentinels have replaced errors — `AnnotationGangInfo.GetMinMember` returns `-1` on parse failure, `WorkloadGangInfo.GetMinMember` returns `0` for "not found", and `doInitByGangInfo` treats `< 0` as failure and `0` as a valid gang size. Two sources, two conventions, neither distinguishable from a real value.

Either have the getters return `(T, error)`, or give `GangInfo` a single `Parse() (*gangSpec, error)` and keep the reporting in `doInitByGangInfo` where it has the gang name to log against. The second is less churn and keeps error handling in one place.

### 3. `return` inside the PodGroup loop abandons the remaining gangs

`onWorkloadAdd:487` and `onWorkloadUpdate:522` both do `if gangCache.handle == nil { return }` inside `for _, pg := range workload.Spec.PodGroups`. A Workload with three PodGroups, `handle == nil`:

```
pg-a -> HasGangInit=true MinRequiredNumber=0
pg-b -> NOT IN CACHE
pg-c -> NOT IN CACHE
```

Not a corner case. `isGangWorthRequeue()` is `HasGangInit && len(Children) >= MinRequiredNumber`, and `MinRequiredNumber` is 0 for any PodGroup without `Policy.Gang`, so `0 >= 0` holds on the very first add and the guard is reached immediately. Hoist the nil check above the loop, or `continue`.

### 4. A PodGroup with no gang policy becomes a gang of zero

`WorkloadGangInfo.GetMinMember` returns 0 when no `PodGroups[i].Policy.Gang` matches, and `onWorkloadAdd` registers a `Gang` for every entry in `Spec.PodGroups` unconditionally. `Policy.Gang == nil` is legal upstream and means the PodGroup is not gang-scheduled:

```
pg-nogang: minMember=0 totalMember=0 gangGroups=[ns/pg-nogang ns/pg-gang]
pg-gang:   minMember=4 totalMember=4 gangGroups=[ns/pg-nogang ns/pg-gang]
missing:   minMember=0
```

So a Workload that mixes gang and non-gang PodGroups produces a permanently-satisfied gang, and — see 5 — pulls the real gang into a group with it. Skip PodGroups without a gang policy instead.

### 5. Every PodGroup in a Workload becomes one GangGroup

`GetGangGroups` returns `util.GetId(ns, pg.Name)` for all of `Spec.PodGroups`, which makes the whole Workload one all-or-nothing unit. That is a semantic decision, not a mapping, and it is exactly what @saintube left open on #3101 — the inline at `:345`, "How to represent the gang group across multiple gangs?". `WorkloadMaxPodGroups` is 8 and #3101 lists GangGroup-larger-than-8 as blocked on upstream, so the two PRs are answering the same question in different places. Please let the proposal settle it.

### 6. Enablement by API discovery instead of a feature gate

`NewPodGroupManager` calls `handle.ClientSet().Discovery().ServerResourcesForGroupVersion("scheduling.k8s.io/v1alpha1")` at plugin construction, drops the error, and latches `workloadAPIEnabled` for the process lifetime.

- A transient failure, an apiserver not yet ready, or a missing discovery permission all resolve to "off" permanently. Recovery requires a restart.
- Nothing is logged either way, so an operator cannot tell which mode the scheduler is in.
- When the API *is* served, koord-scheduler starts claiming Workloads with no arbitration against upstream's own gang plugin. That is precisely the ownership conflict #3101's Ownership Resolution section exists to resolve, and it would land before the proposal does.

`GenericWorkload` and `GangScheduling` are both alpha and off in 1.35, and #3101 specifies a gate, default off. Please use one. Presence of an API is not consent to drive it.

### 7. No RBAC for `workloads.scheduling.k8s.io`

There is no rule for it anywhere under `config/rbac`, and this PR adds none — the diff is ten Go files. Discovery needs no permission, so on a cluster where the API is served, `workloadAPIEnabled` goes true, the informer starts, and then fails `list`/`watch` with Forbidden on a retry loop for the lifetime of the process. Needs the RBAC marker plus `make manifests`, and the corresponding chart-side ClusterRole in koordinator-sh/charts.

## Non-blocking

- **`onPodGroupAdd` reorder.** `SetGangGroupInfo` moved from after the `workloadAuditor` block to before it. That changes the existing PodGroup path, in a PR that says it preserves it. I suspect it is deliberate and correct — the auditor wants `GangGroupInfo` populated — but it needs a sentence in the description, or its own commit.
- **Auditor key inconsistency.** `onWorkloadAdd` passes `gang.GangGroupId`, `onWorkloadDelete` passes `gang.GangGroupInfo.GangGroupId`. Pick one; the second also dereferences a field that is nil until `SetGangGroupInfo` has run.
- **Dead interface surface.** `doInitByGangInfo` never calls `GetName()` or `GetNamespace()` — gang identity still comes from the cache key. `WorkloadGangInfo.GetName()` returning `podGroupName` has no reader at all.
- **`WorkloadGangInfo` hardcodes mode, wait time and match policy.** Reasonable for a first cut, but say so in a comment: native gangs are Strict with the plugin default timeout, not configurable.
- **`UpdateByGangInfo` has no `HasGangInit` guard**, so a Workload update can rewrite `MinRequiredNumber` while children are already waiting. `tryInitByPodGroup` behaved the same way, so this is not new, but the Workload update path is, and in-flight gangs deserve a thought.

## Merge order and rebase

Merge base is `6dfc4d78` (Aug 6), 26 behind. Please rebase. `core/core.go` also moves in #2900, and #2900 is on hold with findings of its own, so expect one conflict there.

More importantly: this should follow #3101. Findings 5 and 6 are both cases of this PR answering, by implementation, questions the proposal has open — and I left two must-fixes on #3101 today, one of which (the ownership rule contradicting the Koordinator-only guardrail) determines what finding 6 should become. Landing the annotation/PodGroup unification first, with the Workload half split out, would unblock the useful part immediately; that split is my suggestion, not a requirement.

Findings 1, 3 and 4 are all invisible to the current suite while it passes. Whatever else changes, please add cases that fail without the fixes — an illegal match policy asserting the defaulted value, a multi-PodGroup Workload asserting every gang reaches the cache, and a `Policy.Gang == nil` PodGroup asserting no gang is created.

/cc @saintube @hormes

ZiMengSheng:
/hold

(the hold in my review above did not register — it was inside backticks.)

ZiMengSheng:
Two corrections to my review above, @Dev10-sys — both about precision of my citations, not about the findings themselves. Flagging them because you'd otherwise go looking for code at line numbers where it isn't.

**1. Line-number baseline.** In blocking finding #1 I described the post-change behaviour but cited `main`'s line numbers. Since this PR removes 186 lines from `gang.go`, they don't line up on your branch. The correct pairs:

| what | on `main` | on this branch (`refs/pull/2969/head`) |
|---|---|---|
| `isGangValidForPermit` / its `default:` arm | `gang.go:557` / `:570` | `gang.go:443` / `:456` |
| the four explicit `== GangMatchPolicyOnceSatisfied` tests | `core.go:228, 317, 397, 592` | `core.go:262, 351, 431, 626` |

The finding is unchanged: the `default:` arm still implements OnceSatisfied semantics with the permissive `|| isGangOnceResourceSatisfied()`, while those four sites test the constant explicitly, so an unvalidated policy value splits the plugin against itself. Only my citations were off. (The `gang.go:145-150` / `:228-233` citations for the *deleted* validation are `main` line numbers, which is correct there since that code only exists on `main`.)

**2. The overlap with #2900 is wider than I said.** I wrote that you collide with #2900 on `core/core.go`. It's three files:

```
pkg/scheduler/plugins/coscheduling/core/core.go
pkg/scheduler/plugins/coscheduling/core/network_topology_workflow_test.go
pkg/scheduler/plugins/coscheduling/core/preemption_test.go
```

The two test files matter for sequencing more than the source file does — #2900 is currently on hold with its own findings, and whichever of you rebases second will be resolving conflicts in test files that encode the *other* PR's semantics. That reinforces rather than changes my earlier suggestion: land #3101's direction first, then #2900, then this.

Dev10-sys:
Thanks for the detailed review @ZiMengSheng.

I've rebased onto latest main and scoped this PR to the Annotation and PodGroup unification via GangInfo as suggested. Restored the match-policy validation, error logs, and added unit tests for the edge cases.

Please take another look when you get a chance!

PR #2970, opened 2026-05-16
[scheduler] implement queueing hints for core plugins

Dev10-sys:
This PR adds SchedulerQueueingHints support for core scheduler plugins to avoid unnecessary pod re-queueing on unrelated cluster events.

Implemented queue hint handling for:

Reservation
DeviceShare
NodeNUMAResource
LoadAwareScheduling

Changes included:

added QueueingHintFn support in EventsToRegister()
added conservative filtering logic for resource-related updates
added EnableQueueHint plugin configuration
added queueing hint unit tests for all affected plugins
documented Coscheduling as an intentional no-op for queueing hints

The queueing hints now only re-queue pods when resource availability or scheduling conditions become more favorable, instead of re-queueing on every update event.

Related:

#2852
#2851

### DIFF
--- a/pkg/scheduler/apis/config/types.go
+++ b/pkg/scheduler/apis/config/types.go
@@ -80,6 +80,8 @@ type LoadAwareSchedulingArgs struct {
 	// and only set up in custom node annotations,
 	// we should pass these resource names in plugin args explicitly.
 	SupportedResources []corev1.ResourceName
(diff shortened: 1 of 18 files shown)


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 15 with no activity since. Note the queueing-hints direction is actively pursued in #2927 (reservation plugin, part of the #2852 umbrella); any revival of the core-plugin hints should build on that pattern.

PR #2973, opened 2026-05-16
 fix: correct off-by-one bounds check in GetNPUCount

### LINKED ISSUE #2972: [BUG] GetNPUCount bounds check is off-by-one — panics when "Total Count" line has exactly 4 fields
**What happened:**
`GetNPUCount` in `pkg/device-daemon/resource/utils.go` parses the output of `npu-smi info -l` to find the total NPU count. After splitting the "Total Count" line by whitespace it guards with `if len(fidlds) < 4` (line 200) but immediately accesses `fidlds[4]` (line 203). The guard requires at least 4 elements (indices 0–3); the access requires at least 5 elements (index 4). Any "Total Count" line with exactly 4 whitespace-separated fields bypasses the guard and causes an index-out-of-range panic at runtime.

**How to reproduce:**
Visible in source at `pkg/device-daemon/resource/utils.go:200–203`. No runtime reproduction needed.

The guard is `if len(fidlds) < 5` so that a 4-element line correctly returns the "no total count field" error instead of panicking.

**What actually happened:**
`if len(fidlds) < 4` allows a 4-element slice to pass the check. `fidlds[4]` on a 4-element slice panics with `index out of range`.

- `pkg/device-daemon/resource/utils.go:200` — change `< 4` to `< 5`.
- The variable is also misspelled (`fidlds` instead of `fields`) on the same line 199; fixing both in one commit is appropriate.
- No other `GetXPU*` function has this pattern — only the NPU branch of `GetXPUCount` is affected.

mdryaan:

  `GetXPUCount` in `pkg/device-daemon/resource/utils.go` parses the `npu-smi info -l` output to find the total NPU count. The "Total Count"  line is split by whitespace and the count is read from index 4 (the fifth  element). The bounds guard was `if len(fields) < 4`, which only ensures four elements exist (indices 0–3); it does not protect the subsequent
  access at index 4. Any "Total Count" line with exactly four  whitespace-separated fields bypasses the guard and panics with an index  out of range. This PR fixes the guard to `< 5` and also corrects the  misspelling of the local variable (`fidlds` → `fields`).

Fixes #2972 

  - `pkg/device-daemon/resource/utils.go:200` was changed from   `if len(fidlds) < 4` to `if len(fields) < 5`.
  - `pkg/device-daemon/resource/utils.go:199` was changed from   `fidlds := strings.Fields(...)` to `fields := strings.Fields(...)`.
  - `pkg/device-daemon/resource/utils.go:203` was updated from `fidlds[4]`    to `fields[4]` to match the rename.
  - All checks pass: `go mod tidy`, `make fmt`, `make vet`, `make fast-test`,    `make lint`.

  The only behavioral change is the stricter bounds guard. A four-field  "Total Count" line now returns the existing `"get npu count err, no total  count field"` error instead of panicking. The rename of `fidlds` to  `fields` is cosmetic and does not affect runtime behavior.

  ### V. Checklist

### DIFF
--- a/pkg/device-daemon/resource/utils.go
+++ b/pkg/device-daemon/resource/utils.go
@@ -196,11 +196,11 @@ func GetXPUCount(deviceType string) (int, error) {
 		if totalCountLine == "" {
 			return 0, fmt.Errorf("get npu count err, no total count line")
 		}
-		fidlds := strings.Fields(totalCountLine)
-		if len(fidlds) < 4 {
+		fields := strings.Fields(totalCountLine)
+		if len(fields) < 5 {
 			return 0, fmt.Errorf("get npu count err, no total count field")
 		}
-		count, err := strconv.Atoi(fidlds[4])
+		count, err := strconv.Atoi(fields[4])
 		if err != nil {
 			return 0, fmt.Errorf("get npu count err, failed to convert field to integer, %w", err)
 		}


mdryaaan:
Hello @ZiMengSheng, could you please have a look whenever you get a chance?
Thanks!

zqzten:
@ZhuZhezz PTAL

PR #2984, opened 2026-05-17
fix: correct off-by-one bounds check and typo in GetNPUCount

### LINKED ISSUE #2972: [BUG] GetNPUCount bounds check is off-by-one — panics when "Total Count" line has exactly 4 fields

madmecodes:
## Summary
- Fixed off-by-one bounds check in `GetNPUCount` (`pkg/device-daemon/resource/utils.go`): changed `len(fidlds) < 4` to `len(fields) < 5`, since the code accesses index 4 which requires at least 5 elements. A 4-element slice previously bypassed the guard and caused an index-out-of-range panic.
- Fixed variable name typo from `fidlds` to `fields`.

Fixes #2972

## Test plan
- [ ] Verify the bounds check now correctly rejects slices with fewer than 5 elements
- [ ] Confirm no other references to the misspelled variable remain in the codebase
- [ ] Existing tests continue to pass

### DIFF
--- a/pkg/device-daemon/resource/utils.go
+++ b/pkg/device-daemon/resource/utils.go
@@ -196,11 +196,11 @@ func GetXPUCount(deviceType string) (int, error) {
 		if totalCountLine == "" {
 			return 0, fmt.Errorf("get npu count err, no total count line")
 		}
-		fidlds := strings.Fields(totalCountLine)
-		if len(fidlds) < 4 {
+		fields := strings.Fields(totalCountLine)
+		if len(fields) < 5 {
 			return 0, fmt.Errorf("get npu count err, no total count field")
 		}
-		count, err := strconv.Atoi(fidlds[4])
+		count, err := strconv.Atoi(fields[4])
 		if err != nil {
 			return 0, fmt.Errorf("get npu count err, failed to convert field to integer, %w", err)
 		}


mdryaaan:
Hey @madmecodes , this is a duplicate of #2973. I had already opened a PR for this issue shortly after creating it.
Please check the PRs mentioned on the issue before starting work on it to help avoid duplicate efforts.

zqzten:
dup with #2973

PR #2986, opened 2026-05-17
koordlet: fix kubelet restart failure by setting cpu.idle at pod level

### LINKED ISSUE #2732: [BUG] kubelet restart failed when  enable cpu qos feature

Ashutoshx7:

I ran into issue #2732 while testing CPU QoS with core scheduling. The root cause is a subtle Linux kernel constraint — when `cpu.idle=1` is set on a cgroup directory, the kernel rejects any `cpu.shares` writes to that **same directory**. 

The problem is that koordinator was setting `cpu.idle` on the **QoS-class root cgroup** (e.g. `kubepods-besteffort.slice/`), which is the exact same directory kubelet writes `cpu.shares` to during startup. So every time kubelet restarted, it crashed with:

```
Failed to start ContainerManager: failed to update top level BestEffort QOS cgroup:
failed to write "2": write .../kubepods-besteffort.slice/cpu.shares: invalid argument
```

Kernel reference: https://github.com/torvalds/linux/blob/v6.18/kernel/sched/fair.c#L13526

**The fix:** move `cpu.idle` reconciliation from the QoS-class root directory down to individual pod cgroup directories. Kubelet never writes to pod-level cgroups, so there's no conflict.

```
# Before (broken)
kubepods.slice/kubepods-besteffort.slice/cpu.idle = 1
↑ kubelet also writes cpu.shares here → EINVAL on restart

# After (fixed)  
kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-podXXX.slice/cpu.idle = 1
↑ kubelet never touches this → safe
```

The scheduling behavior per-pod is identical — only the cgroup level where `cpu.idle` is applied changes.

**Changes:**
- Replaced the `KubeQOSLevel` CPUIdle reconciler with a `PodLevel` one using a new `SetPodQOSCPUIdle` handler
- Added `SetPodQOSCPUIdle()` — operates on individual pod cgroups instead of the QoS root
- Added `IsPodQOSCPUIdle()` rule method — looks up the koordinator `podQOSParams` map
- Removed the QoS-class-level `cpu.idle` loop from `refreshForAllPods()`, added per-pod reconciliation inside the existing pod loop
- `SetKubeQOSCPUIdle` is kept intact — nothing that calls it directly will break
- Uses `ptr.To[int64]()` (consistent with the rest of the file, non-deprecated)

Fixes #2732

1. Enable CPU QoS with `cpuPolicy: coreSched` in NodeSLO
2. Restart kubelet — should start successfully without the `invalid argument` error
3. Verify `cpu.idle` is set at pod level, not the class root:

```bash
# QoS-class root should remain 0 (untouched)
cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.idle
# expected: 0

# Individual BE pod cgroup should have cpu.idle=1
cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/<pod-cgroup>/cpu.idle
# expected: 1
```

- `SetKubeQOSCPUIdle` is preserved — only the reconciler registration is replaced. No breaking changes.
- `IsPodQOSCPUIdle` reads from `podQOSParams` (koordinator QoS classes: LSE/LSR/LS/BE) — consistent with how `SetPodQOSCPUIdle` resolves QoS class from pod labels/annotations.
- Updated `Test_ruleUpdateCb` expectation: BestEffort root stays `0` since we no longer write `cpu.idle` there.
- Added `TestRule_IsPodQOSCPUIdle` with 10 test cases covering all QoS classes, uninitialized rules, disabled rules, and edge cases (empty/unknown QoS class).

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
+++ b/pkg/koordlet/runtimehooks/hooks/coresched/core_sched.go
@@ -116,9 +116,10 @@ func (p *Plugin) Register(op hooks.Options) {
 	reconciler.RegisterCgroupReconciler(reconciler.SandboxLevel, sysutil.VirtualCoreSchedCookie,
 		"set core sched cookie to process groups of sandbox container specified",
 		p.SetContainerCookie, reconciler.PodQOSFilter(), podQOSConditions...)
-	// TODO: support host application
-	reconciler.RegisterCgroupReconciler(reconciler.KubeQOSLevel, sysutil.CPUIdle, "reconcile QoS level cpu idle",
-		p.SetKubeQOSCPUIdle, reconciler.NoneFilter())
+	// Register pod-level cpu.idle reconciler to avoid conflicting with kubelet's
+	// cpu.shares writes at the QoS-class root cgroup directory.
+	reconciler.RegisterCgroupReconciler(reconciler.PodLevel, sysutil.CPUIdle,
+		"reconcile pod level cpu idle", p.SetPodQOSCPUIdle, reconciler.NoneFilter())
 	p.Setup(op)
 }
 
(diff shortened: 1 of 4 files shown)


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 15 with no activity since. Note this overlaps with the actively-reviewed #2775 (same kubelet-restart-on-cpuqos concern, #2732) — a revival should coordinate with that thread.

Issue #1704, opened 2023-10-11
[proposal] Add E2E tests for koordlet


saintube:
**What is your proposal**:

Currently, the end-to-end test for the koordlet is missing. Though the basic metrics reporting and batch overcommitment are tested in the slo-controller's E2E, we expect to check the QoS strategies provided by the koordlet like BECPUSuppress, BEMemoryEvict, BECPUEvict, CPUBurst, CPU QoS/Memory QoS, and so on.

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

After #1573 and #1688, the E2E tests are running on the CI pipeline of GitHub Actions. We can enable to run new E2E tests by adjusting the commands in `.github/workflows/e2e-k8s-xxx.yaml`.

aclfe:
我来帮忙把这些 TODO 的 tests 完成
多谢

ZiMengSheng:
## Progress ledger (maintainer triage, Sep 2026)

Recording state here so contributors don't duplicate work — this proposal has attracted **8 PRs** and they all create the same missing scaffolding.

### Current state on main
`test/e2e/koordlet/` **does not exist yet**. Existing E2E suites are only `test/e2e/slocontroller`, `scheduling`, and `quota`. Every koordlet E2E PR therefore has to bring its own framework/suite registration, which means **they conflict with each other by construction**.

### Ordering dependency (important for contributors and reviewers)
**#3006 should land first** — it is the only open PR that includes `test/e2e/koordlet/framework.go` (the shared harness) alongside its BECPUSuppress + BECPUEvict cases. The remaining PRs should then rebase on top of it and only add their own test files:

| PR | Coverage | Status |
|---|---|---|
| **#3006** | BECPUSuppress + BECPUEvict **+ framework.go** | land first (base) |
| #3012 | CPU QoS (cpu.shares / cpuset isolation) | rebase onto #3006 |
| #3014 | BlkIOQoS | rebase onto #3006 |
| #3016 | resctrl | rebase onto #3006 |

### Closed for inactivity — code still reusable
These were closed in a stale sweep (authors inactive since May 2026), but the work is **not obsolete** and none of it is covered on main. Anyone picking up a gap below is encouraged to start from these diffs:

- **BEMemoryEvict** → #3002
- **BECPUSuppress** → #3007 (superseded by #3006, useful as a scenario reference)
- **psi_collector** → #3008
- **HugePageReport** → #3021

[...]

PR #2987, opened 2026-05-17
test: add E2E tests for koordlet GroupIdentity (CPU BVT) plugin

Ashutoshx7:

This adds the first koordlet-specific E2E tests, starting with the **GroupIdentity (CPU BVT)** plugin. It's part of the roadmap in #1704.

The GroupIdentity plugin sets `cpu.bvt_warp_ns` on pod cgroup directories to give LS pods scheduling priority and throttle BE pods. There was no E2E coverage for this at all — this PR adds it.

Three test cases under `test/e2e/koordlet/`:

1. LS pod gets `cpu.bvt_warp_ns=2` when GroupIdentity is enabled
2. BE pod gets `cpu.bvt_warp_ns=-1` when GroupIdentity is enabled
3. BVT resets to `0` after GroupIdentity is disabled

The test enables GroupIdentity via `slo-controller-config`, creates a pod with the right koordinator QoS label, waits for it to run, then execs into the container to read the cgroup file directly. It always restores the original config afterward via defer, and skips gracefully on nodes where `cpu.bvt_warp_ns` doesn't exist (cgroup v2 / non-Anolis kernels).

Part of #1704

Deploy koordinator on an Anolis OS node (cgroup v1 with BVT support), then run:

```bash
go test ./test/e2e/... -v --ginkgo.focus="GroupIdentity"
```

- Skips cleanly on nodes without `cpu.bvt_warp_ns` so it won't fail on standard upstream kernels
- Polls for 60s to give koordlet's reconcile loop time to apply the value
- Config is always rolled back in defer — no side effects on the cluster after the test

### V. Checklist

- [x] Tests are self-cleaning (config rollback via defer)

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/koordinator-sh/koordinator/test/utils/image"
 
 	// test sources
+	_ "github.com/koordinator-sh/koordinator/test/e2e/koordlet"
 	_ "github.com/koordinator-sh/koordinator/test/e2e/quota"
 	_ "github.com/koordinator-sh/koordinator/test/e2e/scheduling"
 	_ "github.com/koordinator-sh/koordinator/test/e2e/slocontroller"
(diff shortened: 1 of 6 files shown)


ZiMengSheng:
Closing this stale PR — marked lifecycle/stale on Aug 15 with no activity since. The E2E coverage for GroupIdentity would still be welcome as a fresh PR if you return to it.

PR #2989, opened 2026-05-17
koordlet: apply CPUBurst policy to init containers

### LINKED ISSUE #2988: [BUG] koordlet: CPUBurst skips init containers — cpu.cfs_burst_us and CFS-quota burst never applied during pod startup
## What happened

`applyCPUBurst` and `applyCFSQuotaBurst` in `pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go` iterate only over `pod.Status.ContainerStatuses`, completely skipping `pod.Status.InitContainerStatuses`.

During pod startup the currently-running init container has its own cgroup directory (the same layout as regular containers).  Because neither function visits `InitContainerStatuses`, the init container's cgroup files are never touched:

- `cpu.cfs_burst_us` is left at its default (0 or whatever the container runtime wrote), so the init container cannot burst above its CPU limit even when CPUBurst is enabled.
- `cpu.cfs_quota_us` is never scaled by the CFSQuotaBurst strategy, so the init container is always throttled at its base quota regardless of node state.

This can cause measurable pod startup slowdowns when init containers perform CPU-intensive work (e.g. copying files, running DB migrations).

## How to reproduce

Visible in source at:

```
pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go

// applyCPUBurst – line 573 (before fix)
for i := range pod.Status.ContainerStatuses {   // InitContainerStatuses never visited
    ...
}

// applyCFSQuotaBurst – line 347 (before fix)
for i := range pod.Status.ContainerStatuses {   // InitContainerStatuses never visited
    ...
}
```

## What you expected to happen

Init containers should receive the same `cpu.cfs_burst_us` and CFS-quota burst scaling as regular containers while they are running.

## Environment

- koordlet version: main (confirmed in current HEAD)
- Affected: all nodes with CPUBurst feature gate enabled and pods that use init containers

## Fix

Extract the per-status iteration into a closure and call it for both `ContainerStatuses` and `InitContainerStatuses` in each function.  The pod-level `cpu.cfs_burst_us` should also accumulate contributions from init containers.

A fix with two regression tests is ready: https://github.com/Ashutoshx7/koordinator/tree/fix/cpuburst-missing-init-containers

Ashutoshx7:

Fixes #2988

`applyCPUBurst` and `applyCFSQuotaBurst` iterated only over `pod.Status.ContainerStatuses`, silently skipping `pod.Status.InitContainerStatuses`.

During pod startup, the active init container runs in its own cgroup directory.  Without this fix:
- `cpu.cfs_burst_us` is never written to the init container cgroup → init containers cannot burst above their CPU limit even when CPUBurst is enabled.
- `cpu.cfs_quota_us` is never scaled by the CFS-quota burst strategy → init containers are always throttled at their base quota regardless of node state.

This causes measurable pod startup slowdowns when init containers perform CPU-intensive work (DB migrations, large file copies, warmup logic).

### Root cause

```go
// Before — both functions only touched regular containers
for i := range pod.Status.ContainerStatuses { ... }
// pod.Status.InitContainerStatuses was never visited
```

### Fix

Extract the per-status loop into a closure and invoke it for both `ContainerStatuses` and `InitContainerStatuses`:

```go
// After
scaleCFSForContainerStatuses(pod.Status.ContainerStatuses, containerMap)
scaleCFSForContainerStatuses(pod.Status.InitContainerStatuses, initContainerMap)
```

The pod-level `cpu.cfs_burst_us` now also accumulates contributions from init containers.

## Ⅱ. Does this PR introduce a user-facing change?

```
koordlet: cpu.cfs_burst_us and CFS-quota burst are now applied to init
container cgroups during pod startup, consistent with regular containers.
```

## Ⅲ. Tests

Two regression tests added:
- `TestCPUBurst_applyCPUBurst_initContainers` — verifies `cpu.cfs_burst_us` is written to init container cgroups
- `TestCPUBurst_applyCFSQuotaBurst_initContainers` — verifies `cpu.cfs_quota_us` is scaled for init container cgroups under nodeBurstOverload

### DIFF
--- a/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
+++ b/pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
@@ -343,84 +343,98 @@ func (b *cpuBurst) applyCFSQuotaBurst(burstCfg *slov1alpha1.CPUBurstConfig, podM
 		container := &pod.Spec.Containers[i]
 		containerMap[container.Name] = container
 	}
+	// also index init containers so we can look them up by status name
+	initContainerMap := make(map[string]*corev1.Container)
+	for i := range pod.Spec.InitContainers {
+		c := &pod.Spec.InitContainers[i]
+		initContainerMap[c.Name] = c
+	}
 
-	for i := range pod.Status.ContainerStatuses {
-		containerStat := &pod.Status.ContainerStatuses[i]
-		container, exist := containerMap[containerStat.Name]
-		if !exist || container == nil {
-			klog.Warningf("container %s/%s/%s not found in pod spec", pod.Namespace, pod.Name, containerStat.Name)
-			continue
-		}
+	// scaleCFSForContainerStatuses applies cfs quota burst to a slice of container statuses
+	// using the provided spec-container lookup map.
+	scaleCFSForContainerStatuses := func(statuses []corev1.ContainerStatus, specMap map[string]*corev1.Container) {
+		for i := range statuses {
+			containerStat := &statuses[i]
+			container, exist := specMap[containerStat.Name]
+			if !exist || container == nil {
+				klog.Warningf("container %s/%s/%s not found in pod spec", pod.Namespace, pod.Name, containerStat.Name)
+				continue
+			}
 
-		if containerStat.State.Running == nil {
-			klog.V(6).Infof("skip container %s/%s/%s, because it is not running", pod.Namespace, pod.Name, containerStat.Name)
-			continue
-		}
+			if containerStat.State.Running == nil {
+				klog.V(6).Infof("skip container %s/%s/%s, because it is not running", pod.Namespace, pod.Name, containerStat.Name)
+				continue
+			}
 
-		containerBaseCFS := koordletutil.GetContainerBaseCFSQuota(container)
-		if containerBaseCFS <= 0 {
-			continue
-		}
-		containerPath, err := koordletutil.GetContainerCgroupParentDir(podMeta.CgroupDir, containerStat)
-		if err != nil {
-			klog.Infof("get container %s/%s/%s cgroup path failed, err %v",
-				pod.Namespace, pod.Name, containerStat.Name, err)
-			continue
-		}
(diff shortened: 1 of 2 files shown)


pkg/koordlet/qosmanager/plugins/cpuburst/cpu_burst.go
```diff
@@ -568,52 +582,69 @@ func (b *cpuBurst) applyCPUBurst(burstCfg *slov1alpha1.CPUBurstConfig, podMeta *
...
+
+			containerCFSBurstVal := calcStaticCPUBurstVal(container, burstCfg)
+			containerDir, burstPathErr := koordletutil.GetContainerCgroupParentDir(podMeta.CgroupDir, containerStat)
+			if burstPathErr != nil {
+				klog.Warningf("get container dir %s/%s/%s failed, dir %v, error %v",
+					pod.Namespace, pod.Name, containerStat.Name, containerDir, burstPathErr)
+				continue
+			}
+
+			total += containerCFSBurstVal
```
mdryaaan, line 618: `applyBurstForStatuses` accumulates `total` for any container with a non-empty `ContainerID`, including terminated init containers. `State.Terminated != nil` still passes the `ContainerID == ""` guard  on line 604, so completed init containers inflate the pod-level  `cpu.cfs_burst_us` on every reconcile.

`scaleCFSForContainerStatuses` already handles this correctly with `if containerStat.State.Running == nil { continue }`. Same guard needed here before the `total +=` line.

mdryaan (commented):
Hi @Ashutoshx7, I have reviewed ths PR, overall the fix looks good,extending both functions to cover init containers is the right approach and the cgroup path construction is correct. Left one inline comment on a missed termination check in `applyBurstForStatuses`.

cc @saintube @ZiMengSheng PTAL

ZiMengSheng:
Closing due to inactivity (last update May 2026). Applying CPUBurst policy to init containers is still not covered on main. A revival should rebase on top of the recent cpuburst normalization changes (#3176) — feel free to reopen if you return to it.

PR #2996, opened 2026-05-18
[scheduler] add required NRI plugin annotations for CPUSet and GPUShare workloads

Dev10-sys:
## What does this PR do?

This PR implements support for NRI required-plugin annotations (`required-plugins.noderesource.dev/container.<container-name>`) for workloads that rely on specific Koordinator resources.

It selectively injects `["koordlet_nri"]` during the `PreBind` phase only for:
1. **CPUSet Workloads** (via `NodeNUMAResource`)
2. **GPUShare Workloads** (via `DeviceShare`)

Unrelated pods and ephemeral containers are untouched to minimize blast radius and ensure runtime stability.

## Key Changes
* Added `apis/extension/nri.go` helper to easily annotate Pod containers.
* Integrated helper into `NodeNUMAResource` and `DeviceShare` `PreBind` flows.
* Updated unit tests for both plugins and verified the annotations persist correctly.

## Verification
Ran all scheduler tests successfully:
`go test ./pkg/scheduler/plugins/...`

### DIFF
--- a/apis/extension/nri.go
+++ b/apis/extension/nri.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extension
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// AnnotationRequiredPluginsContainerPrefix is the prefix of required plugins annotation key for a container.
+	AnnotationRequiredPluginsContainerPrefix = "required-plugins.noderesource.dev/container."
+
+	// DefaultNRIPluginName is the default NRI plugin name of koordlet runtime hooks.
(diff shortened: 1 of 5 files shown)


apis/extension/nri.go
```diff
@@ -0,0 +1,52 @@
...
+	if pod == nil {
+		return
+	}
+	annotations := pod.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	for _, container := range pod.Spec.InitContainers {
+		key := AnnotationRequiredPluginsContainerPrefix + container.Name
+		annotations[key] = DefaultRequiredPluginsValue
```
mdryaaan, line 45: This overwrites any existing value at this key. If another plugin or webhook already set this annotation, it gets silently replaced.  should check if the key exists first and append to the existing JSON array instead of replacing.

mdryaan (commented):
Hi @Dev10-sys, i noticed three issues, can you take a look

1. Nothing reads this annotation. koordlet's NRI server doesn't look  for required-plugins.noderesource.dev/container.* anywhere, and the standard NRI annotation domain is nri.io/* not noderesource.dev. the scheduler sets the annotation but nothing acts on it.

2. SetPodRequiredNRIPlugins overwrites existing annotation values see inline comment.

3. DefaultRequiredPluginsValue hardcodes "koordlet_nri" as a raw JSON string separately from DefaultNRIPluginName. If someone runs koordlet with --runtime-hooks-nri-plugin-name set to a custom name, the annotation won't match.

cc @saintube @ZiMengSheng

ZiMengSheng:
Closing due to inactivity (last update May 2026). The NRI plugin annotations for CPUSet/GPUShare workloads (apis/extension/nri.go + plugin injection) are not on main — a revived version would still be welcome. Feel free to reopen or send a fresh PR.

PR #3005, opened 2026-05-20
test: fix TestSyncedEventHandler hanging indefinitely in unit tests

### LINKED ISSUE #2957: Failing Github Action -  TestSyncedEventHandler times out in unit tests
I've noticed that the `TestSyncedEventHandler` consistently times out and fails during the CI runs for my PRs (and observed in the main workflow as well). This seems to block progress on changes and might indicate a deeper issue with the test itself.

**Steps to Reproduce:**

Open any recent PR (example: #2956)
Observe the unit-tests job ([workflow run link](https://github.com/koordinator-sh/koordinator/actions/runs/25930868437/job/76224102689?pr=2956))
The TestSyncedEventHandler test times out after 10 minutes

Failing workflow: [CI run #2956](https://github.com/koordinator-sh/koordinator/actions/runs/25930868437/job/76224102689?pr=2956)

**Request:**
Would appreciate any guidance or fixes from the maintainers/colleagues who might have seen this before!

Dev10-sys:
Fixes #2957

The TestSyncedEventHandler and TestSyncedEventHandlerWithReplace tests were
using wait.NeverStop as the stop channel for wait.PollUntil. This meant that
if the fake clientset's watch event was ever delayed or not delivered (which
can happen under CI load), the poll would run forever with no way out, causing
the entire unit-tests job to hang until the runner was killed after 10 minutes.

Replaced wait.NeverStop with a context.WithTimeout(10s).Done() channel. The
tests still pass immediately when the watch event arrives normally (~1s), but
now fail fast and cleanly instead of blocking the CI queue when something goes
wrong.

Tested locally — both tests pass consistently.

### DIFF
--- a/pkg/scheduler/frameworkext/helper/forcesync_eventhandler_test.go
+++ b/pkg/scheduler/frameworkext/helper/forcesync_eventhandler_test.go
@@ -72,7 +72,16 @@ func TestSyncedEventHandler(t *testing.T) {
 	defer close(stopCh)
 	sharedInformerFactory.Start(stopCh)
 
-	wg.Wait()
+	wgDone := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(wgDone)
+	}()
+	select {
+	case <-wgDone:
+	case <-time.After(10 * time.Second):
+		t.Fatal("timed out waiting for informer to deliver initial list events")
+	}
 
 	mu.Lock()
 	for _, v := range addTimes {
@@ -92,12 +101,14 @@ func TestSyncedEventHandler(t *testing.T) {
 	node.ResourceVersion = "100"
 	_, err = fakeClientSet.CoreV1().Nodes().Update(context.TODO(), node, metav1.UpdateOptions{})
 	assert.NoError(t, err)
+	pollCtx, pollCancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer pollCancel()
 	err = wait.PollUntil(1*time.Second, func() (done bool, err error) {
 		node, err := nodeInformer.Lister().Get("node-0")
 		assert.NoError(t, err)
 		assert.NotNil(t, node)
 		return node.ResourceVersion == "100", nil
-	}, wait.NeverStop)
+	}, pollCtx.Done())
 	assert.NoError(t, err)
 }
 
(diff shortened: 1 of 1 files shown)


IITI-tushar:
@Dev10-sys 
Thanks for working on this, failing fast instead of hanging the entire CI job is definitely an improvement.

One thing I was wondering though: this seems to mainly solve the “infinite wait” symptom, but maybe not the underlying synchronization issue itself.

From what I understand, the test is still polling informer cache state and assuming the watch/update event will eventually arrive. If the informer cache is not fully synced yet, or if the fake client watch delivery is delayed under CI load, **the test could still remain flaky** and **now just fail after 10s** instead of hanging forever.

Would it make sense to also:

* explicitly wait for `cache.WaitForCacheSync(...)`, or
* use a more event-driven signal/channel instead of polling `Lister().Get(...)` repeatedly?

I couldn’t reproduce locally either, so I’m not fully sure about the exact root cause yet, but I feel there may still be some timing/synchronization dependency underneath the test.

IITI-tushar:
@Dev10-sys 
What I'm saying is right now the timeout acts more like a safety fuse, but the root cause (missed sync/event ordering/race) may still exist underneath.

Dev10-sys:
hi @IITI-tushar,
The test already calls WaitForCacheSync right after the initial AddFunc handler syncs. The root cause of the indefinite CI hang wasn't just the polling, but the blocking wg.Wait() before the cache sync, which would deadlock the test entirely if the fake clientset dropped or delayed an initial event under high load.

I just pushed an updated commit that properly wraps wg.Wait() in a select-channel timeout as well.

Using a bounded 10s timeout for both the wg.Wait() sync and the Lister() polling is the standard safety fuse to ensure we fail fast rather than hanging the entire GitHub runner for 15+ minutes when the fake client events get throttled. Check the latest commit fc7dd1f, it addresses the exact underlying sync deadlock.

ZiMengSheng:
Closing due to inactivity (last update May 2026). The flaky/hanging `TestSyncedEventHandler` in `pkg/scheduler/frameworkext/helper/forcesync_eventhandler_test.go` is still present on main, so a refreshed fix would still be welcome — feel free to reopen or send a new PR.

PR #3006, opened 2026-05-21
test(koordlet): add e2e tests for BECPUSuppress and BECPUEvict plugins

Vatsalpatni73:

This PR introduces End-to-End (E2E) test coverage for the remaining Koordlet QoS plugins:
- **BECPUSuppress**: Verifies that Koordlet dynamically throttles `cpu.cfs_quota_us` for BestEffort (BE) pods when node CPU usage thresholds are exceeded, and correctly restores it to unlimited (`-1`) when the feature is disabled.
- **BECPUEvict**: Verifies that CPU-hungry BestEffort pods are correctly evicted (moving to `Failed/Evicted` status) when their CPU satisfaction drops below the configured threshold, and asserts that they remain running when the feature is disabled.

Both tests utilize the E2E test framework to safely patch `slo-controller-config` configs via defer rollbacks, spawn dummy BE workload pods, and poll cgroup properties or pod eventual states.

Part of #1704

Run the E2E tests for the koordlet packages:
```bash
go test ./test/e2e/... -v --ginkgo.focus="BECPUSuppress"
go test ./test/e2e/... -v --ginkgo.focus="BECPUEvict"
```

- Both tests use aggressive thresholds (1% suppress, 90% satisfaction) to guarantee triggering in CI without needing to fully saturate node resources.
- Config rollback is always performed via `defer`, no cluster side effects remain after the test regardless of pass or fail.
- `waitForPodRunning` and the cgroup path discovery pattern are reused from the existing koordlet test suite.

### V. Checklist

### DIFF
--- a/test/e2e/koordlet/becpuevict.go
+++ b/test/e2e/koordlet/becpuevict.go
@@ -1,0 +1,182 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package koordlet
+
+import (
+	"context"
+
+	"github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/koordinator-sh/koordinator/apis/configuration"
+	apiext "github.com/koordinator-sh/koordinator/apis/extension"
+	"github.com/koordinator-sh/koordinator/test/e2e/framework"
+)
+
+var _ = SIGDescribe("BECPUEvict", func() {
+	f := framework.NewDefaultFramework("becpuevict")
+
+	framework.KoordinatorDescribe("BECPUEvict [koordlet]", func() {
+		framework.ConformanceIt("should evict a CPU-hungry BE pod when BECPUEvict is enabled", func() {
+			c := f.ClientSet
+
+			// Patch config
+			cm, err := c.CoreV1().ConfigMaps(framework.TestContext.KoordinatorComponentNamespace).Get(context.TODO(), "slo-controller-config", metav1.GetOptions{})
+			framework.ExpectNoError(err, "failed to get slo-controller-config")
+			oldData := cm.DeepCopy().Data
+			defer func() {
+				cmToRestore, err := c.CoreV1().ConfigMaps(framework.TestContext.KoordinatorComponentNamespace).Get(context.TODO(), "slo-controller-config", metav1.GetOptions{})
+				framework.ExpectNoError(err)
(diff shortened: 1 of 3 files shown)


ZiMengSheng:
### Pre-review: this PR is the ordering gate for the whole #1704 e2e effort

First, thanks for this — the design here is right, and I want to be explicit about why this PR matters more than its +428 lines suggest.

**Why this one goes first.** `test/e2e/koordlet/` does not exist on `main` today. There are currently **4 open PRs** (#3006, #3012, #3014, #3016) each creating that directory *and its own* `framework.go`, plus 4 more that were closed as stale. They are structurally mutually exclusive — whichever lands first defines the base, and the rest must rebase onto it. This PR is the only one carrying `framework.go`, so it is the base. I've recorded the full ordering in #1704.

That also means the bar here is higher: any defect in the base is inherited by three more PRs.

**The design is good.** `SIGDescribe` with the `[koordlet]` prefix mirrors `test/e2e/slocontroller/framework.go:22-24` exactly, and you reuse the existing `test/e2e/framework` rather than reinventing it. Keeping the base at 36 lines is the correct instinct. The config-patch-with-defer-rollback pattern in both test files is also right.

---

### Blocking: the tests are currently dead code — they never execute

CI is green on this PR, but that green is meaningless, because **nothing in it ever ran these tests.** Two pieces of wiring are missing:

**1. Missing blank import in `test/e2e/e2e_test.go`**

Suites are linked into the e2e binary via blank imports (lines 45-47 on `main`):

```go
_ "github.com/koordinator-sh/koordinator/test/e2e/quota"
_ "github.com/koordinator-sh/koordinator/test/e2e/scheduling"
_ "github.com/koordinator-sh/koordinator/test/e2e/slocontroller"
```

Without `_ "github.com/koordinator-sh/koordinator/test/e2e/koordlet"`, the package is never linked, so your `ginkgo.Describe` blocks never register.

**2. Missing `koordlet` in the workflow focus filter**

Even once registered, the ginkgo invocation filters them out. From `.github/workflows/e2e-k8s-1.28.yaml:113`:

```
./bin/ginkgo -timeout 60m -v --focus='slo-controller|\] Reservation|\] Preemption' ...
```

`[koordlet]` doesn't match that focus expression. This needs a `koordlet` alternative added (and the same applies to the other `e2e-k8s-*.yaml` files you want this to run in).

Please add both, then confirm in the PR description that you've seen the specs actually execute in CI — not just that CI passed. The fix is small, but until it's in, this PR and the three that rebase onto it would all be inert while looking healthy.

**One thing that is fine, for the record:** the runtime prerequisites hold. `hack/deploy_kind.sh` does deploy koordlet as a DaemonSet and the workflow waits for it to be ready, so cgroup assertions against a live koordlet are viable in the kind environment. No concern there.

---

/cc @saintube — flagging this one for you specifically, since it sets the koordlet e2e base that #3012 (CPU QoS), #3014 (BlkIOQoS) and #3016 (resctrl) will all inherit. Two things worth your eyes beyond the wiring above:

1. **Is a 36-line base the right size?** The three followers will each need config patching and cgroup polling. Right now that logic is duplicated per test file. Worth pulling `patchSLOConfig` / cgroup-poll helpers into `framework.go` now, or better to let it duplicate and extract later once we see three real consumers?
2. **BECPUSuppress/BECPUEvict thresholds in CI.** These drive real node CPU pressure and assert on `cpu.cfs_quota_us` and eviction. On a shared GitHub runner, are the thresholds and poll timeouts here realistic, or will this be flaky? You have the most context on how these two plugins behave under a noisy kind node.

I'd rather hold this until you weigh in on (1), because changing the base shape after three PRs rebase onto it is the expensive path.

Vatsalpatni73:
Ni Hao @ZiMengSheng , thanks for the review. Fixed both:

1. Added blank import for test/e2e/koordlet in e2e_test.go.
2. Rebased onto main and added |koordlet to the --focus filter in all 5 e2e-k8s-*.yaml workflows.

PR #3010, opened 2026-05-21
ci: fix E2E slo-controller flakiness on resource-constrained CI runners

Dev10-sys:
The slo-controller E2E tests were failing intermittently because the CI
workflow was starting the test runner before koordlet had fully come up.
The readiness polling loop only waited ~54s, which isn't enough on slower
GitHub Actions runners.

Also, the batch resource check inside the test required 70% of nodes to
be ready within 3 minutes — too tight for a 2-node Kind cluster where
even a single slow koordlet causes a failure.

- Increased the pod readiness wait from 9 to 19 iterations across all
  e2e workflow files (~54s → ~114s)
- Raised the batch resource check timeout from 180s to 300s
- Lowered the minimum node passing ratio from 0.7 to 0.5


ZiMengSheng:
Closing due to inactivity (last update May 2026). If the E2E slo-controller flakiness on resource-constrained runners recurs, feel free to reopen — the CI timeout relaxation idea can be re-evaluated then.

Issue #2306, opened 2024-12-27
[BUG] CPUBurst cannot co-work with CPUNormalization


saintube:
**What happened**:

Koordlet supports two features CPUBurst and CPUNormalization which can change the `cpu.cfs_quota_us`/`cpu.max` of the LS pods and containers. However, these features can not co-work for the cfs_quota adjustment. When both features are enabled, the QoS plugin of the CPUBurst modifies the cfs_quota due to the cfs throttled ratio of pods. In contrast, the runtime hook plugin of the CPUNormalization may overwrite the cfs_quota according to the CPU-normalisation ratio of the node. Then neither of them can achieve their reconciliation.

**What you expected to happen**:

To make the CPUBurst feature co-work with the CPUNormalization, the CPUBurst QoS plugin should be aware of the CPU-normalization ratio of the node. When both features are enabled, the CPUBurst QoS plugin is responsible for adjusting the `cpu.cfs_quota_us`/`cpu.max` according to the scaling ratio and the CPU-normalization ratio, and the runtime hook of the CPUNormalization should stop its reconciliation to avoid conflicts.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- App version: v1.5.0
- Kubernetes version (use `kubectl version`):
- Install details (e.g. helm install args):
- Node environment (for koordlet/runtime-proxy issue):
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
  - Cgroup driver: cgroupfs/systemd
- Others:

nakul-py:
Hi @saintube 
I came across this issue and would like to work on resolving it. Could you please assign it to me?

saintube:
> Hi @saintube I came across this issue and would like to work on resolving it. Could you please assign it to me?

OK.

nakul-py:
I cannot build or install

pratik-mahalle:
Anyone is working or should I take this issue ? @saintube

saintube:
> Anyone is working or should I take this issue ? [@saintube](https://github.com/saintube)

@pratik-mahalle OK, welcome to contribute!

PR #3011, opened 2026-05-21
chore: refine eventhandler, workloadauditor and gang cache

saintube:
### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
+++ b/pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
@@ -334,8 +334,11 @@ func (w *workloadAuditorImpl) RecordPodScheduleResult(pod *corev1.Pod, recordTyp
 		klog.V(4).Infof("WorkloadAuditor delete(podScheduled): workloadKey=%s %s, attempts=%d",
 			record.WorkloadKey, record.labelDetail, record.Attempts)
 		finalizeWorkloadRecord(record, outcomeScheduled)
-		record.mu.Unlock()
+		// Delete the entry from the records map before releasing the per-record
+		// lock to prevent another goroutine that already obtained the same record
+		// reference from mutating an already-finalized record after deletion.
 		w.records.Delete(workloadKey)
+		record.mu.Unlock()
 		return
 	}
 	record.mu.Unlock()
@@ -360,8 +363,11 @@ func (w *workloadAuditorImpl) RecordGangScheduleResult(gangKey string, recordTyp
 		klog.V(4).Infof("WorkloadAuditor delete(gangScheduled): workloadKey=%s %s, attempts=%d",
 			record.WorkloadKey, record.labelDetail, record.Attempts)
 		finalizeWorkloadRecord(record, outcomeGangScheduled)
-		record.mu.Unlock()
+		// Delete the entry from the records map before releasing the per-record
+		// lock to prevent a concurrent goroutine that already obtained the same
+		// record reference from appending data to an already-finalized record.
 		w.records.Delete(gangKey)
+		record.mu.Unlock()
 		return
 	}
 	record.mu.Unlock()
--- a/pkg/scheduler/plugins/coscheduling/core/gang_cache.go
+++ b/pkg/scheduler/plugins/coscheduling/core/gang_cache.go
@@ -322,11 +322,15 @@ func (gangCache *GangCache) onPodGroupUpdate(oldObj interface{}, newObj interfac
 	}
 
 	// When PodGroup transitions from a pending phase to a non-pending phase,
-	// delete the gang record from the workload auditor.
+	// delete the gang record from the workload auditor. The reverse transition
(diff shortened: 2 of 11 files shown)


pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
```diff
@@ -360,8 +363,11 @@ func (w *workloadAuditorImpl) RecordGangScheduleResult(gangKey string, recordTyp
 		klog.V(4).Infof("WorkloadAuditor delete(gangScheduled): workloadKey=%s %s, attempts=%d",
 			record.WorkloadKey, record.labelDetail, record.Attempts)
 		finalizeWorkloadRecord(record, outcomeGangScheduled)
-		record.mu.Unlock()
+		// Delete the entry from the records map before releasing the per-record
+		// lock to prevent a concurrent goroutine that already obtained the same
+		// record reference from appending data to an already-finalized record.
 		w.records.Delete(gangKey)
+		record.mu.Unlock()
```
ZiMengSheng, line 370: Could this cause a deadlock?

pkg/scheduler/plugins/coscheduling/core/gang_cache.go
```diff
@@ -323,10 +323,19 @@ func (gangCache *GangCache) onPodGroupUpdate(oldObj interface{}, newObj interfac
...
 	// When PodGroup transitions from a pending phase to a non-pending phase,
 	// delete the gang record from the workload auditor.
+	// When PodGroup transitions back from a non-pending phase to a pending phase
+	// (e.g., after a failed scheduling round re-queues the gang), re-add the
+	// record so subsequent scheduling attempts can be tracked.
 	if gangCache.workloadAuditor != nil {
 		oldPg, ok := oldObj.(*v1alpha1.PodGroup)
-		if ok && isPodGroupPendingPhase(oldPg.Status.Phase) && !isPodGroupPendingPhase(pg.Status.Phase) {
-			gangCache.workloadAuditor.DeleteGangGroup(gang.GangGroupId)
+		if ok {
```
ZiMengSheng: It's better to ignore it to avoid potential record leaks.

pkg/scheduler/frameworkext/workloadauditor/workload_auditor.go
saintube: No deadlock risk. w.records is a sync.Map whose Delete operation is lock-free (implemented with atomic CAS + lazy sweep).The ordering finalizeWorkloadRecord → records.Delete → mu.Unlock is intentional: it closes a race window where another goroutine could obtain the same record reference (via Load) after we've finalized it but before we've deleted it, and then append data to an already-finalized record. Reversing the order would introduce a correctness bug, not prevent one.(The lock graph is: per-record mu only — no second lock is acquired while holding mu, so there is no cycle.)

pkg/scheduler/plugins/coscheduling/core/gang_cache.go
saintube: Agreed — removed the else if !oldPending && newPending branch. The reverse transition is rare and onPodGroupAdd already seeds the record when the PodGroup is first observed in pending state; re-adding it here could leave stale/partial records behind. Simplified to only handle the pending → non-pending cleanup and updated TestGangCache_onPodGroupUpdate_PhaseTransitions accordingly (cases 2/3 now assert no-op).

ZiMengSheng:
/remove-lifecycle stale

@saintube this was the last remaining stale PR in the repo — I've cleared the label rather than letting the bot close it, since a +793 refactor of the eventhandler / workloadauditor / gang cache isn't something we want auto-reaped.

It's been idle since 2026-08-23. Could you say whether you intend to carry it forward? Either is fine, I just want it off the ambiguous pile:

- **Still relevant** → I'll review it. Worth noting it touches `frameworkext/eventhandlers` and `frameworkext/cross_scheduler_nominator`, which overlaps the reservation-nomination cluster I've just sequenced on #3069 (#3084 → #3156 → #3114). If your refactor lands first, those three will need to rebase onto it; if it lands after, you will. Your call which order is less painful, but it should be a deliberate choice.
- **Superseded** → close it, and I'll note anything worth salvaging on the relevant issues.

Separately, there are four things waiting on you where I've already written up the specific checkpoints, listed roughly by how much they're blocking others:

1. **#3006** — koordlet e2e base. Most urgent: it gates #3012 / #3014 / #3016, and I found the specs never actually execute (missing blank import + `--focus` filter). I need your call on whether the 36-line base should absorb the config-patch / cgroup-poll helpers *before* three PRs rebase onto it.
2. **#2863** — SignPlugin across core plugins, already `/lgtm` from me, KEP-5598 core.
3. **#3176** — CPUBurst × CPUNormalization, already `/lgtm`, fixes #2306.
4. **#3195** — letting the stale bot cover milestone items so this backlog doesn't regrow.

No rush on any single one, but #3006 is holding up four PRs from mentees, so that's the one I'd prioritise.

PR #3019, opened 2026-05-25
fix: stop hostReconciler cleanly after stopCh closes

### LINKED ISSUE #3018: [BUG] koordlet/runtimehooks: hostReconciler does not exit after stopCh closes
**What happened**:
`hostReconciler.reconcile()` continues running even after `stopCh` is closed.

Inside the reconcile loop, the stop case logs shutdown information but does not exit the goroutine:

```
case <-stopCh:
    klog.Infof(...)

```
Since the loop continues after the select finishes, the `reconciler goroutine` never terminates cleanly.

`hostReconciler.reconcile()` should exit immediately once `stopCh` is closed.

Add the following unit test:

```go
func TestHostReconcilerStopsOnStopCh(t *testing.T) {
    r := &hostReconciler{
        appUpdated:        make(chan struct{}, 1),
        reconcileInterval: 10 * time.Millisecond,
    }

    stopCh := make(chan struct{})
    done := make(chan struct{})

    go func() {
        r.reconcile(stopCh)
        close(done)
    }()

    close(stopCh)

    select {
    case <-done:
        // ok
    case <-time.After(200 * time.Millisecond):
        t.Fatal("hostReconciler.reconcile did not stop after stopCh closed")
    }
}

```
Run:

```bash
go test ./pkg/koordlet/runtimehooks/reconciler \
  -run TestHostReconcilerStopsOnStopCh \
  -count=1 \
  -v
```

Before the fix:

test times out and fails

After the fix:

test passes successfully

The issue is caused by missing termination logic in the stop channel branch.

Suggested fix:
```

case <-stopCh:
    klog.Infof(...)
    return

```
This ensures the reconciler goroutine exits cleanly during shutdown.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

IITI-tushar:

This PR fixes a goroutine lifecycle issue in hostReconciler.reconcile().
Previously, when stopCh was closed, the reconciler logged the shutdown event but continued looping instead of exiting, As a result, the reconcile goroutine could remain alive indefinitely after shutdown.

Fix:

- add return in the stop channel case so the goroutine exits cleanly
- add a unit test verifying the reconciler stops correctly after stopCh closes

The new test reproduces the issue deterministically and validates the shutdown behavior.

fixes #3018 

Run:

```bash
go test ./pkg/koordlet/runtimehooks/reconciler \
  -run TestHostReconcilerStopsOnStopCh \
  -count=1 \
  -v
```

Expected result:

before fix: test fails by timeout
after fix: test passes successfully

- This change only affects shutdown behavior.
- No reconcile logic or scheduling behavior was modified.
- The fix prevents leaked background goroutines during shutdown and testing.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler.go
+++ b/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler.go
@@ -98,6 +98,7 @@ func (r *hostReconciler) reconcile(stopCh <-chan struct{}) {
 			timer.Reset(r.reconcileInterval)
 		case <-stopCh:
 			klog.V(1).Infof("stop reconcile for host application")
+			return
 		}
 	}
 }
--- a/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler_test.go
+++ b/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler_test.go
@@ -335,3 +335,26 @@ func TestNewHostAppReconciler(t *testing.T) {
 	defer close(stop)
 	assert.NoError(t, r.Run(stop))
 }
+func TestHostReconcilerStopsOnStopCh(t *testing.T) {
+	r := &hostReconciler{
+		appUpdated:        make(chan struct{}, 1),
+		reconcileInterval: 10 * time.Millisecond,
+	}
+
+	stopCh := make(chan struct{})
+	done := make(chan struct{})
+
+	go func() {
+		r.reconcile(stopCh)
+		close(done)
+	}()
+
+	close(stopCh)
+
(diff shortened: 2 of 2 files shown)


ZiMengSheng:
Closing due to inactivity (last update May 2026). **The goroutine leak is confirmed still on main**: `runtimehooks/reconciler/host_app_reconciler.go` `reconcile()` only logs in the `<-stopCh` case without returning, so both goroutines leak on shutdown. Anyone is welcome to pick this up — happy to fast-track the review.

PR #3028, opened 2026-05-27
webhook: validate quota on pod update when quota label changes

TaoYang526:

  Support quota admission validation on pod update when the quota label changes.

  When a pod is created, the webhook validates its quota usage via `evaluateQuota`. However, the quota name is derived from the pod label `quota.scheduling.koordinator.sh/name`, which can be modified via Update
  operations. Currently, `evaluateQuota` only handles Create and skips Update, allowing users to bypass quota limits by changing the quota label after creation — moving a pod into a full quota without any admission
   check.

  This PR:
  - Adds `admissionv1.Update` handling in `evaluateQuota`: decodes old/new pod, compares quota labels, and evaluates the target quota capacity if the label changed.
  - Extends `Handles()` in the evaluator to accept Update operations.
  - Introduces a new `EnableQuotaAdmissionOnUpdate` feature gate (Alpha, default off) to control the behavior independently.

NONE

  1. Run existing tests to confirm no regression:
     ```bash
     go test ./pkg/webhook/pod/validating/ -run TestEvaluateQuota -v
     go test ./pkg/webhook/quotaevaluate/ -v
  2. New test cases cover:
    - Update with feature gate disabled — allowed (skip)
    - Update with quota label unchanged — allowed (skip)
    - Update with quota label changed, target quota has capacity — allowed
    - Update with quota label changed, target quota is full — denied

  - The feature is gated behind EnableQuotaAdmissionOnUpdate (Alpha, default off), so it has zero impact when not explicitly enabled.
  - The Handles() method in the evaluator accepts Update unconditionally; the feature gate check is performed by the caller (evaluateQuota) before constructing the Attributes. A comment documents this contract.
  - The scheduler's GroupQuotaManager.OnPodUpdate already handles quota label migration (adjusts used/request accounting), so the quota state will reconcile correctly after the webhook validates the update.

### V. Checklist

### DIFF
--- a/pkg/webhook/quotaevaluate/evaluator.go
+++ b/pkg/webhook/quotaevaluate/evaluator.go
@@ -448,8 +448,11 @@ func (e *quotaEvaluator) initCtx(ctx *quotaCheckContext, quota *v1alpha1.Elastic
 	return ctx
 }
 
+// Handles returns true if the evaluator should process this operation.
+// For Update, the caller (evaluateQuota) has already checked the EnableQuotaAdmissionOnUpdate feature gate
+// before constructing the Attributes, so no redundant gate check is needed here.
 func (e *quotaEvaluator) Handles(a *Attributes) bool {
-	if a.Operation == admissionv1.Create {
+	if a.Operation == admissionv1.Create || a.Operation == admissionv1.Update {
 		return true
 	}
 	return false
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -91,6 +91,9 @@ const (
 	// EnableQuotaAdmission enables quota admission.
 	EnableQuotaAdmission featuregate.Feature = "EnableQuotaAdmission"
 
+	// EnableQuotaAdmissionOnUpdate enables quota admission on pod update when quota label changes.
+	EnableQuotaAdmissionOnUpdate featuregate.Feature = "EnableQuotaAdmissionOnUpdate"
+
 	// EnablePodEnhancedValidator enables enhanced validator for pods with configurable rules.
 	EnablePodEnhancedValidator featuregate.Feature = "EnablePodEnhancedValidator"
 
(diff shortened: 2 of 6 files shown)


TaoYang526:
/assign @shaloulcy @saintube 
Please help review this PR, thanks!

pkg/webhook/quotaevaluate/evaluator.go
```diff
@@ -270,8 +270,11 @@ func (e *quotaEvaluator) checkQuota(quota *v1alpha1.ElasticQuota, admissionAttri
 	e.checkQuota(newQuota, admissionAttributes, remainingRetries-1)
 }
 
+// Handles returns true if the evaluator should process this operation.
+// For Update, the caller (evaluateQuota) has already checked the EnableQuotaAdmissionOnUpdate feature gate
+// before constructing the Attributes, so no redundant gate check is needed here.
 func (e *quotaEvaluator) Handles(a *Attributes) bool {
-	if a.Operation == admissionv1.Create {
+	if a.Operation == admissionv1.Create || a.Operation == admissionv1.Update {
```
saintube, line 455: need to update the webhook configuration under the config directory

TaoYang526: @saintube Thanks for the review.
The webhook configuration in [config/webhook/manifests.yaml](https://github.com/koordinator-sh/koordinator/blob/main/config/webhook/manifests.yaml#L178-L198) already includes UPDATE in the operations list for /validate-pod, so no additional changes are needed.

TaoYang526:
Rebased from main to resolve conflict issue.

TaoYang526:
The last UT failure seems unrelated to this PR. Updating the commit message to re-trigger the CI checks.

saintube (approved):
/lgtm

PR #3034, opened 2026-05-28
fix: report all missing EstimatedScalingFactors in ValidateLoadAwareSchedulingArgs

### LINKED ISSUE #3033: [BUG] ValidateLoadAwareSchedulingArgs stops after the first missing EstimatedScalingFactor instead of reporting all
**What happened:**
`pkg/scheduler/apis/config/validation/validation_pluginargs.go:55–59` cross-validates that every key in `ResourceWeights` also appears in `EstimatedScalingFactors`. The loop uses `break` after appending the first `field.NotFound` error, so it stops as soon as it finds one missing entry. If an operator configures weights for `cpu`, `memory`, and `gpu` but forgets to add all three to `EstimatedScalingFactors`, they see only one error on the first failed run. They must fix and restart the scheduler repeatedly to discover each subsequent missing entry.

**How to reproduce:**
Visible in source at `pkg/scheduler/apis/config/validation/validation_pluginargs.go:58`. No runtime reproduction needed.

```go
for resourceName := range args.ResourceWeights {
    if _, ok := args.EstimatedScalingFactors[resourceName]; !ok {
        allErrs = append(allErrs, field.NotFound(...))
        break  // stops after first missing entry — should be removed
    }
}
```

The loop runs to completion and reports a `field.NotFound` error for every resource weight that has no corresponding scaling factor, matching the behavior of all other validation loops in the same function and in the Kubernetes ecosystem (`field.ErrorList` is designed to accumulate multiple errors).

**What actually happened:**
Only the first missing scaling factor is reported per validation call. Operators with multiple missing entries must fix and retry multiple times to discover all configuration errors.

- `pkg/scheduler/apis/config/validation/validation_pluginargs.go:58` — remove the `break` statement. No other change needed.
- Every other validation loop in the same file (e.g. `validateResources`, `validateResourceWeights`) uses no `break` and collects all errors. This is the only loop with the premature `break`.
- A unit test for `ValidateLoadAwareSchedulingArgs` with two weights both missing from `EstimatedScalingFactors` demonstrates the bug: before fix, `allErrs` has length 1; after fix, length 2.

mdryaan:

`ValidateLoadAwareSchedulingArgs` in `pkg/scheduler/apis/config/validation/validation_pluginargs.go` validates that every key in `ResourceWeights` has a corresponding entry in `EstimatedScalingFactors`. The loop had a `break` after the first `field.NotFound` error, so validation stopped at the first missing entry. Operators with multiple missing scaling factors had to fix and restart the scheduler repeatedly to discover each one.

Removes the `break` so the loop runs to completion and all missing entries are reported in a single validation call, consistent with every other loop in the same function and with `field.ErrorList` semantics across the Kubernetes ecosystem.

fixes #3033 

```bash
go test ./pkg/scheduler/apis/config/validation/... -v -run TestValidateLoadAwareSchedulingArgs
```

New test `TestValidateLoadAwareSchedulingArgs_ReportsAllMissingScalingFactors` configures `ResourceWeights` with both `cpu` and `memory` while leaving `EstimatedScalingFactors` empty. Before the fix only one resource name appeared in the error; after the fix both appear. All existing tests continue to pass.

One `break` removed — no other logic change. The existing `"resourceWeight missing estimatedScalingFactor"` table case only checks `wantErr: true` so it passes either way; the new test is what specifically exercises the all-errors path.

### V. Checklist

### DIFF
--- a/pkg/scheduler/apis/config/validation/validation_pluginargs.go
+++ b/pkg/scheduler/apis/config/validation/validation_pluginargs.go
@@ -55,7 +55,6 @@ func ValidateLoadAwareSchedulingArgs(args *config.LoadAwareSchedulingArgs) error
 	for resourceName := range args.ResourceWeights {
 		if _, ok := args.EstimatedScalingFactors[resourceName]; !ok {
 			allErrs = append(allErrs, field.NotFound(field.NewPath("estimatedScalingFactors"), resourceName))
-			break
 		}
 	}
 
--- a/pkg/scheduler/apis/config/validation/validation_pluginargs_test.go
+++ b/pkg/scheduler/apis/config/validation/validation_pluginargs_test.go
@@ -212,6 +212,21 @@ func TestValidateLoadAwareSchedulingArgs(t *testing.T) {
 	}
 }
 
+func TestValidateLoadAwareSchedulingArgs_ReportsAllMissingScalingFactors(t *testing.T) {
+	args := &config.LoadAwareSchedulingArgs{
+		ResourceWeights: map[corev1.ResourceName]int64{
+			corev1.ResourceCPU:    1,
+			corev1.ResourceMemory: 1,
+		},
+		EstimatedScalingFactors: map[corev1.ResourceName]int64{},
+	}
+	err := ValidateLoadAwareSchedulingArgs(args)
+	assert.Error(t, err)
+	errStr := err.Error()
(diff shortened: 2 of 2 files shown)


mdryaaan:
Hello @ZiMengSheng PTAL , whenvever you get a chance. Thanks

PR #3042, opened 2026-05-28
fix(descheduler): fail PodMigrationJob when target pod is missing in arbitrator

### LINKED ISSUE #3039: [BUG] descheduler arbitrator passes PMJ when target pod cannot be loaded
**What happened**:

When the arbitrator cannot load the target pod for a `PodMigrationJob` (pod deleted, RBAC error, transient API failure, or `PodRef == nil`), it treats the PMJ as having passed arbitration and annotates it `descheduler.koordinator.sh/passed-arbitration=true`. The PMJ then proceeds toward eviction even though there is no live target pod. This bypasses all safety filters (nonRetryablePodFilter, retryablePodFilter, namespace allowlists, owner checks).

If the target pod cannot be loaded, arbitration should fail (or at least retry), and the PMJ should not be annotated as passed. A missing target should never bypass safety filters.

1. Start a kind cluster and deploy Koordinator (including `koord-descheduler`).
2. Create a namespace and a deployment with 2 replicas.
```bash
kubectl create ns pmj-test

kubectl create deployment pmj-demo -n pmj-test --image=busybox:1.36 -- sleep 3600
kubectl scale deploy pmj-demo -n pmj-test --replicas=2

kubectl get pod -n pmj-test -l app=pmj-demo -w
# wait until both pods are Running, then CTRL+C
```

3. Tail descheduler logs to observe arbitration results.
```bash
kubectl logs -n koordinator-system deploy/koord-descheduler -c descheduler -f --since=5m \
  | grep -E "arbitrator|failed to get Pod of PodMigrationJob|passed-arbitration"
```

4. create a PMJ manifest with non-existent pod.
```bash
cat <<EOF > /tmp/pmj-demo.yaml
kind: PodMigrationJob
metadata:
  name: pmj-demo
spec:
  mode: EvictDirectly
  podRef:
    namespace: pmj-test
    name: does-not-exist
EOF
```
5. Apply the manifest and observe the logs
```bash
kubectl apply -f /tmp/pmj-demo.yaml
```

6. Verify the PMJ gets passed-arbitration=true even though the pod does not exist.
```bash
kubectl get pmj pmj-demo -o jsonpath='{.metadata.annotations.descheduler\.koordinator\.sh/passed-arbitration}{"\n"}'
```
Expected (buggy) result: the command prints `true` even though logs show:
```bash
"failed to get Pod of PodMigrationJob" err="Pod \"does-not-exist\" not found" PodMigrationJob="pmj-demo"
```

**Environment**:

- App version: main
- Kubernetes version (use kubectl version): v1.35.0
- Install details (e.g. helm install args): kind + `deploy_kind.sh`
  - Containerd/Docker version: 29.3.0 (Docker)
  - OS version: Ubuntu 22.04.5 LTS
  - Kernal version: 6.8.0-106-generic
- Others:

Apolo151:
When the migration arbitrator cannot resolve the target pod, fail the PodMigrationJob with MissingPod instead of approving it for migration.

The migration controller’s arbitrator resolves each PodMigrationJob’s target pod via getPodForJob. When the pod is missing (deleted, not yet created, or PodRef cannot be resolved), the job is absent from podOfJob, so filtering receives nil.

Previously, filtering(nil) skipped filter checks and returned isPassed = true, which led to updatePassedJob and set descheduler.koordinator.sh/passed-arbitration. That incorrectly approved migration for a job with no resolvable pod.

This PR:

1. `filtering`: Treats a `nil` pod as failure (`isFailed = true`) and returns early, so the job never receives the passed-arbitration annotation.
2. `updateFailedJob`: When `pod == nil`, sets `Status.Phase = Failed`, `Status.Reason = MissingPod`, and a message naming the pod reference (aligned with the migration controller’s existing missing-pod handling). Filter failures on a real pod still use `ForbiddenMigratePod`.
3. Tests: Updates `TestFiltering` for the nil-pod case and adds `TestDoOnceArbitrateMissingPodFails` for the end-to-end arbitrate path.

fixes #3039

Run the relevant tests

```bash
go test ./pkg/descheduler/controllers/migration/arbitrator -run 'TestFiltering|TestDoOnceArbitrateMissingPodFails' -v
```

- **Scope**: Only the arbitrator path when `getPodForJob` leaves `podOfJob[job]` unset (`PodRef` nil, `Get` error, or pod not in API). Transient API errors still omit the pod from the map; this change fails the job instead of approving it. If retry-on-transient-error is desired, that would be a separate follow-up.
- **Consistency**: `PodMigrationJobReasonMissingPod` is already used in `pkg/descheduler/controllers/migration/controller.go` for missing pods; this PR aligns arbitrator behavior with that.
- **Event reason**: `updateFailedJob` now uses `job.Status.Reason` in the event (so missing-pod jobs emit `MissingPod`, not always `ForbiddenMigratePod`).

### V. Checklist

### DIFF
--- a/pkg/descheduler/controllers/migration/arbitrator/arbitrator.go
+++ b/pkg/descheduler/controllers/migration/arbitrator/arbitrator.go
@@ -155,16 +155,18 @@ func (a *arbitratorImpl) sort(jobs []*v1alpha1.PodMigrationJob, podOfJob map[*v1
 
 // filtering calls nonRetryablePodFilter and retryablePodFilter to filter one PodMigrationJob.
 func (a *arbitratorImpl) filtering(pod *corev1.Pod) (isFailed, isPassed bool) {
-	if pod != nil {
-		markPodArbitrating(pod)
-		if a.filter.nonRetryablePodFilter != nil && !a.filter.nonRetryablePodFilter(pod) {
-			isFailed = true
-			return
-		}
-		if a.filter.retryablePodFilter != nil && !a.filter.retryablePodFilter(pod) {
-			isPassed = false
-			return
-		}
+	if pod == nil {
+		isFailed = true
+		return
+	}
+	markPodArbitrating(pod)
+	if a.filter.nonRetryablePodFilter != nil && !a.filter.nonRetryablePodFilter(pod) {
+		isFailed = true
+		return
+	}
+	if a.filter.retryablePodFilter != nil && !a.filter.retryablePodFilter(pod) {
+		isPassed = false
+		return
 	}
 	isPassed = true
 	return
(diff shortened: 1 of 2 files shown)


mdryaan (commented):
I have reviewed it, LGTM.
cc @ZiMengSheng

PR #3043, opened 2026-05-29
test: add unit tests for pkg/util/selector.go

Vatsalpatni73:

This PR adds unit tests for `pkg/util/selector.go`, specifically testing the `GetFastLabelSelector` helper function which previously had 0% coverage. A new table-driven test `TestGetFastLabelSelector` was introduced in `pkg/util/selector_test.go`. It verifies the fast path for match-labels-only selectors, the fallback path for `MatchExpressions`, consistent behavior for empty selectors, and expected error handling for invalid match expressions. 

NONE

Run the targeted tests locally to verify they pass and check the coverage improvement:
```bash
go test ./pkg/util -v -run TestGetFastLabelSelector
go test ./pkg/util -coverprofile=coverage.out && go tool cover -func=coverage.out
```

### V. Checklist

### DIFF
--- a/pkg/util/selector_test.go
+++ b/pkg/util/selector_test.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


ZiMengSheng:

/lgtm

PR #3046, opened 2026-05-30
scheduler: remove unused nodeNUMAResouece lock

ZiMengSheng:
### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/preempt.go
+++ b/pkg/scheduler/plugins/nodenumaresource/preempt.go
@@ -127,7 +127,6 @@ func (p *Plugin) AddPod(_ context.Context, cycleState fwktype.CycleState, preemp
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/nodenumaresource/preempt.go
```diff
@@ -127,7 +127,6 @@ func (p *Plugin) AddPod(_ context.Context, cycleState fwktype.CycleState, preemp
 		return nil
 	}
 
-	state.schedulingStateData.lock.Lock()
```
saintube, line 130: Why is it safe to write the states.preemptibleState lockless?

ZiMengSheng: Also deep copy it when clone

saintube: Recommended to add a comment about this assumption.

PR #3053, opened 2026-06-01
test(scheduler): add unit tests for `pkg/scheduler/plugins/elasticquota/metrics.go`

### LINKED ISSUE #3052: [Test] `pkg/scheduler/plugins/elasticquota/metrics.go` has no unit tests
**What happened**:

`pkg/scheduler/plugins/elasticquota/metrics.go` defines Prometheus metrics for the ElasticQuota scheduler plugin (`elastic_quota_spec`, `elastic_quota_status`, `update_elastic_quota_status_duration_seconds`) and helper functions to record and delete gauge series (`RecordElasticQuotaMetric`, `DeleteElasticQuotaMetric`, plus internal `recordResourceMetric` / `deleteResourceMetric`).

It had no dedicated unit tests. Existing coverage only exercised metrics indirectly via `Test_syncElasticQuotaMetrics` and `Test_deleteElasticQuotaMetrics` in `controller_test.go`, and only through `ElasticQuotaStatusMetric` on full quota sync paths. 

Important behavior was untested in isolation:

- CPU values are stored in **millicores** (`MilliValue`) while other resources use **raw `Value()`**
- `DeleteElasticQuotaMetric` partial vs. full cleanup (stale series after quota delete, related to [#2491](https://github.com/koordinator-sh/koordinator/issues/2491))
- Label injection (`resource`, `field`) on top of caller-provided quota labels
- Shared `labels` map mutation across multi-resource loops
- Metric registration via `init()` for all three package-level instruments

Add `metrics_test.go` covering:

- `RecordElasticQuotaMetric`: CPU MilliValue conversion, non-CPU/GPU Value conversion, multi-resource recording, empty `ResourceList`
- `DeleteElasticQuotaMetric`: full delete (zero series remain) and subset delete (only matching series removed)
- `RecordElasticQuotaMetric` label behavior: distinct `resource`/`field` per series when reusing one labels map
- `TestMetricsRegistered`: `ElasticQuotaSpecMetric`, `ElasticQuotaStatusMetric`, and `UpdateElasticQuotaStatusLatency` are non-nil and usable after registration

Apolo151:

Adds `pkg/scheduler/plugins/elasticquota/metrics_test.go` to give direct unit coverage for `metrics.go`, which previously had none.

The new tests exercise:

- **`RecordElasticQuotaMetric`**: CPU is recorded in millicores (`CPU(2)` → `2000`), memory and GPU use raw `Value()`, multi-resource lists emit one series per resource with correct values, and an empty `ResourceList` emits no series. Base quota labels (`name`, `tree`, `is_parent`, `parent`) are preserved with `resource` and `field` injected per series.
- **`DeleteElasticQuotaMetric`**: Full delete clears all recorded series; subset delete removes only matching resources (e.g. CPU removed while memory/GPU remain), guarding against stale Prometheus series after partial cleanup.
- **`RecordElasticQuotaMetric` label behavior**: A single shared `labels` map across a multi-resource list produces distinct `resource`/`field` per series with no duplicates.
- **`TestMetricsRegistered`**: Package-level `ElasticQuotaSpecMetric`, `ElasticQuotaStatusMetric`, and `UpdateElasticQuotaStatusLatency` are non-nil; `ElasticQuotaSpecMetric` records after `init()` registration. Global gauges are reset in test cleanup.

Tests use per-test registered `metrics.GaugeVec` instances (k8s component-base gauges are noop until registered via `koordschedulermetrics.RegisterMetrics`) and the same channel-collect pattern as `controller_test.go`, so global metric state is not polluted between cases.

No production code changes.

fixes #3052 

Run the new tests:

```bash
go test -count=1 -v ./pkg/scheduler/plugins/elasticquota/ -run 'TestRecordElasticQuotaMetric|TestDeleteElasticQuotaMetric|TestMetrics'
```

Run the full package (including existing controller metrics tests):

```bash
go test -count=1 ./pkg/scheduler/plugins/elasticquota/
```

- `newRegisteredTestGaugeVec` assigns a unique metric name per test via an atomic counter so gauges can be registered without colliding in parallel or repeated subtests.
- `TestMetricsRegistered` touches `ElasticQuotaSpecMetric` globally but calls `Reset()` in `t.Cleanup` to avoid leaking state into other tests in the package.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/elasticquota/metrics_test.go
+++ b/pkg/scheduler/plugins/elasticquota/metrics_test.go
@@ -1,0 +1,275 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package elasticquota
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"sync/atomic"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/component-base/metrics"
+
+	koordschedulermetrics "github.com/koordinator-sh/koordinator/pkg/scheduler/metrics"
+)
+
+var testGaugeCounter atomic.Uint64
+
+// newRegisteredTestGaugeVec returns a GaugeVec registered with the scheduler metrics
+// registry. k8s component-base GaugeVecs are lazily initialized and noop until registered.
+func newRegisteredTestGaugeVec(t *testing.T) *metrics.GaugeVec {
+	t.Helper()
+	id := testGaugeCounter.Add(1)
+	gauge := metrics.NewGaugeVec(
+		&metrics.GaugeOpts{
+			Subsystem: "test",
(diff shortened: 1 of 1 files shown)


ZiMengSheng:

/lgtm

PR #3056, opened 2026-06-03
test: add unit tests for pkg/util/parallelize/parallelism.go

Vatsalpatni73:

This PR adds missing unit test coverage for `pkg/util/parallelize/parallelism.go`. The test file `parallelism_test.go` exercises the pure utility logic in the package:
- `chunkSizeFor(n int)`: Tested against zero, small, exact-boundary, and large `n` values to ensure optimal chunk boundaries.
- `Until(...)`: Tested to ensure it executes each work item exactly once, safely handles `pieces == 0` without invoking the worker, and properly respects context cancellation.

NONE

Run the added unit tests for the package:
```bash
go test -v ./pkg/util/parallelize
```

### V. Checklist

### DIFF
--- a/pkg/util/parallelize/parallelism_test.go
+++ b/pkg/util/parallelize/parallelism_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


ZiMengSheng:

/lgtm

PR #3061, opened 2026-06-05
koordlet: fix perf group cpu error accumulation

### LINKED ISSUE #3060: [BUG]: koordlet perf-group collector loses earlier per-CPU errors and publishes partial CPI metrics
**what happened**:

`GetContainerPerfResult` can lose errors while collecting cpi metrics across multiple cpus:

```go
if err = pc.collect(collector.valueCh); err != nil {
	err = multierr.Append(err, err)
}
```

if one cpu fails and a later cpu succeeds, the later call overwrites `err` with `nil`. the function can then return partial cycle and instruction totals as a successful result. it also appends the error to itself instead of preserving errors from each cpu.

**what you expected to happen**:

errors from every cpu should be preserved. any collection failure should make the function return an error so incomplete cpi metrics are not recorded.

```go
if collectErr := pc.collect(collector.valueCh); collectErr != nil {
	err = multierr.Append(err, collectErr)
}
```

**how to reproduce it (as minimally and precisely as possible)**:

add a unit test with two mocked cpu collectors. make the first collection fail and the second succeed, then call `GetContainerPerfResult`.

the current code returns `nil` because the second result overwrites the first error. the expected result is a non-nil error containing the first failure.

**anything else we need to know?**:

this affects the libpfm4 path when `CPICollector` and `Libpfm4` are enabled. incomplete values can reach the metric cache and prometheus because the caller only rejects a measurement when this function returns an error.

i searched open and closed issues and pull requests and found no existing report for this behavior. issue #1568 is different; it covered cgroup file cleanup and zero cycle values.

**screenshots**:

`GetContainerPerfResult` overwrites `err`, so a later successful CPU can hide an earlier collection failure.

Image

**environment**:

- app version: current `main` (`d30ca464`)
- kubernetes version (use `kubectl version`): kubectl client `v1.35.0`; server version unavailable because the configured `kind-mcp-gateway` cluster is not running
- install details (e.g. helm install args): not reproduced in a running koordinator installation
- node environment (for koordlet/runtime-proxy issue):
  - containerd/docker version: docker desktop `4.51.0`, docker engine `28.5.2`, containerd `1.7.29`
  - os version: macos `26.4.1` on arm64
  - kernel version: darwin `25.4.0`; docker vm uses linuxkit `6.12.54`
  - cgroup driver: `cgroupfs`, cgroup v2 inside the docker vm
- others:
  - helm version: `v4.1.4`
  - this bug was found through code inspection and was not reproduced on this macos machine

**reproduction requirements**:

- a linux node where koordlet can use perf events
- `CPICollector=true`
- `Libpfm4=true`
- more than one cpu available to the container collector
- one cpu collection must fail while a later cpu collection succeeds

the repository's `config/manager/koordlet.yaml` enables all alpha and beta feature gates, including these two alpha features. this does not mean they are currently running on this machine.

mugiwaraluffy56:

Preserves per-cpu collection errors in the perf-group cpi collector.

`GetContainerPerfResult` was reusing the same `err` variable inside the cpu loop. If one cpu collection failed and a later cpu collection succeeded, the later success could overwrite the earlier error. That made it possible to return partial cycle and instruction totals as if collection had fully succeeded.

This keeps the accumulated error separate from each per-cpu `collectErr`, so any cpu collection failure is still returned to the caller.

fixes #3060

```bash
docker run --rm \
  -v "$PWD":/workspace \
  -w /workspace \
  golang:1.25-bookworm \
  bash -c 'apt-get update >/dev/null && \
    apt-get install -y libpfm4-dev >/dev/null && \
    go test ./pkg/koordlet/util/perf_group \
      -run TestGetContainerPerfResultPreservesErrors \
      -count=1 \
      -v && \
    go test ./pkg/koordlet/util/perf_group -count=1'
```

The perf-group package is linux-only and depends on libpfm4, so I verified the test in a linux container.

No screenshots are needed for this PR. The regression test covers the failing path directly.

### V. Checklist

### DIFF
--- a/pkg/koordlet/util/perf_group/perf_group_linux.go
+++ b/pkg/koordlet/util/perf_group/perf_group_linux.go
@@ -246,8 +246,8 @@ func GetContainerPerfResult(collector *PerfGroupCollector) (map[string]float64, 
 	var err error
 	for _, cpu := range collector.cpus {
 		if pc, ok := collector.perfCollectors[cpu]; ok {
-			if err = pc.collect(collector.valueCh); err != nil {
-				err = multierr.Append(err, err)
+			if collectErr := pc.collect(collector.valueCh); collectErr != nil {
+				err = multierr.Append(err, collectErr)
 			}
 		}
 	}
--- a/pkg/koordlet/util/perf_group/perf_group_linux_test.go
+++ b/pkg/koordlet/util/perf_group/perf_group_linux_test.go
@@ -141,3 +141,81 @@ func Test_GetContainerCyclesAndInstructions(t *testing.T) {
 	//assert.Nil(t, err)
 	//LibFinalize()
 }
+
+func TestGetContainerPerfResultPreservesErrors(t *testing.T) {
+	InitBufferPool(map[int]struct{}{1: {}})
+	oldNew := perfValuePool.New
+	perfValuePool.New = func() interface{} {
+		return &value{}
+	}
+	defer func() {
+		perfValuePool.New = oldNew
+	}()
+
+	tempDir := t.TempDir()
+	cgroupFile, err := os.OpenFile(tempDir, os.O_RDONLY, os.ModeDir)
+	assert.NoError(t, err)
+
+	failedCPUFile := newPerfValueFile(t, tempDir, "failed-cpu", 1, 10)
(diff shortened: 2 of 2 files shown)


mugiwaraluffy56:
@saintube @zwzhang0107 can u review this, thanks!

Issue #3058, opened 2026-06-05
[feature] support schedule-admission in pod label


KunWuLuan:
**What is your proposal**:

Add a plugin in scheduler to block pod scheduling by label.

**Why is this needed**:

SchedulingGates is a mechanism used by controllers to mark Pods as not ready for scheduling. However, schedulingGates can only be set at Pod creation time and cannot be added later.

koord-queue is a queueing system designed for job scheduling. It needs the ability to temporarily prevent certain Pods from being scheduled, for example when a higher-priority job should be scheduled first. In some cases, these Pods also cannot be evicted once created. Therefore, koord-queue needs a label or similar marker to indicate that a Pod is not yet ready to be scheduled.

**Is there a suggested solution, if so, please add it**:

PR #3062, opened 2026-06-08
feat: add schedule-admission PreEnqueue plugin for label-based pod gating

KunWuLuan:

Implement a schedule-admission mechanism that allows multiple controllers to gate pod scheduling via labels with prefix
"scheduling.koordinator.sh/schedule-admission-". A pod is only schedulable when all such labels are removed, similar to Kubernetes schedulingGates.

#3058 

### V. Checklist


cmd/koord-scheduler/main.go
```diff
@@ -52,6 +53,7 @@ var koordinatorPlugins = map[string]frameworkruntime.PluginFactory{
 	defaultprebind.Name:          defaultprebind.New,
 	noderesourcesfitplus.Name:    noderesourcesfitplus.New,
 	scarceresourceavoidance.Name: scarceresourceavoidance.New,
+	scheduleadmission.Name:      scheduleadmission.New,
```
ZiMengSheng: @KunWuLuan PTAL

apis/extension/schedule_admission.go
```diff
@@ -0,0 +1,60 @@
...
+	// LabelScheduleAdmissionPrefix is the label key prefix used by controllers to gate pod scheduling.
+	// Each controller adds a label with this prefix and a unique suffix (gate name) with value "true".
+	// A pod is only schedulable when no labels with this prefix exist.
+	// Example: scheduling.koordinator.sh/schedule-admission-quota-check: "true"
+	LabelScheduleAdmissionPrefix = SchedulingDomainPrefix + "/schedule-admission-"
+)
+
+// HasScheduleAdmissionLabels returns true if the pod has any label with the schedule-admission prefix,
+// meaning the pod is gated and should not be scheduled.
+func HasScheduleAdmissionLabels(pod *corev1.Pod) bool {
```
saintube: This may be costly with many pod labels. Could we add a fast path for full-context matching without the prefix overhead?

KunWuLuan: Do you mean only 1 valid label for check?

pkg/scheduler/plugins/scheduleadmission/schedule_admission.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (pl *Plugin) PreEnqueue(ctx context.Context, pod *corev1.Pod) *fwktype.Status {
+	if extension.HasScheduleAdmissionLabels(pod, pl.enablePrefixMatch) {
+		gates := extension.GetScheduleAdmissionGates(pod, pl.enablePrefixMatch)
+		return fwktype.NewStatus(fwktype.UnschedulableAndUnresolvable,
+			fmt.Sprintf("pod has schedule-admission gates: %v", gates))
+	}
+	return fwktype.NewStatus(fwktype.Success, "")
+}
+
```
ZiMengSheng, line 95: Should we also provide a PreFilter extension point? This way, Pods accumulating in ActiveQ can still be intercepted even after the controller has applied the label.

KunWuLuan: Good point. Added a PreFilter extension point in the latest commit: it shares the same admission logic with PreEnqueue (extracted into a common admit() method), so pods that entered the ActiveQ before the controller applied the schedule-admission label are now rejected with UnschedulableAndUnresolvable during the scheduling cycle. The existing UpdatePodLabel queueing hint still handles re-enqueueing once all gate labels are removed. PTAL.

apis/extension/schedule_admission.go
KunWuLuan: Resolved in 3dc6803: the plugin now defaults to an O(1) exact match on the fixed label scheduling.koordinator.sh/schedule-admission (fast path), without iterating pod labels. The prefix scan over schedule-admission-* labels is only performed when enablePrefixMatch is explicitly set to true in the plugin args, for users who need multiple independent gates. PTAL.

PR #3072, opened 2026-06-12
opt: move nodenumaresource NUMA hint checking to Filter phase(3070)

### LINKED ISSUE #3070: [Optimization] Move NUMA hint checking earlier in the scheduling cycle to save CPU
**What happened**

While reviewing the `nodenumaresource` scheduler plugin, I noticed a `FIXME` comment in `pkg/scheduler/plugins/nodenumaresource/plugin.go` indicating that `FilterByNUMANode` should be moved earlier in the scheduling cycle.

Currently, NUMA topology compatibility is checked relatively late (during resource allocation/reservation-related stages). As a result, the scheduler may perform resource allocation calculations for nodes that are later rejected because their NUMA topology does not satisfy the Pod's requirements.

**What you expected to happen**

Nodes that do not satisfy a Pod's NUMA topology requirements should ideally be filtered out as early as possible during scheduling. Moving the `FilterByNUMANode` check into the scheduler's `Filter` phase may help eliminate unsuitable nodes earlier and avoid unnecessary allocation calculations.

**How to reproduce it (as minimally and precisely as possible)**

1. Open `pkg/scheduler/plugins/nodenumaresource/plugin.go`.
2. Locate the existing `FIXME` comment related to moving `FilterByNUMANode` into the `Filter` phase.
3. Review the scheduling flow and observe that NUMA topology validation currently occurs after some resource allocation computations have already been performed.

The code already contains a `FIXME` describing the desired optimization. Since the change appears to involve moving existing validation logic to an earlier scheduling phase, this may be a good candidate for a `good first issue` and/or `help wanted` label, subject to maintainer review.

**Environment**

* App version: Latest `main` branch
* Kubernetes version: N/A (code-level observation)
* Installation method: N/A
* Runtime environment: N/A
* Other relevant details: None

krsatyamthakur-droid:

This PR resolves a `FIXME` comment in `pkg/scheduler/plugins/nodenumaresource/plugin.go`. 

Previously, the plugin waited until relatively late in the scheduling cycle to check if a node actually matches a Pod's NUMA topology hints using `FilterByNUMANode`. Because this check happened so late, the scheduler would perform heavy resource allocation calculations for nodes that were ultimately going to be thrown out anyway.

I shifted the `FilterByNUMANode` check earlier into the `Filter` phase (specifically right after the `designatedAllocation` short-circuit). Now, nodes that fail the NUMA topology check are dropped immediately, saving the scheduler from doing wasted math and speeding up scheduling throughput.

fixes #3070

You can run the existing unit tests for the nodenumaresource plugin. Everything passes cleanly, which confirms that the new placement still perfectly respects the `designatedAllocation` logic while dropping unsuitable nodes earlier.

This is a straightforward cleanup that just moves the existing validation logic up to resolve the `FIXME`!

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/plugin.go
+++ b/pkg/scheduler/plugins/nodenumaresource/plugin.go
@@ -369,6 +369,7 @@ func (p *Plugin) Filter(ctx context.Context, cycleState fwktype.CycleState, pod 
 	if err != nil {
 		return fwktype.NewStatus(fwktype.UnschedulableAndUnresolvable, ErrNotMatchNUMATopology)
 	}
+
 	if state.designatedAllocation != nil {
 		status = p.allocate(ctx, cycleState, pod, nodeInfo.Node(), numaTopologyPolicy)
 		if !status.IsSuccess() {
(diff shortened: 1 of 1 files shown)


krsatyamthakur-droid:
Hey @ZiMengSheng  — small cleanup, all tests pass. Would love a review when you get a chance!

pkg/scheduler/plugins/nodenumaresource/plugin.go
```diff
@@ -434,17 +450,6 @@ func (p *Plugin) Filter(ctx context.Context, cycleState fwktype.CycleState, pod
 		}
 	}
 
-	// FIXME: move it ahead the resourceManager.Allocate so that we can check with NUMA hints almost in the Filter
```
ZiMengSheng, line 437: maybe this comment is misleading, I haven't yet seen the benefit of putting him up front. Since the NUMA check has higher time complexity and is more time-consuming, placing it first might actually make the situation worse?

PR #3074, opened 2026-06-13
scheduler: fix swapped total/free args in deviceshare shared-GPU scoring

### LINKED ISSUE #3073: [BUG] deviceshare: scoreDevice receives total/free in swapped order for shared-GPU scoring
**What happened**:

In the GPU topology allocator, `allocateFromScope` scores each candidate GPU minor for shared-GPU (`koordinator.sh/gpu-shared`) requests via:

```go
contextOfDevice.score = allocateContext.allocationScorer.scoreDevice(
    requirements.requestsPerGPU, freeResources, totalResources)
```

(`pkg/scheduler/plugins/deviceshare/allocator_gpu.go:412`)

But `scoreDevice`'s signature is `scoreDevice(podRequest, total, free corev1.ResourceList)` (`pkg/scheduler/plugins/deviceshare/scoring.go:201`). The `total` and `free` arguments are passed in **swapped order**. The only other caller, `pkg/scheduler/plugins/deviceshare/device_resources.go:176`, passes them correctly: `scoreDevice(podRequest, totalResources[minor], free)`.

Because of the swap, inside `scoreDevice` the `totalQuantity.Cmp(freeQuantity) >= 0` branch and the `allocatable`/`requested` values are computed from the wrong amounts. For any partially-used minor the computed `requested` ends up equal to `allocatable`, so the score collapses to a constant ("fully allocated") value and the scorer can no longer distinguish minors at different utilization.

For shared-GPU scheduling, the scoring strategy (e.g. MostAllocated / LeastAllocated) should rank candidate GPU minors by their real utilization and pick accordingly (MostAllocated → most-utilized minor, to bin-pack). Instead, among multiple eligible partially-used minors the scores tie and selection falls back to minor iteration order, so the wrong GPU minor is chosen.

Node with two shareable GPU minors at different utilization (both with enough free memory for the request), DeviceShare scoring strategy = `MostAllocated`. Submit a shared-GPU pod: it is placed on the lower-indexed minor rather than the more-utilized one.

A focused unit test demonstrating this (`TestAllocateFromScopeSharedGPUScoreUsesTotalAndFree`) is included in the linked PR — it fails on the current code (selects minor 6) and passes after the fix (selects the more-utilized minor 7).

One-line fix in `pkg/scheduler/plugins/deviceshare/allocator_gpu.go`. PR attached.

**Environment**:
- App version: master (the logic is present regardless of release)
- Others: affects the koord-scheduler `deviceshare` plugin shared-GPU scoring path

namansh70747:

`allocateFromScope` scored shared-GPU candidate minors by calling `scoreDevice(requestsPerGPU, freeResources, totalResources)`, but the method signature is `scoreDevice(podRequest, total, free)` — `total` and `free` were passed in swapped order. The correct usage is in `device_resources.go`. With the swap, every partially-used minor produces an identical "fully allocated" score, so the scoring strategy (MostAllocated / LeastAllocated) can no longer differentiate minors and selection degenerates to iteration order, placing shared-GPU pods on the wrong minor.

This PR passes the arguments in the correct order and adds a regression test.

fixes #3073

```
go test ./pkg/scheduler/plugins/deviceshare/ -run TestAllocateFromScopeSharedGPUScoreUsesTotalAndFree
```

The new test sets up two shared-GPU minors (minor 6: 75/100 free, minor 7: 50/100 free) under MostAllocated and asserts the more-utilized minor (7) is chosen. It fails on the current code (selects minor 6) and passes with this fix.

Single-line production change; the rest is the regression test.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/allocator_gpu.go
+++ b/pkg/scheduler/plugins/deviceshare/allocator_gpu.go
@@ -409,7 +409,7 @@ func allocateFromScope(requirements *GPURequirements, scope *GPUTopologyScope, a
 			_, belongToTotal := allocateContext.deviceTotal[minor]
 			contextOfDevice.satisfied = contextOfDevice.satisfied && belongToTotal
 			if contextOfDevice.satisfied && requirements.gpuShared && allocateContext.allocationScorer != nil {
-				contextOfDevice.score = allocateContext.allocationScorer.scoreDevice(requirements.requestsPerGPU, freeResources, totalResources)
+				contextOfDevice.score = allocateContext.allocationScorer.scoreDevice(requirements.requestsPerGPU, totalResources, freeResources)
 			}
 		}
 		if !contextOfDevice.satisfied {
--- a/pkg/scheduler/plugins/deviceshare/allocator_gpu_test.go
+++ b/pkg/scheduler/plugins/deviceshare/allocator_gpu_test.go
@@ -1690,6 +1690,58 @@ func TestAllocateSharedGPU(t *testing.T) {
 	}
 }
 
+func TestAllocateFromScopeSharedGPUScoreUsesTotalAndFree(t *testing.T) {
+	// Two shared-GPU minors in the same scope, both able to satisfy the request
+	// but at different utilization levels:
+	//   minor 6: 75 free out of 100 (25 used)
+	//   minor 7: 50 free out of 100 (50 used)
+	// With the MostAllocated strategy the scorer must prefer the more-utilized
+	// minor (7) to bin-pack shared GPUs. This only holds when scoreDevice is
+	// called as scoreDevice(request, total, free); if total and free are swapped,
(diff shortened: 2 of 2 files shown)


ZiMengSheng:
Verified against `main`. This is a real bug and the fix is correct.

The signature is `scoreDevice(podRequest corev1.ResourceList, total, free corev1.ResourceList)` (`scoring.go:201`), and `allocator_gpu.go:412` was calling it as `scoreDevice(requirements.requestsPerGPU, freeResources, totalResources)` — total and free swapped. Every shared-GPU minor was being scored against an inverted utilization ratio, so bin-packing would prefer exactly the wrong devices. Silent, and invisible without a test that uses asymmetric total/free values, which is what makes it worth catching.

Your test comment spells out the reasoning rather than just asserting a number, which is the right call for a fix like this — it's what stops someone from "simplifying" the argument order back later.

CI green.

/lgtm

PR #3076, opened 2026-06-13
apis: fix ExactMatchReservation returning early on absent resource

### LINKED ISSUE #3075: [BUG] ExactMatchReservation returns true too early when a resource is missing on both sides
**What happened**:

The function ExactMatchReservation in apis/extension/reservation.go checks every resource name listed in the spec. When a resource is missing from both the pod and the reservation the function returns true right away. This ends the loop early so any resource listed after it is never checked.

Because of this a pod can be treated as an exact match even when a later resource does not actually match.

When a resource is missing on both sides it should simply be skipped and the function should keep checking the rest of the list. It should return true only when all the listed resources match.

Set the spec resource names to memory and then cpu. Let memory be absent on both the pod and the reservation. Let the pod request cpu 2 while the reservation has cpu 1. Today the function returns true even though cpu does not match. The expected result is false.

A small unit test that shows this is included in the linked pull request. It fails before the change and passes after.

It looks like a one line change. I have prepared a fix and would be glad to send it if this looks right to you. Could you please confirm. Thank you for taking a look.

cc @ZiMengSheng

**Environment**:
- App version: master
- Others: koord scheduler reservation matching path

namansh70747:

ExactMatchReservation checks every resource name in the spec and must confirm each one matches. When a resource was absent on both the pod and the reservation it returned true immediately, which ended the loop and skipped the resources listed after it. A pod could then be treated as an exact match even when a later resource did not match.

This change skips the absent on both sides resource with continue so the rest of the list is still checked, and adds a regression test.

fixes #3075

```
go test ./apis/extension/ -run TestExactMatchReservation
```

The new case lists memory then cpu, with memory absent on both sides and cpu mismatched. It expects false. The case fails on the current code and passes after this change.

One line production change. The rest is the test.

### V. Checklist

### DIFF
--- a/apis/extension/reservation.go
+++ b/apis/extension/reservation.go
@@ -262,7 +262,10 @@ func ExactMatchReservation(podRequests, reservationAllocatable corev1.ResourceLi
 		request, existsInPod := podRequests[resourceName]
 		if !existsInReservation || !existsInPod {
 			if !existsInReservation && !existsInPod {
-				return true
+				// This resource is absent on both sides, so it trivially
+				// matches. Continue checking the remaining resources instead
+				// of returning early, otherwise a later mismatch is missed.
+				continue
 			}
 			return false
 		}
--- a/apis/extension/reservation_test.go
+++ b/apis/extension/reservation_test.go
@@ -251,6 +251,22 @@ func TestExactMatchReservation(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "resource absent on both sides must not skip a later mismatch",
+			spec: &ExactMatchReservationSpec{
+				ResourceNames: []corev1.ResourceName{
+					corev1.ResourceMemory,
+					corev1.ResourceCPU,
+				},
+			},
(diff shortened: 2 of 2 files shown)


ZiMengSheng:
Verified against `main`. Real bug, correct fix.

```go
if !existsInReservation || !existsInPod {
    if !existsInReservation && !existsInPod {
        return true        // ← skips every remaining resource name
    }
    return false
}
```

`ExactMatchReservation` must confirm that *every* name in `spec.ResourceNames` matches. Returning `true` on the first name that's absent from both sides abandons the rest of the loop, so with `ResourceNames: [memory, cpu]` a pod with no memory request would match a reservation whose CPU doesn't match at all. Exact-match reservations silently stop being exact.

`continue` is the right fix, and your test case name — "resource absent on both sides must not skip a later mismatch" — describes the actual failure mode rather than the code shape. That ordering (absent resource first, mismatching resource second) is what makes the test meaningful; a reversed order would pass even with the bug.

CI green.

/lgtm

PR #3078, opened 2026-06-13
scheduler: key required device resources by minor value in deviceshare

### LINKED ISSUE #3077: [BUG] deviceshare calcRequiredDeviceResources keys the map by slice index not minor value
**What happened**:

In apis deviceshare the function calcRequiredDeviceResources builds a map for a reservation that has no capacity left. To do this it walks the device minors with a loop like this

```go
for minor := range minors.UnsortedList() {
    resources[minor] = corev1.ResourceList{}
}
```

UnsortedList returns a slice of minor values. A range loop with one variable over a slice gives the index of each item and not the value. So the map ends up keyed by 0 1 2 and so on instead of the real device minors.

When the device minors are not a simple run starting from zero the map points at the wrong devices.

The map should be keyed by the actual minor values that come from the reservation.

Use a reservation whose GPU minors are not contiguous, for example minor 3 and minor 5, with no capacity remaining. The required map comes back keyed by 0 and 1 instead of 3 and 5.

A small unit test that shows this is included in the linked pull request. It fails before the change and passes after.

It looks like a one line change, just adding the value variable to the range loop. I have prepared a fix and would be glad to send it if this looks right to you. Could you please confirm. Thank you for taking a look.

cc @ZiMengSheng @zqzten

**Environment**:
- App version: master
- Others: koord scheduler deviceshare reservation path

namansh70747:

calcRequiredDeviceResources builds an empty capacity required map for a fully consumed reservation. It ranged over minors.UnsortedList() with a single loop variable, which iterates the slice indices rather than the minor values, so the map was keyed by 0..n-1 instead of the real device minors. With non contiguous minors the reservation constraint was then applied to the wrong devices.

This change ranges over the minor values and adds a regression test.

fixes #3077

```
go test ./pkg/scheduler/plugins/deviceshare/ -run TestCalcRequiredDeviceResourcesKeyedByMinor
```

The new test uses a reservation with non contiguous GPU minors 3 and 5 and asserts the required map is keyed by 3 and 5. It fails on the current code (keys 0 and 1) and passes after this change.

One line production change. The rest is the test.

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/reservation.go
+++ b/pkg/scheduler/plugins/deviceshare/reservation.go
@@ -445,7 +445,7 @@ func calcRequiredDeviceResources(alloc *reusableAlloc, preemptibleInRR map[sched
 		}
 		for deviceType, minors := range minorHints {
 			resources := deviceResources{}
-			for minor := range minors.UnsortedList() {
+			for _, minor := range minors.UnsortedList() {
 				resources[minor] = corev1.ResourceList{}
 			}
 			required[deviceType] = resources
--- a/pkg/scheduler/plugins/deviceshare/reservation_test.go
+++ b/pkg/scheduler/plugins/deviceshare/reservation_test.go
@@ -1464,3 +1464,29 @@ func Test_isDeviceAllocationsInclude(t *testing.T) {
 		})
 	}
 }
+
+func TestCalcRequiredDeviceResourcesKeyedByMinor(t *testing.T) {
+	// A fully consumed reservation produces an empty-capacity required map.
+	// The map must be keyed by the real device minors. With non-contiguous
+	// minors (3 and 5) this only holds if we range over the minor values; if
+	// we range over the slice itself we get the indices 0 and 1 instead, which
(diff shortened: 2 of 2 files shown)


ZiMengSheng:
Verified against `main`. Real bug, correct fix.

`reservation.go:445` was `for minor := range minors.UnsortedList()`. `UnsortedList()` returns a slice, so a single-variable `range` yields **indices**, not minor values. For a fully consumed reservation the required map ends up keyed `0, 1, 2...` instead of by real device minor.

This is benign only when minors happen to be contiguous from zero. With non-contiguous minors — a node where GPUs 3 and 5 are allocated to the reservation — the zero-capacity constraint gets attached to minors 0 and 1, which means the *wrong* devices are marked exhausted and the actually-exhausted ones look free. Partial GPU allocation on such a node would be allowed when it shouldn't be.

Your test picks minors 3 and 5 specifically, which is what makes it a real test: with 0 and 1 it would pass even with the bug present. The comment explaining why non-contiguous values matter is worth keeping.

CI green.

/lgtm

PR #3080, opened 2026-06-14
test: add unit tests for pkg/webhook/util/writer/atomic/atomic_writer.go

### LINKED ISSUE #3079: [test improvement] Add unit tests for pkg/webhook/util/writer/atomic/atomic_writer.go
**What happened:**
`pkg/webhook/util/writer/atomic/atomic_writer.go` contains the package `atomic` used by the koord-manager webhook cert writer to project cert / CA / key files into a target directory atomically. It exports the `Writer` type, the `FileProjection` struct (`Data []byte`, `Mode int32`), and two methods: `NewAtomicWriter(targetDir string) (*Writer, error)` (returns an error if `targetDir` does not exist) and `(*Writer).Write(payload map[string]FileProjection) error` (the 11-step atomic projection algorithm). Internally it has nine unexported helpers: `validatePayload`, `validatePath`, `shouldWritePayload`, `shouldWriteFile`, `pathsToRemove`, `newTimestampDir`, `writePayloadToDir`, `createUserVisibleFiles`, `removeUserVisiblePaths`. None of these has a unit test — `pkg/webhook/util/writer/atomic/` contains only `atomic_writer.go`. The behavior of `validatePath` (rejects empty / absolute / `..`-containing / > 4096-char / > 255-char-filename / starts-with-`..` paths) and the no-op-vs-rewrite branch of `shouldWritePayload` are entirely untested.

**How to reproduce:**
```bash
cd <repo-root>
go test ./pkg/webhook/util/writer/atomic/... -v
```
Expected output: `[no test files]` — confirming zero tests exist.

Table-driven tests covering:

- `NewAtomicWriter`: success with an existing directory; error when the target directory does not exist.
- `validatePath`: empty path, absolute path, path containing `..` as an element, single component equal to `..`, path starting with `..foo` (3+ chars), path longer than 4096 chars, filename longer than 255 chars, valid relative path is accepted.
- `shouldWriteFile` (via `Write` / `shouldWritePayload`): returns `true` for a missing file; returns `false` for identical content; returns `true` for different content.
- `Write` end-to-end against a `t.TempDir()`:
  - First write of `{ "tls.crt": {...}, "tls.key": {...} }` creates the `..data` symlink, the timestamped data dir, and the user-visible `tls.crt` / `tls.key` symlinks; reading via the symlink returns the projected bytes.
  - Second write with the same payload is a no-op (no new timestamped dir created).
  - Third write with changed `tls.crt` data updates the symlink and the new bytes are visible; a new timestamped dir replaces the old one.
  - Fourth write with `tls.key` removed from the payload deletes the `tls.key` symlink and the file is no longer readable via the target dir, while `tls.crt` remains.
  - `Write` with a path equal to `""`, absolute, or containing `..` returns the validation error and does not touch the filesystem.
- `Write` followed by a second `Write` after `os.RemoveAll(targetDir/..data)` still succeeds and recreates the projection from scratch.
- `Write` on a nested path like `"ca/bundle.crt"` creates the `ca` user-visible symlink and writes the file under the timestamped dir.

**What actually happened:**
No `*_test.go` file exists in `pkg/webhook/util/writer/atomic/`. `Writer.Write` and all unexported helpers have zero test coverage. A regression in `validatePath` (e.g. permitting `..`) or in the no-op branch of `shouldWritePayload` would be caught only at runtime in the webhook cert rotation path.

**Relevant log output:**
Screenshot of test output goes here.

**Environment:**
- App version: main
- Linux Ubuntu 24.04

- `pkg/webhook/util/writer/atomic/atomic_writer.go` — all 11 functions / methods (1 exported constructor, 1 exported `Write`, 9 unexported helpers).
- Tests are filesystem-bound; use `t.TempDir()` for the target directory and clean up between sub-tests. No Kubernetes API or fake clients needed.
- Reference test pattern for table-driven tests using only `testing` and `github.com/stretchr/testify/assert`: `pkg/webhook/cm/plugins/sloconfig/checker_test.go` (pure-function assertions) and `pkg/webhook/elasticquota/plugin_check_quota_meta_validate_test.go` (table-driven cases for webhook validation logic).
- For the `Write` end-to-end tests, capture the previous `..data` symlink target with `os.Readlink` before and after the call to assert the timestamped dir was rotated only when an actual change occurred.
- The path-validation rules in `validatePath` are documented in the function's comment block (rules 1–5); each rule should have at least one test case.

Digvijay-x1:

Adds unit tests for `pkg/webhook/util/writer/atomic/atomic_writer.go`, which had zero test coverage. The package implements the atomic file projection used by the koord-manager webhook cert writer.

fixes #3079

```bash
go test ./pkg/webhook/util/writer/atomic/... -v
go test ./pkg/webhook/util/writer/atomic/... -cover
```

Expected: all tests pass, ~74% statement coverage

image

No production code changes. Tests use only testing, github.com/stretchr/testify/assert, and the standard library. No Kubernetes API, fake clients, or cluster access required.

### V. Checklist

### DIFF
--- a/pkg/webhook/util/writer/atomic/atomic_writer_test.go
+++ b/pkg/webhook/util/writer/atomic/atomic_writer_test.go
@@ -1,0 +1,437 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package atomic
+
+import (
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewAtomicWriter(t *testing.T) {
+	t.Run("existing directory", func(t *testing.T) {
+		dir := t.TempDir()
+		w, err := NewAtomicWriter(dir)
+		assert.NoError(t, err)
+		assert.NotNil(t, w)
+		assert.Equal(t, dir, w.targetDir)
+	})
+
+	t.Run("non-existent directory", func(t *testing.T) {
+		dir := filepath.Join(t.TempDir(), "does-not-exist")
+		w, err := NewAtomicWriter(dir)
+		assert.Error(t, err)
+		assert.True(t, os.IsNotExist(err))
+		assert.Nil(t, w)
+	})
+}
+
+func TestValidatePath(t *testing.T) {
+	longFileName := strings.Repeat("a", maxFileNameLength+1)
+	longPath := strings.Repeat("a/", (maxPathLength/2)+1)
+
+	tests := []struct {
+		name      string
+		path      string
+		wantError bool
+	}{
+		{
+			name:      "empty path",
+			path:      "",
+			wantError: true,
+		},
+		{
+			name:      "absolute path",
+			path:      "/etc/passwd",
+			wantError: true,
+		},
+		{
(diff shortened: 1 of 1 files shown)


ZiMengSheng:

/lgtm

PR #3083, opened 2026-06-15
scheduler: fix wrong type names in noderesourcefitplus and scarceresourceavoidance args error

### LINKED ISSUE #3082: [BUG] noderesourcefitplus and scarceresourceavoidance print the wrong type in their args error
## What happened

While reading through the scheduler plugins I noticed that two of them, `noderesourcefitplus` and `scarceresourceavoidance`, give you a confusing error when their plugin args are the wrong type. Both build the error inside their `New()` constructor, and both get it wrong in the same way.

Here is the `noderesourcefitplus` one (`pkg/scheduler/plugins/noderesourcefitplus/node_resources_fit_plus.go:48`):

```go
func New(_ context.Context, args runtime.Object, handle fwk.Handle) (fwk.Plugin, error) {
	nodeResourcesFitPlusArgs, ok := args.(*config.NodeResourcesFitPlusArgs)

	if !ok {
		return nil, fmt.Errorf("want args to be of type NodeResourcesArgs, got %T", nodeResourcesFitPlusArgs)
	}
	...
}
```

And `scarceresourceavoidance` does the same thing (`pkg/scheduler/plugins/scarceresourceavoidance/scarce_resource_avoidance.go:50`):

```go
func New(_ context.Context, args runtime.Object, handle fwk.Handle) (fwk.Plugin, error) {
	scarceResourceAvoidanceArgs, ok := args.(*config.ScarceResourceAvoidanceArgs)

	if !ok {
		return nil, fmt.Errorf("want args to be of type ResourceTypesArgs, got %T", scarceResourceAvoidanceArgs)
	}
	...
}
```

There are actually two separate mistakes packed into that one line in each plugin.

The first one is the `%T`. It is being handed the variable that comes out of the failed type assertion (`nodeResourcesFitPlusArgs` / `scarceResourceAvoidanceArgs`). When a comma-ok assertion fails, that variable is just a typed nil of the type you were asserting to, so `%T` will always print the plugin's own expected type. The whole reason you reach for `%T` here is to tell the operator what they actually passed in, and this version can never do that.

The second one is simpler: the type name written into the message text is plain wrong. `noderesourcefitplus` expects `NodeResourcesFitPlusArgs` but the message says `NodeResourcesArgs`, and `scarceresourceavoidance` expects `ScarceResourceAvoidanceArgs` but the message says `ResourceTypesArgs`.

## What you expected to happen

The error should tell you the right expected type and show you what was actually handed in. Every other scheduler plugin already does this correctly by passing the original `args` (or `obj`) to `%T`. For example `loadaware` (`pkg/scheduler/plugins/loadaware/load_aware.go:85`):

```go
return nil, fmt.Errorf("want args to be of type LoadAwareSchedulingArgs, got %T", args)
```

`coscheduling`, `elasticquota`, `deviceshare`, `reservation`, `schedulinghint`, and `nodenumaresource` all follow that same shape. So the two broken plugins really just need to match it:

```go
return nil, fmt.Errorf("want args to be of type NodeResourcesFitPlusArgs, got %T", args)
```

```go
return nil, fmt.Errorf("want args to be of type ScarceResourceAvoidanceArgs, got %T", args)
```

## How to reproduce it

Call either plugin's `New()` with the wrong args type, say a `*config.LoadAwareSchedulingArgs`. You get back:

```
want args to be of type NodeResourcesArgs, got *config.NodeResourcesFitPlusArgs
```

The expected type is wrong (`NodeResourcesArgs`), and the `got` part shows the plugin's own type instead of the `*config.LoadAwareSchedulingArgs` you actually passed. In real life you hit this when a scheduler config profile wires the wrong `pluginConfig` args under one of these two plugins, which is exactly the moment a clear error would help the most.

## Anything else we need to know?

Both plugins landed together in #2302, and the bug has been there since. The reason nobody caught it is that neither plugin's test ever calls `New()` with a bad type:

* `pkg/scheduler/plugins/noderesourcefitplus/node_resources_fit_plus_test.go` only exercises the happy path.
* `pkg/scheduler/plugins/scarceresourceavoidance/scarce_resource_avoidance_test.go` only exercises the happy path.

`schedulinghint` already has a nice "wrong args type" test case to copy from at `pkg/scheduler/plugins/schedulinghint/scheduling_hint_test.go:303`, so adding coverage is straightforward.

This is a tiny fix: one line in each file, plus a wrong-args test for each plugin so it does not regress.

## Environment

* App version: main (commit 02575ef)
* Component: koord-scheduler

PRAteek-singHWY:

This cleans up the "wrong args type" error in two scheduler plugins, `noderesourcefitplus` and `scarceresourceavoidance`, so the message actually tells you something useful when the args don't match.

Both plugins had the same two slip-ups on a single line. First, the `%T` was being given the variable from the failed type assertion, which is just a typed nil of the expected type, so the error always printed the plugin's own type instead of whatever was really passed in. Second, the type name in the text was wrong: `NodeResourcesArgs` instead of `NodeResourcesFitPlusArgs`, and `ResourceTypesArgs` instead of `ScarceResourceAvoidanceArgs`.

The fix is small. I pass the original `args` to `%T` and correct the type name, which is exactly what the other plugins (`loadaware`, `coscheduling`, `elasticquota`, `deviceshare`, `reservation`, `schedulinghint`, `nodenumaresource`) already do.

Before:

```go
// noderesourcefitplus
return nil, fmt.Errorf("want args to be of type NodeResourcesArgs, got %T", nodeResourcesFitPlusArgs)
// scarceresourceavoidance
return nil, fmt.Errorf("want args to be of type ResourceTypesArgs, got %T", scarceResourceAvoidanceArgs)
```

After:

```go
// noderesourcefitplus
return nil, fmt.Errorf("want args to be of type NodeResourcesFitPlusArgs, got %T", args)
// scarceresourceavoidance
return nil, fmt.Errorf("want args to be of type ScarceResourceAvoidanceArgs, got %T", args)
```

I also added a "wrong args type" test case to each plugin so this stays fixed, following the same pattern `schedulinghint` already uses.

fixes #3082

Run the tests for the two plugins:

```bash
go test ./pkg/scheduler/plugins/noderesourcefitplus/... ./pkg/scheduler/plugins/scarceresourceavoidance/...
```

The new cases hand `New()` an args value of the wrong type (a `*config.LoadAwareSchedulingArgs`) and check that the error now names the right expected type and reports the type that was actually passed.

Nothing changes at runtime beyond the wording of one error string in each plugin. The rest is just the new test cases.

### V. Checklist


PRAteek-singHWY:
Hi @saintube @honpey @ZiMengSheng Please Have a look when u are here.
Happy to iterate.

ZiMengSheng:
### Review：这个 PR 比标题说的"类型名笔误"值钱

我原本把它归进"低风险杂项清理"批量过一遍，读完发现里面藏了两个 bug 而不是一个，PR 描述把它讲清楚了，我核实一遍确认无误。

**Bug 1（表面的）**：期望类型名写错了。`noderesourcefitplus` 说自己要 `NodeResourcesArgs`（这个类型根本不存在），`scarceresourceavoidance` 说自己要 `ResourceTypesArgs`（同样不存在）。

**Bug 2（真正的那个）**：`%T` 的参数传错了，导致 "got" 那一半永远是错的。

```go
nodeResourcesFitPlusArgs, ok := args.(*config.NodeResourcesFitPlusArgs)
if !ok {
    return nil, fmt.Errorf("want args to be of type NodeResourcesArgs, got %T", nodeResourcesFitPlusArgs)
    //                                                                          ^^^^ 断言失败时它是 typed nil
}
```

类型断言失败时 `ok == false`，而那个变量拿到的是目标类型的零值，也就是 `(*config.NodeResourcesFitPlusArgs)(nil)`——一个**带类型的 nil**。`%T` 打印的是静态类型而不是值，所以它永远输出 `*config.NodeResourcesFitPlusArgs`，也就是**期望类型**。结果这条错误信息变成"我要 A，我拿到了 B"，其中 B 恒等于真正的期望类型，A 是个不存在的名字——两半都在骗人，谁配错了 args 都定位不到。

改成传 `args` 之后才是真正传进来的那个对象的类型。

**我对照 main 确认了这是这两个插件独有的问题**，不是仓库风格：

```
coscheduling.go:82          got %T", obj      ✅
defaultprebind/plugin.go:82 got %T", args     ✅
deviceshare/plugin.go:828   got %T", obj      ✅
elasticquota/plugin.go:123  got %T", args     ✅
loadaware/load_aware.go:85  got %T", args     ✅
nodenumaresource/plugin.go  got %T", args     ✅
reservation/plugin.go:119   got %T", args     ✅
scheduleadmission           got %T", args     ✅
schedulinghint              got %T", args     ✅
noderesourcefitplus:52      got %T", nodeResourcesFitPlusArgs   ❌
scarceresourceavoidance:54  got %T", scarceResourceAvoidanceArgs ❌
```

11 个插件里就这 2 个写错，PR 描述里"matching the other scheduler plugins"这句是准确的。

**测试选得也对。** 两个 `TestNew_WrongArgsType` 都分别断言了错误信息的两半：

```go
assert.Contains(t, err.Error(), "want args to be of type NodeResourcesFitPlusArgs")  // 修 Bug 1
assert.Contains(t, err.Error(), "config.LoadAwareSchedulingArgs")                    // 修 Bug 2
```

第二条断言是关键——它传了一个**别的**插件的 args 类型进去，所以如果有人把 `%T` 的参数改回那个 typed nil，这条断言立刻红。用 `LoadAwareSchedulingArgs` 当"错误输入"也很聪明：它是仓库里真实存在的另一个 args 类型，比造一个假类型更能反映实际配错的场景。

顺便说一句，这类"错误信息本身是错的"的 bug 特别难被发现，因为它只在出错路径上暴露，而出错路径通常没有测试覆盖。你不但修了，还补了回归测试钉住，做法完全正确。

CI 全绿。

/lgtm

PR #3084, opened 2026-06-15
scheduler: persist Reservation.status.nominatedNodeName to fix preemption churn across restarts (fixes #3069) 

### LINKED ISSUE #3069: [BUG] Reservation does not support NominatedNodeName
**What happened**:

Reservation is a mechanism that Koordinator schedules as a pod not handled by the container runtime.

Currently, the koord-scheduler supports using a reservation to preempt a low-priority pod. However, the NNN (nominatedNodeName) attribute is not defined in the Reservation API, so the scheduler may lose the nomination of the preemption result once the object is updated. In addition, the upcoming feature like [KEP-5278](https://github.com/kubernetes/kubernetes/issues/132381) is missing for reservations without the NNN field.

Add the field NominatedNodeName into the Reservation API.

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

harshakumar25:
## Root Cause Analysis

`Reservation` is scheduled as a synthetic Pod by `koord-scheduler`. When it 
triggers preemption via `PostFilter`, the scheduler assigns a `nominatedNodeName` 
**exclusively in-memory** inside the nominator map 
(`pkg/scheduler/plugins/reservation/nominator.go`). 

This in-memory state has two critical failure points:

1. **Scheduler restart:** The nominator map is ephemeral; after any pod/leader restart the map is empty. The same Reservation re-enters the scheduling cycle, `PostFilter` runs again, and it tries to preempt pods that may have already been evicted. This is the **preemption churn loop** described in #3069.

2. **KEP-5278 incompatibility:** Upstream Kubernetes expects `nominatedNodeName` to be readable from the object's `.status` field so that other controllers and the scheduler itself can re-hydrate nominator state from the API server without requiring a live cache hit.

**Upstream comparison:** For regular `Pod` objects, `pod.status.nominatedNodeName` is set by `kube-apiserver` during `PostFilter`. Koordinator's `Reservation` type had no equivalent field, making the two semantically inconsistent.

---

## Code Audit Summary

| Layer | Finding |
|-------|---------|
| `reservation_types.go` L141–165 | `ReservationStatus` has no `NominatedNodeName` field |
| `reservation_handler.go` L306 | `addNominatedReservation()` writes in-memory only; `updateReservationStatus()` never sets the field |
| `eventhandler.go` (plugin) L46–56 | `OnAdd` populates cache for `IsReservationActive` but never restores nominator from persisted status |
| `reservation.go` L64 | `NewReservePod` converts the Reservation to a synthetic Pod but originally did not copy `status.nominatedNodeName` to the Pod status. |
| `reservation.go` L299, 338, 366 | Lifecycle helpers (`SetReservationExpired/Succeeded/Available`) don't clear stale nomination |
| `zz_generated.deepcopy.go` L955 | `*out = *in` — all scalar fields including new `string` are copied automatically; no deepcopy regen needed |

---

## Solution Design

All changes are guarded by a new **alpha feature gate** `ReservationNominatedNodeName` (default: `false`), ensuring zero runtime impact when disabled and enabling safe progressive rollout.

---

### `apis/scheduling/v1alpha1/reservation_types.go`
Added `NominatedNodeName string` to `ReservationStatus`:
```go
// NominatedNodeName is set when the Reservation preempts other pods on the node but cannot be
// scheduled right away (e.g. waiting for preempted pods to terminate). The scheduler uses this
// field to persist the preemption nomination across restarts and to avoid repeated preemption
// loops, maintaining compatibility with upstream KEP-5278 scheduler semantics.
// +optional
NominatedNodeName string `json:"nominatedNodeName,omitempty" protobuf:"bytes,7,opt,name=nominatedNodeName"`
```
- Protobuf tag `7` — next sequential after `Allocated` (tag `6`); no collision
- `omitempty` — field absent from JSON when empty; backward-compatible wire format
- `zz_generated.deepcopy.go` requires **no change** — `*out = *in` at line 955 copies all value types

### `config/crd/bases/scheduling.koordinator.sh_reservations.yaml`
CRD manifest updated via `make manifests generate` — `nominatedNodeName` added to the `status` schema with its doc comment and `type: string`.

### `pkg/features/scheduler_features.go`
```go
// owner: @harshakumar25  alpha: v1.9
ReservationNominatedNodeName featuregate.Feature = "ReservationNominatedNodeName"

// in defaultSchedulerFeatureGates:
ReservationNominatedNodeName: {Default: false, PreRelease: featuregate.Alpha},
```

### `pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go`
- `handleReservationSchedulingFailure`: reads `nominatingInfo.Mode()` to extract the correct `nominatedNodeName` (handles both `ModeOverride` and `ModeNoop` cases)
- `updateReservationStatus`: new `nominatedNodeName string` parameter; persists field to `curR.Status` inside the existing `RetryOnConflictOrTooManyRequests` retry loop

### `pkg/util/reservation/reservation.go`
- **Synthetic Pod Status Restoration**: Updated `NewReservePod` to copy the `NominatedNodeName` from the Reservation status to the synthetic Pod status when the feature gate is enabled:
  ```go
  if k8sfeature.DefaultFeatureGate.Enabled(features.ReservationNominatedNodeName) {
      reservePod.Status.NominatedNodeName = r.Status.NominatedNodeName
  }
  ```
  *Why this is critical*: The failure handler (`handleReservationSchedulingFailure`) extracts the nominated node from the synthetic Pod's status during `ModeNoop` (noop/unchanged nomination). Without this propagation, a subsequent scheduling cycle in `ModeNoop` would read an empty string from the synthetic pod and overwrite/clear the persisted nomination in the API server.
- **Terminal/Active Cleanup**: Feature-gate-guarded `r.Status.NominatedNodeName = ""` added to:
  - `SetReservationAvailable` — Reservation bound to a real node; nomination obsolete
  - `SetReservationExpired`   — TTL exceeded; nomination must not be carried forward
  - `SetReservationSucceeded` — Owner allocated; nomination obsolete

### `pkg/scheduler/plugins/reservation/eventhandler.go`
`OnAdd` and `OnUpdate` both now restore the in-memory nominator from `r.Status.NominatedNodeName` for pending Reservations (not `Active`, `Failed`, or `Succeeded`). Guard conditions prevent re-nomination of already-scheduled or terminal Reservations.

---

### Test run results

```
ok  github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1                  (cached)
ok  github.com/koordinator-sh/koordinator/pkg/features                              0.791s
ok  github.com/koordinator-sh/koordinator/pkg/util/reservation                     3.596s
ok  github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/eventhandlers 4.829s
ok  github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/reservation        22.381s
```
All packages build clean. Zero regressions across the full `./pkg/scheduler/...` test suite.

---

## Upgrade Safety

- Feature gate defaults to `false` — **zero behavioral change** on upgrade
- All new code paths are behind `k8sfeature.DefaultFeatureGate.Enabled(features.ReservationNominatedNodeName)`
- New API field uses `omitempty` — existing Reservation objects without the field remain valid
- Protobuf tag `7` does not conflict with any existing field
- `RetryOnConflictOrTooManyRequests` wrap already handles optimistic locking races for `UpdateStatus`

## How to Enable

```
--feature-gates=ReservationNominatedNodeName=true
```

## For checking Test cases use:
please use this command for tracking all test cases;
```bash
go test -v ./apis/scheduling/v1alpha1/... ./pkg/features/... ./pkg/scheduler/frameworkext/eventhandlers/... ./pkg/scheduler/plugins/reservation/... ./pkg/util/reservation/...
```

Fixes #3069


harshakumar25:
hi @saintube , @ZiMengSheng 
i have wrote everything in  description which are related to this pr , please followup that and let me know if i can help on some more feature and issues .
Thanks !

pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
```diff
@@ -305,16 +305,32 @@ func handleReservationSchedulingFailure(sched *scheduler.Scheduler,
 		//   from the informer cache.
 		addNominatedReservation(fwk, podInfo, nominatingInfo)
 
+		// Extract the nominated node name from the preemption result so it can be
+		// persisted to the Reservation status. This prevents state loss across scheduler
+		// restarts and avoids repeated preemption loops (KEP-5278 compatibility).
+		// ModeOverride: PostFilter explicitly nominated a new node.
+		// ModeNoop: the pod already has an existing nominatedNodeName in its status.
+		var nominatedNodeName string
+		if k8sfeature.DefaultFeatureGate.Enabled(features.ReservationNominatedNodeName) {
```
saintube: This appears to be mandatory, as I don't see a clear benefit from making this option configurable.

harshakumar25: sure @saintube ,
i have removed ReservationNominatedNodeName feature gate entirely ,
also added  nomination persistence logic unconditional across reservation_handler.go and eventhandler.go,
also resolved nil check bug,  i think now everyhting is perfect, please let  me if there is something need my input else please merge this one ,
thanks !

harshakumar25:
hi @saintube ,
pr update ,i have modified according to copilet suggestion and from your insights please check .
Thanks !

harshakumar25:
hi @saintube 😊,
i have completed review and resolve issue which was already raised , just quick check need from your side now 
Thanks !

pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
```diff
@@ -305,16 +305,34 @@ func handleReservationSchedulingFailure(sched *scheduler.Scheduler,
 		//   from the informer cache.
 		addNominatedReservation(fwk, podInfo, nominatingInfo)
 
+		// Extract the nominated node name to persist to Reservation.status.
+		// Default to the currently-persisted value so that non-preemption failures
+		// (nominatingInfo == nil) do not accidentally clear an existing nomination.
+		// ModeOverride: PostFilter explicitly set a new nominated node.
+		// ModeNoop:     the pod already carries the existing nominatedNodeName in its status.
+		nominatedNodeName := cachedR.Status.NominatedNodeName
+if nominatingInfo != nil {
```
saintube: incorrect indent

pkg/scheduler/plugins/reservation/eventhandler.go
```diff
@@ -80,6 +96,26 @@ func (h *reservationEventHandler) OnUpdate(oldObj, newObj interface{}) {
...
 		h.rrNominator.DeleteReservePod(reservationutil.NewReservePod(newR))
 	}
+	// On status updates, re-sync the in-memory nominator from the persisted NominatedNodeName.
+	// This handles the case where the failure handler has written the nomination to the API
+	// server and the informer update arrives (including after a scheduler restart).
+	// We only act on pending Reservations that are not yet fully scheduled or terminated.
+	newNom := newR.Status.NominatedNodeName
+	if oldR.Status.NominatedNodeName != "" && oldR.Status.NominatedNodeName != newNom {
+		// nomination was cleared or changed; remove the old in-memory record first
+		h.rrNominator.DeleteReservePod(reservationutil.NewReservePod(oldR))
```
saintube, line 107: This update operation doesn't appear to be atomic.
I suggest aligning the implementation with how pod NNN is handled.

harshakumar25: @saintube ,
I've updated OnUpdate to mirror the upstream ` UpdateNominatedPod ` pattern, ensuring the old nomination is atomically deleted before the new one is added under a single mutex

pkg/scheduler/frameworkext/eventhandlers/reservation_nomination_leak_test.go
```diff
@@ -0,0 +1,158 @@
+/*
```
saintube: There is no need for a separate file for this unit test. Please merge it into the existing test file.

harshakumar25: updated , 
i have merged testcase in exsiting codebase , copilot suggestion are false positive so i ignore them .
please check it now .

ZiMengSheng:
### Pre-review: the direction is right, and I want this to land as the base of the cluster

This is the most complete of the four PRs currently touching reservation nomination, and the design holds up:

- The three `nominatingInfo` modes are handled correctly — `nil` keeps the existing value (so a non-preemption failure doesn't silently wipe a nomination), `ModeOverride` takes the new node, `ModeNoop` reads back from `pod.Status.NominatedNodeName`. That defaulting choice is the subtle one and you got it right.
- Clearing is wired at all three exits: unschedulable, fully allocated, and successfully scheduled. A persisted nomination that never clears would be worse than no persistence at all, so this matters.
- `NewReservePod` propagating `r.Status.NominatedNodeName` onto the reserve pod is what actually makes the persisted value visible to the scheduler.
- protobuf tag `7` is correct — `ReservationStatus` currently ends at `6` (`Allocated`). Field is `+optional`, so the CRD change is backward compatible.

Aligning with upstream KEP-5278 semantics here is the right call: `Pod.Status.NominatedNodeName` is persisted upstream, and a Reservation scheduled as a synthetic pod should behave the same way. #3069 is a real bug.

### Blocking (mechanical)

- **DCO failing** — needs `git commit --amend -s` / rebase with sign-off
- **`unit-tests (Verify gofmt)` failing** — run `make fmt`

### One thing I checked so you don't have to worry about it

I initially suspected this would add API write amplification, since `handleReservationSchedulingFailure` fires on every scheduling failure. It doesn't: `SetReservationUnschedulable` on `main` already refreshes `LastProbeTime: metav1.Now()` unconditionally, so every failure already produces a real `UpdateStatus` write. **You're adding a field to a write that was already happening**, not adding a write.

Worth noting as debt though: `status` now also carries nomination state, so nomination churn rides along on that existing write path. That's relevant to #2853 (Non-blocking API Calls) but not a reason to hold this PR.

### Coordination — please read before pushing the fixup

Three other PRs touch this area, and the ordering matters:

| PR | Scope | Relationship |
|---|---|---|
| **#3084** (this) | API + CRD + persistence | **Base.** Owns the `Status.NominatedNodeName` field |
| #3114 (yours) | Case 5 handover cleanup | **Currently cannot compile** — asserts on `Status.NominatedNodeName` without carrying the API change. I've asked you there to decouple it |
| #3156 (@thc1006) | `nominator.go` rewrite (+279), phantom-hold cleanup | Overlaps this PR in `reservation_handler.go` and `plugins/reservation/eventhandler.go`. Will need rebase after this lands |
| #2885 (@adity1raut) | `pod_eventhandler.go` + `preemption.go` | Independent — no file overlap. Can proceed in parallel |

Fix DCO and gofmt and I'll do a full line-level review. Ordering recorded on #3069.

ZiMengSheng:
Cross-reviewed this together with #3114, #3156 and #2885, because all four end up manipulating the same `reservation` nominator. Posting the shared analysis here since this PR is the one that has to reconcile with the others.

## The cluster

Four open PRs, four different trigger points, one shared structure:

| PR | trigger | effect on the nominator |
|---|---|---|
| #2885 | pod's `NominatedNodeName` cleared; right after a successful reservation preemption | removes |
| #3114 | reservation `schedulerName` handover (`oldResponsible && !newResponsible`) | removes |
| #3156 | reservation UID replaced / entering deletion / spec generation or labels/annotations changed | removes |
| **#3084 (this one)** | scheduler restart, informer resync, `status.NominatedNodeName` change | **adds** |

So three PRs remove stale nominations at three different points, and this one is the only writer that puts them back from persisted state. None of them is a duplicate of another — I checked the diffs, not just the titles.

Mechanically they compose: I merged each pair in both orders against current `main` and **every production file merges cleanly** in all six orderings. The only conflicts are in `reservation_handler_test.go` and `eventhandler_test.go`, where you and thc1006 append cases at the same spot — trivial to resolve.

## But this PR and #3156 contradict each other semantically

Your re-sync block sits *after* the `if / else if` chain in `OnUpdate` and is unconditional. #3156 adds two new `else if` branches to that same chain. Merged, the resulting order is:

```go
if IsReservationActive(newR)            { ... delete ... }
else if Failed || Succeeded             { ... delete ... }
else if !newR.DeletionTimestamp.IsZero(){ DeleteReservePod(newR) }          // #3156
else if Generation/Labels/Annotations changed { DeleteReservePodIfStale(newR) } // #3156
// then, unconditionally, from this PR:
if newNom != "" && !Active && !Failed && !Succeeded {
    AddNominatedReservePod(podInfo, newNom)
}
```

Two concrete cases where the combination is wrong:

1. **Reservation entering deletion.** `deletionTimestamp` set, phase still Pending, `status.NominatedNodeName` non-empty (a finalizer keeps the object around). #3156 deletes the nomination; your block re-adds it on the very same event. The node stays locked by an object that will never be scheduled.
2. **Spec change / `schedulerName` handover.** A spec edit bumps `Generation`, so #3156 calls `DeleteReservePodIfStale`; your block then re-adds from `status.NominatedNodeName`, which is exactly the value that just became stale.

Your gate is `!IsReservationActive && !IsReservationFailed && !IsReservationSucceeded`. That admits both cases above, because a terminating-but-pending reservation is none of the three, and a generation bump doesn't change the phase either.

This isn't an argument for dropping either PR — persisting the nomination across restarts (#3069) is a real gap, and so is dropping invalidated nominations. But whichever of the two lands second has to make the re-sync condition subordinate to the invalidation logic, not run after it. Concretely, the re-sync needs to at minimum also require `newR.DeletionTimestamp.IsZero()`, and needs to not fire when the persisted `NominatedNodeName` predates the current `Generation` — which is information the current `Reservation.status` doesn't carry.

That last point is worth thinking about before this API field is frozen: `status.NominatedNodeName` alone can't be validated against the spec it was computed for. `status.observedGeneration`, or stamping the generation alongside the nomination, would make the restore path checkable instead of trusting.

## Smaller points on this PR

- The mode switch you add in `handleReservationSchedulingFailure` (`ModeOverride` → `nominatingInfo.NominatedNodeName`, `ModeNoop` → `pod.Status.NominatedNodeName`) duplicates, line for line, what `addNominatedReservation` does three lines above. After both land there are two copies of the same rule that have to be kept in sync. Please extract it, or have `addNominatedReservation` return the node it used.
  Note this is not cosmetic: #3156 is currently *changing* which pod is correct on the `ModeNoop` path. You happen to pass `pod`, which matches what #3156 argues is right, but a second copy of the rule is exactly how that fix gets half-applied.
- `updateReservationStatus` now writes `curR.Status.NominatedNodeName = nominatedNodeName` unconditionally, including the empty string. Combined with the caller defaulting to `cachedR.Status.NominatedNodeName`, a non-preemption failure preserves the old value — good — but a `ModeNoop` where `pod.Status.NominatedNodeName` happens to be empty will now clear a previously persisted nomination. Worth an explicit test.
- Your merge-base is 60 commits behind `main`. Please rebase; the tree merges cleanly today but the test files will collide with #3114 / #3156.

@harshakumar25 @thc1006 — you two own the two halves of this. Could you agree on which lands first and who folds in the reconciliation? I'd suggest #3156's invalidation paths first (they're pure removals, so they can't create a bad state on their own), then this PR layered on top with the gating above. Happy to review in whichever order you settle on.

saintube:
@harshakumar25 any update for the remaining issues :)

PR #3090, opened 2026-06-18
fix(koordlet): exit runtimehooks reconcile goroutines on stopCh

namansh70747:

The host application reconcile loop (`host_app_reconciler.go`) and the kubeqos cgroup reconcile loop (`reconciler.go`) log on `stopCh` but never `return`, so the `for`/`select` keeps running after shutdown is signalled. Once `stopCh` is closed the closed channel is always ready, so the select fires the `<-stopCh` case on every iteration — the goroutine busy-spins instead of exiting.

This adds the missing `return`, matching the sibling loops in the same package (`doHostAppCgroup` and `reconcilePodCgroup`) that already return on `stopCh`. These two were the only `case <-stopCh:` in `pkg/` that didn't return.

NONE

```
go test ./pkg/koordlet/runtimehooks/reconciler/ -run exitsOnStop
```

The added tests close `stopCh` and assert the loop returns within a deadline. They time out on the old behaviour and pass once the loops return.

Minimal change: one `return` per loop plus the tests.

### V. Checklist

### DIFF
--- a/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler.go
+++ b/pkg/koordlet/runtimehooks/reconciler/host_app_reconciler.go
@@ -98,6 +98,7 @@ func (r *hostReconciler) reconcile(stopCh <-chan struct{}) {
 			timer.Reset(r.reconcileInterval)
 		case <-stopCh:
 			klog.V(1).Infof("stop reconcile for host application")
+			return
 		}
 	}
 }
--- a/pkg/koordlet/runtimehooks/reconciler/reconcile_stop_test.go
+++ b/pkg/koordlet/runtimehooks/reconciler/reconcile_stop_test.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2025 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


ZiMengSheng:
Verified against `main`. Both defects are real, and I appreciate that you fixed **both** loops rather than only the one reported in #3018.

`reconciler.go:401-403`:
```go
case <-stopCh:
    klog.V(1).Infof("stop reconcile kube qos cgroup")
```
No `return`, so the `for`/`select` keeps spinning after shutdown. `host_app_reconciler.go` has the identical shape. Both are launched as goroutines from `reconciler.Run` (lines 365-366), so both leak — and since koordlet is a DaemonSet, leaked reconcile loops keep touching cgroups after the component is supposed to be done.

Adding `reconcile_stop_test.go` to assert the exit path is the part that keeps this from silently regressing; a fix like this is easy to undo during a later refactor without a test pinning it.

**One thing to fix:** `license-check` is failing — the new test file needs the Apache license header. Copy the block from any neighbouring file in that package. Everything else is green.

Push that and I'll lgtm.

For context: I had labelled #3018 as a `good first issue` earlier today, not realising your PR was already open on it. I've removed those labels and noted that this PR supersedes it — apologies for the confusion.

PR #3092, opened 2026-06-22
descheduler: add ranker and unit tests for scaledownbinpack

Vatsalpatni73:

This PR adds the initial `Scale-Down Binpack` ranker implementation under `pkg/descheduler/framework/plugins/scaledownbinpack/`.

It introduces:

- `RankPods(...)` to compute a per-node evacuation score.
- Node grouping for target pods and non-target pods.
- Node-level sorting based on evacuation score, target pod count, and node name for deterministic ordering.
- Pod-level sorting within each node by weighted size, then creation timestamp, then namespace/name.
- A global rank assignment for all target pods, which can later be used for pod-deletion-cost annotation or direct eviction flow.

This logic is designed to improve scale-down packing by preferring victims whose removal helps consolidate remaining workload onto fewer nodes.

Refers #2790

Run the unit tests for the new ranker package:
```bash
go test ./pkg/descheduler/framework/plugins/scaledownbinpack/...
```

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
+++ b/pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
@@ -78,9 +78,9 @@ func RankPods(nodes []*corev1.Node, eligibleTargetPods []*corev1.Pod, skippedTar
 
 		var num, den float64
 		for resName, weight := range resourceWeights {
-			capacity := node.Status.Capacity[resName]
-			capMilli := capacity.MilliValue()
-			if capMilli <= 0 {
+			allocatable := node.Status.Allocatable[resName]
+			allocMilli := allocatable.MilliValue()
+			if allocMilli <= 0 {
 				continue
 			}
 
@@ -98,8 +98,8 @@ func RankPods(nodes []*corev1.Node, eligibleTargetPods []*corev1.Pod, skippedTar
 				ntReq += q.MilliValue()
 			}
 
-			uNt := float64(ntReq) / float64(capMilli)
-			uT := float64(tReq) / float64(capMilli)
+			uNt := float64(ntReq) / float64(allocMilli)
+			uT := float64(tReq) / float64(allocMilli)
 
 			num += weight * uNt
 			den += weight * uT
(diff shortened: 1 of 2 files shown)


pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return globalRanked
+}
+
+func getPodSize(pod *corev1.Pod, weights map[corev1.ResourceName]float64) float64 {
+	var size float64
+	for resName, weight := range weights {
+		q := util.GetPodRequest(pod, resName)[resName]
+		size += weight * float64(q.MilliValue())
```
songtao98: `MilliValue()` values from different resources are not comparable: 1 CPU is 1000, while 1 GiB memory is about 1e12 in milli-units. With the default 1:1 weights, memory completely dominates this ordering. Please normalize requests, e.g. against node allocatable. A test with CPU-heavy versus memory-heavy pods would help catch this.

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,184 @@
...
+			// Skip nodes with zero eligible target pods
+			continue
+		}
+
+		skippedPods := skippedByNode[node.Name]
+		ntPods := nonTargetByNode[node.Name]
+
+		var num, den float64
+		for resName, weight := range resourceWeights {
+			capacity := node.Status.Capacity[resName]
```
songtao98: Consider using Allocatable but not Capacity. For scale-down binpacking, the closer scheduler reference is NodeResourcesFit/MostAllocated: each resource is normalized as requested / allocatable before weights are combined. Therefore, this ranker would be better to use node.Status.Allocatable, not node.Status.Capacity. See https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/most_allocated.go#L30-L45

songtao98 (changes requested):

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Thanks for the review!
I've updated the logic to normalize the requested resources against `Allocatable` so memory no longer dominates the score. I also added a test case (`TestRankPods_ResourceDomination`) for the CPU vs. memory-heavy scenario to verify it works as expected.
Could you please check if this looks good?

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Thanks for the review!
I've swapped out `Capacity` for `Allocatable` across the board.
Could you please check if this looks good?

songtao98 (approved):

PR #3093, opened 2026-06-22
descheduler: add FragmentationAware balance plugin

Vatsalpatni73:

This PR introduces the **FragmentationAware** descheduler plugin to mitigate intra-node resource fragmentation caused by resource-type imbalance (e.g., high CPU utilization but low memory utilization on the same node). Such imbalance prevents additional workloads from being scheduled even when significant resources remain available, reducing overall cluster utilization.

The plugin identifies nodes whose resource utilization imbalance exceeds a configurable **ImbalanceThreshold**, where imbalance is measured using the standard deviation of per-resource utilization. For each eligible node, it evaluates candidate pod evictions by calculating the theoretical standard deviation after removing each pod and selects the pod that provides the greatest reduction in fragmentation.

Key features include:

* Detects resource imbalance using configurable thresholds.
* Selects the optimal pod eviction candidate based on maximum standard deviation improvement.
* Supports a configurable **MinImprovementThreshold** to avoid low-impact evictions.
* Integrates with the existing descheduler Balance framework.
* Honors `DryRun` and `Paused` states.
* Performs `NodeFit` validation to ensure evicted pods have alternative scheduling destinations.
* Applies standard `podFilter` and `PreEvictionFilter` safeguards before eviction.

Refers #2332

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/fragmentationaware/fragmentation_aware.go
+++ b/pkg/descheduler/framework/plugins/fragmentationaware/fragmentation_aware.go
@@ -1,0 +1,233 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fragmentationaware
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/descheduler/framework/plugins/fragmentationaware/fragmentation_aware.go
```diff
@@ -0,0 +1,233 @@
...
+
+var _ framework.BalancePlugin = &FragmentationAware{}
+
+type FragmentationAware struct {
+	handle    framework.Handle
+	podFilter framework.FilterFunc
+	args      *deschedulerconfig.FragmentationAwareArgs
+}
+
+func NewFragmentationAware(ctx context.Context, args runtime.Object, handle framework.Handle) (framework.Plugin, error) {
```
songtao98, line 49: A new plugin needs to be registered in NewInTreeRegistry(), which is the registry used by the real descheduler startup path.

songtao98 (changes requested):

pkg/descheduler/framework/plugins/fragmentationaware/fragmentation_aware.go
Vatsalpatni73: Thanks for pointing that out. I've registered the new plugin in `NewInTreeRegistry()` and pushed the change.

songtao98 (approved):

Issue #2749, opened 2025-12-22
[proposal] Support shared states between multiple profiles in the single scheduler


saintube:
**What is your proposal**:

Currently, most scheduler plugins maintain private caches for each framework or profile. As a result, when multiple profiles are configured in the KubeSchedulerConfiguration, each profile registration leads to duplicated plugin caches. However, the SchedulerCache and PodNominator provided by the scheduling framework are singletons within a single scheduler instance.
In this context, we explore how to unify plugin-level caches and the reservation nominator into a shared state across all profiles within the same scheduler instance.

**Why is this needed**:

**Is there a suggested so? If so, please add it**:

nikunjkumar05:
Hi @ZiMengSheng  @saintube I'm interested in this project I'm planning to apply it through LFX 
Am I right this comes under LFX program ?
Thank you currently exploring codebase and picking up some good first issue to get familiar with codebase

ishanvaibhav:
Hi @saintube and @ZiMengSheng,

I'm very interested in this proposal for the LFX mentorship program. To answer your question @nikunjkumar05 — yes, this is part of the LFX program, and it's a fantastic opportunity to work on scheduler architecture.

The problem statement resonates deeply with me. Having each profile maintain its own duplicated plugin caches is inefficient, especially in multi-tenant clusters where operators configure separate profiles for different QoS tiers (e.g., one for LSR pods, another for BE pods). When both profiles register plugins like NodeReservation, ReservationScore, or LoadAwareScheduling, we end up with redundant in-memory caches that consume significant memory and, more critically, can lead to stale or inconsistent scheduling decisions when a reservation is created under one profile but needs to be visible to another.

Here is my initial thinking on the approach:

1. Shared Cache Layer
Instead of each profile instantiating its own plugin cache, introduce a scheduler-scoped shared cache injected via the Handle or a custom context. Plugins would register once against this shared state rather than per-profile. The SchedulerCache and PodNominator being singletons already sets the precedent — we just need to extend this pattern downward to plugin-level caches.

2. Profile-Aware but Cache-Shared
The key challenge is that different profiles may still need profile-scoped views of the same underlying data. For example, Profile A might only care about reservations with scheduleTimeout annotations, while Profile B cares about all reservations. The shared cache should store the raw data once, with profiles applying lightweight filters or projections at query time rather than maintaining separate copies.

3. ReservationNominator Unification
[...]

PR #3095, opened 2026-06-23
docs: add proposal for shared plugin cache and unified event dispatcher

mdryaan:

Adds a design proposal for the **Shared Plugin Cache and Unified Event Dispatcher** mechanism for Koordinator scheduler profiles.

Koordinator's out-of-tree scheduler plugins each maintain their own cache. When multiple scheduling profiles are configured, each plugin's `New()` is called once per profile, creating N independent cache instances that all track the same cluster-wide state, and registering N independent informer event handler sets.

This proposal defines a `SharedPluginCache` interface that allows plugins to opt into a single shared cache instance across all profiles, with lifecycle management and event dispatch centralized in `FrameworkExtenderFactory`.

**Problems addressed:**

- **P1 — Per-profile cache duplication:** DeviceShare's `nodeDeviceCache`, NodeNUMAResource's `resourceManager`/`topologyOptionsManager`, and Reservation's `reservationCache` are each instantiated N times (N = number of profiles), all tracking the same cluster-wide state.
- **P2 — Event handler timing inconsistency:** Independent informer handlers fire asynchronously relative to the in-tree scheduler cache, causing brief windows where the two caches disagree — leading to double-counted device resources, incorrect quota accounting, and preemption failures.

**Design summary:**

- `SharedPluginCache` interface with `Start(ctx)` + `OnPod{Add,Update,Delete}` + `OnNode{Add,Update,Delete}` methods.
- `FrameworkExtenderFactory.GetOrRegisterSharedCache(key, createFn)` — guarantees exactly-once initialization across all profiles.
- `FrameworkExtenderFactory.StartSharedCaches(ctx)` — called once after all profiles are built; starts GC goroutines and registers plugin-specific CRD informer handlers.
- Unified dispatcher routes pod/node events to all registered caches in a defined order.
- Optional `Snapshottable` extension for per-cycle immutable cache snapshots (P3 bonus, DeviceShare prototype).
- Migration order: DeviceShare → NodeNUMAResource → Reservation. EQ/LoadAware deferred.

Part of the LFX Mentorship project: *Plugin-level Cache and Nominator Shared Across Scheduler Profiles* (Term 2 2026).

Part of #2749

Review the proposal document at `docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md`. No code changes in this PR.

- The exact method signatures in the `SharedPluginCache` interface are intentionally marked as "to be refined during PR review" — please suggest adjustments.
- The interface is limited to pod and node events because those are the common denominator across all three target plugins. Plugin-specific CRD events (e.g., `Device`) are registered inside `Start()`.
- `GetOrRegisterSharedCache` is exposed through `ExtendedHandle` so plugins call it from `New()` without per-plugin wiring in `server.go`.
- ElasticQuota (`GroupQuotaManager`) and LoadAware are explicitly out of scope for this phase.

### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
+++ b/docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
@@ -1,0 +1,270 @@
+---
+title: Shared Plugin Cache and Unified Event Dispatcher for Scheduler Profiles
+authors:
+  - "@mdryaan"
+reviewers:
+  - "@saintube"
+  - "@ZiMengSheng"
+creation-date: 2026-06-23
+last-updated: 2026-06-23
+status: provisional
+---
+
+<!-- TOC -->
+
+- [Shared Plugin Cache and Unified Event Dispatcher for Scheduler Profiles](#shared-plugin-cache-and-unified-event-dispatcher-for-scheduler-profiles)
+    - [Summary](#summary)
+    - [Motivation](#motivation)
+        - [Goals](#goals)
+        - [Non-Goals/Future Work](#non-goalsfuture-work)
+    - [Proposal](#proposal)
+        - [User Stories](#user-stories)
+            - [Story 1 — Per-profile cache duplication wastes resources and causes divergent views](#story-1--per-profile-cache-duplication-wastes-resources-and-causes-divergent-views)
+            - [Story 2 — Independent event handlers cause timing races between in-tree and out-of-tree caches](#story-2--independent-event-handlers-cause-timing-races-between-in-tree-and-out-of-tree-caches)
+            - [Story 3 — No snapshot causes lock contention and inconsistent reads under high throughput](#story-3--no-snapshot-causes-lock-contention-and-inconsistent-reads-under-high-throughput)
+        - [Design Details](#design-details)
+            - [SharedPluginCache interface](#sharedplugincache-interface)
+            - [Registration and lifecycle](#registration-and-lifecycle)
+            - [Event dispatch](#event-dispatch)
+            - [Plugin migration plan](#plugin-migration-plan)
+            - [Snapshot (Phase 3, bonus)](#snapshot-phase-3-bonus)
+    - [Alternatives](#alternatives)
+    - [Implementation History](#implementation-history)
+
+<!-- /TOC -->
+
+# Shared Plugin Cache and Unified Event Dispatcher for Scheduler Profiles
+
+## Summary
+
+Koordinator's out-of-tree scheduler plugins each maintain their own cache of cluster state. When
+multiple scheduling profiles are configured, each plugin's `New()` is called once per profile,
+creating N independent cache instances that all track the same cluster-wide state. Additionally,
+these per-plugin caches register their own informer event handlers independently, which fires
+asynchronously from in-tree handlers and from each other, causing brief inconsistencies between
+what the in-tree scheduler cache and out-of-tree plugin caches believe about the cluster.
+
+This proposal introduces a `SharedPluginCache` interface that allows plugins to opt into a
+single shared cache instance across all scheduler profiles, with lifecycle management and event
+dispatch centralized in the framework extender.
+
+## Motivation
+
+The kube-scheduler supports multiple scheduling profiles, each with its own plugin set. Koordinator
+builds on this with `FrameworkExtender` and `PluginFactoryProxy`, which call each plugin's `New()`
+once per profile. This creates three concrete problems:
+
+**P1 — Per-profile cache duplication.** With N profiles, DeviceShare's `nodeDeviceCache`,
+NodeNUMAResource's `resourceManager`/`topologyOptionsManager`, and Reservation's `reservationCache`
+are each instantiated N times, all tracking the same cluster-wide state. Since a single scheduler
+instance serializes scheduling cycles across profiles, this duplication serves no correctness
+purpose — it wastes memory, CPU, and spawns redundant GC goroutines and event handler sets.
+
+**P2 — Event handler timing inconsistency.** Each plugin registers its own informer handlers in
+`New()`, independent from the in-tree scheduler cache handlers and from other plugins. When a pod
+is deleted, the in-tree cache may already reflect the deletion while a Koordinator plugin is still
+processing an update for the same pod. This race causes double-counted device resources, incorrect
+quota accounting, and preemption failures.
+
+**P3 — No snapshot (bonus).** In-tree plugins read a per-cycle immutable `NodeInfo` snapshot.
+Out-of-tree plugins read live maps protected by `sync.RWMutex`. Under high scheduling throughput
+(~200 pods/s), this causes reader lock contention and inconsistent reads within a single scheduling
+cycle.
+
+### Goals
+
+1. Define a `SharedPluginCache` interface that plugins can opt into, giving the framework
+   structured control over cache lifecycle and event dispatch.
+2. Ensure each opted-in plugin cache is created exactly once per scheduler instance and shared
+   across all profiles.
+3. Route pod and node events to all registered caches through a centralized dispatcher, ensuring
+   consistent ordering relative to the in-tree scheduler cache.
+4. Migrate DeviceShare, NodeNUMAResource, and Reservation caches onto the new mechanism.
+5. No upstream Kubernetes changes — all work stays within `koordinator-sh/koordinator`.
+6. Opt-in only — plugins that do not implement `SharedPluginCache` continue to work unchanged.
+7. Each phase is independently mergeable.
+
+### Non-Goals/Future Work
+
+- Migrating ElasticQuota (`GroupQuotaManager`) and LoadAware in this phase; they are a follow-up
+  after the mechanism is stable.
+- Fixing issue #1959 (nominated-pod Nominator) — that is a separate effort that will build on
+  top of the shared cache foundation.
+- Modifying kube-scheduler's internal event handler ordering or any upstream code.
+- Full snapshot implementation for all plugins (P3 targets DeviceShare as a prototype only).
+
+## Proposal
+
+### User Stories
+
+#### Story 1 — Per-profile cache duplication wastes resources and causes divergent views
+
+A cluster operator runs koord-scheduler with three scheduling profiles (e.g., one for
+latency-sensitive workloads, one for batch, one for GPU jobs). DeviceShare's `nodeDeviceCache` is
+instantiated three times, each listening on the same Device and Pod informers, each running its own
+GC goroutine. A `Reserve` call on profile A records a GPU allocation in profile A's cache. Until
+the pod informer event arrives (typically milliseconds later), profile B and C caches do not see
+the allocation. In a burst-scheduling scenario this assume-window gap allows a pod scheduled by
+profile B to be placed on a node that profile A has already fully allocated.
+
+#### Story 2 — Independent event handlers cause timing races between in-tree and out-of-tree caches
+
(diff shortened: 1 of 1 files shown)


docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,270 @@
...
+    // Called exactly once per scheduler instance, before the first scheduling cycle.
+    // Plugin-specific informer handlers for non-pod/node resources (e.g., Device CRDs) are
+    // registered here, since the framework only dispatches common pod/node events centrally.
+    Start(ctx context.Context)
+
+    // The following methods are called by the framework's unified event dispatcher when
+    // pod/node events arrive from the shared informer. They replace per-plugin informer handler
+    // registration for pod and node events, ensuring consistent ordering.
+
+    OnPodAdd(pod *corev1.Pod)
```
saintube, line 152: It seems to miss the handling for Reserve/Unreserve.

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,270 @@
...
+
+As a bonus phase after the shared cache migration is complete, we will add snapshot support.
+A `Snapshottable` optional interface extends `SharedPluginCache`:
+
+```go
+type Snapshottable interface {
+    SharedPluginCache
+    // Snapshot returns an immutable point-in-time copy of the cache for use during a
+    // single scheduling cycle. The returned snapshot is safe to read concurrently without locks.
+    Snapshot() CacheSnapshot
```
saintube, line 597: Supporting snapshots for out-of-tree plugins presents a key challenge that lies not in defining the snapshot method itself, but in designing a comprehensive data structure to govern snapshot derivation and replace the existing cache (e.g. how to support the Reserve and the AddPod/RemovePod).

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,270 @@
...
+plugin-supplied string. A plugin registers its cache from within `New()`:
+
+```go
+// In FrameworkExtenderFactory — registration point available to all plugins via ExtendedHandle.
+// If a cache for the given key already exists (registered by an earlier profile), the existing
+// instance is returned and create() is not called. This guarantees exactly-once initialization.
+func (f *FrameworkExtenderFactory) GetOrRegisterSharedCache(key string, create func() SharedPluginCache) SharedPluginCache
+```
+
+After all profiles have been initialized (all `New()` calls complete), the factory calls `Start()`
```
saintube, line 186: What is the initialization order between the plugin cache, the plugin (and its controller), and the shared informer factory?

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,270 @@
...
+#### Registration and lifecycle
+
+`FrameworkExtenderFactory` maintains a registry of `SharedPluginCache` instances keyed by a
+plugin-supplied string. A plugin registers its cache from within `New()`:
+
+```go
+// In FrameworkExtenderFactory — registration point available to all plugins via ExtendedHandle.
+// If a cache for the given key already exists (registered by an earlier profile), the existing
+// instance is returned and create() is not called. This guarantees exactly-once initialization.
+func (f *FrameworkExtenderFactory) GetOrRegisterSharedCache(key string, create func() SharedPluginCache) SharedPluginCache
```
saintube: The parameters required by the shared cache, such as extendedHandle and informerFactory, should be explicitly defined.

mdryaaan:
Thanks for the review @saintube. PTAL
- Added the initialization order section, explicit handle/factory params for GetOrRegisterSharedCache, and expanded the snapshot section to cover the layered state challenge you mentioned.
- For Reserve/Unreserve  kept them out of the interface since they need plugin-specific CycleState, but added a section explaining the direct-call model. Let me know if you meant something different.

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
mdryaaan: Addressed in [4c02ba8](https://github.com/koordinator-sh/koordinator/pull/3095/commits/4c02ba894f20e80fd37a383fe6930df4189c93a1)

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
mdryaaan: Addressed in [4c02ba8](https://github.com/koordinator-sh/koordinator/pull/3095/commits/4c02ba894f20e80fd37a383fe6930df4189c93a1)

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
mdryaaan: Addressed in [4c02ba8](https://github.com/koordinator-sh/koordinator/pull/3095/commits/4c02ba894f20e80fd37a383fe6930df4189c93a1)

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
saintube: Check the ExtendedHandle API to see what it has exposed (e.g., SharedInformerFactory).

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -214,6 +247,28 @@ all Koordinator plugin caches see pod/node events in a defined order. Coordinati
...
+to apply or roll back assumed allocations. The `SharedPluginCache` interface does not include
+these methods, and the framework does not dispatch them centrally, for two reasons:
+
+- Reserve and Unreserve require plugin-specific state from `framework.CycleState` that differs
+  across plugins and cannot be generalized into a shared interface.
+- Concurrent Reserve calls from different scheduling cycles are already serialized by the
+  per-node lock present in each cache (e.g., `nodeDevice.lock` in DeviceShare). No additional
+  synchronization layer is needed at the interface level.
+
+A correctness benefit of the shared model: because Reserve writes to the single shared cache
```
saintube: If we decide not to add the Reserve/Unreserve methods, we should establish a proper pattern enabling these phases to cooperate with the new shared cache. Note that while our goal is to unify the plugin cache, certain plugins still modify their cache during the Reserve phase, which has a chance to break the consistency without limitations.

mdryaaan:
@saintube, Updated both simplified the signature to create func(handle ExtendedHandle) since the handle already exposes all the informer factories.

for Reserve/Unreserve added an optional CacheReserver interface with AssumePod/ForgetPod. The double-count risk you pointed out is real: DeviceShare's OnPodAdd reads allocations from pod annotations and writes to the cache, so a pod that was already Reserved would get applied twice. ForgetPod before applying event state is the fix. DeviceShare and Reservation need to implement it and NodeNUMAResource needs analysis during migration.

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,378 @@
...
+    // informer event supersedes it.
+    AssumePod(pod *corev1.Pod, nodeName string) error
+    // ForgetPod removes the assumed entry for pod. Called on Unreserve to roll back
+    // assumed allocation state. Must be idempotent — safe to call even if AssumePod
+    // was never called for this pod.
+    ForgetPod(pod *corev1.Pod) error
+}
+```
+
+**Invariants that `CacheReserver` implementations must uphold:**
```
saintube, line 288: Could you please provide an example plugin refactored to the new interfaces? This would help us demonstrate and validate whether the approach make sense.

mdryaaan:
Hi @saintube PTAL,
Added a DeviceShare example to the proposal - nodeDeviceCache implementing both SharedPluginCache and CacheReserver, with Start, all six event handlers, AssumePod/ForgetPod, and the updated Reserve/Unreserve call sites.

Main thing writing it confirmed: the anti-double-count logic is small and self-contained  - a types.UID -> nodeName map, one helper (consumeAssumedIfMatch), and one guard in OnPodAdd/OnPodUpdate. Existing updatePod/deletePod paths stay untouched, the per-node nodeDevice.lock model is preserved, no new global write lock introduced.

Also added a note that Reservation follows the same pattern and NodeNUMAResource will be analyzed during PR-2 to determine if it needs CacheReserver at all. Happy to sketch Reservation here too if that would help, or we can leave it for the next migration PR.

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
```diff
@@ -0,0 +1,572 @@
...
+    n.lock.Lock()
+    defer n.lock.Unlock()
+    delete(n.assumedPods, pod.UID)
+    return nil
+}
+
+// consumeAssumedIfMatch atomically clears the assumed marker for pod and returns true
+// if the pod was in the assumed set. Callers use the result to skip re-applying an
+// allocation that Reserve has already applied to the cache.
+func (n *nodeDeviceCache) consumeAssumedIfMatch(pod *corev1.Pod) bool {
```
saintube: What if the event object differs from the assumed result?

saintube:
PTAL /cc @ZiMengSheng for the DeviceShare adaptation example.

docs/proposals/scheduling/20260623-shared-plugin-cache-and-unified-event-dispatcher.md
mdryaaan: Good catch, "skip if assumed" is only correct when the event agrees with what Reserve wrote, and that's not guaranteed. If the pod ends up bound to a different node than Reserve assumed, or an admission webhook rewrites its device annotations after Reserve, skipping the event leaves a phantom allocation on the assumed node and never records the real one. Deletion before binding has the same shape.

mdryaaan:
@saintube Updated the proposal: AssumePod now snapshots (nodeName, allocations) instead of just the UID, and OnPodAdd/OnPodUpdate/OnPodDelete reconcile - roll the snapshot back, apply the event's authoritative state. Net-zero in the common case where they agree, converges on the event when they don't. This mirrors what schedulerCache.addPod does with assumed pods upstream.

saintube:
Overall LGTM. The details around the framework startup sequencing and plugin registration can be refined during the implementation PRs. For a refactor of this scope, we should also make sure that comprehensive regression tests are in place.

saintube:
/lgtm
PTAL /cc @ZiMengSheng

mdryaaan:
> Overall LGTM. The details around the framework startup sequencing and plugin registration can be refined during the implementation PRs. For a refactor of this scope, we should also make sure that comprehensive regression tests are in place.

Thanks for the review. I'II make sure regression tests are thorough in the implementation PRs.

saintube:
@mdryaan While we continue refining the remaining details, I'd suggest starting the framework code in parallel.

mdryaaan:
@saintube Ya, i'm already working on it, will open PR very soon.

PR #3097, opened 2026-06-23
test/perf: add scheduler benchmark skeleton 

Shivam-1827:

This PR implements the `koord-scheduler` end-to-end performance benchmark tool at `test/perf/`, as discussed and approved in PR #3066 with @ZiMengSheng and @saintube.

This PR delivers the core foundation (Phase 1): the project structure, interface contracts, kwok node provider, scenario registry, and a complete benchmark engine that provisions simulated nodes, creates pods through a bounded worker pool, watches `PodScheduled` events, computes scheduling latency percentiles (P50/P90/P99), calculates throughput, tracks `FailedScheduling` events, enforces a configurable run timeout, and writes structured benchmark reports.

**What is added:**

| Path | What it is |
|---|---|
| `pkg/types/` | Shared structs (`ScenarioConfig`, `BenchmarkResult`, `NodeSpec`, `PodLatency`, `Thresholds`) — leaf package imported by all components without creating dependency cycles; includes `ScenarioConfig.Validate()` for early configuration validation and `TimeoutDuration()` with `DefaultRunTimeout` fallback |
| `pkg/framework/engine.go` | Full benchmark orchestration engine — node provisioning → scenario setup → pod burst → scheduling watcher → metrics → report generation; bounded by `context.WithTimeout`; writes partial report with `timedOut: true` on deadline |
| `pkg/framework/watcher.go` | Watch-based `PodScheduled` observer that records per-pod scheduling latency from `CreationTimestamp` until scheduling completion; uses `ready chan struct{}` handshake so the burst cannot start before the stream is established |
| `pkg/framework/failure_watcher.go` | Watches `FailedScheduling` events for the run's pod set; tracks distinct pods affected and total event count; scoped by pod name set (Events don't carry pod labels so a label selector cannot be used) |
| `pkg/framework/metrics.go` | Metric helpers implementing `ComputeLatencyPercentiles()` (P50/P90/P99) and `ComputeThroughput()` |
| `pkg/framework/report.go` | JSON report generation and terminal summary output |
| `pkg/nodeprovider/provider.go` | `NodeProvider` interface — `CreateNodes`, `DeleteNodes`, `WaitReady` |
| `pkg/nodeprovider/kwok/` | kwok implementation using parallel node creation (`errgroup` + semaphore with cancellation-safe select) and runtime YAML node template loading via `NodeTemplateFile` |
| `pkg/scenarios/scenario.go` | `Scenario` interface — `Name`, `Setup`, `Pods`, `Teardown` |
| `pkg/scenarios/registry.go` | Factory-based scenario registry returning a fresh scenario instance for every benchmark run |
| `pkg/scenarios/basic/` | Baseline scenario generating plain pod bursts with kwok toleration and node selector |
| `configs/scenarios/basic-1k.yaml` | Declarative benchmark configuration for 1,000 pods, 100 kwok nodes, configurable parallelism, `koord-scheduler`, and a `timeout: 10m` hard ceiling |
| `cmd/benchmark/main.go` | CLI entry point supporting `--config`, `--output`, `--kubeconfig`, and `--baseline`; validates configuration before contacting the cluster |
| `hack/setup-cluster.sh` | kind + kwok + Koordinator bootstrap |
| `hack/teardown-cluster.sh` | kind cluster teardown |
| `Makefile` | `make benchmark`, `make setup`, `make teardown`, `make test`; supports `CONFIG=` for any config path and `SCENARIO=` shorthand |
| `docs/README.md` | Full reference documentation — quickstart, scenario config table, CLI reference, output field reference, troubleshooting, and new-scenario guide |
| `pkg/framework/metrics_test.go` | Unit tests for `ComputeLatencyPercentiles` and `ComputeThroughput` |
| `pkg/framework/report_test.go` | Unit tests for `CompareToBaseline` covering breach and no-breach cases |
| `baselines/basic-1k.json` | First committed baseline captured from a kind + kwok run (45.57 pods/sec, P99=13s) |

**How the benchmark works:**

```text
CreateNodes(runID, nodeSpec, count)         ← parallel, errgroup + semaphore
WaitReady(runID, 60s timeout)
scenario.Setup(...)                          ← namespace creation, CRDs, etc.
└─ defer scenario.Teardown(context.Background())
└─ defer provider.DeleteNodes(context.Background())

runCtx, cancel = context.WithTimeout(ctx, cfg.TimeoutDuration())   ← hard ceiling

watcher.Start(runCtx) ────── goroutine ─────────────────────────────────────┐
failureWatcher.Start(failureCtx) ── goroutine ──────────────────────────┐   │
<-watcher.Ready(); <-failureWatcher.Ready()   ← wait before burst       │   │
burstStart = time.Now()                                                  │   │
errgroup + semaphore(cfg.Concurrency) → Create all cfg.PodCount pods    │   │
g.Wait() → apiCreationDuration                                           │   │
<-watcherErrCh → totalDuration           ◄──────────────────────────────┼───┘
failureCancel() → <-failureWatcherErrCh  ◄──────────────────────────────┘

ComputeLatencyPercentiles(watcher.Latencies()) → P50, P90, P99
failureWatcher.Stats() → schedulingFailureCount, schedulingFailureRate
CompareToBaseline(...) → thresholdBreached
WriteReport(result, outputPath)
```

`Teardown()` and `DeleteNodes()` are deferred with `context.Background()` so cleanup continues even if the benchmark context is cancelled or times out. On `DeadlineExceeded`, a partial report with `timedOut: true` is written so whatever numbers were collected before the deadline are not lost.

**Verified output from running `make benchmark` against a live kind + kwok cluster:**

```
I0708 00:17:33 "Starting benchmark" scenario="basic" runID="32d28728-aed5-44af-bcb8-867beb91a33d" timeout="10m0s"
I0708 00:17:33 "API server reachable" scenario="basic"
I0708 00:17:33 "Creating nodes" count=100 workers=20
I0708 00:17:33 "Nodes ready"
I0708 00:17:33 "Starting pod burst" pods=1000 concurrency=50
I0708 00:17:42 "API creation phase complete" duration="8.89s"
I0708 00:17:42 "Waiting for all pods to be scheduled"
I0708 00:17:55 "Benchmark result" scenario="basic" throughputPodsPerSec="45.57"
               apiCreationDurationSec="8.89s" totalDurationSec="21.94s"
               p50LatencySec="6.00s" p90LatencySec="12.00s" p99LatencySec="13.00s"
               output="results/result.json"
```

```json
{
  "name": "basic",
  "koordinatorVersion": "e4169896-dirty",
  "throughputPodsPerSec": 45.57,
  "totalDurationSec": 21.94,
  "latencyP50Sec": 6,
  "latencyP90Sec": 12,
  "latencyP99Sec": 13,
  "thresholdBreached": false,
  "timedOut": false,
  "schedulingFailureCount": 0,
  "schedulingFailureRate": 0
}
```

The reported numbers are from a development environment using the default-scheduler on a single-node kind cluster. They are intended only to validate correctness of the benchmark implementation and are not representative performance numbers for `koord-scheduler`.

---

NONE — this is a new feature addition, not a bug fix.

---

**Option A — No cluster needed (verify it builds successfully):**

```bash
# Build all benchmark packages
go build ./test/perf/...

# Build the CLI binary
go build -o /tmp/kwok-bench ./test/perf/cmd/benchmark/

# Run unit tests
make -C test/perf test
```

```bash
# Create kind cluster + install kwok + install Koordinator
make -C test/perf setup

# Run benchmark
make -C test/perf benchmark

# Tear down when done
make -C test/perf teardown
```

---

### Ⅳ. Special notes for reviewers

**The two interface contracts are the most important things to review.** Everything in the remaining implementation phases builds directly on top of them. If the shapes need to change, now is the cheapest time to do it.

**1. `NodeProvider` — `pkg/nodeprovider/provider.go`**

Three methods: `CreateNodes`, `DeleteNodes`, `WaitReady`.

The engine only depends on this interface and never references any kwok-specific implementation details. Future providers (for example, Kubemark) can be added under `pkg/nodeprovider/` by implementing the same interface.

**2. `Scenario` — `pkg/scenarios/scenario.go`**

Four methods: `Name`, `Setup`, `Pods`, `Teardown`.

The engine invokes them in order for every benchmark run. The `basic` scenario implements a minimal namespace setup, while future scenarios (`gang`, `elasticquota`, etc.) will provision additional Koordinator resources before starting the pod burst.

**3. `pkg/types` — why it exists**

All shared structs live in `pkg/types` rather than `pkg/framework` to keep the package dependency graph acyclic.

Both `pkg/nodeprovider` and `pkg/scenarios` require `ScenarioConfig` and `NodeSpec`, while `pkg/framework` imports both packages.

Keeping shared types in `pkg/types` avoids the dependency cycles:

- `framework → nodeprovider → framework`
- `framework → scenarios → framework`

and keeps the architecture modular.

**4. Design decisions confirmed by maintainers**

- `test/perf/` stays in the root module — no separate `go.mod`
- Baselines committed to the repo; path is configurable via `--baseline`
- Benchmark cluster setup remains self-contained in `hack/setup-cluster.sh`
- `KoordinatorVersion` injected via `ldflags` from `git describe` at build time
- Forward-looking fields (`GangCompletionP50/P99`, pprof artifacts) kept as placeholders
- `FailureWatcher` scoped by pod name set rather than label selector — Events don't carry pod labels, making name-based filtering the correct approach
- Timeout defaults to `10m`; configurable per scenario via the `timeout:` YAML field

**What is intentionally not in this PR:**

- `gang`, `elasticquota`, `reservation`, `loadaware` scenarios → Phase 2
- GitHub Actions CI workflow → Phase 2

---

### Ⅴ. Checklist

  - `docs/README.md` covers quickstart, scenario config reference table, CLI reference, output field reference, troubleshooting, and new-scenario guide
  - All exported types and interfaces have Go doc comments
  - `ScenarioConfig.Validate()` validates benchmark configuration before cluster interaction
  - `metrics_test.go`: `ComputeLatencyPercentiles` (empty, single, sorted, unsorted), `ComputeThroughput`
  - `report_test.go`: `CompareToBaseline` (no baseline, within thresholds, throughput breach, P99 breach, missing file)
  - `go build ./test/perf/...` passes cleanly with zero errors
  - `make test` (`go test ./...`) passes — unit tests cover metrics, report, and baseline comparison


Shivam-1827:
@saintube and @ZiMengSheng 
PTAL

Shivam-1827:
@saintube ,
PTAL, I have made changes as per above comments

Shivam-1827:
@saintube 
PTAL, made changes as per the review

test/perf/pkg/nodeprovider/kwok/node_template.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/koordinator-sh/koordinator/test/perf/pkg/types"
+)
+
+const RunIDLabel = "benchmark.koordinator.sh/run-id"
+
+// buildKwokNode constructs a fake Node object that the kwok controller simulates as Ready.
+// Uses stable kubernetes.io/os and kubernetes.io/arch labels instead of the
+// deprecated beta.kubernetes.io/* equivalents.
+func buildKwokNode(name, runID string, spec types.NodeSpec) (*corev1.Node, error) {
```
saintube, line 36: Why not use the kwok api (kwokctl CLI/kwokconfiguration CRD) for node creation?

Shivam-1827: Thanks for the question. The direct Nodes().Create() approach is intentional for programmatic benchmark use.

kwokctl is a cluster-management CLI (analogous to kind) — it manages the kwok controller lifecycle and is the right tool for setup-cluster.sh, where we already use it implicitly via kwok.yaml. It does not expose a Go API suitable for dynamically provisioning nodes during a benchmark run.

KwokConfiguration / Stage CRDs control how the kwok controller simulates node behaviour (heartbeat, conditions, readiness transitions) — we already apply stage-fast.yaml in setup-cluster.sh, which is what makes a plain Node object appear Ready immediately. That is the kwok API layer, and we are using it.

Creating Node objects directly via client.CoreV1().Nodes().Create() is the documented pattern for programmatic kwok usage and is what upstream kubernetes/kubernetes scheduler-perf tests do. The kwok controller watches for those objects and handles the rest via the stage configuration.

Shivam-1827:
@saintube 

PTAL

test/perf/pkg/nodeprovider/kwok/node_template.go
saintube: OK.

test/perf/pkg/nodeprovider/kwok/kwok.go
```diff
@@ -0,0 +1,122 @@
...
+func (p *Provider) CreateNodes(ctx context.Context, runID string, spec types.NodeSpec, count int) error {
+	if p.client == nil {
+		fmt.Printf("[kwok stub] Would create %d nodes for runID=%s\n", count, runID)
+		return nil
+	}
+	runIDPrefix := runID
+	if len(runIDPrefix) > 8 {
+		runIDPrefix = runIDPrefix[:8]
+	}
+	for i := 0; i < count; i++ {
```
saintube, line 65: Would it be feasible to build the nodes in parallel?

test/perf/configs/scenarios/basic-1k.yaml
```diff
@@ -0,0 +1,18 @@
...
+clientBurst: 200
+qosClass: LS
+resourceRequests:
+  cpu: "500m"
+  memory: "512Mi"
+labels:
+  koordinator.sh/priority: "100"
+thresholds:
+  throughputDropPct: 10
+  p99IncreasePct: 20
```
saintube: A trailing newline for new files is recommended to comply with POSIX standards.

test/perf/pkg/nodeprovider/kwok/node_template.go
saintube: We need to explore approaches that allow node template modifications without rebuilding binaries or images.

test/perf/configs/scenarios/basic-1k.yaml
Shivam-1827: Fixed in `basic-1k.yaml`. Also found and fixed the same issue in `hack/teardown-cluster.sh` while reviewing nearby files.

test/perf/pkg/nodeprovider/kwok/kwok.go
Shivam-1827: Yes, implemented. Used `errgroup.WithContext` with a semaphore channel to bound concurrency. `name` is declared inside the loop body so closure capture is safe. All errors are collected and the first one returned via `g.Wait()`.

The worker count is not hardcoded — it is configurable via `nodeCreationWorkers` in the scenario YAML (defaulting to 20 when unset) so users can tune it for their cluster size and API server capacity.

Shivam-1827:
`We need to explore approaches that allow node template modifications without rebuilding binaries or images. `

-> Added a `nodeTemplateFile` field to both `ScenarioConfig` and `NodeSpec`. When set in the scenario YAML, the node provider loads a complete `corev1.Node` object from that file at runtime instead of using the built-in defaults. The required kwok labels, annotation, and taint are always overlaid on top of whatever the file specifies, so the file only needs to describe the parts that differ between scenarios.

This means changing node capacity, adding topology labels, simulating GPU taints, or any other node shape change is a YAML edit with no binary rebuild involved.

The field is optional — existing scenario configs that omit `nodeTemplateFile` continue to work exactly as before.

test/perf/pkg/nodeprovider/kwok/node_template.go
Shivam-1827: Here are some approaches:

 Added a `nodeTemplateFile` field to both `ScenarioConfig` and `NodeSpec`. When set in the scenario YAML, the node provider loads a complete `corev1.Node` object from that file at runtime instead of using the built-in defaults. The required kwok labels, annotation, and taint are always overlaid on top of whatever the file specifies, so the file only needs to describe the parts that differ between scenarios.

This means changing node capacity, adding topology labels, simulating GPU taints, or any other node shape change is a YAML edit with no binary rebuild involved.

The field is optional — existing scenario configs that omit `nodeTemplateFile` continue to work exactly as before.

ZiMengSheng:
```
All throughput and latency values in the output JSON are 0 — intentional
for the skeleton. Real numbers come once the engine loop is implemented.
```

Look forward to your implementation of the engine loop!

test/perf/pkg/nodeprovider/kwok/node_template.go
saintube: LGTM

saintube (commented):
Given that `ScenarioConfig` is a flat struct with many parameters, we should add a validation step to catch invalid values.

Shivam-1827:
> Given that `ScenarioConfig` is a flat struct with many parameters, we should add a validation step to catch invalid values.

Added a Validate() error method on ScenarioConfig in pkg/types/types.go. It checks every field that can meaningfully be invalid:

name non-empty
nodeCount, podCount, concurrency, clientQPS, clientBurst all > 0
nodeCreationWorkers >= 0 (zero means "use default", so it's allowed)
thresholds.throughputDropPct in [0, 100]
thresholds.p99IncreasePct >= 0
Validate() is called in main.go immediately after YAML unmarshalling, so bad configs are rejected with a clear error before any cluster connection or benchmark work begins.

Shivam-1827:
> ```
> All throughput and latency values in the output JSON are 0 — intentional
> for the skeleton. Real numbers come once the engine loop is implemented.
> ```
> 
> Look forward to your implementation of the engine loop!

sure

Shivam-1827:
Hi @saintube and @ZiMengSheng, I have a few design questions before moving forward
with the next phase. Would appreciate your guidance on these:

**1. Should `test/perf/` have its own `go.mod`?**
The benchmark currently lives inside the main Koordinator module, so dependencies
like `github.com/google/uuid` and `golang.org/x/sync` land in the root `go.mod`.
As the tool grows (pprof parsers, reporting libraries, etc.), should we isolate it
under `test/perf/go.mod` to avoid polluting the main module graph, or is staying
in the root module preferred for simplicity?

**2. Where should baseline result files live for regression detection?**
`breachesThresholds()` is currently a stub. Implementing real regression detection
requires a baseline `result.json` to compare against. Should baselines be:
- Committed to the repo (e.g. `test/perf/baselines/basic-1k.json`)?
- Stored in a CI artifact store and downloaded at run time?
- Updated manually after each intentional performance change?

**3. Should the benchmark reuse `test/e2e` cluster helpers or keep its own `hack/setup-cluster.sh`?**
We currently have a standalone `hack/setup-cluster.sh` that creates a kind cluster
and installs kwok. The `test/e2e` directory already has cluster setup helpers.
Should we share that infrastructure, or is keeping the benchmark cluster setup
isolated preferred since it requires a specific kwok + koord-scheduler configuration?

**4. How should `KoordinatorVersion` be populated in `BenchmarkResult`?**
The output JSON has a `koordinatorVersion` field that is always empty today. Should it be:
- Injected via `ldflags` at build time from a `VERSION` file?
- Passed as a `--version` CLI flag by the caller?
- Queried from the running koord-scheduler pod's version endpoint at runtime?

**5. Should `GangCompletion` and pprof fields stay in `BenchmarkResult` or be removed until implemented?**
`BenchmarkResult` has four fields that are always `nil`/empty right now:
`gangCompletionP50Sec`, `gangCompletionP99Sec`, `pprofCPUArtifact`, `pprofHeapArtifact`.
These need gang scheduling watcher logic and pprof scraping support respectively.
Should we remove them from the struct until those features are implemented, or keep
them as forward-compatible placeholders?

**6. Should the kwok provider and the engine share a single k8s client?**
Currently `main.go` builds a k8s client and passes it to `kwokprovider.New()`,
while `framework.NewEngine()` independently builds a second client from the same
kubeconfig. Both clients are needed (the provider creates nodes, the engine creates
pods and namespaces), but two clients sharing the same QPS budget effectively halves
the available throughput. Should the engine expose its client and inject it into the
provider, or is having two separate clients acceptable?

Shivam-1827:
@saintube and @ZiMengSheng 
PTAL

test/perf/pkg/framework/engine.go
```diff
@@ -0,0 +1,250 @@
...
+// the API server even if the run's ctx has been cancelled or timed out.
+func (e *Engine) Run(ctx context.Context, cfg types.ScenarioConfig, outputPath string) error {
+	scenario, ok := scenarios.Get(cfg.Name)
+	if !ok {
+		return fmt.Errorf("scenario %q not registered; available: %v",
+			cfg.Name, scenarios.List())
+	}
+
+	runID := uuid.New().String()
+	fmt.Printf("Starting benchmark: scenario=%s runID=%s\n", cfg.Name, runID)
```
saintube: use `klog/v2` for structured logging?

saintube:
> Hi @saintube and @ZiMengSheng, I have a few design questions before moving forward with the next phase. Would appreciate your guidance on these:
> 
> **1. Should `test/perf/` have its own `go.mod`?** The benchmark currently lives inside the main Koordinator module, so dependencies like `github.com/google/uuid` and `golang.org/x/sync` land in the root `go.mod`. As the tool grows (pprof parsers, reporting libraries, etc.), should we isolate it under `test/perf/go.mod` to avoid polluting the main module graph, or is staying in the root module preferred for simplicity?
> 
> **2. Where should baseline result files live for regression detection?** `breachesThresholds()` is currently a stub. Implementing real regression detection requires a baseline `result.json` to compare against. Should baselines be:
> 
> * Committed to the repo (e.g. `test/perf/baselines/basic-1k.json`)?
> * Stored in a CI artifact store and downloaded at run time?
> * Updated manually after each intentional performance change?
> 
> **3. Should the benchmark reuse `test/e2e` cluster helpers or keep its own `hack/setup-cluster.sh`?** We currently have a standalone `hack/setup-cluster.sh` that creates a kind cluster and installs kwok. The `test/e2e` directory already has cluster setup helpers. Should we share that infrastructure, or is keeping the benchmark cluster setup isolated preferred since it requires a specific kwok + koord-scheduler configuration?
> 
> **4. How should `KoordinatorVersion` be populated in `BenchmarkResult`?** The output JSON has a `koordinatorVersion` field that is always empty today. Should it be:
> 
> * Injected via `ldflags` at build time from a `VERSION` file?
> * Passed as a `--version` CLI flag by the caller?
> * Queried from the running koord-scheduler pod's version endpoint at runtime?
> 
> **5. Should `GangCompletion` and pprof fields stay in `BenchmarkResult` or be removed until implemented?** `BenchmarkResult` has four fields that are always `nil`/empty right now: `gangCompletionP50Sec`, `gangCompletionP99Sec`, `pprofCPUArtifact`, `pprofHeapArtifact`. These need gang scheduling watcher logic and pprof scraping support respectively. Should we remove them from the struct until those features are implemented, or keep them as forward-compatible placeholders?
> 
> **6. Should the kwok provider and the engine share a single k8s client?** Currently `main.go` builds a k8s client and passes it to `kwokprovider.New()`, while `framework.NewEngine()` independently builds a second client from the same kubeconfig. Both clients are needed (the provider creates nodes, the engine creates pods and namespaces), but two clients sharing the same QPS budget effectively halves the available throughput. Should the engine expose its client and inject it into the provider, or is having two separate clients acceptable?

@Shivam-1827 
1. I prefer not. Maintaining a separate module adds overhead that outweighs the isolation benefit right now.
2. Committed to the repo. This keeps baselines versioned, reviewable, and diffable alongside code changes. And we should make the baseline path configurable.
3. Keeping an isolated one seems to be more maintainable.
4. Let's use the git tag or short commit SHA.
5. Keeping them is fine.
6. I prefer sharing k8s client.

Shivam-1827:
> > Hi @saintube and @ZiMengSheng, I have a few design questions before moving forward with the next phase. Would appreciate your guidance on these:
> > **1. Should `test/perf/` have its own `go.mod`?** The benchmark currently lives inside the main Koordinator module, so dependencies like `github.com/google/uuid` and `golang.org/x/sync` land in the root `go.mod`. As the tool grows (pprof parsers, reporting libraries, etc.), should we isolate it under `test/perf/go.mod` to avoid polluting the main module graph, or is staying in the root module preferred for simplicity?
> > **2. Where should baseline result files live for regression detection?** `breachesThresholds()` is currently a stub. Implementing real regression detection requires a baseline `result.json` to compare against. Should baselines be:
> > 
> > * Committed to the repo (e.g. `test/perf/baselines/basic-1k.json`)?
> > * Stored in a CI artifact store and downloaded at run time?
> > * Updated manually after each intentional performance change?
> > 
> > **3. Should the benchmark reuse `test/e2e` cluster helpers or keep its own `hack/setup-cluster.sh`?** We currently have a standalone `hack/setup-cluster.sh` that creates a kind cluster and installs kwok. The `test/e2e` directory already has cluster setup helpers. Should we share that infrastructure, or is keeping the benchmark cluster setup isolated preferred since it requires a specific kwok + koord-scheduler configuration?
> > **4. How should `KoordinatorVersion` be populated in `BenchmarkResult`?** The output JSON has a `koordinatorVersion` field that is always empty today. Should it be:
> > 
> > * Injected via `ldflags` at build time from a `VERSION` file?
> > * Passed as a `--version` CLI flag by the caller?
> > * Queried from the running koord-scheduler pod's version endpoint at runtime?
> > 
> > **5. Should `GangCompletion` and pprof fields stay in `BenchmarkResult` or be removed until implemented?** `BenchmarkResult` has four fields that are always `nil`/empty right now: `gangCompletionP50Sec`, `gangCompletionP99Sec`, `pprofCPUArtifact`, `pprofHeapArtifact`. These need gang scheduling watcher logic and pprof scraping support respectively. Should we remove them from the struct until those features are implemented, or keep them as forward-compatible placeholders?
> > **6. Should the kwok provider and the engine share a single k8s client?** Currently `main.go` builds a k8s client and passes it to `kwokprovider.New()`, while `framework.NewEngine()` independently builds a second client from the same kubeconfig. Both clients are needed (the provider creates nodes, the engine creates pods and namespaces), but two clients sharing the same QPS budget effectively halves the available throughput. Should the engine expose its client and inject it into the provider, or is having two separate clients acceptable?
> 
> @Shivam-1827
> 
> 1. I prefer not. Maintaining a separate module adds overhead that outweighs the isolation benefit right now.
> 2. Committed to the repo. This keeps baselines versioned, reviewable, and diffable alongside code changes. And we should make the baseline path configurable.
> 3. Keeping an isolated one seems to be more maintainable.
> 4. Let's use the git tag or short commit SHA.
> 5. Keeping them is fine.
> 6. I prefer sharing k8s client.

Hi @saintube  and @ZiMengSheng, thank you for the quick responses! Here is a
summary of what has been implemented based on your guidance:

**1. Separate go.mod — No change needed.**
Understood, staying in the root module. No changes made.

**2. Baselines committed to repo with configurable path **
- Added `test/perf/baselines/basic-1k.json` — the first baseline captured
  from a successful run against a kind + kwok cluster (45.57 pods/sec, P99=13s).
- Implemented `CompareToBaseline()` in `pkg/framework/report.go` that loads
  the baseline JSON, computes throughput drop % and P99 increase %, and sets
  `ThresholdBreached: true` if either exceeds the thresholds defined in the
  scenario YAML.
- Added `--baseline` flag to the CLI so the path is fully configurable:
  `make benchmark BASELINE=baselines/basic-1k.json`

**3. Isolated `hack/setup-cluster.sh` — No change needed.**
Understood, keeping the standalone setup script.

**4. Git tag or short commit SHA **
- Added `var Version = "dev"` in `pkg/framework/engine.go`, injected at
  build time via `-ldflags`.
- Updated the `Makefile` to derive `GIT_VERSION` from
  `git describe --tags --always --dirty` and pass it through:
  `go run -ldflags="-X .../framework.Version=$(GIT_VERSION)"`
- Every `result.json` now carries the exact commit SHA, e.g.:
  `"koordinatorVersion": "e4169896-dirty"`

**5. Keep gang/pprof fields as placeholders — No change needed.**
Fields remain in `BenchmarkResult` for Phase 2.

**6. Share one k8s client **
- `Engine` now exposes `Client()` and `SetProvider()`.
- `main.go` builds the engine first, then constructs the kwok provider
  using `engine.Client()` — one connection pool per run, not two.

Also added unit tests for `ComputeLatencyPercentiles` and `CompareToBaseline`
as part of this week's work (`pkg/framework/metrics_test.go` and
`pkg/framework/report_test.go`).

The full benchmark runs end-to-end and produces real numbers:

Throughput:  45.57 pods/sec
P50:         6s
P90:         12s
P99:         13s
koordinatorVersion: e4169896-dirty


Please let me know if any of the implementations need adjustment before
I proceed to the next phase.

test/perf/pkg/framework/engine.go
Shivam-1827: sure

saintube (commented):
/lgtm
As a heads-up for the next iteration, the full implementation still needs a benchmark-level timeout mechanism and a collection of FailedScheduling events for scheduling failure rate tracking.

Shivam-1827:
> /lgtm As a heads-up for the next iteration, the full implementation still needs a benchmark-level timeout mechanism and a collection of FailedScheduling events for scheduling failure rate tracking.

@saintube 
Thanks for the heads-up! Both of these are already implemented in the latest commit :

Benchmark-level timeout (`test/perf/pkg/types/types.go`, `engine.go`):
- Added `Timeout string` field to `ScenarioConfig` (e.g. `timeout: 10m` in YAML)
- `TimeoutDuration() `method parses it, falling back to a `DefaultRunTimeout` of 10 minutes
- `Run()` now wraps the entire scenario in `context.WithTimeout`, so a stuck scheduler cannot hang the job indefinitely
- On `DeadlineExceeded`, a `timeoutAwareReport()` helper writes a partial JSON result with `timedOut: true` so whatever numbers were collected before the deadline are preserved

FailedScheduling event collection (`test/perf/pkg/framework/failure_watcher.go`):
- `FailureWatcher` watches the Events API with field selector `reason=FailedScheduling,involvedObject.kind=Pod`
- Scoped by pod name set rather than a label selector (Events don't carry pod labels, so name-based filtering is the correct approach)
- Runs under a child context that is cancelled as soon as the main watcher returns, so it doesn't outlive the scheduling phase
- `Stats()` returns distinct-pod count and total-event count, both written to `BenchmarkResult` as `schedulingFailureCount` and `schedulingFailureRate`

Both watchers use the `ready chan struct{} `handshake so the pod burst cannot start before both watch streams are established.

Shivam-1827:
@saintube PTAL

Shivam-1827:
@saintube PTAL

saintube (approved):
/lgtm
PTAL /cc @ZiMengSheng

Shivam-1827:
Hi @saintube and @ZiMengSheng , 
Thanks for the review! Looking at the Phase 1 plan, it looks like the only piece we haven't explicitly closed out is — README, Makefile, and general polish on the basic scenario, with what topic would like to work next?

Shivam-1827:
Hi @saintube and @ZiMengSheng,

Before I start working on the `README` and `Makefile`, I have a couple of clarifications:

1. Would you like me to further refine the `basic` scenario? For example, by making the pod templates more realistic, adding configurable resource profiles, or incorporating any other improvements you have in mind.
2. Would you prefer a formal Phase 1 sign-off or checklist before I proceed to Phase 2?

Please let me know your preference. Thanks!

ZiMengSheng (approved):
/lgtm
/approve

ZiMengSheng:
> Hi @saintube and @ZiMengSheng,
> 
> Before I start working on the `README` and `Makefile`, I have a couple of clarifications:
> 
> 1. Would you like me to further refine the `basic` scenario? For example, by making the pod templates more realistic, adding configurable resource profiles, or incorporating any other improvements you have in mind.
> 2. Would you prefer a formal Phase 1 sign-off or checklist before I proceed to Phase 2?
> 
> Please let me know your preference. Thanks!

@Shivam-1827  Nothing, can we trigger ci now once this PR merged.

Shivam-1827:
> > Hi @saintube and @ZiMengSheng,
> > Before I start working on the `README` and `Makefile`, I have a couple of clarifications:
> > 
> > 1. Would you like me to further refine the `basic` scenario? For example, by making the pod templates more realistic, adding configurable resource profiles, or incorporating any other improvements you have in mind.
> > 2. Would you prefer a formal Phase 1 sign-off or checklist before I proceed to Phase 2?
> > 
> > Please let me know your preference. Thanks!
> 
> @Shivam-1827 Nothing, can we trigger ci now once this PR merged.

@ZiMengSheng 
sure

PR #3099, opened 2026-06-25
Scaledownbinpack (descheduler): add args, defaults, and validation

Vatsalpatni73:
- Adds the `ScaleDownBinPackStrategy` enum (`CalculateOnly` and `EvictDirectly`) and `ScaleDownBinPackPodSelector` types.
- Introduces `ScaleDownBinPackArgs` to both the internal and versioned (`v1alpha2`) plugin configurations.
- Wires up the default configuration behavior (e.g. `Paused` defaults to `false`, `Strategy` defaults to `CalculateOnly`, missing resources default to `[cpu, memory]`, and missing weights default to `1.0`).
- Adds rigorous configuration validation logic and unit tests to ensure that the strategies are valid, `MaxPodsToEvict` limits are respected, and resource weights are positive and finite.
- Includes auto-generated updates for deepcopy, defaulters, and type conversions.

Part of #2790 

- Review the newly added unit tests in `pkg/descheduler/apis/config/validation/validation_scaledownbinpack_test.go`.
- Run `make test` to ensure that all validation test cases correctly pass.
- Verify that `ScaleDownBinPackArgs` is properly recognized when unmarshaling config files in a local test cluster.

- The `zz_generated.*` files (deepcopy, defaults, and conversions) were auto-regenerated successfully after adding these types.
- When the `Strategy` is set to `EvictDirectly`, the validation strictly enforces that `MaxPodsToEvict` is a positive integer.

### V. Checklist


pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-;
```
songtao98: The license header here seems not right. Please check and fix it. Should be http://www.apache.org/licenses/LICENSE-2.0

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,182 @@
...
+				EvacuationScore: info.EvacuationScore,
+			})
+			globalRank++
+		}
+	}
+
+	return globalRanked
+}
+
+func getPodResourceRequest(pod *corev1.Pod, resourceName corev1.ResourceName) int64 {
```
songtao98: This function does not consider InitContainers / Sidecars or Pod overhead. You can just use util.GetPodRequest（pkg/util/pod_resources_utils.go:64

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,182 @@
...
+			// Skip nodes with zero eligible target pods
+			continue
+		}
+
+		skippedPods := skippedByNode[node.Name]
+		ntPods := nonTargetByNode[node.Name]
+
+		var num, den float64
+		for resName, weight := range resourceWeights {
+			cap := node.Status.Capacity[resName]
```
songtao98: `cap` is a built-in golang function. I suggest use `capacity` to name it.

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
```diff
@@ -0,0 +1,182 @@
...
+			uNt := float64(ntReq) / float64(capMilli)
+			uT := float64(tReq) / float64(capMilli)
+
+			num += weight * uNt
+			den += weight * uT
+		}
+
+		score := 0.0
+		if den > 0 {
+			score = num / den
```
songtao98, line 110: The if den > 0 guard is correct in that it avoids NaN (Go floating-point division doesn't panic — 0.0/0.0 yields NaN, which would break sort.Slice's comparator transitivity). 

But it also lumps in the num > 0 sub-case: when a node has eligible target pods whose weighted request total is 0 (den == 0) while its non-target load is non-zero (num > 0), score falls back to 0.0 — which sorts it as the most preferred node to drain. That's the opposite of the intended semantics: a node carrying non-target "tax" but no target utilization should be drained last.

So if den == 0 && num > 0, score = math.Inf(1) should fix this. 

Worth adding a unit test covering a node with a zero-request (e.g. BestEffort) eligible target pod plus a non-target pod

songtao98 (changes requested):

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Done

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Done

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Sure

pkg/descheduler/framework/plugins/scaledownbinpack/ranker.go
Vatsalpatni73: Done

songtao98:
Nice Job! @Vatsalpatni73 Please squash your commits into 1 single commit then I can merge it.

songtao98:
@Vatsalpatni73 The newest commit looks strange. Seems like some non-releated files are changed and there are some conflicts. Would you like to rebase your branch onto the newest main branch and commit again?

Vatsalpatni73:
Hello @songtao98 !!
Thanks for the review. I've rebased my branch onto the latest main, resolved the conflicts, removed the unrelated changes, and pushed the updated commits. Could you please take another look?

songtao98 (approved):
/lgtm

PR #3101, opened 2026-06-26
docs: adding native gang scheduling proposal

aclfe:

Defines concept mapping, coexistence strategy, preemption forward-compatibility, and Beta-gated migration plan between Koord Coscheduling and k8s 1.35+ native GangScheduling/Workload API.

Design-only proposal that aligns with Maintainer's needs. They mentioned https://github.com/koordinator-sh/koordinator/issues/2856#issuecomment-4430028670, thus there's a big focus on future compatibility. 

Addresses for https://github.com/koordinator-sh/koordinator/issues/2856
Parent issue: https://github.com/koordinator-sh/koordinator/issues/2851

### V. Checklist


aclfe:
@saintube @ZiMengSheng @hormes please review. If you suggest, maybe this needs further discussions in the bi-weekly?

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
```diff
@@ -0,0 +1,575 @@
...
+`Workload` rework) is contained to one file rather than spread across the plugin.
+
+```go
+// Three implementations exist:
+//   - annotationGangInfo (existing; parses gang.scheduling.koordinator.sh/*)
+//   - podGroupGangInfo  (existing; derives from PodGroup CRD)
+//   - workloadGangInfo  (new; backed by upstream Workload/PodGroup API)
+type GangInfo interface {
+    Identity() string
+    Members() []*corev1.Pod
```
saintube, line 344: How to integrate with the upstream WorkloadManager?

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
```diff
@@ -0,0 +1,575 @@
...
+    MinQuorum() int32
```
saintube, line 345: How to represent the gang group across multiple gangs?

saintube (commented):
PTAL /cc @ZiMengSheng

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
```diff
@@ -0,0 +1,575 @@
...
+### Compatibility Matrix
+
+| Koord feature | Classification | Notes |
+|---|---|---|
+| PodGroup CRD gang | bridgeable | Maps to a single-PodGroup Workload |
+| Annotation/lightweight-label gang | bridgeable | Maps to a single-PodGroup Workload |
+| `MinCount` quorum | bridgeable | Direct field mapping |
+| GangGroup (<= 8 members) | bridgeable | Maps to a multi-PodGroup Workload |
+| GangGroup (> 8 members) | blocked-on-upstream | Exceeds `WorkloadMaxPodGroups` |
+| Strict mode | partial / analogy | Proposed analogy to `DisruptionMode.All` (KEP-5710); see caveat below -- different axis |
```
ZiMengSheng: I don't know if these two are equivalent.

ZiMengSheng:
The WorkloadSchedulingCycle and WorkloadManager in Kubernetes are excellent designs—could Koordinator's implementation potentially evolve in this direction? However, certain differing implementations and Koordinator's client-facing protocols—such as the preemption algorithm, network topology-aware scheduling algorithm, WorkloadAuditor, and Gang scheduling protocol—should remain consistent with the current approach.

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
aclfe: GangGroupInfo can be reused (from ganggroup.go), we can do one Workload = one GangGroup (podGroups[] ≤8 = siblings)

I'm updating the plan to include details

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
aclfe: Kubernetes is expected to rename some of these pieces before this becomes stable, but Workload object itself can tell us **which pods should join, how many are needed**. Now **which pods have shown up, whether we have enough yet** come from a separate live tracker of Kubernetes keeps
Native plugin already read from this

docs/proposals/scheduling/20260626-native-gang-scheduling-compatibility.md
aclfe: Let me fix it, I've taken the comparison out

aclfe:
> The WorkloadSchedulingCycle and WorkloadManager in Kubernetes are excellent designs—could Koordinator's implementation potentially evolve in this direction? However, certain differing implementations and Koordinator's client-facing protocols—such as the preemption algorithm, network topology-aware scheduling algorithm, WorkloadAuditor, and Gang scheduling protocol—should remain consistent with the current approach.

the previous plan might have been too big, I've made changes to better match what is expected
/cc @ZiMengSheng @saintube

ZiMengSheng:
Reviewed as a reference document, which means I checked the load-bearing upstream claims against the vendored tree rather than just reading for shape. Almost all of them hold exactly, which is unusual and worth saying. Two things need fixing before this lands: one upstream fact is wrong in a way that invalidates a scoping decision, and the ownership rule contradicts the Koord-only guardrail.

Both earlier review threads are addressed: @saintube's "how to represent the gang group across multiple gangs" is answered concretely (one Workload = one GangGroup, `podGroups[]` are the siblings, reuse `core.GangGroupInfo`), his "how to integrate with WorkloadManager" is answered by splitting the seam into static config from the `Workload` lister vs runtime state from `WorkloadManager`, and my `:261` objection was resolved the right way — by deleting the strict/non-strict ↔ `DisruptionMode` analogy and classifying it `koord-only`, with an explanation of why the two live at different lifecycle stages. That is the correct answer, not a patch over it.

### What I verified

| Claim | Result |
|---|---|
| `GangScheduling: {GenericWorkload}` dependency | ✅ `kube_features.go:2249` |
| `GangScheduling` alpha/off at 1.35 | ✅ `:1317-1319` |
| `GenericWorkload` alpha/off at 1.35 | ✅ `:1321-1323` |
| `WorkloadMaxPodGroups = 8` | ✅ `k8s.io/api@v0.35.2 scheduling/v1alpha1/types.go:111` |
| `Workload` has `Spec` only, no `Status` | ✅ `:82-95` |
| `PodGroup` is `Name` + `Policy` only (so no `DisruptionMode`) | ✅ `:154-165` |
| `WorkloadReference` = `Name`+`PodGroup`+`PodGroupReplicaKey`, no `PodGroupKind` | ✅ `core/v1/types.go:4568-4593` |
| gang plugin implements `PreEnqueue`/`Reserve`/`Unreserve`/`Permit`, no `PostFilter` | ✅ `gangscheduling.go:120/163/177/201` |
| `PodGroupInfo` tracks all/unscheduled/assumed/assigned | ✅ `workloadmanager/podgroupinfo.go:54-63` |
| 5-minute scheduling timeout | ✅ `podgroupinfo.go:32` (`var`, not `const` — overridable in-process, though not via API; your conclusion stands) |
| `WorkloadSchedulingCycle` absent from 1.35 | ✅ zero hits under `pkg/` |
| only `pkg/descheduler/informers/generic.go` imports `scheduling/v1alpha1` | ✅ single hit on `upstream/main` |
| `extension.IsGangPod` exists | ✅ `apis/extension/coscheduling.go:112` |
| `tryInitByPodConfig` only when the pod carries no `PodGroupLabel` | ✅ `gang_cache.go:145-146` — the condition is literally `if pod.Labels[v1alpha1.PodGroupLabel] == ""` |
| `GangGroupInfo` / `gangGroupInfoMap` / `getGangGroupInfo(id, group, createIfNotExist)` | ✅ `ganggroup.go:22`, `gang_cache.go:43/66` |
| coscheduling `EnablePreemption` default off | ✅ `apis/config/v1/defaults.go` |

The 1.36/1.37 forward-looking claims (KEP-4671/5710 Beta targets, `CompositePodGroup`, the decoupling rework) are outside the dependency tree and I did not verify them. Please date-stamp that whole block so a future reader knows it is a snapshot rather than a checked fact.

### 1. `OpportunisticBatching` is Beta and **on** by default, not "alpha, off" — and it is not orthogonal to gang (must fix)

The Glossary says:

> **OpportunisticBatching**: an upstream scheduler feature gate (`OpportunisticBatching`, alpha, off, present in `kube_features.go@v1.35.2`) ... It is orthogonal to gang enforcement and out of scope here

The vendored tree says otherwise:

```go
// kube_features.go:1588
OpportunisticBatching: {
        {Version: version.MustParse("1.35"), Default: true, PreRelease: featuregate.Beta},
},
```

Beta, default **true**. It is registered into the global gate by `pkg/features` `init()` (`kube_features.go:2584-2586`), and it is live in `schedule_one.go:450/469/512` and `framework/runtime/framework.go:337/433`. koord-scheduler builds on that same `framework/runtime` and the same `k8s.io/apiserver/pkg/util/feature` `DefaultFeatureGate`, so this gate is **already on in koord-scheduler today** — it is not something we would be "enabling by default" as the Non-Goals section implies.

That matters beyond accuracy, because the mechanism is not orthogonal to gang at all. `framework/runtime/batch.go` caches filter and score results keyed by `fwk.PodSignature` obtained from `h.SignPod(...)`, and reuses the previously top-scored node for the *next consecutive pod with the same signature*. Gang members are the canonical population of consecutive identical-signature pods, so gang scheduling is precisely the hot path for this cache, and the cache is only valid "in a 1-pod-per-node manner" (`batch.go` comment) — which is exactly the assumption Koord's topology-aware gang placement does not always satisfy. `SignPod` is also the `SignPlugin` interface that #2863 implements, and #2854 tracks Opportunistic Batching as a sibling of this work under umbrella #2851.

Please (a) correct the maturity, (b) drop it from Non-Goals or reword it, since we are not choosing whether to enable it, and (c) replace "orthogonal, out of scope" with a short subsection stating the actual interaction and cross-referencing #2854 / #2863. It can still be out of scope for *this* design, but it has to be out of scope for a true reason.

### 2. The ownership rule and the Koord-only guardrail contradict each other (must fix)

Ownership Resolution rule 3:

> A pod carrying both a `WorkloadRef` and Koord gang identity is a configuration error; `WorkloadRef` wins and a warning event is emitted.

Koordinator-only Extensions:

> before `resolveGangOwner` may return `GangOwnerNative`, the candidate is checked against these Koord-only markers, and any match forces `GangOwnerKoord`.

For a pod carrying `WorkloadRef` plus, say, a network-topology annotation, rule 3 says native wins and the guardrail says Koord wins. The pseudo-code implements rule 3 and not the guardrail — it emits the event and then returns `GangOwnerNative` unconditionally:

```go
if pod.Spec.WorkloadRef != nil && nativeGangPluginEnabled(handle) {
    if extension.IsGangPod(pod) { emitMisconfigEvent(...) }
    return GangOwnerNative
}
```

Given the Risks section commits to "silent loss of Koord-only semantics: mitigated by the ownership rule routing any pod that needs GangGroup / topology / reservation / quota features to Koord", the guardrail is the intended behavior and rule 3 plus the pseudo-code are the bug. Please make the precedence explicit and single-valued — my suggestion: Koord-only marker present ⇒ `GangOwnerKoord` and warn; `WorkloadRef` plus a *plain* Koord gang identity ⇒ native wins and warn — and update the pseudo-code to match. This is worth getting right now precisely because later code will be written against it.

### 3. Scheduler gates and apiserver gates are different components (should fix)

Rule 1 keys off "both the `GenericWorkload` and `GangScheduling` feature gates are enabled", then notes a profile-inspecting alternative as "more robust". They are not two implementations of one check. Whether `WorkloadRef` is persisted is decided by the **apiserver's** `GenericWorkload`; whether the gang is enforced is decided by the **scheduler's** profile. Reading `utilfeature.DefaultFeatureGate` inside koord-scheduler tells you the scheduler's opinion of a gate that only the apiserver acts on, and in a managed cluster those are set independently. The profile check is not the more robust variant, it is the only correct one; the gate-reading version should be deleted rather than presented as primary. Note also that `pod.Spec.WorkloadRef != nil` is already sufficient evidence that the apiserver persisted the field, which is the cleaner formulation.

### 4. `alphav2` vs `v1beta1` inconsistency (nit)

Migration says "The decoupled PodGroup API (PodGroupTemplates + v1beta1 group) landed as alphav2 in 1.36", stage 2 says "the decoupled PodGroupTemplate **will land at** `scheduling.k8s.io/v1beta1`", and References says "The group **moves to** `scheduling.k8s.io/v1beta1` at Beta". The first sentence can't be reconciled with the other two. Pick one: alphav2 landed in 1.36 at the existing group, and the move to `v1beta1` is a Beta-time event.

### 5. Metadata (nit)

`last-updated: 2026-06-26` but the document was substantially revised on 07-09; `reviewers: TBD`; Implementation History has only the initial draft, missing the revision that removed the `DisruptionMode` analogy and added the GangGroup / WorkloadManager answers. Please refresh all three. `status: provisional` is appropriate and should stay.

### 6. Worth adding to References

The native plugin also registers `EventsToRegister` with two queueing hints (`isSchedulableAfterPodAdded`, `isSchedulableAfterWorkloadAdded`, `gangscheduling.go:68-118`). The References entry lists only the four gating extension points. That omission is worth fixing because it is directly relevant to our own hint work — #2927 just landed `QueueingHint` support for reservation and #3145 tracks the remaining event sources, so the upstream gang plugin's hint design is a live reference rather than trivia.

### Verdict

The design is sound and the shape is right: single seam, default-off, `WorkloadManager` authoritative for native-owned pods so no pod is counted by two state machines, GangGroup > 8 and GangGroup-as-preemption-unit honestly classified `blocked-on-upstream`, and gang-level priority flagged as a real behavior change rather than smuggled in as an alignment no-op. The `GangCache`-ingestion insertion point is the right one and matches how `gang_cache.go` actually works.

Not approving yet only because 1 and 2 are substantive: a reference document that gets a feature gate's maturity backwards will be trusted and repeated, and the ownership contradiction is what future code would be written against. Fix those two plus the nits and I will `/lgtm` on the next push — this should land before the gang implementation PRs (#2969, #2907) so it can fix direction, which is the whole point of merging proposals first.

/cc @saintube @hormes

Issue #3032, opened 2026-05-27
[proposal]Improve scheduling throughput with resource reservations


saintube:
**What is your proposal**:

Enabling reservations in `koord-scheduler` currently degrades throughput and latency. This creates a bottleneck for high-performance use cases like LLM inference and agent sandboxes.

To address this, we plan to implement the following optimizations:

1.  **Index reservations** for pods with `reservation-affinity` to speed up lookup: https://github.com/koordinator-sh/koordinator/pull/3009
2.  **Relax restoration constraints** for node-level reservations to reduce overhead: https://github.com/koordinator-sh/koordinator/pull/2882
3.  **Enable early termination** in reservation filtering when global optimality is not mandatory.

We also welcome any other optimization suggestions or discussions. Please feel free to share your ideas in the comments below.

**Why is this needed**:

**Is there a suggested solution, if so, please add it**:

anshikasharmaa1517:
Hi, I'd like to pick this up.

Went through the reservation plugin code and both PRs. #3009 makes sense, indexing reservation-affinity so we stop iterating over everything. #2882 skips restore for node-scoped reservations where it's just wasted work.

For the early termination part, from what I can see the Filter phase goes through all candidate reservations on a node looking for the best match. But a lot of pods just need any valid reservation, they don't care about optimal. So we could short-circuit on the first valid match instead of scoring all of them. Something like a FirstFit mode you can toggle via plugin arg or pod annotation.

A couple of other things I noticed that might be worth looking at:

- In `BeforePreFilter`, the plugin restores resources for all unmatched reservations across nodes. Even with #2882 handling the node-level case, for cluster-wide reservations we're still doing work on nodes that `PreFilterResult` will eliminate anyway. Could we defer restoration to only the nodes that survive PreFilter?
- The reservation status gets synced every cycle even when nothing changed. If we tracked a dirty flag or a generation counter on reservation state, we could skip the sync when there's no actual update. Might help under high pod churn.

I'd start by tracing `filterReservation` for the early termination, then look into the deferred restore idea. Would also want to make sure everything composes cleanly with the affinity index from #3009.

Few questions:
- For the FirstFit toggle, global plugin arg or per-pod? I was thinking plugin arg with per-pod override but open to whatever fits better.
- Are #3009 and #2882 in a mergeable state or still need work? Happy to help review if useful.

ZiMengSheng:
@anshikasharmaa1517 
[...]

PR #3103, opened 2026-06-27
feat: Enable early termination in reservation

aclfe:

Adding early termination
off-by-default feature gate ReservationFirstFitNomination

Limitation: maybe a score-based best-fit selection? This would have to be discussed with maintainer's expectation. 

Addresses third part of #3032

### V. Checklist

### DIFF
--- a/pkg/scheduler/plugins/reservation/nominator.go
+++ b/pkg/scheduler/plugins/reservation/nominator.go
@@ -375,6 +375,18 @@ func (pl *Plugin) NominateReservation(ctx context.Context, cycleState fwktype.Cy
 		return nil, fwktype.AsStatus(fmt.Errorf("not implemented frameworkext.FrameworkExtender"))
 	}
 
+	// When global optimality is not mandatory, return the first feasible reservation instead of filtering and
+	// scoring every matched reservation to pick the best one.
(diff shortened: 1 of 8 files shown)


aclfe:
@saintube

pkg/scheduler/plugins/reservation/nominator.go
```diff
@@ -375,6 +375,18 @@ func (pl *Plugin) NominateReservation(ctx context.Context, cycleState fwktype.Cy
 		return nil, fwktype.AsStatus(fmt.Errorf("not implemented frameworkext.FrameworkExtender"))
 	}
 
+	// When global optimality is not mandatory, return the first feasible reservation instead of filtering and
```
saintube, line 378: Most reservation traversal overhead comes from matching and restoring reserved resources in transform.go, not NominateReservation.

aclfe: Restoring is already being worked at #2882, I made changes for matching in transform and kept the previous changes for nominator.go. Benchmark tests show good results for the new changes

aclfe:
@saintube I've made some changes, please let me know if this aligns with what's expected

PR #3106, opened 2026-06-29
koordlet: add PSI QoS reconciler for cgroup v2

AutuSnow:

Add a PSI-based QoS reconciler for Koordlet on cgroup v2 nodes.

This PR introduces:
- `PSIReconcile` feature gate and NodeSLO PSI strategy configuration.
- PSI exporter that updates pod pressure conditions from `cpu.pressure`, `memory.pressure`, and `io.pressure`.
- Memory suppression based on `memory.high`.
- CPU group share and budget balance based on `cpu.weight`.
- cgroup v2 helpers for PSI, CPU, memory, and IO resources, reusing the shared `resourceexecutor` and `system` cgroup utilities.
- Reconciler hardening for operator lifecycle sync, deterministic operator execution, cgroup write error propagation, shared PSI reads, and safe IO device handling.

Current IO per-device BPS/IOPS handling only uses per-device stats when the device is explicit or uniquely inferable from `io.stat`; multi-device cgroups fall back to pressure-only behavior to avoid incorrect device selection.

NONE

### V. Checklist


pkg/koordlet/qosmanager/plugins/psi/cgroup/cgroup.go
```diff
@@ -0,0 +1,241 @@
...
+type ResourceMask int
+
+const (
+	ResourceMaskCpu ResourceMask = 1 << iota
+	ResourceMaskMemory
+	ResourceMaskIO
+	ResourceMaskAll = ResourceMaskCpu | ResourceMaskMemory | ResourceMaskIO
+)
+
+func NewCgroup(path string, dev [2]int64) *Cgroup {
```
saintube, line 48: We'd better reuse the pkg resource executor and the type CgroupResource.

AutuSnow: Updated.
﻿
The PSI cgroup wrappers now reuse the registered v2 `CgroupResource`s from `pkg/koordlet/util/system`, and the write path goes through the common `resourceexecutor` helpers instead of PSI-private file bindings.

Issue #2879, opened 2026-04-28
[proposal] Native support for AI Agent orchestration and high-concurrency Sandbox scheduling


ZiMengSheng:
#### 1. Motivation
With the rapid evolution of AI Agentic Workflows (e.g., multi-agent collaboration frameworks, automated evaluation systems like SWE-bench), traditional Kubernetes workload management faces new challenges. Agents frequently spawn and destroy large numbers of short-lived **Sandboxes** to execute specific Skills or Tools. This requires the scheduling system to provide extremely low latency, robust resource isolation, and high throughput.

We propose that Koordinator provides native support for AI Agent scenarios, leveraging its strengths in colocation and fine-grained scheduling to optimize resource efficiency for sandbox environments.

#### 2. Proposed Scenarios & Features

**A. Agent Sandbox Colocation & Hardening**
* **Best Practice Support:** Native integration of Agent Sandboxes (e.g., gVisor, Wasm, Kata Containers) with Koordinator QoS classes (`LSR`, `LS`, `BE`).
* **Resource Hardening:** Provide precise resource isolation templates for sandbox runtimes (e.g., managing the overhead of the `runsc` process in gVisor) to ensure zero resource contention between the Agent "Brain" and the Skill execution environment in high-density deployments.

**B. Scheduling Optimization for High-Concurrency Tasks**
* **Equivalence Class Scheduling:** Implement equivalence class logic for massive, near-identical sandbox requests (e.g., the same Skill image running on different datasets). This allows the scheduler to skip redundant Predicate and Priority calculations, significantly boosting throughput.
* **APIServer Pressure Relief:** Mitigate APIServer bottlenecks caused by high-frequency Pod lifecycle events through batch processing or optimized caching within the Koordinator scheduler.

**C. Customized Sandbox Pipeline**
* **Pipeline Orchestration:** Introduce a dedicated Sandbox Pipeline mechanism to manage the entire lifecycle from environment preparation to task execution.
* **Pre-warming Mechanism:** Support pre-pulling images or pre-creating sandbox environments based on Agent execution hints.
* **Automated Environment Provisioning:** Automatically handle resource mapping (e.g., creating directory `Symlinks` for large datasets) and Capacity reservations within the pipeline flow.

**D. Automated Diagnostics & Feedback**
[...]

PR #3112, opened 2026-07-01
docs: add Native Support for AI Agent Orchestration and Sandbox Scheduling proposal.

tan90github:

This PR adds a proposal for native support of AI Agent orchestration and high-concurrency sandbox scheduling in Koordinator.

The proposal defines the overall framework:

- Pod-shaped sandboxes opt in through the `scheduling.koordinator.sh/sandbox=true` label on the final Pod or PodTemplate.
- Non-Pod sandboxes are represented by accounting-only Reservations so their resource usage is visible to Koordinator.
- A dedicated sandbox scheduling framework is proposed for high-throughput sandbox workloads.
- The initial performance target is 5k sandboxes/s sustained bind throughput in a 1000 fake-node benchmark.
- Future work includes deeper QoS integration, sandbox failure diagnosis, and integration with sandbox prewarming APIs.

Related to #2879

Documentation-only change. 

### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20260626-native-support-for-ai-agent-sandbox-scheduling.md
+++ b/docs/proposals/scheduling/20260626-native-support-for-ai-agent-sandbox-scheduling.md
@@ -10,9 +10,9 @@ last-updated: 2026-07-01
 status: provisional
 see-also:
   - "https://github.com/koordinator-sh/koordinator/issues/2879"
-  - "docs/proposals/scheduling/20220609-resource-reservation.md"
-  - "docs/proposals/scheduling/20221227-node-resource-reservation.md"
-  - "docs/proposals/koordlet/20220615-qos-manager.md"
+  - "/docs/proposals/scheduling/20220609-resource-reservation.md"
+  - "/docs/proposals/scheduling/20221227-node-resource-reservation.md"
+  - "/docs/proposals/koordlet/20220615-qos-manager.md"
 ---
 
 # Native Support for AI Agent Orchestration and Sandbox Scheduling
@@ -102,12 +102,12 @@ Overall workflow:
           |                                   |
           |                                   v
           |                           accounting-only Reservation
-          |                           sandbox=true label
+          |                           scheduling.koordinator.sh/sandbox=true label
           |                                   |
           +-----------------+-----------------+
                             |
                             v
-             koordinator sandbox scheduling framework
+             Koordinator sandbox scheduling framework
                     (high-throughput path)
                             |
               +-------------+-------------+
@@ -137,7 +137,7 @@ Koordinator can provide three main benefits:
 - Define the conversion protocol from Non-Pod sandboxes to accounting-only Reservations.
 - Complete the minimum loop for sandbox creation, scheduling or resource accounting, startup, exit, and release.
 - Introduce a dedicated sandbox scheduling framework to support efficient sandbox scheduling.
-- Define a reproducible benchmark profile, use sustained bind throughput of 5k sandboxes/s on 1000 fake nodes as the first-stage target, and use later benchmarks to evaluate long-term targets such as 10k+ sandboxes/s or higher.
+- Define a reproducible benchmark profile, use sustained bind throughput of 2k sandboxes/s on 1000 fake nodes as the first-stage target, and use later benchmarks to drive toward a 5k sandboxes/s long-term target.
 
 ### Non-Goals/Future Work
 
@@ -203,7 +203,7 @@ Koordinator MUST provide a dedicated sandbox scheduling framework or pipeline fo
 
 ##### NFR1
 
-High-throughput sandbox scheduling is a primary requirement for production adoption. The implementation SHOULD use explicit benchmark profiles and staged performance targets. The first concrete optimization target is 5k sandboxes/s bind throughput in a 1000 fake-node homogeneous sandbox burst benchmark. Long-term targets such as 10k+ sandboxes/s require separate benchmark evidence and correctness validation.
+High-throughput sandbox scheduling is a primary requirement for production adoption. The implementation SHOULD use explicit benchmark profiles and staged performance targets. The first concrete optimization target is 2k sandboxes/s bind throughput in a 1000 fake-node homogeneous sandbox burst benchmark, so 2k sandboxes/s is the first-stage baseline and 5k sandboxes/s is the long-term target. The 5k target requires separate benchmark evidence and correctness validation.
 
 ##### NFR2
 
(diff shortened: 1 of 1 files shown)


tan90github:
PTAL @saintube @ZiMengSheng , thanks~

docs/proposals/scheduling/20260626-native-support-for-ai-agent-sandbox-scheduling.md
```diff
@@ -0,0 +1,474 @@
...
+
+### Test Plan [optional]
+
+TBD
+
+### Open Questions
+
+- Should the sandbox scheduling framework run inside the existing koord-scheduler process, or should it be split into a separate sandbox scheduler? The decision criteria include throughput targets, fault isolation, scheduler cache / Reservation reuse, deployment complexity, and operational cost.
+- Does accounting-only Reservation need explicit new API / plugin semantics to guarantee that it can never be claimed by Pods?
+- If benchmarks prove that Filter / Score is the main bottleneck, which scheduler plugins can safely enable equivalence-class reuse in the first version?
```
saintube, line 464: Certainly, we should reuse the equivalence-class algorithm. Additionally, for sandbox reservations/pods sharing the same template, we can add a fast path to derive their signature using the template ID.

saintube (commented):
Good work! This is a significant undertaking that may span multiple versions. Let's set milestones and focus on the core deliverables within the LFX timeline. Looking forward to your future contributions.

tan90github:
> Good work! This is a significant undertaking that may span multiple versions. Let's set milestones and focus on the core deliverables within the LFX timeline. Looking forward to your future contributions.

Thanks @saintube. I agree this proposal may span multiple versions, so we should narrow the LFX-scope milestones and keep the first implementation focused.

One scope question needs mentor/community confirmation: should the LFX implementation include the Non-Pod sandbox adapter, or should we keep that part as design/future work and focus first on sandbox scheduling throughput optimization?

My concern is that the Non-Pod adapter path needs at least one concrete target runtime/sandbox type to validate lifecycle mapping, admission constraints and the complete scheduling process. If the community has a preferred first target, I can use it to make this milestone concrete. Otherwise, I suggest regarding the adapter as a follow-up task, at minimum not implementing it within LFX.

By the way, during my research, I found that Non-pod sandboxes mostly do not expose APIs for scheduling target nodes or launching on specified nodes (such as e2b). This means it will become difficult for these Non-pod sandboxes to be scheduled via reservations and launched on designated nodes. Feel free to leave a comment if the community has more relevant information.


Tentative LFX milestones:

| Milestone | Scope |
|---|---|
| M1 | Finalize the Pod-shaped sandbox scheduling contract and opt-in label. |
| M2 | Build the dedicated sandbox scheduling framework/path while preserving existing behavior for non-sandbox Pods. |
| M3 | Trim unnecessary plugins and tune basic scheduling parameters for sandbox workloads. |
| M4 | Establish reproducible benchmarks and optimize toward the first-stage 5k sandboxes/s target. |

Future/TBD work includes Non-Pod sandbox adapters, equivalence-class reuse, binder pipeline optimization, 10k+ target validation, QoS integration, and prewarming integration.

Does this split match your expectation for the LFX timeline?

tan90github:
@ZiMengSheng @saintube 
All Copilot review feedback has been fixed. Could you take a look when you get a chance, and also respond to the questions I left in [comments](https://github.com/koordinator-sh/koordinator/pull/3112#issuecomment-4864003518)? Thanks!

saintube:
@tan90github [Custom Workflow](https://github.com/koordinator-sh/koordinator/blob/main/cmd/koord-scheduler/app/workflow.go)

miantalha45:
Thanks for the detailed proposal. especially the clear separation between Pod-shaped and Non-Pod sandboxes, plus the benchmark-first performance plan.
One implementation detail seems important to clarify before the first milestone: `cmd/koord-scheduler/app/workflow.go` says that when a `CustomWorkflow` is enabled, the default scheduler workflow does not run.
Since the proposal requires sandbox workloads to use a dedicated path while non-sandbox Pods retain existing `koord-scheduler` behavior, could the proposal define the intended compatibility model?
- Will sandbox workloads use a dedicated scheduler deployment/profile, while ordinary Pods remain on the default scheduler workflow?
- Or will one custom workflow dispatch both sandbox and ordinary Pods safely?
- How will the `scheduling.koordinator.sh/sandbox=true` label select the intended path without affecting existing workloads?

I think documenting this boundary would make the initial LFX scope much more concrete. It may also support a focused first milestone: label-based Pod sandbox routing, correctness/compatibility tests for ordinary Pods, and a baseline benchmark before adding equivalence-class reuse or binding-pipeline optimizations.
For the Non-Pod adapter, deferring implementation until a concrete target runtime and lifecycle contract are agreed seems reasonable, since node selection, lifecycle reconciliation, and Reservation cleanup need end-to-end validation.

saintube:
@tan90github Any update on the CustomWorkflow?

tan90github:
> @tan90github Any update on the CustomWorkflow?


@saintube, I have mainly been working on optimizing and testing the upstream Kubernetes scheduling framework.

Last week, I posted [[this comment](https://github.com/koordinator-sh/koordinator/pull/3135#issuecomment-5130611414)](https://github.com/koordinator-sh/koordinator/pull/3135#issuecomment-5130611414) about using a simple plugin at the Score extension point to spread pods across nodes. After adding the Score plugin, the scheduling throughput dropped from 2,000 pods/s to 1,700 pods/s.

During testing, I found that the Kubernetes 1.35 scheduling framework supports [[opportunistic batching](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5598-opportunistic-batching/README.md)](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5598-opportunistic-batching/README.md), and I am currently looking into it. This may help further improve the scheduling throughput. So far, it seems that all in-tree Kubernetes scheduler plugins have implemented `SignPlugin`, while the Koordinator plugins have not, so they cannot benefit from batch scheduling acceleration yet.

Or should we stop optimizing the Kubernetes scheduling framework at 1,700 pods/s and focus our efforts on CustomWorkflow instead?

saintube:
@tan90github Adopting the upstream opportunistic batching is the long-term task of Koordinator. Unfortunately, the current mechanism is designed specifically for the one-pod-per-node scenario and does not improve the throughput for the multiple-pod-per-node case. Actually, we do not impose constraints on the methods used to optimize performance. However, since our objective is sandbox awareness, we tend to pick the optimization strategies that hit the sweet spot for sandbox scenarios.

tan90github:
> @tan90github Adopting the upstream opportunistic batching is the long-term task of Koordinator. Unfortunately, the current mechanism is designed specifically for the one-pod-per-node scenario and does not improve the throughput for the multiple-pod-per-node case. Actually, we do not impose constraints on the methods used to optimize performance. However, since our objective is sandbox awareness, we tend to pick the optimization strategies that hit the sweet spot for sandbox scenarios.

@saintube Okay, I got it, thank you so much for sharing your valuable feedback with me. I’m going to develop the sandbox scheduling features that rely on CustomWorkflow.

PR #3114, opened 2026-07-02
scheduler: fix reservation nomination leak on schedulerName handover

### LINKED ISSUE #3091: [BUG] In-Memory Reservation Nomination Leak & Resource Locking on Scheduler Name Handover
# Root Cause:
When an active reservation's responsibility is transferred from one scheduler profile to another (oldResponsible && !newResponsible), the event handler correctly removes the reservation from the old scheduler's queue. However, it fails to retract the corresponding resource nomination from the old scheduler's rrNominator.

## Impact:
The leaked nomination persists as a "ghost" resource lock on the nominated node. The old scheduler profile will perpetually see an inflated resource allocation on that node, causing it to incorrectly reject other pods that would otherwise fit. This leads to scheduling failures and reduced cluster resource utilization.

## Remediation:
The fix is to invoke DeleteNominatedReservePodOrReservation on the nominator object during the responsibility handover. This ensures that when a scheduler is no longer responsible for a reservation, it cleans up both the scheduling queue and any active nominations, maintaining accurate resource accounting.

Image

harshakumar25:
**Fixes a silent memory leak and "ghost lock" vulnerability that occurs when an active Reservation's responsibility is transferred between scheduler profiles (e.g., from `koordinator-scheduler` to a descheduler process).**

## Root Cause:
When a reservation's `schedulerName` changes (`oldResponsible && !newResponsible` in Case 5 of `updateReservation`), the event handler correctly removes the reservation from the old scheduler's queue, but failed to retract the corresponding resource nomination from the old scheduler's `ReservationNominator`. 

## Impact:
The leaked nomination persists as an invisible resource lock on the nominated node in the old profile's memory. This artificially inflates resource accounting on that node, causing the old scheduler to perpetually reject valid pods with `Insufficient cpu/memory` until it is restarted.
## Remediation:
Introduced `deleteNominatedReservationFromProfile` to strictly clean up the `ReservationNominator` during handover. The implementation validates the framework, extender, and nominator exist in O(1) time before executing the deep-copy allocation required for the cleanup, preserving scheduler latency budgets.
### Special note :- 

I have added `TestUpdateReservation_NominationLeak` which accurately models the multi-process scheduler topology (where `newResponsible=false` due to the new profile not existing in the current instance) to ensure this leak cannot regress.

### DIFF
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
@@ -485,7 +485,11 @@ func updateReservation(sched *scheduler.Scheduler, schedAdapter frameworkext.Sch
 		} else if !oldResponsible && newResponsible {
 			addReservationToSchedulingQueue(schedAdapter, newR)
 		} else if oldResponsible && !newResponsible {
+			// The reservation is no longer managed by the current scheduler.
+			// Remove from the scheduling queue and retract the in-memory nomination
+			// so this profile no longer holds a ghost resource lock on the nominated node.
 			deleteReservationFromSchedulingQueue(sched, schedAdapter, oldR)
+			deleteNominatedReservationFromProfile(sched, oldR)
 		}
 		return
 	}
@@ -669,6 +673,30 @@ func deleteReservationFromSchedulingQueue(sched *scheduler.Scheduler, schedAdapt
 	if fwk != nil {
 		fwk.RejectWaitingPod(reservePod.UID)
 	}
+}
+
+// deleteNominatedReservationFromProfile retracts the in-memory reservation nomination
+// held by the old scheduler profile during a scheduler-name handover. Nominations
+// that are not cleaned up here become ghost resource locks on the nominated node,
+// causing the old profile to inflate its resource accounting and reject other pods.
+func deleteNominatedReservationFromProfile(sched *scheduler.Scheduler, r *schedulingv1alpha1.Reservation) {
+	schedulerName := reservationutil.GetReservationSchedulerName(r)
+	fwk := sched.Profiles[schedulerName]
+	if fwk == nil {
+		return
+	}
+	extender, ok := fwk.(frameworkext.FrameworkExtender)
+	if !ok {
+		return
+	}
+	nominator := extender.GetReservationNominator()
+	if nominator == nil {
+		return
+	}
+	reservePod := reservationutil.NewReservePod(r)
+	nominator.DeleteNominatedReservePodOrReservation(reservePod)
+	klog.V(4).InfoS("Deleted reservation nomination on scheduler handover",
+		"reservation", klog.KObj(r), "schedulerName", schedulerName)
 }
 
 func isResponsibleForReservation(profiles profile.Map, r *schedulingv1alpha1.Reservation) bool {
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler_test.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler_test.go
@@ -1829,3 +1829,234 @@ func Test_generatePodEventOnReservationLevel(t *testing.T) {
 		})
 	}
 }
+
+
+func TestMakeReservationErrorHandler_NominatedNodeName_Persistence(t *testing.T) {
+	testR := &schedulingv1alpha1.Reservation{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "reserve-pod-persistence-test",
+			UID:  "persistence-uid",
+		},
+		Spec: schedulingv1alpha1.ReservationSpec{
+			Template: &corev1.PodTemplateSpec{},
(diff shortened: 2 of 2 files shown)


ZiMengSheng:
### This PR cannot compile — it references a field that only exists in your other PR

Thanks for splitting #3091 out as its own fix; the diagnosis in the description is accurate and the leak is real. But there's a structural problem that explains why CI is red across the board (`Run Go test`, `govet`, `gofmt`, `golangci-lint` all failing):

Your test asserts on `updatedR.Status.NominatedNodeName`:

```go
rCopy.Status.NominatedNodeName = tt.initialNomination
assert.Equal(t, tt.expectPersisted, updatedR.Status.NominatedNodeName)
```

**That field does not exist on `main`.** It is introduced by your other PR #3084 (`apis/scheduling/v1alpha1/reservation_types.go`, the new `NominatedNodeName` with protobuf tag 7). This PR's base is `main` and it does not carry that API change, so the package doesn't build.

This looks like the branch was cut from your #3084 work and then lost the API commit. Two ways out, and the choice matters:

**Option 1 — decouple (recommended).** The actual fix for #3091 is `deleteNominatedReservationFromProfile`, and that function needs **no API change at all** — it only retracts in-memory state via `nominator.DeleteNominatedReservePodOrReservation`. Drop `TestMakeReservationErrorHandler_NominatedNodeName_Persistence` from this PR (it belongs in #3084), keep a test that asserts the nominator no longer holds the reserve pod after handover, and this becomes a small self-contained fix that can merge independently.

**Option 2 — stack it.** Rebase onto #3084 and say so explicitly in the description. This makes #3091 wait on the API discussion, which I'd rather avoid since the leak is a real bug worth fixing sooner.

I'd go with Option 1.

### One more thing: overlap with #3156

@thc1006's #3156 also touches `reservation_handler.go` and rewrites `nominator.go` (+279) to remove stale nominator holds when a reservation stops being able to use a node. Your Case 5 handover cleanup and his phantom-hold cleanup are **different triggers for the same class of bug** — a nomination that outlives its validity.

Please read #3156 before reworking this. If his `nominator.go` changes already provide the retraction path you need, your fix may reduce to wiring Case 5 into it rather than adding a parallel helper. Worth the two of you agreeing on one cleanup mechanism instead of two.

I've recorded the ordering for this whole cluster on #3069.

ZiMengSheng:
Reviewed as part of a cross-check over the four PRs touching the reservation nominator (#2885, #3084, #3156, this one). Cluster analysis: https://github.com/koordinator-sh/koordinator/pull/3084#issuecomment-5564939241

**The bug is real and I verified the fix is correctly wired.** On `main`, `reservationEventHandler.OnUpdate` only drops a nomination when the reservation is Active, Failed or Succeeded; a still-unscheduled reservation whose `schedulerName` moves away hits none of those, and Case 5 in `updateReservation` only removes it from the scheduling queue. So the nomination does survive the handover. I also checked the call you use rather than assuming it: `DeleteNominatedReservePodOrReservation` routes on `IsReservePod`, and `NewReservePod` sets `AnnotationReservePod`, so it reaches `deleteReservePod` — which also clears `nominatedPreAllocatable` for that UID. And `sched.Profiles[name]` is the same lookup `HandlesSchedulerName` performs, so the profile you fetch is the one that held the entry.

Two things before this can go in:

1. **#3156 covers the same leak, differently, and one of you should decide which shape we keep.** Its `OnUpdate` gains a branch on `Generation`/labels/annotations changing, and its comment names this exact case: "on a schedulerName handover the framework handler only removes the reserve pod from the scheduling queue, never from this nominator". The difference is not cosmetic — yours fires precisely at the handover and removes unconditionally; #3156's fires on any spec change and removes only if the informer store agrees the entry is stale. On a handover the reservation demonstrably no longer belongs to this profile, so unconditional is the stronger and more appropriate guarantee there. My inclination is that both survive: #3156 as the general staleness backstop, yours as the exact-trigger removal. But I'm not merging two overlapping fixes without you and @thc1006 agreeing on that. Please sync on #3156.

2. **`oldActive && newTerminated` (Case 6) has the same shape and you didn't touch it.** There, `oldResponsible` is true and the reservation had a nomination, and only the queue removal happens. It's covered today by the plugin handler's `Failed || Succeeded` branch, so it isn't a live bug — but it means the invariant "leaving this scheduler's control implies dropping the nomination" is enforced in two unrelated files depending on which transition you take. Worth a sentence in the commit message about why Case 5 was the only one that needed it, so the next person doesn't have to re-derive that.

Smaller: is there an issue or a reproduction behind this? The description asserts the ghost-lock symptom (old profile inflating its accounting and rejecting other pods) but nothing links it to an observed failure, and `reservation_handler_test.go` tests the new function rather than the symptom. A test that shows the old profile still counting the node's resources after the handover would make the case much harder to argue with.

Your merge-base is 49 commits behind `main`. It merges cleanly today, and cleanly against #3084 / #3156 in either order — the only collisions are in `reservation_handler_test.go`, where all three of you append cases at the same spot. Please rebase.

PR #3117, opened 2026-07-05
descheduler: add ScaleDownBinPack plugin and registration

Vatsalpatni73:

Adds the plugin wrapper, side-effect executor, registration, and unit tests for the `ScaleDownBinPack` descheduler plugin.

**`scale_down_binpack.go`**: Implements `framework.BalancePlugin`:
- `NewScaleDownBinPack` factory validates args and builds a composed pod filter from `PodSelectors`, `EvictableNamespaces`, and `handle.Evictor().Filter`.
- `Balance()` filters nodes by `NodeSelector`, retrieves pods per node via the cached `handle.GetPodsAssignedToNodeFunc()` (no kube-apiserver list calls), partitions them into eligible targets / skipped targets / non-target pods, and calls the existing `RankPods()` ranker. It then dispatches to the execution strategy: `CalculateOnly` or `EvictDirectly`.
- `isTargetPod()` distinguishes skipped targets (match selectors but fail eviction constraints) from non-target workloads, ensuring skipped targets contribute to the non-target tax in the evacuation score.

**`executor.go`**: Handles the execution strategies dispatched by `Balance()`:
- `patchDeletionCosts()` (for `CalculateOnly` strategy) maps the pod's global rank to its native `controller.kubernetes.io/pod-deletion-cost` annotation. It also assigns `math.MaxInt32` to skipped target pods so that native controllers scale them down last. 
- `evictRankedPods()` (for `EvictDirectly` strategy) sequentially routes the ranked pods to the framework's existing evictor flow, respecting the `MaxPodsToEvict` threshold.

**`registry.go`**: Registers `ScaleDownBinPack` in `NewInTreeRegistry()`.

**Tests (`scale_down_binpack_test.go` and `executor_test.go`)**: Contains 24 tests covering:
- **Execution & Patching:** verifies ranked pods are assigned correct deletion costs, skipped pods get `MaxInt32`, and the framework evictor limits are respected. Tests use an injected fake clientset to test side-effects without panic.
- **Node & Pod Retrieval:** Cached pod function is called (not kube-apiserver `Pods().List()`), exactly once per node. Node target counts match cached pods; nodes with zero eligible targets are skipped.
- **Constraints & Edge Cases:** Paused plugin exits immediately. Completed (Succeeded/Failed) pods are excluded. Node selector filtering, namespace filtering, and evictor filter integration behave correctly.

Part of #2790

```bash
# Run the plugin unit tests
go test -v -count=1 ./pkg/descheduler/framework/plugins/scaledownbinpack/...

# Verify the build
go build ./pkg/descheduler/framework/plugins/...
```

- `Balance()` orchestrates the decision and then calls the `executor.go` helpers. The framework's existing migration controller files perform the lower-level eviction flow (PMJs, etc.) after the plugin calls the framework evictor.
- Pod retrieval uses the cached `GetPodsAssignedToNodeFunc` exclusively. The fake `podsLister` in tests tracks call counts to verify no API server calls are made.
- The plugin follows established patterns from `LowNodeLoad` and `CustomPriority` plugins for filter composition, node selection, and framework integration.

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
+++ b/pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scaledownbinpack
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/pkg/descheduler/framework"
+)
+
+const (
+	podDeletionCostAnnotation = "controller.kubernetes.io/pod-deletion-cost"
+)
+
+func (pl *ScaleDownBinPack) patchDeletionCosts(ctx context.Context, ranked []RankedPod, skipped []*corev1.Pod) *framework.Status {
+	if pl.handle.IsDryRun() {
+		klog.V(4).InfoS("DryRun: would patch deletion costs", "rankedPods", len(ranked), "skippedPods", len(skipped))
+		return nil
+	}
+
+	var errs []error
+	client := pl.handle.ClientSet()
+
+	// Patch eligible ranked pods with their computed rank.
+	for _, rp := range ranked {
+		costStr := fmt.Sprintf("%d", rp.Rank)
+		if err := pl.patchPodAnnotation(ctx, client, rp.Pod, podDeletionCostAnnotation, costStr); err != nil {
+			errs = append(errs, err)
+			klog.ErrorS(err, "Failed to patch pod deletion cost for ranked pod", "pod", klog.KObj(rp.Pod))
+		}
+	}
+
+	// Skipped target pods (matched by selector but blocked by constraints) should ideally
+	// not be scaled down. We can assign them the maximum integer cost to ensure native
+	// controllers delete them last if scale-down is forced.
+	for _, pod := range skipped {
+		costStr := fmt.Sprintf("%d", math.MaxInt32)
+		if err := pl.patchPodAnnotation(ctx, client, pod, podDeletionCostAnnotation, costStr); err != nil {
+			errs = append(errs, err)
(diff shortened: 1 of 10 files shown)


pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
```diff
@@ -0,0 +1,106 @@
...
+	"k8s.io/klog/v2"
+
+	"github.com/koordinator-sh/koordinator/pkg/descheduler/framework"
+)
+
+const (
+	podDeletionCostAnnotation = "controller.kubernetes.io/pod-deletion-cost"
+)
+
+func (pl *ScaleDownBinPack) patchDeletionCosts(ctx context.Context, ranked []RankedPod, skipped []*corev1.Pod) *framework.Status {
```
songtao98, line 38: `patchDeletionCosts` writes directly through `ClientSet`, bypassing the framework’s dry-run-aware evictor. As a result, running koord-descheduler with global `dryRun: true` still mutates live Pods and changes future scale-down decisions. 

`evictRankedPods` mode is compatible with dryRun because it calls `pl.handle.Evictor().Evict` which implements dryRun logic globally.


Please make annotation writes dry-run-aware.

songtao98 (changes requested):

pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
Vatsalpatni73: Thanks for pointing this out!

I've pushed an update that fixes this by:
1. Adding a `DryRun` field to `ScaleDownBinPackArgs` in both the internal and `v1alpha2` config APIs.
2. Guarding both `patchDeletionCosts` and `evictRankedPods` with `pl.args.DryRun`. If dry-run is enabled, it logs the intended action (`klog.V(4)`) but returns early without applying the annotation patches or calling the evictor.
3. Adding explicit unit tests (`TestPatchDeletionCosts_DryRun` and `TestEvictRankedPods_DryRun`) to verify no side-effects occur.

songtao98 (changes requested):
The unit test failed. Please check TestSetDefaults_ScaleDownBinPackArgs and fix it.

pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
songtao98: Adding a DryRun field to ScaleDownBinPackArgs cannot fix the problem I mentioned before. Notice that running koord-descheduler with **_global_** dryRun: true still mutates live Pods and changes future scale-down decisions.

songtao98:
> /retest

@Vatsalpatni73 A new commit push will trigger auto-retest. The newest commit needs to fix go fmt.

pkg/descheduler/framework/plugins/scaledownbinpack/executor.go
Vatsalpatni73: @songtao98 , Thanks for catching this!
I've pushed a new commit that fixes this properly:

1. I added IsDryRun() bool to the framework.Handle interface so plugins can inspect the global dry-run state.
2. I updated both patchDeletionCosts and evictRankedPods to check pl.handle.IsDryRun() and return early without mutating if it's enabled.
3. I removed the redundant DryRun field from ScaleDownBinPackArgs and its v1alpha2 types since it's no longer needed, and updated the tests to use frameworkruntime.WithDryRun(true).

Vatsalpatni73:
@songtao98 Thanks for the review. I’ve fixed the code and pushed the changes. Could you please review it again?

songtao98 (approved):

PR #3120, opened 2026-07-09
fix: add webhook validation for quotaLabels in ElasticQuotaProfile

### LINKED ISSUE #3116: [BUG]We have noticed an instance of Incorrect Admission Validator in your repository.
**What happened**:
We observed that the Custom Resource Definition (CRD) does not sufficiently validate user-provided input for one or more label-related fields.

Specifically, the affected field accepts values that do not satisfy the expected label format because the CRD schema does not define an appropriate validation rule (e.g., a pattern constraint), and no admission validation logic rejects the invalid input (e.g., label: test-wec1, test-wec2). As a result, invalid label values can be successfully created and stored by the API server. These invalid values may later cause errors when the application or controller processes the affected fields, leading to reconciliation failures or unexpected runtime behavior.

Locations:

https://github.com/koordinator-sh/koordinator/blob/989ca85c62abcca92b303aa12fd2ccff2ed30fed/config/crd/bases/quota.koordinator.sh_elasticquotaprofiles.yaml#L88

The invalid label value should be rejected during resource creation or update.

This validation can be enforced by:

defining an appropriate validation rule (e.g., a pattern constraint) in the CRD schema, or
implementing admission validation logic (such as a validating admission webhook or equivalent internal validation) to reject invalid values before the resource is persisted.

1. Install the CRD and the corresponding controller.
2. Create a CR containing an invalid value for the label-related field.
3. Apply the resource.
4. Verify that the resource is successfully created.

**Environment**:
- App version: 1.8.0
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

Aryanbhargava18:
Ⅰ. **Describe what this PR does**
This PR adds validating admission webhook logic to reject malformed keys and values in `quotaLabels` for the `ElasticQuotaProfile` CRD. It utilizes `k8s.io/apimachinery/pkg/util/validation` to enforce that keys are qualified names and values are valid label values.

Ⅱ. **Does this pull request fix one issue?**
Fixes #3116

Ⅲ. **Describe how to verify it**
1. Install the CRD and run the webhook locally or in a cluster.
2. Attempt to create an `ElasticQuotaProfile` with an invalid `quotaLabels` format (e.g. `label: test-wec1, test-wec2` or spaces in the label keys).
3. The webhook should successfully reject the resource creation, returning the specific validation error.

Ⅳ. **Special notes for reviews**
This approach validates `quotaLabels` proactively at admission time and leverages the standard `validation.IsQualifiedName` and `validation.IsValidLabelValue` utilities from `k8s.io/apimachinery`. 

V. **Checklist**

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -161,6 +161,27 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
+      path: /validate-quota-koordinator-sh-v1alpha1-elasticquotaprofile
+  failurePolicy: Fail
+  name: velasticquotaprofile.koordinator.sh
+  rules:
+  - apiGroups:
+    - quota.koordinator.sh
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - elasticquotaprofiles
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
       path: /validate-node
   failurePolicy: Ignore
   name: vnode.koordinator.sh
--- a/pkg/webhook/elasticquota/validating/profile_validating_handler.go
+++ b/pkg/webhook/elasticquota/validating/profile_validating_handler.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


Aryanbhargava18:
Hello @saintube @zwzhang0107, just bubbling this up for visibility when you have a moment. 

I've verified the webhook validation logic locally and added the necessary tests to ensure `quotaLabels` are properly enforced when modifying an `ElasticQuotaProfile`. This resolves the underlying bug where invalid quota labels could previously bypass validation.

Please let me know if you need any adjustments or further test coverage for the implementation. Thanks!

PR #3121, opened 2026-07-09
fix: use NUMA node ID instead of slice index in addPodAllocation

### LINKED ISSUE #3081: [BUG] nodenumaresource: NodeAllocation.addPod stores slice index instead of NUMA node ID
What happened:
While going through pkg/scheduler/plugins/nodenumaresource/node_allocation.go, I noticed that addPod (around lines 145-155) builds a new NUMANodeResource entry like this:

for nodeID, numaNodeRes := range request.NUMANodeResources {
    res := n.allocatedResources[numaNodeRes.Node]
    if res == nil {
        res = &NUMANodeResource{
            Node:      nodeID,   // should be numaNodeRes.Node
            Resources: make(corev1.ResourceList),
        }
        n.allocatedResources[numaNodeRes.Node] = res
    }
    res.Resources = quotav1.Add(res.Resources, numaNodeRes.Resources)
}
request.NUMANodeResources is a slice, so nodeID here is just the loop index (0, 1, 2...), not the real NUMA node ID. The map n.allocatedResources is keyed correctly by numaNodeRes.Node, but the value stored under that key gets Node: nodeID instead. So if a pod ends up allocated on NUMA nodes [1, 2] (not starting at 0), you get a map entry with key 1 whose .Node field says 0, and key 2 whose .Node field says 1.

What you expected to happen:
The .Node field on the stored struct should match the map key (numaNodeRes.Node), same as how release() and getAvailableNUMANodeResources() already use numaNodeRes.Node consistently.

How to reproduce it (as minimally and precisely as possible):

Get a pod scheduled with a NUMA allocation that doesn't start at node 0 — e.g. NUMA nodes 1 and 2 on a multi-NUMA machine.
request.NUMANodeResources will look like [{Node:1,...}, {Node:2,...}].
After addPod runs, n.allocatedResources[1].Node == 0 and n.allocatedResources[2].Node == 1 — both wrong, mismatched against the map keys.
Anything else we need to know?:
Right now nothing actually reads .Node back from this map, so there's no visible scheduling bug yet — it's more of a landmine for whoever next writes code that trusts this field (debug endpoints, metrics, logging, etc.). One-line fix: use Node: numaNodeRes.Node instead of Node: nodeID.

Environment:

App version: built from optimization/numa-hint-checking branch, commit aafae9bc
Kubernetes version (use kubectl version): n/a, code-level issue
Install details (e.g. helm install args): n/a
Node environment (for koordlet/runtime-proxy issue):
Containerd/Docker version: n/a
OS version: n/a
Kernal version: n/a
Cgroup driver: cgroupfs/systemd: n/a
Others: pkg/scheduler/plugins/nodenumaresource/node_allocation.go, (*NodeAllocation).addPod, lines 145-155

Aryanbhargava18:
Ⅰ. **Describe what this PR does**
This PR fixes a bug in `NodeAllocation.addPodAllocation` within the `nodenumaresource` scheduler plugin. 

Previously, when iterating over the pod's `NUMANodeResources` slice, the loop used the slice index (`nodeID`) rather than the actual NUMA node ID (`numaNodeRes.Node`) when constructing the new `NUMANodeResource` struct. If a pod was allocated on NUMA nodes `[1, 2]`, the resulting map entries would incorrectly store `Node: 0` and `Node: 1` as their internal IDs, despite being keyed correctly in the `n.allocatedResources` map.

This PR corrects the loop to assign `numaNodeRes.Node` to the `.Node` field, aligning it with the behavior in `release()` and preventing future bugs when this field is read by metrics or debugging endpoints.

Ⅱ. **Does this pull request fix one issue?**
Fixes #3081 

Ⅲ. **Describe how to verify it**
A new unit test `TestNodeAllocation_addPodAllocation` has been added to `pkg/scheduler/plugins/nodenumaresource/node_allocation_test.go`.

Run the unit test to verify the fix:
```bash
go test -v ./pkg/scheduler/plugins/nodenumaresource/... -run TestNodeAllocation_addPodAllocation
```

Ⅳ. **Special notes for reviewers**
None. This is a targeted, single-line bug fix.

Ⅴ. **Checklist**

### DIFF
--- a/pkg/scheduler/plugins/nodenumaresource/node_allocation.go
+++ b/pkg/scheduler/plugins/nodenumaresource/node_allocation.go
@@ -142,11 +142,11 @@ func (n *NodeAllocation) addPodAllocation(request *PodAllocation, cpuTopology *C
 		}
 	}
 
-	for nodeID, numaNodeRes := range request.NUMANodeResources {
+	for _, numaNodeRes := range request.NUMANodeResources {
 		res := n.allocatedResources[numaNodeRes.Node]
 		if res == nil {
 			res = &NUMANodeResource{
-				Node:      nodeID,
+				Node:      numaNodeRes.Node,
 				Resources: make(corev1.ResourceList),
 			}
 			n.allocatedResources[numaNodeRes.Node] = res
--- a/pkg/scheduler/plugins/nodenumaresource/node_allocation_test.go
+++ b/pkg/scheduler/plugins/nodenumaresource/node_allocation_test.go
@@ -96,6 +96,40 @@ func TestNodeAllocationAddCPUs(t *testing.T) {
 	assert.Equal(t, expectAllocatedCPUs, allocationState.allocatedCPUs)
 }
 
+func TestNodeAllocation_addPodAllocation(t *testing.T) {
+	allocationState := NewNodeAllocation("test-node-1")
+	assert.NotNil(t, allocationState)
+
+	podUID := uuid.NewUUID()
+	request := &PodAllocation{
+		UID: podUID,
+		NUMANodeResources: []NUMANodeResource{
+			{
+				Node: 1, // Start from 1 to test the slice index bug
+				Resources: corev1.ResourceList{
+					corev1.ResourceCPU: resource.MustParse("2"),
+				},
+			},
+			{
+				Node: 2,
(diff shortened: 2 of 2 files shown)


Aryanbhargava18:
cc @ZiMengSheng @buptcozy PTAL

PR #3123, opened 2026-07-09
scheduler: add OpenTelemetry tracing to koord-scheduler PreFilter phase

### LINKED ISSUE #3122: [proposal] Add OpenTelemetry distributed tracing to koord-scheduler
**What is your proposal**:

Add OpenTelemetry (OTel) distributed tracing across the koord-scheduler scheduling cycle and Koordinator's custom extension points, using the same `k8s.io/component-base/tracing` mechanism that the Kubernetes API server and kubelet already use (introduced by KEP-1454).

Today koord-scheduler has no distributed tracing. Note that the upstream kube-scheduler itself was never instrumented with tracing either (KEP-1454 covered the API server and kubelet, not the scheduler), so there is nothing to inherit from the embedded upstream scheduler. The only timing primitive in koord-scheduler is a single `k8s.io/utils/trace` call in the `RunPreFilterPlugins` method of `pkg/scheduler/frameworkext/framework_extender.go` that logs a line when a step is slower than a threshold. It does not create spans and cannot be exported to a tracing backend.

Concretely, the proposal is to:

1. Wire `k8s.io/component-base/tracing` and an OTLP exporter into the koord-scheduler startup options, reusing the same `TracingConfiguration` type that the API server and kubelet use.
2. Create a span for each scheduling cycle and propagate `context.Context` through the frameworkext extender.
3. Emit child spans for the Koordinator specific extension points: the PreFilter, Filter and Score transformers, reservation restore, reservation filtering and scoring, NUMA topology admit, and custom preemption.
4. Attach useful attributes (pod, node, plugin, reservation, result) and keep the feature gated and sampled so the overhead is negligible.

**Why is this needed**:

Most of Koordinator's value lives in extension points that are invisible to any existing tracer. When a pod takes seconds to schedule in a colocated or GPU cluster, operators have no span level breakdown of which transformer, reservation step, or plugin dominated the latency.

The Kubernetes API server and kubelet already ship OTel tracing through KEP-1454 using `k8s.io/component-base/tracing`. Bringing that same standard pattern to the scheduler closes an observability gap and fits under the umbrella issue #2851, "Align Koordinator scheduling with Kubernetes 1.35 capabilities".

Two properties make this low risk:

* It is additive. It does not change any scheduling decision, so it cannot mis-place a pod or mis-allocate a device.
* The OpenTelemetry SDK is already present in the dependency graph (pulled in transitively by other Kubernetes libraries), so no heavy new dependency is introduced.

Follow the same `k8s.io/component-base/tracing` design that the API server and kubelet use:

1. Add a tracing configuration file option and build an OpenTelemetry `TracerProvider` in the koord-scheduler command using `k8s.io/component-base/tracing`. It should be configurable for sampling and endpoint, and off by default.
2. Start a root span at the beginning of the scheduling cycle and thread `context.Context` through `frameworkExtenderImpl`.
3. Instrument the Koordinator extension points in `pkg/scheduler/frameworkext/framework_extender.go` with child spans and attributes.
4. Instrument the transformers, the reservation restore path, and custom preemption in the reservation, nodenumaresource, and deviceshare plugins.
5. Add documentation with a Jaeger or OTLP example, an end to end sanity check, and an overhead benchmark.

This can land as a series of small, reviewable PRs. The first PR sets up the tracing infrastructure and the root scheduling cycle span, and later PRs instrument the individual extension points.

PRAteek-singHWY:

This is the first PR of the OpenTelemetry tracing effort for koord-scheduler (issue fixes #3122, part of the Kubernetes 1.35 alignment umbrella #2851). It adds the tracing infrastructure and the root scheduling cycle span, without changing any scheduling behavior.

Specifically it:

* Adds a `--tracing-config-file` option to koord-scheduler and builds an OpenTelemetry `TracerProvider` through `k8s.io/component-base/tracing`. Tracing is disabled by default, and the sampling rate and OTLP endpoint are configurable, using the same mechanism the API server and kubelet use (KEP-1454).
* Starts a root span at the beginning of each scheduling attempt and propagates `context.Context` through the frameworkext extender.
* Instruments the first Koordinator specific extension point (the PreFilter transformers in `pkg/scheduler/frameworkext/framework_extender.go`) with a child span and basic attributes such as pod and node.
* Keeps the existing `k8s.io/utils/trace` log line untouched, so current behavior is preserved.

When tracing is disabled, the added code path uses a no-op span, so there is no measurable overhead and no change to scheduling results.

Part of #3122. This is the first PR of the series and does not close the issue on its own.

1. Run the unit tests:

   ```bash
   make test
   ```

2. Run koord-scheduler locally with tracing enabled, pointed at a local OTLP collector (for example Jaeger running with the OTLP receiver on port 4317):

   ```bash
   ./koord-scheduler --tracing-config-file=/path/to/tracing.yaml
   ```

3. Schedule a test pod, then open the Jaeger UI and confirm that a scheduling cycle span appears for the pod, with a child span for the PreFilter transformers and the expected pod and node attributes.

4. Restart koord-scheduler with tracing disabled (the default) and confirm scheduling works exactly as before and no spans are exported.

* Tracing is off by default and gated. When it is disabled, the extension point uses a no-op span, so the hot path overhead is negligible.
* The scheduler was never instrumented with tracing upstream (KEP-1454 covered the API server and kubelet). This PR reuses their exact `k8s.io/component-base/tracing` mechanism so the configuration surface and behavior are familiar to Kubernetes users.
* This PR only sets up the infrastructure and one instrumented extension point. Instrumentation of the remaining extension points (reservation restore, reservation filtering and scoring, NUMA topology admit, custom preemption) will follow in separate PRs to keep each change small and reviewable.

### V. Checklist

- [x] I have written necessary code comments (user-facing docs will follow in a later PR of the series)
- [x] Verified `go build`, `go vet`, and `go test` pass for the affected packages


PRAteek-singHWY:
Hi @saintube @ZiMengSheng, this is the first PR for OpenTelemetry tracing in koord-scheduler (#3122, under the k8s 1.35 alignment umbrella #2851). It sets up the tracing infrastructure and a span for the PreFilter transformer phase, disabled by default and a no-op when off, so there is no behavior change. I would love your feedback on the approach before I instrument the remaining extension points. Thanks!

pkg/scheduler/frameworkext/framework_extender.go
```diff
@@ -299,6 +306,17 @@ func (ext *frameworkExtenderImpl) GetWorkloadAuditor() workloadauditor.WorkloadA
 func (ext *frameworkExtenderImpl) RunPreFilterPlugins(ctx context.Context, cycleState fwktype.CycleState, pod *corev1.Pod) (*fwktype.PreFilterResult, *fwktype.Status, sets.Set[string]) {
 	trace := utiltrace.New("RunPreFilterPluginTransformers", utiltrace.Field{Key: "namespace", Value: pod.Namespace}, utiltrace.Field{Key: "name", Value: pod.Name})
 	defer trace.LogIfLong(5 * time.Millisecond)
+
+	// Emit an OpenTelemetry span for the PreFilter transformer phase. When no tracer
+	// provider is configured, the global provider is a no-op and this adds negligible
+	// overhead. Reassigning ctx lets downstream extension points nest under this span.
+	ctx, span := otel.GetTracerProvider().Tracer(tracingInstrumentationScope).Start(ctx, "RunPreFilterPluginTransformers",
```
saintube: Why not in the scheduler_monitor?

PRAteek-singHWY: Hey @saintube 
Good point, placing it in scheduler_monitor makes sense, since it already brackets the whole scheduling attempt (StartMonitoring → Complete) and owns the timing/metrics, so the root span belongs there rather than inline in the PreFilter phase. Moved the span into SchedulerMonitor and reverted the inline one in framework_extender.go. Fixed.
Looking forward for your review and happy to make any iterations as required.
Thank You.

ZiMengSheng (commented):
Thanks a lot for kicking this off and welcome to Koordinator! The overall direction — reusing `k8s.io/component-base/tracing` and threading `context.Context` through the extender — is exactly right, and splitting the work into a series of small PRs is appreciated. A few things I'd like to see addressed before we merge this first PR, plus some non-blocking suggestions.

## Must-fix before merge

**1. PR description doesn't match the code (blocker on the description or the code, your choice).**

The PR description says:

> Instruments the first Koordinator specific extension point (the PreFilter transformers in `pkg/scheduler/frameworkext/framework_extender.go`) with a child span and basic attributes such as pod and node.

But the diff to `pkg/scheduler/frameworkext/framework_extender.go` is a single blank line — `RunPreFilterPlugins` still has no `Start(ctx, "PreFilterTransformers", ...)` call. The only real instrumentation lives in `scheduler_monitor.go`, which is the **root** `SchedulingCycle` span, not a PreFilter child span. Please either:

- narrow the PR title/description to "root scheduling cycle span" (and rename the branch/PR title accordingly), or
- actually add the PreFilter transformer child span to `RunPreFilterPlugins` so it matches the description.

I'd lean toward the first option (keep this PR minimal, add child spans in follow-up PRs), but either is fine as long as the description matches what ships.

**2. Span lifecycle leak on timeout and re-entry paths.**

`state.span` is only ended in `Complete()`. Two paths bypass `Complete()` today, so the span never ends and the OTel SDK internal buffers grow:

- **Timeout path**: `monitor.monitor()` → `CompleteMonitor` logs `!!!CRITICAL TIMEOUT!!!` for pods that exceed `schedulingTimeout`, but does not call `Complete()`. The span for that attempt is leaked.
- **Re-entry path**: in `StartMonitoring`, when `scheduleState, exists := m.schedulingPods[pod.UID]` returns `exists == true`, we assign `scheduleState.span = span` unconditionally and overwrite the previous span. The old span reference is dropped without `End()` being called.

Suggested fix:

```go
// In StartMonitoring, before overwriting:
if exists && scheduleState.span != nil {
    scheduleState.span.SetStatus(codes.Error, "scheduling attempt superseded")
    scheduleState.span.End()
}
scheduleState.span = span
```

And in `monitor.monitor()`, when a timeout is detected, end the span with an error status before (or alongside) the `!!!CRITICAL TIMEOUT!!!` log so it isn't leaked.

**3. `span.End()` called while holding the monitor lock.**

In `Complete()`:

```go
m.lock.Lock()
...
if state.span != nil {
    state.span.SetStatus(...)
    state.span.End()      // called under m.lock
}
...
m.lock.Unlock()
```

`End()` is normally cheap but it hands the span off to the exporter's batch processor, which can allocate/queue. Since this is on the scheduling hot path, please lift the span out of the state, unlock, then call `End()` outside the critical section — keeps the lock as tight as it is today.

## Non-blocking suggestions

**4. `WorkloadAuditor.RecordAttemptPod(pod)` doesn't accept `ctx`.**

`scheduleOne` now propagates a scheduling-cycle context, but the very next call — `f.workloadAuditor.RecordAttemptPod(pod)` — still takes only `pod`, so the auditor cannot nest spans under the attempt span later. Either:

- change the signature to `RecordAttemptPod(ctx, pod)` in this PR for consistency, or
- add a note in the PR description saying it's intentionally deferred to a follow-up PR.

**5. Formatting churn in `Config`.**

In `cmd/koord-scheduler/app/config/config.go`, adding `TracingConfigFile` reflowed the existing `InsecureServing` line from the aligned multi-line block to a single-space form. This mixes a semantic change with a whitespace change and makes the diff noisier than it needs to be. Please restore the original alignment for the untouched fields.

**6. `service.name` is hard-coded to `"koord-scheduler"`.**

Reasonable choice. When we eventually run multiple shards / replicas, consider also attaching `service.instance.id` (e.g. `os.Hostname()`) so operators can distinguish instances in Jaeger. Optional for this PR.

**7. No sample `tracing.yaml` shipped.**

The PR verification steps say `--tracing-config-file=/path/to/tracing.yaml` but the repo has no example. A minimal file under `hack/` or `docs/` (with the `endpoint` and `samplingRatePerMillion` fields you've already validated in the test) would make it much easier for reviewers and users to try tracing locally. Full user docs can still come in a later PR.

---

Overall: really nice starting PR — setup + tests are clean, and the choice to reuse `k8s.io/component-base/tracing` is exactly what we want. Once the description matches the code, the span lifecycle holes are closed, and `End()` moves outside the lock, this looks good to me. Happy to re-review as soon as those are addressed. Thanks again for taking this on!

PRAteek-singHWY:
Hi @ZiMengSheng

Thank you for the really detailed review. This was exactly the kind of feedback I was hoping for on a first contribution here, and it caught a few things I'd have otherwise missed. I've pushed commits that work through all of it.

**1. Description mismatch.** I went with your second option and added the real instrumentation rather than watering down the description, since the PreFilter phase is the whole point of the PR. `RunPreFilterPlugins` now opens a `PreFilterTransformers` child span under the `SchedulingCycle` root, and the derived context flows down so the native plugins nest underneath it. One small liberty I took with the wording: I set only `pod.namespace` / `pod.name`, not a node attribute, because at PreFilter there's no chosen node yet, so a node field would just be empty. Node attributes will fit better on the Filter/Score spans later. I also added a unit test that asserts the span and its attributes are emitted.

**2. Lifecycle leaks.** Closed, and I pulled the termination logic into a small `endSpan` helper so every path is consistent. The superseded re-entry in `StartMonitoring` and the pod-deleted case in `RecordNextPod` both end the span with an error status now. On the timeout path, I'll be honest about how I got there. I first tried keeping the timed-out attempt tracked and letting `Complete` stay the single owner that ends the span. That actually worked fine in my testing, since `Complete` is reached in the normal flow, but it leaves the leak open exactly in the case you called out, where the attempt never reaches `Complete`. So I followed your suggestion and made `monitor()` end the span with an error status when the timeout fires, right alongside the `!!!CRITICAL TIMEOUT!!!` log. To keep your point-3 fix intact I end it outside the monitor lock, and I clear the stored span so `Complete` doesn't end it a second time. There's a unit test for the timeout path too.

**3. `End()` under the lock.** Fixed. `Complete()` grabs the state, deletes, unlocks, and only then ends the span, so the critical section stays as tight as before.

**4. `RecordAttemptPod`.** Deferred on purpose, and I called that out in the description. I'd rather thread the context in the same PR that actually instruments the auditor, so the signature change lands next to the code that uses it.

**5-7. Smaller items.** Done too. I restored the `Config` alignment so the diff is purely additive (5), added `service.instance.id` via `os.Hostname()` as a best-effort attribute (6), and shipped a sample `docs/koord-scheduler/tracing.yaml` with the `endpoint` and `samplingRatePerMillion` fields so it's easy to try locally (7).

`go build`, `go vet`, and `go test` all pass for the affected packages. Thanks again for taking the time on this. I really appreciate it, and I'm happy to iterate further on anything.

Thank you.

PRAteek-singHWY:
Hi @ZiMengSheng @saintube, gentle ping on this one.

All the review feedback was addressed in `23ea6f9f` and `16475b23` (PreFilter child span with a test, span lifecycle closed on the superseded / pod-deleted / timeout paths, `End()` moved outside the monitor lock, `service.instance.id`, `Config` alignment restored, sample `docs/koord-scheduler/tracing.yaml`). The branch is rebased on current `main` and CI is green, unit tests, build, govet, gofmt, golangci-lint and codecov all pass.

Summary of what changed since your review, so it's easy to spot-check:

| Review item | Where |
| --- | --- |
| 1. PreFilter child span (description/code mismatch) | `RunPreFilterPlugins` in `pkg/scheduler/frameworkext/framework_extender.go` |
| 2. Span leak on re-entry / pod-deleted / timeout | `endSpan` helper + `StartMonitoring`, `RecordNextPod`, `monitor()` in `pkg/scheduler/frameworkext/scheduler_monitor.go` |
| 3. `End()` called under the monitor lock | `Complete()`,  state lifted out, unlock, then end |
| 4. `RecordAttemptPod(ctx, pod)` | intentionally deferred to the follow-up PR that instruments the auditor, noted in the description |
| 5. `Config` formatting churn | original alignment restored, diff is purely additive |
| 6. `service.instance.id` | added via `os.Hostname()`, best-effort |
| 7. Sample tracing config | `docs/koord-scheduler/tracing.yaml` with `endpoint` and `samplingRatePerMillion` |

Two open questions where I'd value a decision from you, since they shape the follow-up PRs:

1. **Span naming and attribute conventions.** I'm using `SchedulingCycle` as the root span and `PreFilterTransformers` as the child, with `pod.namespace` / `pod.name` attributes. If Koordinator wants to settle on a convention (for example a `koordinator.sh/` attribute prefix, or upstream's `k8s.pod.*` naming), I'd rather adopt it now than rename spans across five follow-up PRs.
2. **Ordering of the remaining extension points.** My plan is reservation restore → reservation filter/score → NUMA topology admit → custom preemption, one PR each. Happy to reorder if some of those are more useful to you first.

No rush, just want to make sure this isn't blocked on anything from my side. If it would help, I'm also glad to split the `SchedulerMonitor` lifecycle changes into their own PR so the tracing infrastructure can land independently.

ZiMengSheng:
Reviewed `5b5df5a4`. All seven items from my earlier round are genuinely addressed — I re-read each one against the code rather than trusting the summary table, and the span lifecycle work in `scheduler_monitor.go` is careful: `endSpan` is uniformly called outside `m.lock`, the timeout path clears the stored span so `Complete` can't double-end it, and the tests pin the superseded / pod-deleted / timeout paths. Locally on your head commit:

```
ok  github.com/koordinator-sh/koordinator/cmd/koord-scheduler/app     0.928s
ok  github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext  3.006s
```

Two things I want to credit explicitly, because they're the reason this is reviewable at all:

- **Zero new dependencies.** The `go.mod` diff is `+3/-3`: `otel`, `otel/sdk`, `otel/trace` move from `// indirect` to direct at the same `v1.41.0` they were already pinned at via `component-base`. No `go.sum` change. For a "new observability subsystem" PR that is unusually cheap.
- **Reusing `k8s.io/component-base/tracing`** instead of hand-rolling provider setup means we inherit `ParentBased` sampling, the batch processor, and the same config schema operators already know from the API server.

Below are three findings and then answers to your two open questions. **None of the three blocks merge on its own**, but I'd like #1 and #2 fixed here rather than deferred, because both get harder to change once five follow-up PRs copy the pattern.

---

### 1. `RecordNextPod` silently drops a live span

`RecordNextPod` builds a **fresh** `podScheduleState` and overwrites the map entry:

```go
scheduleState := podScheduleState{ namespace: ..., name: ..., dequeued: now, ... }
RecordQueuePodInfo(podInfo, &scheduleState)
m.lock.Lock()
m.schedulingPods[pod.UID] = scheduleState   // span field is the zero value: nil
m.lock.Unlock()
```

If a span is stored for that UID at this moment, it is neither ended nor reachable afterwards — `monitor()` can't find it, because the new `state.start` is zero and `isPodUnhandledExceedingTimeout` sends that entry down the skip path. The span is leaked, and the failure is invisible: a leaked span is simply a trace that never appears in the backend, with no log line and no metric.

I tried to talk myself out of this one. In today's flow `Complete` is reached on both terminal paths (`RunPostBindPlugins` on success, the wrapped `FailureHandler` on failure), so `StartMonitoring` → re-pop-without-`Complete` is hard to hit deliberately. But that argument is exactly the one that *didn't* hold for the three paths you just fixed, and the fix here is the same shape you already wrote:

```go
m.lock.Lock()
previous, ok := m.schedulingPods[pod.UID]
m.schedulingPods[pod.UID] = scheduleState
m.lock.Unlock()
if ok {
    endSpan(previous.span, codes.Error, "scheduling attempt requeued")
}
```

Worth doing now: this is the last remaining place where the `span` field is dropped instead of transferred, and once the invariant is "every write to `schedulingPods` either carries the span forward or ends it", the follow-up PRs can rely on it.

### 2. The hot path resolves the tracer from the global provider on every call

Both call sites do `otel.GetTracerProvider().Tracer(scope).Start(...)`. The PR description says this is "a no-op when off" and "negligible overhead"; that's nearly true but not quite, and the gap is measurable. `internal/global.tracerProvider.Tracer()` takes a **process-global mutex unconditionally**, including on the delegate-installed path:

```go
func (p *tracerProvider) Tracer(name string, opts ...trace.TracerOption) trace.Tracer {
	p.mtx.Lock()
	defer p.mtx.Unlock()
	if p.delegate != nil {
		return p.delegate.Tracer(name, opts...)
	}
	...
```

whereas `tracer.Start` itself is lock-free (`t.delegate.Load()`). I benchmarked your exact call shape with **tracing disabled** (no provider installed), `go.opentelemetry.io/otel v1.41.0`, darwin/arm64:

```
BenchmarkTracerPerCall-15            170.6 ns/op    408 B/op    6 allocs/op   # as written today
BenchmarkTracerCached-15             137.7 ns/op    408 B/op    6 allocs/op   # tracer resolved once
BenchmarkTracerPerCallParallel-15    211.6 ns/op    240 B/op    3 allocs/op
BenchmarkTracerCachedParallel-15      75.5 ns/op    240 B/op    3 allocs/op   # 2.8x under contention
BenchmarkTracerGatedDisabled-15        0.2824 ns/op   0 B/op    0 allocs/op   # skipped entirely
```

Two separate costs. The `Tracer()` lookup is ~33 ns serial and ~2.8x under contention. And the `oteltrace.WithAttributes(attribute.String(...), attribute.String(...))` argument is built *before* `Start` is entered, so **6 allocations / 408 B per call are paid even when tracing is off** — about 12 allocs and 800 B of pure garbage per scheduling cycle across your two call sites. Not a latency problem against a multi-millisecond cycle, but it is allocation on the scheduler's hottest path in exchange for nothing when the feature is disabled, which is not what "no-op" implies.

**Recommendation: hold the tracer as a field** on `frameworkExtenderImpl` and `SchedulerMonitor`, resolved once and injectable. That fixes this *and* finding #3 at the same time.

One trap to avoid: do **not** use a package-level `var tracer = otel.GetTracerProvider().Tracer(...)`. `global.SetTracerProvider` propagates to already-created tracers under a `sync.Once` (`delegateTraceOnce`), so a package-init-cached tracer is permanently pinned to whichever provider is installed first — fine in production where `setupTracing` runs once, but it would break any test that swaps providers. A struct field sidesteps it.

### 3. `defer otel.SetTracerProvider(previous)` does not restore anything

Both new tracing tests do:

```go
previous := otel.GetTracerProvider()
otel.SetTracerProvider(tp)
defer otel.SetTracerProvider(previous)
```

That restore is a no-op, and I verified it rather than assuming — a throwaway test that performs the set/restore and then starts a span through `otel.GetTracerProvider()`:

```
spans recorded AFTER restore: 1
  leaked span: AfterRestore
=> restore is INEFFECTIVE: the test provider stays installed for the rest of the package
```

The reason is the same `delegateTraceOnce`: `previous` is the *delegating* global provider whose delegate was already wired to `tp` by the first call, so restoring it restores a provider that still forwards to your recorder. Your recorder stays live for every subsequent test in `pkg/scheduler/frameworkext`.

Nothing fails today, since no other test asserts on spans. But `assert.Empty(t, recorder.Ended(), "span should not be ended before Complete")` is precisely the kind of assertion that starts flaking once a second tracing test exists. Injecting the provider (finding #2) removes the need to touch global state in tests at all.

---

### Answers to your two open questions

**Q1 — span naming and attribute conventions. Adopt OTel semantic conventions for the pod identity attributes; keep your span names.**

`pod.namespace` / `pod.name` are custom keys that no tracing backend knows. Use the semconv constants, which are already in the module graph at the version you depend on — `go.opentelemetry.io/otel/semconv/v1.37.0` — so this costs nothing:

| today | use instead |
|---|---|
| `attribute.String("pod.namespace", …)` | `semconv.K8SNamespaceName(…)` → `k8s.namespace.name` |
| `attribute.String("pod.name", …)` | `semconv.K8SPodName(…)` → `k8s.pod.name` |
| — | `semconv.K8SPodUID(…)` → `k8s.pod.uid`, worth adding; name+namespace is not a stable key across recreation |
| for later Filter/Score spans | `semconv.K8SNodeName(…)` → `k8s.node.name` |

Reserve a `koordinator.` prefix for things with no semconv equivalent (`koordinator.reservation.name`, `koordinator.quota.name`, and so on). Do **not** use `koordinator.sh/` — dots are the OTel attribute separator and a `/` in a key is awkward in several backends.

Span names: keep `SchedulingCycle` as the root, and name each child after the extension point exactly as the framework calls it — `PreFilterTransformers` is good because it says *transformers*, distinguishing our work from the native `PreFilter` plugins nested beneath it. Keep names static and low-cardinality (never interpolate pod or node names into a span name).

**Q2 — ordering of the remaining extension points. Reservation restore first.** Suggested order:

1. **`RunReservationExtensionRestoreReservation`** — highest value by a wide margin. It is per-node work inside PreFilter and it is what people actually report as slow on large reservation counts. This is the span that will justify the whole series.
2. **NUMA topology admit** — second, because it is per-node in Filter and its cost varies with topology in ways we currently can't see.
3. **Reservation filter/score.**
4. **Custom preemption** — last; it is rare per cycle, so it is the least useful place to spend a span budget early.

A useful heuristic for picking the next one: **wherever the code already has a `utiltrace.New(...)` + `LogIfLong`, someone already suspected that spot was slow.** Right now `pkg/scheduler/frameworkext` has exactly one — `RunPreFilterPluginTransformers`, the very site you instrumented. That's a good sign you picked the right first target.

Which raises a policy question I'd rather settle now than after five PRs: **we will have two tracing mechanisms.** Your OTel span and the existing `utiltrace` now wrap the same block, duplicating the same two fields. I don't want them to diverge. My preference: leave `utiltrace` alone in this PR (it is the only thing that works with tracing disabled, and it logs unconditionally on slowness), but as you add spans, put the OTel span *at the same boundary* as any existing `utiltrace`, with matching attributes, so a later PR can retire `utiltrace` mechanically. Please don't add new `utiltrace` calls in the follow-ups.

---

On your offer to split the `SchedulerMonitor` lifecycle changes into their own PR: not necessary. The span field and its lifecycle are inseparable from the root span, and splitting would land a `span` field that nothing sets. Keep it together.

Holding off on `/lgtm` only for findings #1 and #2 — both are small and local. Ping me when they're pushed and I'll re-read just those two. Thanks for the persistence across three rounds on this; the lifecycle reasoning in your last comment (including saying plainly which approach you tried first and why you abandoned it) made this much faster to review.

cc @saintube — the two convention decisions above are the parts worth a second opinion before the follow-up PRs commit to them.

ZiMengSheng:
Addendum — one more non-blocking item I found while reading `component-base` to check your provider setup, which I'd rather add than leave out.

**A mis-typed or minimal tracing config file logs success and then traces nothing.**

Two things compose badly:

1. `loadTracingConfiguration` uses plain `yaml.Unmarshal` (non-strict), so any unknown or misspelled key is silently discarded. `samplingRatePerMillion` typo'd → the field stays `nil`.
2. `tracing.NewProvider` then does:

```go
sampler := sdktrace.NeverSample()
if tracingConfig.SamplingRatePerMillion != nil && *tracingConfig.SamplingRatePerMillion > 0 {
    sampler = sdktrace.TraceIDRatioBased(...)
}
... sdktrace.WithSampler(sdktrace.ParentBased(sampler))
```

`ParentBased(NeverSample())` means **no root span is ever sampled**. Since `SchedulingCycle` is always a root span in the scheduler, the result is a fully wired provider, an OTLP exporter dialing the collector, and exactly zero spans — while the log line reads:

```
Initialized OpenTelemetry tracing for koord-scheduler  configFile=/etc/koord/tracing.yaml
```

An operator debugging "I enabled tracing and see nothing" has no signal to work from. Two cheap fixes, either is fine:

- Log the effective values rather than just the path: `klog.InfoS("Initialized OpenTelemetry tracing", "configFile", configFile, "endpoint", endpoint, "samplingRatePerMillion", rate)`, resolving `nil` to `0` explicitly — then the zero is visible.
- And/or warn when the rate resolves to 0: at that point tracing is enabled in name only.

Also a small doc-accuracy point while you're in there. The comment in `docs/koord-scheduler/tracing.yaml` says the schema is "the same one used by the API server and kubelet (KEP-1454)". The *fields* are, but the API server's `--tracing-config-file` is decoded through `runtime.DecodeInto` with a versioned codec and therefore **requires** an `apiVersion: apiserver.config.k8s.io/v1beta1` / `kind: TracingConfiguration` envelope, which your sample file omits and your non-strict `yaml.Unmarshal` would silently ignore if present. Worth saying "the same *fields* as ..., without the apiVersion/kind envelope" so nobody copies a working API server config across and wonders why it behaves differently.

Neither of these changes behaviour when the config is correct, so treat them as part of the same round as findings #1 and #2 if convenient, or a follow-up if not.

PR #3125, opened 2026-07-10
feat: share NodeDeviceCache across scheduler profiles

mdryaaan:

Koordinator's out-of-tree plugins build their cache once per scheduling profile, so with N profiles DeviceShare's `nodeDeviceCache` is instantiated N times — N handler sets, N GC goroutines, N assume-window views — all tracking the same cluster-wide state. This PR introduces an opt-in `SharedPluginCache` mechanism in `frameworkext` and migrates DeviceShare onto it, so the cache is created exactly once per scheduler instance and shared across all profiles.

The framework extender owns the lifecycle: plugins register a cache from `New()` via `ExtendedHandle.GetOrRegisterSharedCache(key, create)` (exactly-once by key); after all profiles are built, `FrameworkExtenderFactory.StartSharedCaches` invokes `Start(ctx)` on each cache once and wires a single unified pod/node event dispatcher that fans out to every registered cache in registration order. Plugin-specific CRD handlers (DeviceShare's `Device` handler) and the GC goroutine move into `Start()`; per-profile hooks (the `ForgetPod` handler, the node informer) are registered in `New()`, since they are per-extender state.

A `CacheReserver` optional interface adds the `AssumePod`/`ForgetPod` assume/forget contract. `AssumePod` records what `Reserve` wrote as a **rollback marker**. The marker is a pure rollback record: it is consumed only by a **negative** event — `ForgetPod`, or a delete / bound→unassigned / terminate through `OnPodDelete`/`OnPodUpdate` — which rolls back the recorded snapshot. A **positive** (assigned) event for an assumed pod is **not** re-added or reconciled: `spec.nodeName` is not a trustworthy bind signal (multi-scheduler arbitration fakes it via a client-go transform and clears it back to `""`), and the marker must survive for a racing `ForgetPod`. The pre-existing `nodeDevice.isValid` guard keeps the informer-add idempotent with the `Reserve` write. The mechanism is opt-in — plugins that do not implement `SharedPluginCache` are unaffected.

Part of #3095 #2749 — this is the first implementation PR (DeviceShare portion of Phase 1) following the approved design proposal. NodeNUMAResource and Reservation cache migrations follow in subsequent PRs.

Scoped to the changed packages:

```bash
go build ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... ./cmd/koord-scheduler/...
go vet   ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... ./cmd/koord-scheduler/...
go test  ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... -count=1
go test  ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... -race -count=1
./bin/golangci-lint run ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/...
```

All green locally, including `-race` and golangci-lint (0 issues). Coverage:

- **Multi-profile sharing:** two profiles built from one factory share a single `nodeDeviceCache` instance, created exactly once; and the `ForgetPod` handler reaches the shared cache when invoked through a *second* profile's extender (not only the one that constructed the cache).
- **Registry + dispatcher:** same-key/distinct-key semantics; `StartSharedCaches` starts each cache once and is idempotent; dispatcher fans out to all caches in registration order across all six pod/node event types; tombstone unwrapping; registering a new cache after start panics.
- **Assume/forget marker model** (`device_cache_assume_test.go`): AssumePod snapshot atomicity + empty-marker skip + missing-node error; ForgetPod idempotency; positive events do not re-add or clear the marker; negative events (delete / unassign / terminate / framework ForgetPod) roll back the snapshot and clear the marker, idempotent across the async ForgetPod↔unassign pair; delete-before-bind rolls back via the snapshot without relying on annotations; reservation synthetic delete/terminate release the marker.
- **Production dispatch path:** a pre-existing bound pod carrying an allocation annotation is populated into the shared cache through the informer initial-list → dispatcher → `OnPodAdd` path.
- **Startup ordering:** no pod/node event is observed by a cache before its `Start()` returns.
- **Race:** concurrent Reserve/AssumePod + informer events (incl. the framework ForgetPod vs unassign pair) under `-race`, with a no-double-count invariant.

Existing DeviceShare and frameworkext unit tests pass unchanged.

This PR was opened as 7 bisectable commits and has grown to 19 across review rounds 1–8; each commit still builds and passes independently. High-level structure:

- **New mechanism (frameworkext):** the `SharedPluginCache` / `CacheReserver` interfaces; the registry + `StartSharedCaches` + unified pod/node dispatcher on `FrameworkExtenderFactory` (lock-free `atomic.Pointer` snapshot on the per-event path); the `ExtendedHandle` accessor; the startup wiring in `cmd/koord-scheduler` (in `Run()`, after all plugins are built and before the shared informer factory starts, with an ordering guard).
- **DeviceShare migration:** `nodeDeviceCache` implements both interfaces; `New()` becomes profile-independent and registers the per-profile hooks; the Device/Reservation CRD handlers and GC (behind an `AfterAllInformersSynced` hook) move into `Start()`; `Reserve`/`Unreserve` use `AssumePod`/`ForgetPod`. Existing DeviceShare tests were adapted in the same commits as the signature/handler changes so no intermediate commit fails to build.
- **Assumed-ledger semantics** were refined across rounds into the rollback-marker model described in §Ⅰ; `AssumePod` snapshots and publishes the marker under a single cache-lock hold; reservation synthetic events and real pod events share one `OnPod*` code path.

Other notes:
- The no-double-count property leans on the pre-existing `nodeDevice.isValid` guard (skips an add when the pod is already in `allocateSet` after Reserve, and skips a subtract when it is already gone), unchanged by this PR.
- Per-node `nodeDevice.lock` is preserved. The per-event handler path takes no cache-wide write lock; the per-Reserve/Unreserve `AssumePod`/`ForgetPod` calls do hold `nodeDeviceCache.lock` briefly.
- Out of scope here (follow-ups): `gpuSharedResourceTemplatesCache` (also per-profile), NodeNUMAResource + Reservation migrations, the dispatcher's configurable parallel mode + per-handler metrics, and re-scoping the assumed-pods observability now that the marker is lifetime-scoped.

### V. Checklist

### DIFF
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -553,7 +553,9 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions 
 	// StartSharedCaches wires the unified pod/node dispatcher on cc.InformerFactory and
 	// invokes Start on every SharedPluginCache registered by plugins during New(). Must
 	// run before cc.InformerFactory.Start() so no event fires before its handler is wired.
-	frameworkExtenderFactory.StartSharedCaches(ctx, cc.InformerFactory)
+	if err := frameworkExtenderFactory.StartSharedCaches(ctx, cc.InformerFactory); err != nil {
+		return nil, nil, nil, nil, err
+	}
 	workloadauditor.AddEventHandler(sched, workloadAuditor, cc.InformerFactory, cc.KoordinatorSharedInformerFactory)
 	reservationErrorHandler := eventhandlers.MakeReservationErrorHandler(
 		sched,
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -550,6 +550,10 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions 
 	schedAdapter := frameworkExtenderFactory.Scheduler()
 
 	eventhandlers.AddScheduleEventHandler(sched, schedAdapter, cc.InformerFactory, cc.KoordinatorSharedInformerFactory, crossSchedulerNominator)
+	// StartSharedCaches wires the unified pod/node dispatcher on cc.InformerFactory and
+	// invokes Start on every SharedPluginCache registered by plugins during New(). Must
+	// run before cc.InformerFactory.Start() so no event fires before its handler is wired.
+	frameworkExtenderFactory.StartSharedCaches(ctx, cc.InformerFactory)
 	workloadauditor.AddEventHandler(sched, workloadAuditor, cc.InformerFactory, cc.KoordinatorSharedInformerFactory)
 	reservationErrorHandler := eventhandlers.MakeReservationErrorHandler(
 		sched,
--- a/cmd/koord-scheduler/app/server.go
+++ b/cmd/koord-scheduler/app/server.go
@@ -241,6 +241,16 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
 	}
 
 	startInformersAndWaitForSync := func(ctx context.Context) error {
+		// Start the opt-in shared plugin caches before any informer factory: this wires the
+		// unified pod/node dispatcher on cc.InformerFactory and invokes Start on every
+		// SharedPluginCache registered by plugins during New(). Doing it here (rather than in
+		// Setup) keeps all runtime startup in Run, and runs it in both the immediate and the
+		// leader-election paths (StartSharedCaches is idempotent). It must precede
+		// cc.InformerFactory.Start() below so no event fires before its handler is wired.
+		if err := extenderFactory.StartSharedCaches(ctx, cc.InformerFactory); err != nil {
+			return err
+		}
+
 		// Startup order matters for data-race freedom: some plugins register
 		// AfterPluginInformersSynced hooks (via frameworkexthelper) that rebuild
 		// internal state from an initial-list snapshot of their private informers
@@ -272,6 +282,14 @@ func Run(ctx context.Context, cc *schedulerserverconfig.CompletedConfig, sched *
 		}
 
 		// Step 3: start the remaining informer factories.
+		// Defensive invariant: the shared plugin caches (and their unified pod/node
+		// dispatcher registered on cc.InformerFactory) must be started before this factory
+		// starts, or events would be missed and shared caches read while unpopulated — the
+		// regression class this PR guards against. In the current structure StartSharedCaches
+		// above always runs first; this check fails loudly if a future edit ever reorders it.
+		if !extenderFactory.SharedCachesStarted() {
+			return fmt.Errorf("shared plugin caches must be started before the main informer factory (StartSharedCaches not called)")
+		}
 		cc.InformerFactory.Start(ctx.Done())
 		// DynInformerFactory can be nil in tests.
 		if cc.DynInformerFactory != nil {
@@ -550,12 +568,6 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions 
 	schedAdapter := frameworkExtenderFactory.Scheduler()
 
 	eventhandlers.AddScheduleEventHandler(sched, schedAdapter, cc.InformerFactory, cc.KoordinatorSharedInformerFactory, crossSchedulerNominator)
-	// StartSharedCaches wires the unified pod/node dispatcher on cc.InformerFactory and
-	// invokes Start on every SharedPluginCache registered by plugins during New(). Must
-	// run before cc.InformerFactory.Start() so no event fires before its handler is wired.
-	if err := frameworkExtenderFactory.StartSharedCaches(ctx, cc.InformerFactory); err != nil {
-		return nil, nil, nil, nil, err
-	}
 	workloadauditor.AddEventHandler(sched, workloadAuditor, cc.InformerFactory, cc.KoordinatorSharedInformerFactory)
 	reservationErrorHandler := eventhandlers.MakeReservationErrorHandler(
 		sched,
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -628,6 +628,10 @@ func (f *FrameworkExtenderFactory) SharedCachesStarted() bool {
 // goal is parallel-by-default dispatch with an opt-in serial mode, so the shared dispatcher
 // does not regress event-handling latency compared to the original per-profile handler
 // pattern. Deferred to the unified event-dispatcher phase (PR-3).
+//
+// TODO: instrument per-handler dispatch latency (e.g. a histogram labeled by cache key and
+// event type) so the dispatcher's handler performance is observable and regressions are
+// caught once serial/parallel execution modes land. Deferred to the same PR-3 phase.
 func (f *FrameworkExtenderFactory) dispatchCaches() []SharedPluginCache {
 	if p := f.startedCaches.Load(); p != nil {
 		return *p
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -561,40 +561,40 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 		f.sharedCachesMu.Unlock()
 		return nil
 	}
-	f.sharedCachesStarted = true
 	caches := make([]SharedPluginCache, 0, len(f.sharedCachesOrder))
 	for _, key := range f.sharedCachesOrder {
 		caches = append(caches, f.sharedCaches[key])
 	}
 	f.sharedCachesMu.Unlock()
 
-	if len(caches) == 0 {
-		return nil
+	if len(caches) > 0 {
+		// Register the unified pod/node dispatchers before committing any state, so a
+		// registration failure leaves the factory un-started (sharedCachesStarted stays
+		// false, no snapshot published) and returns an error for the caller to fail fast
+		// on, rather than a partially-started state that later calls would skip.
+		if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    f.dispatchPodAdd,
+			UpdateFunc: f.dispatchPodUpdate,
+			DeleteFunc: f.dispatchPodDelete,
+		}); err != nil {
+			return fmt.Errorf("failed to register shared cache pod event handler, err: %w", err)
+		}
+		if _, err := informerFactory.Core().V1().Nodes().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    f.dispatchNodeAdd,
+			UpdateFunc: f.dispatchNodeUpdate,
+			DeleteFunc: f.dispatchNodeDelete,
+		}); err != nil {
+			return fmt.Errorf("failed to register shared cache node event handler, err: %w", err)
+		}
+		// Publish the immutable snapshot for lock-free dispatch. Events cannot arrive until
+		// the informer factory is started (after this returns), so it is safe to publish
+		// after the handlers are registered.
+		f.startedCaches.Store(&caches)
 	}
 
-	// Publish the immutable snapshot before wiring the dispatcher so events (which cannot
-	// arrive until the informer factory is started, after this returns) always read it.
-	f.startedCaches.Store(&caches)
-
-	// Register unified pod/node dispatchers before invoking Start so plugin-specific
-	// CRD handlers registered inside Start observe the same "handlers wired before
-	// factory started" invariant. A registration failure means the dispatcher is not
-	// wired and the caches would never receive pod/node events, so fail fast rather than
-	// starting into a silently broken state.
-	if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    f.dispatchPodAdd,
-		UpdateFunc: f.dispatchPodUpdate,
-		DeleteFunc: f.dispatchPodDelete,
-	}); err != nil {
-		return fmt.Errorf("failed to register shared cache pod event handler, err: %w", err)
-	}
-	if _, err := informerFactory.Core().V1().Nodes().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    f.dispatchNodeAdd,
-		UpdateFunc: f.dispatchNodeUpdate,
-		DeleteFunc: f.dispatchNodeDelete,
-	}); err != nil {
-		return fmt.Errorf("failed to register shared cache node event handler, err: %w", err)
-	}
+	f.sharedCachesMu.Lock()
+	f.sharedCachesStarted = true
+	f.sharedCachesMu.Unlock()
 
 	for _, c := range caches {
 		c.Start(ctx)
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	nrtinformers "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/informers/externalversions"
@@ -173,6 +174,10 @@ type FrameworkExtenderFactory struct {
 	sharedCaches        map[string]SharedPluginCache
 	sharedCachesOrder   []string
 	sharedCachesStarted bool
+	// startedCaches is the immutable, registration-ordered snapshot of shared caches
+	// published once by StartSharedCaches. The unified dispatcher reads it lock-free on
+	// every pod/node event, avoiding a mutex acquisition and slice allocation per event.
+	startedCaches atomic.Pointer[[]SharedPluginCache]
 }
 
 func NewFrameworkExtenderFactory(options ...Option) (*FrameworkExtenderFactory, error) {
@@ -550,11 +555,11 @@ func (f *FrameworkExtenderFactory) getOrRegisterSharedCache(key string, handle E
 // after all profiles are built (all Plugin.New() calls completed) and before
 // informerFactory.Start() so no event is delivered before its handler is registered.
 // Idempotent — subsequent calls after the first are no-ops.
-func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, informerFactory informers.SharedInformerFactory) {
+func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, informerFactory informers.SharedInformerFactory) error {
 	f.sharedCachesMu.Lock()
 	if f.sharedCachesStarted {
 		f.sharedCachesMu.Unlock()
-		return
+		return nil
 	}
 	f.sharedCachesStarted = true
 	caches := make([]SharedPluginCache, 0, len(f.sharedCachesOrder))
@@ -564,40 +569,46 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 	f.sharedCachesMu.Unlock()
 
 	if len(caches) == 0 {
-		return
+		return nil
 	}
+
+	// Publish the immutable snapshot before wiring the dispatcher so events (which cannot
+	// arrive until the informer factory is started, after this returns) always read it.
+	f.startedCaches.Store(&caches)
 
 	// Register unified pod/node dispatchers before invoking Start so plugin-specific
 	// CRD handlers registered inside Start observe the same "handlers wired before
-	// factory started" invariant.
+	// factory started" invariant. A registration failure means the dispatcher is not
+	// wired and the caches would never receive pod/node events, so fail fast rather than
+	// starting into a silently broken state.
 	if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    f.dispatchPodAdd,
 		UpdateFunc: f.dispatchPodUpdate,
 		DeleteFunc: f.dispatchPodDelete,
 	}); err != nil {
-		klog.ErrorS(err, "failed to register shared cache pod event handler")
+		return fmt.Errorf("failed to register shared cache pod event handler, err: %w", err)
 	}
 	if _, err := informerFactory.Core().V1().Nodes().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    f.dispatchNodeAdd,
 		UpdateFunc: f.dispatchNodeUpdate,
 		DeleteFunc: f.dispatchNodeDelete,
 	}); err != nil {
-		klog.ErrorS(err, "failed to register shared cache node event handler")
+		return fmt.Errorf("failed to register shared cache node event handler, err: %w", err)
 	}
 
 	for _, c := range caches {
 		c.Start(ctx)
 	}
+	return nil
 }
 
-func (f *FrameworkExtenderFactory) snapshotCaches() []SharedPluginCache {
-	f.sharedCachesMu.Lock()
-	defer f.sharedCachesMu.Unlock()
-	caches := make([]SharedPluginCache, 0, len(f.sharedCachesOrder))
-	for _, key := range f.sharedCachesOrder {
-		caches = append(caches, f.sharedCaches[key])
+// dispatchCaches returns the immutable snapshot of registered shared caches published by
+// StartSharedCaches. Read lock-free on every event — no mutex, no allocation.
+func (f *FrameworkExtenderFactory) dispatchCaches() []SharedPluginCache {
+	if p := f.startedCaches.Load(); p != nil {
+		return *p
 	}
-	return caches
+	return nil
 }
 
 func (f *FrameworkExtenderFactory) dispatchPodAdd(obj interface{}) {
@@ -605,7 +616,7 @@ func (f *FrameworkExtenderFactory) dispatchPodAdd(obj interface{}) {
 	if !ok {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnPodAdd(pod)
 	}
 }
@@ -619,7 +630,7 @@ func (f *FrameworkExtenderFactory) dispatchPodUpdate(oldObj, newObj interface{})
 	if !ok {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnPodUpdate(oldPod, newPod)
 	}
 }
@@ -635,7 +646,7 @@ func (f *FrameworkExtenderFactory) dispatchPodDelete(obj interface{}) {
 	if pod == nil {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnPodDelete(pod)
 	}
 }
@@ -645,7 +656,7 @@ func (f *FrameworkExtenderFactory) dispatchNodeAdd(obj interface{}) {
 	if !ok {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnNodeAdd(node)
 	}
 }
@@ -659,7 +670,7 @@ func (f *FrameworkExtenderFactory) dispatchNodeUpdate(oldObj, newObj interface{}
 	if !ok {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnNodeUpdate(oldNode, newNode)
 	}
 }
@@ -675,7 +686,7 @@ func (f *FrameworkExtenderFactory) dispatchNodeDelete(obj interface{}) {
 	if node == nil {
 		return
 	}
-	for _, c := range f.snapshotCaches() {
+	for _, c := range f.dispatchCaches() {
 		c.OnNodeDelete(node)
 	}
 }
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -639,15 +639,6 @@ func (f *FrameworkExtenderFactory) SharedCachesStarted() bool {
 
 // dispatchCaches returns the immutable snapshot of registered shared caches published by
 // StartSharedCaches. Read lock-free on every event — no mutex, no allocation.
-//
-// TODO: the dispatch* methods below invoke each cache's handler sequentially. The end-state
-// goal is parallel-by-default dispatch with an opt-in serial mode, so the shared dispatcher
-// does not regress event-handling latency compared to the original per-profile handler
-// pattern. Deferred to the unified event-dispatcher phase (PR-3).
-//
-// TODO: instrument per-handler dispatch latency (e.g. a histogram labeled by cache key and
-// event type) so the dispatcher's handler performance is observable and regressions are
-// caught once serial/parallel execution modes land. Deferred to the same PR-3 phase.
 func (f *FrameworkExtenderFactory) dispatchCaches() []SharedPluginCache {
 	if p := f.startedCaches.Load(); p != nil {
 		return *p
@@ -655,6 +646,15 @@ func (f *FrameworkExtenderFactory) dispatchCaches() []SharedPluginCache {
 	return nil
 }
 
+// The dispatch* methods below invoke each registered cache's handler sequentially.
+//
+// TODO: the end-state goal is parallel-by-default dispatch with an opt-in serial mode, so the
+// shared dispatcher does not regress event-handling latency compared to the original per-profile
+// handler pattern. Deferred to the unified event-dispatcher phase (PR-3).
+//
+// TODO: instrument per-handler dispatch latency (e.g. a histogram labeled by cache key and event
+// type) so the dispatcher's handler performance is observable and regressions are caught once
+// serial/parallel execution modes land. Deferred to the same PR-3 phase.
 func (f *FrameworkExtenderFactory) dispatchPodAdd(obj interface{}) {
 	pod, ok := obj.(*corev1.Pod)
 	if !ok {
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -545,6 +545,13 @@ func (f *FrameworkExtenderFactory) getOrRegisterSharedCache(key string, handle E
 	if c, ok := f.sharedCaches[key]; ok {
 		return c
 	}
+	if f.sharedCachesStarted {
+		// Registering a new shared cache after StartSharedCaches has published the dispatch
+		// snapshot would leave it out of that snapshot: it never gets Start() and never
+		// receives an event. That is a wiring bug (registration must happen during plugin
+		// New(), before StartSharedCaches), not a recoverable state — fail loudly.
+		panic(fmt.Sprintf("frameworkext: shared cache %q registered after StartSharedCaches; register it during plugin New()", key))
+	}
 	c := create(handle)
 	f.sharedCaches[key] = c
 	f.sharedCachesOrder = append(f.sharedCachesOrder, key)
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -45,6 +45,7 @@ import (
 	koordinatorclientset "github.com/koordinator-sh/koordinator/pkg/client/clientset/versioned"
 	koordinatorinformers "github.com/koordinator-sh/koordinator/pkg/client/informers/externalversions"
 	"github.com/koordinator-sh/koordinator/pkg/features"
+	frameworkexthelper "github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/helper"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/indexer"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/networktopology"
 	"github.com/koordinator-sh/koordinator/pkg/scheduler/frameworkext/services"
@@ -572,18 +573,27 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 		// registration failure leaves the factory un-started (sharedCachesStarted stays
 		// false, no snapshot published) and returns an error for the caller to fail fast
 		// on, rather than a partially-started state that later calls would skip.
-		if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-			AddFunc:    f.dispatchPodAdd,
-			UpdateFunc: f.dispatchPodUpdate,
-			DeleteFunc: f.dispatchPodDelete,
-		}); err != nil {
+		//
+		// Use ForceSyncFromInformer (not a bare AddEventHandler) so the dispatcher's
+		// registrations are collected into WaitForHandlersSync: the scheduler then waits
+		// for these handlers to drain their initial pod/node list into the shared caches
+		// before the first scheduling cycle, so a shared cache is never read while still
+		// unpopulated. This preserves the sync guarantee the per-plugin handlers had before
+		// they were centralized here.
+		if _, err := frameworkexthelper.ForceSyncFromInformer(ctx.Done(), informerFactory,
+			informerFactory.Core().V1().Pods().Informer(), cache.ResourceEventHandlerFuncs{
+				AddFunc:    f.dispatchPodAdd,
+				UpdateFunc: f.dispatchPodUpdate,
+				DeleteFunc: f.dispatchPodDelete,
+			}); err != nil {
 			return fmt.Errorf("failed to register shared cache pod event handler, err: %w", err)
 		}
-		if _, err := informerFactory.Core().V1().Nodes().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-			AddFunc:    f.dispatchNodeAdd,
-			UpdateFunc: f.dispatchNodeUpdate,
-			DeleteFunc: f.dispatchNodeDelete,
-		}); err != nil {
+		if _, err := frameworkexthelper.ForceSyncFromInformer(ctx.Done(), informerFactory,
+			informerFactory.Core().V1().Nodes().Informer(), cache.ResourceEventHandlerFuncs{
+				AddFunc:    f.dispatchNodeAdd,
+				UpdateFunc: f.dispatchNodeUpdate,
+				DeleteFunc: f.dispatchNodeDelete,
+			}); err != nil {
 			return fmt.Errorf("failed to register shared cache node event handler, err: %w", err)
 		}
 		// Publish the immutable snapshot for lock-free dispatch. Events cannot arrive until
@@ -604,6 +614,11 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 
 // dispatchCaches returns the immutable snapshot of registered shared caches published by
 // StartSharedCaches. Read lock-free on every event — no mutex, no allocation.
+//
+// TODO: the dispatch* methods below invoke each cache's handler sequentially. The end-state
+// goal is parallel-by-default dispatch with an opt-in serial mode, so the shared dispatcher
+// does not regress event-handling latency compared to the original per-profile handler
+// pattern. Deferred to the unified event-dispatcher phase (PR-3).
 func (f *FrameworkExtenderFactory) dispatchCaches() []SharedPluginCache {
 	if p := f.startedCaches.Load(); p != nil {
 		return *p
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"sync"
 	"time"
 
 	nrtinformers "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/informers/externalversions"
@@ -31,6 +32,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	k8sfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	fwktype "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler"
@@ -165,6 +168,11 @@ type FrameworkExtenderFactory struct {
 	pluginInformerFactories []SharedInformerFactory
 
 	metricsRecorder *metrics.MetricAsyncRecorder
+
+	sharedCachesMu      sync.Mutex
+	sharedCaches        map[string]SharedPluginCache
+	sharedCachesOrder   []string
+	sharedCachesStarted bool
 }
 
 func NewFrameworkExtenderFactory(options ...Option) (*FrameworkExtenderFactory, error) {
@@ -192,6 +200,7 @@ func NewFrameworkExtenderFactory(options ...Option) (*FrameworkExtenderFactory, 
 		crossSchedulerNominator:             handleOptions.crossSchedulerNominator,
 		workloadAuditor:                     handleOptions.workloadAuditor,
 		metricsRecorder:                     metrics.NewMetricsAsyncRecorder(1000, time.Second, wait.NeverStop),
+		sharedCaches:                        map[string]SharedPluginCache{},
 	}, nil
 }
 
@@ -517,5 +526,156 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
 		extenderFactory.updatePlugins(plugin, fw.ProfileName())
 		frameworkExtender.(*frameworkExtenderImpl).updatePlugins(plugin)
 		return plugin, nil
+	}
+}
+
+// getOrRegisterSharedCache is the factory-side implementation behind
+// ExtendedHandle.GetOrRegisterSharedCache. First call for a given key invokes create(handle)
+// and stores the result; subsequent calls return the stored instance and do not call
+// create. Registration order is preserved for deterministic event dispatch.
+func (f *FrameworkExtenderFactory) getOrRegisterSharedCache(key string, handle ExtendedHandle, create func(ExtendedHandle) SharedPluginCache) SharedPluginCache {
+	f.sharedCachesMu.Lock()
+	defer f.sharedCachesMu.Unlock()
+	if c, ok := f.sharedCaches[key]; ok {
+		return c
+	}
+	c := create(handle)
+	f.sharedCaches[key] = c
+	f.sharedCachesOrder = append(f.sharedCachesOrder, key)
+	return c
+}
+
+// StartSharedCaches wires up the unified pod/node event dispatcher on informerFactory and
+// invokes Start(ctx) on every registered SharedPluginCache exactly once. Must be called
+// after all profiles are built (all Plugin.New() calls completed) and before
+// informerFactory.Start() so no event is delivered before its handler is registered.
+// Idempotent — subsequent calls after the first are no-ops.
+func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, informerFactory informers.SharedInformerFactory) {
+	f.sharedCachesMu.Lock()
+	if f.sharedCachesStarted {
+		f.sharedCachesMu.Unlock()
+		return
+	}
+	f.sharedCachesStarted = true
+	caches := make([]SharedPluginCache, 0, len(f.sharedCachesOrder))
+	for _, key := range f.sharedCachesOrder {
+		caches = append(caches, f.sharedCaches[key])
+	}
+	f.sharedCachesMu.Unlock()
+
+	if len(caches) == 0 {
+		return
+	}
+
+	// Register unified pod/node dispatchers before invoking Start so plugin-specific
+	// CRD handlers registered inside Start observe the same "handlers wired before
+	// factory started" invariant.
+	if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    f.dispatchPodAdd,
+		UpdateFunc: f.dispatchPodUpdate,
+		DeleteFunc: f.dispatchPodDelete,
+	}); err != nil {
+		klog.ErrorS(err, "failed to register shared cache pod event handler")
+	}
+	if _, err := informerFactory.Core().V1().Nodes().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    f.dispatchNodeAdd,
+		UpdateFunc: f.dispatchNodeUpdate,
+		DeleteFunc: f.dispatchNodeDelete,
+	}); err != nil {
+		klog.ErrorS(err, "failed to register shared cache node event handler")
+	}
+
+	for _, c := range caches {
+		c.Start(ctx)
+	}
+}
+
+func (f *FrameworkExtenderFactory) snapshotCaches() []SharedPluginCache {
+	f.sharedCachesMu.Lock()
+	defer f.sharedCachesMu.Unlock()
+	caches := make([]SharedPluginCache, 0, len(f.sharedCachesOrder))
+	for _, key := range f.sharedCachesOrder {
+		caches = append(caches, f.sharedCaches[key])
+	}
+	return caches
+}
+
+func (f *FrameworkExtenderFactory) dispatchPodAdd(obj interface{}) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnPodAdd(pod)
+	}
+}
+
+func (f *FrameworkExtenderFactory) dispatchPodUpdate(oldObj, newObj interface{}) {
+	oldPod, ok := oldObj.(*corev1.Pod)
+	if !ok {
+		return
+	}
+	newPod, ok := newObj.(*corev1.Pod)
+	if !ok {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnPodUpdate(oldPod, newPod)
+	}
+}
+
+func (f *FrameworkExtenderFactory) dispatchPodDelete(obj interface{}) {
+	var pod *corev1.Pod
+	switch t := obj.(type) {
+	case *corev1.Pod:
+		pod = t
+	case cache.DeletedFinalStateUnknown:
+		pod, _ = t.Obj.(*corev1.Pod)
+	}
+	if pod == nil {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnPodDelete(pod)
+	}
+}
+
+func (f *FrameworkExtenderFactory) dispatchNodeAdd(obj interface{}) {
+	node, ok := obj.(*corev1.Node)
+	if !ok {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnNodeAdd(node)
+	}
+}
+
+func (f *FrameworkExtenderFactory) dispatchNodeUpdate(oldObj, newObj interface{}) {
+	oldNode, ok := oldObj.(*corev1.Node)
+	if !ok {
+		return
+	}
+	newNode, ok := newObj.(*corev1.Node)
+	if !ok {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnNodeUpdate(oldNode, newNode)
+	}
+}
+
+func (f *FrameworkExtenderFactory) dispatchNodeDelete(obj interface{}) {
+	var node *corev1.Node
+	switch t := obj.(type) {
+	case *corev1.Node:
+		node = t
+	case cache.DeletedFinalStateUnknown:
+		node, _ = t.Obj.(*corev1.Node)
+	}
+	if node == nil {
+		return
+	}
+	for _, c := range f.snapshotCaches() {
+		c.OnNodeDelete(node)
 	}
 }
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -575,6 +575,15 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 	}
 	f.sharedCachesMu.Unlock()
 
+	// INVARIANT: no getOrRegisterSharedCache runs concurrently with this function. sharedCachesStarted
+	// is not set until after the dispatchers register below (deliberately, so a registration failure
+	// leaves the factory un-started for fail-fast), so a cache registered in the window between the
+	// snapshot above and the flag set below would miss BOTH the snapshot and the post-start panic
+	// guard. This is unreachable today because profiles are constructed serially and every
+	// getOrRegisterSharedCache (from Plugin.New()) completes before StartSharedCaches is called.
+	// Future hardening (only if that ordering ever changes) must preserve the fail-fast property —
+	// e.g. fold the flag into a critical section that also covers registration.
+
 	if len(caches) > 0 {
 		// Register the unified pod/node dispatchers before committing any state, so a
 		// registration failure leaves the factory un-started (sharedCachesStarted stays
--- a/pkg/scheduler/frameworkext/framework_extender_factory.go
+++ b/pkg/scheduler/frameworkext/framework_extender_factory.go
@@ -612,6 +612,15 @@ func (f *FrameworkExtenderFactory) StartSharedCaches(ctx context.Context, inform
 	return nil
 }
 
+// SharedCachesStarted reports whether StartSharedCaches has already run. Used by the
+// scheduler startup pipeline to assert that the shared caches (and their pod/node dispatcher)
+// are wired before the main informer factory is started.
+func (f *FrameworkExtenderFactory) SharedCachesStarted() bool {
+	f.sharedCachesMu.Lock()
+	defer f.sharedCachesMu.Unlock()
+	return f.sharedCachesStarted
+}
+
 // dispatchCaches returns the immutable snapshot of registered shared caches published by
 // StartSharedCaches. Read lock-free on every event — no mutex, no allocation.
 //
--- a/pkg/scheduler/frameworkext/shared_plugin_cache.go
+++ b/pkg/scheduler/frameworkext/shared_plugin_cache.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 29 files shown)


mdryaaan:
Hi @saintube @ZiMengSheng, This implements the design from the proposal in #3095
- It adds the opt-in SharedPluginCache mechanism in frameworkext (registry + StartSharedCaches + unified pod/node dispatcher) and migrates DeviceShare's nodeDeviceCache onto it, so the cache is created once per scheduler instance and shared across all profiles instead of one per profile. Reserve/Unreserve use the CacheReserver assume/forget contract with reconcile-on-event.
-  I kept it to 7 small commits so it's easy to review in order  commits 1–3 are the new mechanism (frameworkext only), commit 4 is the DeviceShare migration (bundled with its test adaptations so each commit still builds), and 5–7 are regression coverage: registry/dispatcher, the assume/forget + reconcile paths, and startup-ordering + -race stress. Existing DeviceShare tests pass unchanged; -race and golangci-lint are clean.

PTAL, happy to adjust anything.

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -471,3 +480,154 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
 		return plugin, nil
 	}
 }
+
+// getOrRegisterSharedCache is the factory-side implementation behind
+// ExtendedHandle.GetOrRegisterSharedCache. First call for a given key invokes create(handle)
+// and stores the result; subsequent calls return the stored instance and do not call
+// create. Registration order is preserved for deterministic event dispatch.
+func (f *FrameworkExtenderFactory) getOrRegisterSharedCache(key string, handle ExtendedHandle, create func(ExtendedHandle) SharedPluginCache) SharedPluginCache {
+	f.sharedCachesMu.Lock()
```
saintube, line 542: This introduces a non-negligible overhead on every pod event.

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -471,3 +480,154 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
...
+	// Register unified pod/node dispatchers before invoking Start so plugin-specific
+	// CRD handlers registered inside Start observe the same "handlers wired before
+	// factory started" invariant.
+	if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    f.dispatchPodAdd,
+		UpdateFunc: f.dispatchPodUpdate,
+		DeleteFunc: f.dispatchPodDelete,
+	}); err != nil {
+		klog.ErrorS(err, "failed to register shared cache pod event handler")
+	}
```
saintube: Is this safe to continue while adding the informer failed?

pkg/scheduler/frameworkext/framework_extender_factory.go
mdryaaan: @saintube Good catch. You're right  snapshotCaches() was taking the mutex and allocating a fresh slice on every pod/node event. The registered set is actually frozen by the time events start flowing (all GetOrRegisterSharedCache calls happen during profile construction, before StartSharedCaches), so I now build the ordered slice once in StartSharedCaches and publish it via an atomic.Pointer. The dispatchers read it lock-free now  no mutex and no allocation on the event path.  Fixed in [e6a0ebc](https://github.com/koordinator-sh/koordinator/pull/3125/commits/e6a0ebcf282c52c1dbb4e497a270aa108882debb) PTAL.

pkg/scheduler/frameworkext/framework_extender_factory.go
mdryaaan: Agreed. If AddEventHandler fails the dispatcher never gets wired, so the shared caches would start but never receive any pod/node events for the rest of the scheduler's lifetime  a silent, hard-to-diagnose failure. Changed StartSharedCaches to return an error and propagate the failure up to Setup(), so startup fails fast instead of continuing in a broken state. Updated the call site in server.go accordingly  in [e6a0ebc](https://github.com/koordinator-sh/koordinator/pull/3125/commits/e6a0ebcf282c52c1dbb4e497a270aa108882debb)

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -826,7 +829,6 @@ func New(ctx context.Context, obj runtime.Object, handle fwktype.Handle) (fwktyp
 	if _, err := frameworkexthelper.ForceSyncFromInformer(ctx.Done(), handle.SharedInformerFactory(), nodeInformer, nil); err != nil {
 		return nil, err
 	}
```
ZiMengSheng: **This `ForceSyncFromInformer(nodeInformer, nil)` registration is more load-bearing than the comment states — it is what gates the GC goroutine, and the coupling is fragile.**

The comment frames this purely as "visible to the framework's `WaitForHandlersSync`". But after this PR, `go gcNodeDevice(...)` runs from `Start()` (via `StartSharedCaches`, inside `Setup()`), which executes *before* `startInformersAndWaitForSync` in `cmd/koord-scheduler/app/server.go` — i.e. before `cc.InformerFactory` is started/synced. The only thing preventing GC's first tick from running against an empty node lister is the `frameworkexthelper.WaitForHandlersSync(ctx)` guard at the top of `gcNodeDevice`, and that guard only works because *this* registration is collected here in `New()` (for every profile) before GC is launched.

So the safety invariant is implicit and easy to break:

- `WaitForHandlersSync` returns `true` as soon as `GetRegistrations()` is empty or all-synced. Its correctness here depends on this node registration already being present and *not-yet-synced* at the moment GC first polls.
- Today that holds only because all profiles' `New()` complete (collecting the registration) before `StartSharedCaches` launches GC.
- If this registration is later moved or removed — e.g. someone assumes the new unified pod/node dispatcher already covers node events and deletes it — `WaitForHandlersSync` would observe an all-synced/empty set, GC's first tick would `List()` an empty node lister and evict every warm `nodeDeviceInfos` entry (recreated only on the next Device CRD event). That would be silent, timing-dependent state loss.

Given the repo already provides `helper.RegisterAfterAllInformersSynced` — documented for exactly this case ("kicking off background reconcilers that rely on fully populated caches", run after all informer factories are started+synced and before pod dispatching) — would it be cleaner to launch `gcNodeDevice` from an `AfterAllInformersSynced` hook instead of leaning on the implicit `WaitForHandlersSync` + this-registration coupling? That gives GC an explicit happens-after-sync guarantee independent of where the goroutine is created, and lets this node registration stand on its stated purpose alone.

Not strictly blocking (the invariant does hold as written), but at minimum the comment here should call out that this registration is gating GC startup.

ZiMengSheng (commented):
One inline note on the DeviceShare `New()` node-informer registration re: GC startup ordering — see the comment on plugin.go. Non-blocking.

pkg/scheduler/plugins/deviceshare/plugin_test.go
```diff
@@ -158,6 +158,17 @@ type pluginTestSuit struct {
...
+
+// start drives the framework's shared-cache startup path (FrameworkExtenderFactory.
+// StartSharedCaches), which invokes Start(ctx) on each registered SharedPluginCache —
+// registering DeviceShare's Device/Reservation CRD handlers and the GC goroutine. In
+// production this is called by cmd/koord-scheduler after all plugins are built; plugin
+// unit tests that rely on cache population must call it after proxyNew and before starting
+// the shared informer factory, matching the four-phase initialization order.
+func (s *pluginTestSuit) start(ctx context.Context) {
+	_ = s.extenderFactory.StartSharedCaches(ctx, s.Framework.SharedInformerFactory())
 }
```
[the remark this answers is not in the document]
mdryaaan: fixed in [6464930](https://github.com/koordinator-sh/koordinator/pull/3125/commits/6464930f26ff21274905ae121f5f29b9232b46dd) . pluginTestSuit.start() now surfaces the error via assert.NoError(s.t, ...) instead of discarding it, so a shared-cache startup failure in a test fails at the source rather than as a confusing downstream assertion.

mdryaaan:
@saintube Went through the Copilot review  addressed two points and left one intentionally.

`StartSharedCaches` now registers the pod/node dispatchers first and only commits state (marks started + publishes the snapshot) once registration succeeds, so a failure leaves the factory un-started and fails fast,
instead of the partial-start state Copilot flagged. Fixed the test helper to assert `NoError` on startup so a failure surfaces at the source.

On the Reserve/AssumePod rollback suggestion — left as-is on purpose: when `Reserve` returns a failure the framework runs `Unreserve`, which already does `updateCacheUsed(..., add=false)` + `ForgetPod`, so a rollback inside
`Reserve` itself would double-subtract. Both `Reserve` and `Unreserve` also guard `nodeDeviceInfo == nil` before locking, so the GC-race case is panic-safe.

pkg/scheduler/plugins/deviceshare/plugin.go
mdryaaan: @ZiMengSheng, Good catch, GC was leaning on an implicit coupling, with `WaitForHandlersSync` only being load-bearing because of the node registration collected in `New()`. Moved `gcNodeDevice` behind a
`RegisterAfterAllInformersSynced` hook, so it now launches after all informer factories are started and synced  an explicit guarantee independent of that registration. `WaitForHandlersSync` stays as defense-in-depth (comment updated), and reworded the `New()` comment so the node registration stands on its actual purpose rather than the "gates GC" framing.

cmd/koord-scheduler/app/server.go
```diff
@@ -545,6 +545,12 @@ func Setup(ctx context.Context, opts *options.Options, outOfTreeRegistryOptions
 	schedAdapter := frameworkExtenderFactory.Scheduler()
 
 	eventhandlers.AddScheduleEventHandler(sched, schedAdapter, cc.InformerFactory, cc.KoordinatorSharedInformerFactory, crossSchedulerNominator)
+	// StartSharedCaches wires the unified pod/node dispatcher on cc.InformerFactory and
+	// invokes Start on every SharedPluginCache registered by plugins during New(). Must
+	// run before cc.InformerFactory.Start() so no event fires before its handler is wired.
+	if err := frameworkExtenderFactory.StartSharedCaches(ctx, cc.InformerFactory); err != nil {
```
saintube: If you want to start the cache before the informerFactory, why not start it in the `Run()` instead of `Setup()`?

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -813,20 +816,20 @@ func New(ctx context.Context, obj runtime.Object, handle fwktype.Handle) (fwktyp
...
-	extendedHandle.RegisterForgetPodHandler(deviceCache.deletePod)
-	// Register the node informer synchronously during New so that the registration
-	// is visible to the framework's WaitForHandlersSync. If we registered it inside
-	// the gcNodeDevice goroutine, the framework might start scheduling before the
-	// handler registration is collected. Using a no-op handler because gcNodeDevice
-	// only reads from the node lister.
+	sharedCache := extendedHandle.GetOrRegisterSharedCache(Name, func(h frameworkext.ExtendedHandle) frameworkext.SharedPluginCache {
+		return newNodeDeviceCache(h)
+	})
+	deviceCache := sharedCache.(*nodeDeviceCache)
```
saintube:
```suggestion
	deviceCache, ok := sharedCache.(*nodeDeviceCache)
	if !ok { // ... }
```

pkg/scheduler/frameworkext/shared_plugin_cache.go
```diff
@@ -0,0 +1,54 @@
...
+// CacheReserver is an optional interface for SharedPluginCache implementations that write
+// assumed allocations during the Reserve scheduling phase. Implementers must uphold the
+// assume/forget contract: AssumePod records what Reserve wrote, ForgetPod (or the
+// reconcile path in OnPodAdd/Update/Delete) rolls it back. The informer event is always
+// authoritative — implementations reconcile rather than skip when it arrives for an
+// assumed pod, so the cache converges on the event's state even when the event's pod
+// object differs from what Reserve assumed.
+type CacheReserver interface {
+	SharedPluginCache
+	AssumePod(pod *corev1.Pod, nodeName string) error
```
saintube, line 75: We'd better add metrics and a debuggable service for this in-memory state in case some bugs make it leak.

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -452,18 +480,174 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
...
+		return nil
+	})
+}
+
+// OnPodAdd implements frameworkext.SharedPluginCache. If the pod was previously assumed,
+// reconciles against the authoritative event instead of skipping — see the CacheReserver
+// invariants in the proposal.
+func (n *nodeDeviceCache) OnPodAdd(pod *corev1.Pod) {
+	if assumed, ok := n.takeAssumed(pod.UID); ok {
+		n.reconcileAssumed(assumed, pod)
```
saintube: Is this thread-safe without a lock?

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -471,3 +485,160 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
...
+		caches = append(caches, f.sharedCaches[key])
+	}
+	f.sharedCachesMu.Unlock()
+
+	if len(caches) > 0 {
+		// Register the unified pod/node dispatchers before committing any state, so a
+		// registration failure leaves the factory un-started (sharedCachesStarted stays
+		// false, no snapshot published) and returns an error for the caller to fail fast
+		// on, rather than a partially-started state that later calls would skip.
+		if _, err := informerFactory.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
saintube: Is this safe without `ForceSyncFromInformer`?

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -471,3 +485,160 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
...
+	}
+	return nil
+}
+
+func (f *FrameworkExtenderFactory) dispatchPodAdd(obj interface{}) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return
+	}
+	for _, c := range f.dispatchCaches() {
```
saintube, line 662: The dispatcher currently invokes each handler sequentially. The end-state goal is parallel-by-default with an opt-in serial mode, so that the shared dispatcher does not regress latency compared to the original per-profile event handler pattern. A TODO marking this for a follow-up is fine at the phase.

pkg/scheduler/plugins/deviceshare/device_cache_assume_test.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	apiext "github.com/koordinator-sh/koordinator/apis/extension"
+	schedulingv1alpha1 "github.com/koordinator-sh/koordinator/apis/scheduling/v1alpha1"
+)
+
```
saintube, line 31: Since this PR touches multiple packages, it is recommended to add some UTs and benchmark tests for concurrency correctness and performance, including both the new device cache and the new shared cache mechanism.

saintube (commented):
Since this PR touches

cmd/koord-scheduler/app/server.go
mdryaaan: @saintube  `StartSharedCaches` only registers the dispatcher handlers, right after `AddScheduleEventHandler` which is also done in `Setup()`  Setup wires, Run starts the informers, and the registration-before-`InformerFactory.Start()` ordering is exactly the four-phase design. Happy to move it into `Run()` (before `startInformersAndWaitForSync`) if you'd prefer, but it'd be equivalent.

pkg/scheduler/plugins/deviceshare/plugin.go
mdryaaan: Done, switched to the two-value form with an error return, as suggested.

pkg/scheduler/frameworkext/shared_plugin_cache.go
mdryaaan: Agreed that the `assumedPods` state is worth making observable so a leak is easy to spot. Given how large this PR already is, would you prefer I add the metric + debug endpoint (extending DeviceShare's existing
`APIServiceProvider` service, plus a gauge via the scheduler metrics pkg) here in this PR, or track it as a follow-up issue right after this lands?

pkg/scheduler/plugins/deviceshare/device_cache.go
mdryaaan: It's safe, the `OnPod*` handlers are delivered serially by the shared informer, and every shared field is individually locked (`assumedPods`/ node-map under `n.lock`, per-node allocations under `nodeDevice.lock`);
`takeAssumed` atomically claims the marker so exactly one path reconciles.  Not holding one coarse lock is deliberate (the proposal's no-global-write-lock invariant). Added a comment documenting this — the `-race` stress test exercises concurrent Reserve+OnPodAdd+Unreserve+OnPodDelete on the same pod.

pkg/scheduler/frameworkext/framework_extender_factory.go
mdryaaan: Good catch  -  moved the dispatcher's pod/node registration to `ForceSyncFromInformer` so it's collected into `WaitForHandlersSync`. The scheduler now waits for the dispatcher to drain the initial pod/node list into the shared caches before the first scheduling cycle, restoring the sync guarantee the per-plugin handlers had before centralization.

pkg/scheduler/frameworkext/framework_extender_factory.go
mdryaaan: Added a TODO marking parallel-by-default dispatch as a follow-up for the event-dispatcher phase (PR-3).

pkg/scheduler/plugins/deviceshare/device_cache_assume_test.go
mdryaaan: Added benchmarks for the device cache (concurrent Reserve/reconcile) andthe shared dispatcher (serial + parallel fan-out). Concurrency correctness is already covered by the `-race` stress tests; these add the performance dimension -  dispatch is ~7ns/op with 0 allocs.

cmd/koord-scheduler/app/server.go
saintube: Let's move into `Run()`. We should explicitly guard against the regression where informers get started before the scheduler is fully ready. It also reads more naturally that everything kicks off runtime behavior lives in `Run`.

pkg/scheduler/frameworkext/shared_plugin_cache.go
saintube: I'd vote for adding it in this PR. Given how complex Koordinator's scheduling plugins are, this first implementation PR doesn't even need to complete the full ledger migration — but every refactor of the core logic must be debuggable and observable. So let's add the necessary metrics on the ledger state here; migrations that don't affect existing behavior or efficiency can be deferred to follow-ups.

cmd/koord-scheduler/app/server.go
mdryaaan: @saintube  Done - moved `StartSharedCaches` into `Run()` as the first step of `startInformersAndWaitForSync`, so it runs in both the immediate and leader-election paths (it's idempotent) and always before `cc.InformerFactory.Start()`.            Also added a `SharedCachesStarted()` check right before the main informer factory starts that fails loudly if the caches aren't up yet, so the ordering invariant is self-enforcing against future reordering, which is the regression class we've hit a couple of times here. Nothing after the old call site depended on the caches being started, so the move was clean.

pkg/scheduler/frameworkext/shared_plugin_cache.go
mdryaaan: Added both. There's now a `koord_scheduler_shared_cache_assumed_pods` gauge labeled by cache name — updated on every AssumePod/ForgetPod/ reconcile, so a leak in the ledger shows up directly, and future PR-2 caches (NodeNUMAResource, Reservation) can reuse the same metric with their own label. 
For live inspection I added a `GET /assumedAllocations` endpoint to DeviceShare's existing debug service, returning the current ledger keyed by pod UID (node + allocation per pod), plus a unit test. Kept it scoped as you suggested,  the broader cache migration stays
deferred to follow-ups.

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -452,16 +480,209 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
...
+	n.recordAssumedPodsSizeLocked()
+	return assumed, true
+}
+
+// reconcileAssumed rolls back what AssumePod recorded on assumed.nodeName, then applies
+// the informer event's authoritative state from pod.Spec.NodeName / pod.Annotations. When
+// the event agrees with Reserve this is a net-zero operation; when it diverges the cache
+// converges on the event's state.
+func (n *nodeDeviceCache) reconcileAssumed(assumed *assumedAllocation, pod *corev1.Pod) {
+	n.rollbackAssumed(assumed, pod)
```
saintube: This seems not atomic and has a chance to under count.

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -452,16 +480,209 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
...
+// write itself so behavior for non-CacheReserver code paths is unchanged.
+func (n *nodeDeviceCache) ForgetPod(pod *corev1.Pod) error {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+	delete(n.assumedPods, pod.UID)
+	n.recordAssumedPodsSizeLocked()
+	return nil
+}
+
+func (n *nodeDeviceCache) takeAssumed(uid types.UID) (*assumedAllocation, bool) {
```
saintube, line 749: What if the pod is just updated with the PreBind's Patch and has not been bound to a node?

We should take care of the concurrency between the scheduling cycle, binding cycle, and the event handlers, which is the common cause of the ledge leak.

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -452,16 +480,209 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
...
+	n.lock.Lock()
+	defer n.lock.Unlock()
```
saintube, line 761: For most update events, the assumed states will not change. So the WLock is kind of heavy here.

pkg/scheduler/plugins/deviceshare/device_cache.go
mdryaaan: The under-count came from `reconcileAssumed` releasing the per-node lock between the subtract (rolling back Reserve's assumed write) and the re-add (crediting the pod's real allocation) a concurrent reader could catch the node with neither applied. Fixed by making the same-node reconcile atomic: it takes `nodeDevice.lock` once and does subtract-then-add under the single hold (subtract-first is required by the `isValid` double-add guard).

One thing I want to flag explicitly rather than leave implicit: your comment is about the same-node case, but while tracing it I found the cross-node path (assumed on node A, pod actually bound to node B) had the mirror-image gap. I extended the fix there as well - credit the real node first, then roll back the phantom on the assumed node, so a reader never observes the allocation missing from both nodes. That cross-node reorder is my own reasoned extension of the point you raised, not something the comment asked for; calling it out so it's clear in review what was requested vs. what I added on top.

Pushed in b58e9dce.

pkg/scheduler/plugins/deviceshare/device_cache.go
mdryaaan: Good catch, this was the leak. The PreBind annotation patch (`ApplyPatch`) fires an `OnPodUpdate` while the pod is still unbound (`Spec.NodeName == ""`), before Bind runs. The old code consumed the assumed marker and reconciled on that event, which rolled back Reserve's write and left the node under-counted until Bind landed (or indefinitely if it never did). Fix: `OnPodAdd`/`OnPodUpdate` now only consume the marker and reconcile once the pod is actually bound (`Spec.NodeName != ""`); the PreBind patch event falls through to the normal `updatePod` path, which no-ops on an unbound pod.

On whether the marker can leak if Bind never completes, I traced kube-scheduler's binding cycle: `handleBindingCycleError` calls `RunReservePluginsUnreserve` on every PreBind/Bind failure path, and DeviceShare's `Unreserve` calls `ForgetPod`, so the marker is cleared on all failure paths. Added `Test_nodeDeviceCache_UnreserveAfterPreBind_NoDoubleCount` to pin that the Unreserve-after-PreBind sequence subtracts exactly once, and `Test_nodeDeviceCache_PreBindPatchEvent_KeepsAssumedMarker` to pin that the unbound patch event doesn't prematurely consume the marker.

Pushed in b58e9dce.

pkg/scheduler/plugins/deviceshare/device_cache.go
mdryaaan: Agreed. `takeAssumed` now peeks under `RLock` first and only escalates to `WLock` on a hit (re-checking the map after the upgrade in case another goroutine claimed the entry in between). Since the overwhelming majority of update events are for pods with no assumed marker, they now take only the read lock and never contend on the write lock.

Pushed in b58e9dce.

mdryaaan:
@saintube, Addresses the three review threads on `device_cache.go`'sassumed-ledger reconcile path, same commit since the fixes interlock.

- Atomicity/under-count 
- PreBind patch before bind
- WLock heavy on updates

New tests cover the PreBind-unbound case, the Unreserve-after-PreBind sequence, and terminated-at-bind. `-race` clean across deviceshare, frameworkext, and metrics.

saintube:
@mdryaan 
Nice iteration. Some lifecycle gaps to close before merge, both stemming from Koordinator's in-memory spec.nodeName semantics (the pod can be unassigned from spec.nodeName="" to spec.nodeName="" in some scenarios).
1. The framework's ForgetPod doesn't clear the assumed marker. We've registered the deletePod as the framework ForgetPodHandler, which removes the ledger entry but leaves the assumedPods marker, whereas the CacheReserver ForgetPod only clears the marker. So a Reserved pod forgotten via the framework path leaks a stale marker; if a late event consumes it, reconcileAssumed subtracts a snapshot deletePod already removed -> double subtract -> negative used -> over-allocation. So please unify the framework's forget path to be idempotent.
2. Don't treat the spec.nodeName!="" as the real bind. We may use the transform mechanism of the client-go to fake the pod's spec.nodeName as a nominated node in the coming feature multi-scheduler arbitration. And on the arbitration failure, the node is cleared back to "". We previously adapted plugins like Reservation and DeviceShare for these unassigned events and ensured that ledger rollback for native K8S plugins is handled correctly through ForgetPod. Please make sure this PR does not introduce regressions.
3. Since one of the original goals is to enable plugin sharing across multiple scheduler profiles, it would be helpful to add some UT cases to verify the cross-profile sharing behavior, e.g., for the DeviceShare plugin.

saintube:
@mdryaan Please double-check whether the current patch handles different Pod ledger events consistently:
1. Reserve
2. Unreserve
3. Add assigned
4. Update to bound
5. Update to unassigned (special case for multi-scheduler arbitration)
6. Update to terminated (regarded as a Delete in the vanilla framework)
7. Deleted
8. Annotation of allocation updated
9. ForgetPodHandler
10. AddPod (invoked by nominators/preemption plugins)
11. RemovePod (invoked by reservation transformers/nominators/preemption plugins)

mdryaaan:
@saintube — following up on point 2 before I touch the marker-consumption logic, I want to confirm one thing about the arbitration-failure path, since the code it depends on isn't in this checkout yet and I'd rather not guess at the ordering.

When multi-scheduler arbitration fails and the faked `spec.nodeName` is cleared back to `""`, what drives the ledger rollback for a plugin like DeviceShare?

1. `ext.ForgetPod` firing the registered `ForgetPodHandler`, or
2. the informer delivering the pod update with `spec.nodeName` cleared (the plugin's own `OnPodUpdate` → unassigned branch), or
3. both?

It matters because if it's (3) the plugin has to make the two paths idempotent so the reserved allocation isn't subtracted twice. In this checkout `ext.ForgetPod` (the wrapper that fires the registered handlers) currently has no production caller, every real `ForgetPod` goes through the raw scheduler cache or the `Unreserve` path ,so I can't read the intended invocation off the tree. Knowing whether arbitration failure rolls back through `ForgetPod`, the informer unassign event, or both (and in what order) tells me exactly where the marker must be cleared and lets me keep the two rollback paths from colliding.

saintube:
@mdryaan `ext.ForgetPod` will be invoked while arbitrating failed, and the update event should not re-add the pod to the cache. These two events are handled asynchronously, so we cannot just ignore the unassigned event during the update.

mdryaaan:
@saintube 

**1. Framework forget path idempotency** — good catch. The two forget paths were asymmetric: the registered `ForgetPodHandler` (`deletePod`) reverted the ledger from the pod annotation but left the `assumedPods` marker, while `CacheReserver.ForgetPod` (the Unreserve path) cleared the marker but not the ledger (Unreserve already reverts it). I unified the framework path to be marker-aware and idempotent — it now mirrors `OnPodDelete`: roll back the assumed snapshot if the marker is present (works even before PreBind wrote the annotation, closing the reverse leak where a pod forgotten pre-PreBind left both the marker and Reserve's write dangling), otherwise fall back to the annotation-based `deletePod`.

Scope note: in the current tree `ext.ForgetPod` (the wrapper that fires the registered handlers) has no production caller — every real `ForgetPod` goes through the raw scheduler cache or the Unreserve path. So this hardens the assume/forget contract for the callers that do fire it: the batch path and, per your note, the coming arbitration-failure path. Verified with a direct unit test on the forget handler (assume → forget → no double-subtract, idempotent on repeat).

**2. Don't treat `spec.nodeName != ""` as the real bind** — you're right, and thanks for the arbitration detail; it reshaped the fix. This was a direct consequence of the `spec.nodeName != ""` gate I added last round, which consumed the marker on any transition to a non-empty nodeName — exactly wrong when arbitration fakes the nodeName via the client-go transform.

Given your note that `ext.ForgetPod` fires on arbitration failure and races the unassigned update, I reworked the marker into a pure rollback record:
- It is **never** consumed on a positive (assigned) event. A positive event for an assumed pod is a no-op: it does not re-add the pod (Reserve already accounted it) and leaves the marker in place, so `ext.ForgetPod` can still roll back on arbitration failure.
- It is consumed **only** on negative events — delete, bound→unassigned, terminate, or the framework `ForgetPod` — all funneled through one `releaseAssumed` path that rolls back from the snapshot. The unassigned update is not ignored: it rolls back on its own, and it and `ForgetPod` are idempotent in either order via the existing `isValid` guard (whichever runs second finds the pod already gone from the ledger and subtracts nothing).

This removed the `reconcileAssumed` re-add path (and last round's cross-node reorder) entirely — the marker only ever rolls back now.

Tracing the negative paths for consistency also surfaced a gap in `OnPodAdd`: unlike `OnPodUpdate` it had no negative branch, so a watch reconnect that relists an already-terminated Reserved pod as an Add would leak Reserve's write and leave the marker dangling — more likely now that the marker lives until a terminal event. Folded it into the same `releaseAssumed` path with a regression test.

One consequence worth flagging: since positive events no longer clear the marker, `assumedPods` now holds an entry for every running Reserved pod, not just the in-flight window — so the round-5 `SharedCacheAssumedPods` gauge and `/assumedAllocations` endpoint shift from "in-flight only" to "all Reserved". I kept the marker terminal-only deliberately (clearing it early on annotation-present would be unsafe if a losing scheduler wrote the annotation before arbitration failure). Happy to re-scope that observability in a follow-up once the arbitration/PreBind ordering lands.

**3. Cross-profile UT** — added `Test_New_SharedCacheAcrossProfiles`: two profiles built from one `FrameworkExtenderFactory` share a single `nodeDeviceCache` pointer (checked against a two-factory negative control to make sure it has teeth).

**On the 11-event checklist** — the model above makes it consistent: 1–2 (Reserve/Unreserve) unchanged; 3/4/8 (add-assigned / update-to-bound / annotation-updated) are positive → no-op while assumed, idempotent `updatePod` otherwise; 5/6/7/9 (unassign / terminate / delete / ForgetPod) are negative → single `releaseAssumed` rollback, idempotent across the async pair; 10/11 (AddPod/RemovePod) are read-only on the ledger, so they can't diverge.

Checks: `-race` clean on deviceshare/frameworkext, golangci-lint 0 issues. New tests cover faked-nodeName, unassign, terminate, the relist-Add, both-orders idempotency, and the `-race` stress now interleaves the framework `ForgetPod` with the unassigned update.

pkg/scheduler/plugins/deviceshare/plugin.go
```diff
@@ -543,8 +543,11 @@ func (p *Plugin) Reserve(ctx context.Context, cycleState fwktype.CycleState, pod
 	}
 	logAllocationContext(pod, nodeName, nodeDeviceInfo, state.designatedAllocation, state.allocationResult)
 	nodeDeviceInfo.lock.Lock()
-	defer nodeDeviceInfo.lock.Unlock()
 	nodeDeviceInfo.updateCacheUsed(state.allocationResult, pod, true)
+	nodeDeviceInfo.lock.Unlock()
+	if err := p.nodeDeviceCache.AssumePod(pod, nodeName); err != nil {
+		return fwktype.AsStatus(err)
```
saintube: The framework won't invoke the Unreserve for the Reserve plugin which returns an error.
So please make sure there is not leakage with the Unreserve.

pkg/scheduler/plugins/deviceshare/device_cache.go
```diff
@@ -452,16 +480,264 @@ func (n *nodeDevice) split(requestsPerInstance corev1.ResourceList, deviceType s
...
+	if info == nil {
+		return fmt.Errorf("nodeDevice for %s is missing when assuming pod %s", nodeName, klog.KObj(pod))
 	}
+	info.lock.RLock()
+	allocations := info.copyPodAllocations(pod)
+	info.lock.RUnlock()
+
+	n.lock.Lock()
+	if n.assumedPods == nil {
+		n.assumedPods = make(map[types.UID]*assumedAllocation)
```
saintube, line 667: What if the pod allocation is updated or removed between L631-L633.

pkg/scheduler/frameworkext/framework_extender_factory.go
```diff
@@ -518,3 +533,183 @@ func PluginFactoryProxy(extenderFactory *FrameworkExtenderFactory, factoryFn fra
...
```
saintube, line 662: Also need some metrics to profile the handlers' performance. We can leave a TODO here in this iteration.

pkg/scheduler/plugins/deviceshare/plugin.go
mdryaaan: Thanks @saintube. you're right that we shouldn't rely on this. Tracing it to be precise: at the Reserve extension point, `RunReservePluginsReserve` itself returns without Unreserving - it's the orchestrator that runs `RunReservePluginsUnreserve` afterward (vanilla `schedulingCycle` at `schedule_one.go:211-213`, and the batch engine via `CleanupAssumedPods`), and both currently do cover the failing plugin, so this isn't an active leak today. But that guarantee lives in the orchestrator, not the Reserve contract, so I've made `Reserve` self-clean: on an `AssumePod` error it now rolls back its own cache write instead of depending on Unreserve  which is what matters for the coming arbitration path.

To correct my own earlier note plainly: I'd previously reasoned that a Reserve-side rollback would double-subtract  that was wrong. The `isValid` guard already skips the second subtract, so the inline rollback is safe even when Unreserve also runs (pinned by a test). Pushed in c986a98c.

pkg/scheduler/plugins/deviceshare/device_cache.go
mdryaaan: Good catch. `AssumePod` was snapshotting the pod's allocation under the per-node lock, releasing it, then taking the cache lock to publish the marker — so a concurrent event handler could mutate the pod's allocation in that gap and leave the marker desynced from the ledger. Fixed by snapshotting and publishing the marker under a single `n.lock` hold (n-before-node order, matching `invalidateNodeDevice`/`updateNodeDevice`), so no handler can interleave between the two — the marker always equals the pod's cache contribution at publish time. Also skip publishing an empty marker for the concurrent-removal case, so no orphaned entry lingers. Pushed in c986a98c

pkg/scheduler/frameworkext/framework_extender_factory.go
mdryaaan: Added a TODO here for per-handler dispatch latency metrics (a histogram labeled by cache key and event type), deferred to the PR-3 dispatcher phase alongside the serial/parallel execution work. Pushed in 3544d95e

saintube:
PTAL /cc @ZiMengSheng for the DeviceShare's change

ZiMengSheng (commented):
Reviewed at head `e6f4132`, reading the branch locally against `6dfc4d78`. The layering is good — `SharedPluginCache` is opt-in and non-invasive, the `atomic.Pointer` snapshot removes the per-event lock/allocation, and moving GC onto `RegisterAfterAllInformersSynced` replaces the implicit "node registration happens to be collected by WaitForHandlersSync" coupling with an explicit guarantee. Comments below are ordered by severity; line refs are against `e6f4132`.

---

### 1. `ForgetPodHandler` is registered on a single profile's handle (correctness, multi-profile)

[`device_cache.go:524`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/device_cache.go#L524) registers the forget handler as `n.handle.RegisterForgetPodHandler(n.forgetPod)`, and `n.handle` is the `ExtendedHandle` of whichever profile happened to construct the shared cache first. `forgetPodHandlers` is per-extender state ([`framework_extender.go:62`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/frameworkext/framework_extender.go#L62), appended at [L890](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/frameworkext/framework_extender.go#L890), fanned out at [L911](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/frameworkext/framework_extender.go#L911)), so `ExtendedHandle.ForgetPod` invoked through any *other* profile's extender never reaches the shared cache.

Before this PR every profile's `New()` registered the handler, which is also the convention every other plugin still follows (`elasticquota/plugin.go:207`, `nodenumaresource/pod_eventhandler.go:46`, `reservation/pod_eventhandler.go:44`, `loadaware/load_aware.go:106`). The consequence in a multi-profile scheduler is exactly the case the PR's own comment at `device_cache.go:596-600` says it closes: on a forget through another profile (e.g. multi-scheduler arbitration failure) Reserve's write is not reverted and the marker dangles — and since positive events are suppressed while the marker exists, the reserved node keeps a phantom allocation for the rest of the pod's lifetime.

Suggestion: keep registering it per profile in `New()` (`n.forgetPod` is already idempotent), or make forget handlers factory-scoped. More generally, it would help to state in the `SharedPluginCache` doc that `Start()` may only use factory-scoped dependencies, since the same trap awaits the NodeNUMAResource / Reservation migrations.

### 2. Reservation events bypass the assumed ledger → `assumedPods` entries leak

[`eventhandler_pod.go:37-45`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/eventhandler_pod.go#L37-L45) still wires the reservation-to-pod synthetic events to the legacy `onPodAdd/onPodUpdate/onPodDelete` (→ `updatePod`/`deletePod`), which never consult `assumedPods`. A reservation's reserve pod does go through `Plugin.Reserve` → `AssumePod` (marker keyed by the reserve pod's UID), but its delete event goes down the annotation path, so `releaseAssumed` is never called: the ledger entry survives forever. That leaks a map entry per device-requesting reservation, makes `shared_cache_assumed_pods` drift upward permanently, and fills `/assumedAllocations` with zombie rows — i.e. it degrades exactly the observability this PR adds. Routing those synthetic events through `OnPodAdd/OnPodUpdate/OnPodDelete` would keep one code path.

### 3. The "marker survives every positive event" design deserves an explicit decision + docs

After `d4a26bb` the marker is consumed only by a negative event, so for a pod that binds successfully it lives for the pod's whole lifetime ([`device_cache.go:558-595`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/device_cache.go#L558-L595)). Three consequences that I don't think are fully captured yet:

- **The metric semantics no longer match its doc.** [`metrics.go:174`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/metrics/metrics.go#L174) says "A steadily climbing value signals an assume/forget leak", but by design the gauge now tracks "pods this process scheduled that are still alive", which climbs during normal operation and therefore hides a real leak (e.g. the one in point 2).
- **A divergent-node event leaves the real node undercounted.** `Test_nodeDeviceCache_PositiveEvent_DivergentNode_NotCredited` ([`device_cache_assume_test.go:296`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/device_cache_assume_test.go#L296)) pins that the event's node is not credited. If the pod really is bound elsewhere, the cache under-counts that node; clearing the marker via `ForgetPod` does not apply the event's state either, so convergence depends on some *later* pod update arriving. Under-counting is the dangerous direction (over-allocation). Replaying `updatePod(nil, pod)` after a successful `releaseAssumed` when the pod still exists and is assigned would close that gap.
- **Post-assume annotation changes never converge**, and a delete rolls back the Reserve snapshot rather than the annotation. That is a behavior change vs. the annotation-authoritative model and is worth stating explicitly in the interface docs.

### 4. Docs contradict the implementation (please fix before further review rounds)

The `CacheReserver` doc comment ([`shared_plugin_cache.go:41-50`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/frameworkext/shared_plugin_cache.go#L41-L50)) still says "The informer event is always authoritative — implementations reconcile rather than skip when it arrives for an assumed pod, so the cache converges on the event's state", which is the design `d4a26bb` replaced with the opposite rule. Sections Ⅰ and Ⅲ of the PR description have the same problem ("rollback-then-apply", "event on a different node (phantom cleared, event's node credited)", "mutated annotations (converge on event)"), and Ⅳ still describes 7 commits where the branch now has 19. A reviewer following the description will build the wrong mental model of the code.

### 5. Smaller items

- **`AssumePod` failure turns a benign race into a scheduling error.** [`plugin.go:548`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/plugin.go#L548) fails Reserve when the `nodeDeviceInfos` entry disappeared ([`device_cache.go:624-632`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/device_cache.go#L624-L632)). The inline rollback is right, but not publishing a marker only degrades to the pre-migration annotation semantics, so a warning log + `return nil` may be the friendlier failure mode than failing the pod.
- **"No global write lock is introduced" isn't quite accurate.** `AssumePod` / `ForgetPod` / `takeAssumed` all take the cache-wide `n.lock` in write mode, and `AssumePod` additionally deep-copies the allocations and reports the Prometheus gauge inside it ([`device_cache.go:655`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/device_cache.go#L655)). It's per-Reserve so probably immaterial, but the metric call could move out of the critical section and the claim in the description softened.
- **Registration after start is a silent no-op.** [`getOrRegisterSharedCache`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/frameworkext/framework_extender_factory.go#L541) still accepts new keys after `StartSharedCaches` ran; such a cache is absent from the published snapshot and never gets `Start()`, so it silently receives nothing. Returning an error (or panicking) once started would make the lifecycle contract enforceable rather than documented.
- **The production dispatch path has no plugin-level coverage.** `registerPodEventHandlerForTest` ([`eventhandler_pod_test.go:42`](https://github.com/koordinator-sh/koordinator/blob/e6f4132d942e1f256f73547a0445f62cf329d8ca/pkg/scheduler/plugins/deviceshare/eventhandler_pod_test.go#L42)) keeps every existing deviceshare test on the legacy wiring, so nothing in that package exercises dispatcher → `OnPod*` → cache population. One case driven through `suit.start()` plus a real informer (with pre-existing pods carrying allocation annotations) would lock in the new wiring, including the initial-list path.

---

Nothing above touches the core idea, which I think is sound. Items 1 and 2 look like real bugs to me and would be worth fixing in this PR; 3 and 4 are mostly "decide and document", and the rest are polish.

mdryaaan:
Thanks for the thorough review @ZiMengSheng - items 1 & 2 were real, fixed in this round (`72a2545f` behavior + `d023dd0f` docs).

**1 (forget handler):** you're right - `forgetPodHandlers` is per-extender and `Start()` only saw the first profile's handle. Moved the registration to `New()` (per profile; `forgetPod` is idempotent), matching the convention the other plugins follow, and documented on `SharedPluginCache` that `Start()` may only use factory-scoped dependencies. Added a test that invokes `ForgetPod` through a *second* profile's extender and asserts the shared cache's marker is cleared.

**2 (reservation ledger leak):** confirmed - the synthetic reservation events were on the legacy `onPod*`→`updatePod`/`deletePod` path, so `releaseAssumed` never fired and the marker leaked. Routed them through `OnPodAdd/OnPodUpdate/OnPodDelete`. I traced `NewReservePod` to confirm the swap is clean: the reserve pod's UID = the reservation UID (matches the marker key), `Spec.NodeName` = `status.nodeName` (empty until scheduled), and an expired/failed reservation surfaces as `Phase=Failed` — so this also fixes a marker leak on reservation *expiry*, not just delete.

**3 (marker survives every positive event):** documented all three consequences on the `CacheReserver` interface, and corrected the metric doc  - it now describes "assumed pods still alive" rather than a leak gauge (a leak now shows up as entries for pods that no longer exist, best detected by diffing the ledger against live pods). On **3b (divergent-node undercount)** I'd rather flag a design tension than pick a side unilaterally: the "never re-add on a positive event" rule comes from @saintube's point that `spec.nodeName` may be a faked arbitration node that must not be trusted, so crediting the event's node on a positive event is exactly what that rule avoids — at the cost of the undercount you note if the bind is real. @saintube @ZiMengSheng — how would you like to resolve this? Options I see: (a) keep suppress + document (current); (b) credit the event's node only once `ForgetPod`/a negative event has confirmed the reserved node is released; (c) a trusted "real bind" signal. Happy to implement whichever you agree on.

**4 (docs contradict code):** fixed the `CacheReserver` comment and the PR description — §Ⅰ/§Ⅲ rewritten to the rollback-marker model, and §Ⅳ updated (incl. the 7→19 commit count).

**5 (smaller items):** AssumePod failure now logs a warning and returns nil (degrade to annotation semantics) instead of failing the pod; `getOrRegisterSharedCache` panics on registration after `StartSharedCaches`; added a plugin-level test driving the real informer's initial list → dispatcher → `OnPodAdd` → cache population. On the metric-in-critical-section nit  I softened the "no global write lock" wording; I left the gauge call where it is since it's a per-Reserve cost, but happy to move it out if you'd prefer.

saintube:
@mdryaaan 
On the multi-scheduler arbitration scenario - what I meant is: yes, spec.nodeName may point to an arbitrating node, but the right reaction isn't to skip accounting. The new transition from a non-empty nodeName to an empty nodeName should be considered. And when the nodeName is reset, the plugin cache should forget the pod.
This is analogous to preemption nomination in the Kubernetes scheduling framework: the nominated node is accounted for, and cancelling the nomination should trigger the rollback - rather than not accounting for the nomination at all.
So the "divergent bind node" that worries about does not occur by design: the positive event's node always equals the Reserve's node, which means "never re-add on a positive event" is exactly right. It seems (a) is better - keep suppress and document. The failure path is already covered by the negative events, and rare coalesced A->B case falls back to the plugin's existing nodeName-change handling.

The core principle here: the plugin already handles nodeName transition today, so this refactor should stay behavior-compatible rather than inventing a new "real bind" signal. We can revisit (c) only if the arbitration design later introduces a genuine node change.

mdryaaan:
Following up to close the item 3b design tension I raised in https://github.com/koordinator-sh/koordinator/pull/3125#issuecomment-5310555034

Thanks @saintube, that clears it up, the preemption-nomination framing is the right mental model and resolves the tension cleanly: since the arbitrating node always equals the Reserve node, there's no divergent bind, so "never re-add on a positive event" stands and the undercount can't arise.

The forget-on-reset mechanism you describe is already in place  `OnPodUpdate` treats a `spec.nodeName` reset from non-empty back to `""` as a negative event and forgets the pod:

```go
unassigned := oldPod != nil && oldPod.Spec.NodeName != "" && newPod.Spec.NodeName == ""
if unassigned || util.IsPodTerminated(newPod) {
    if !n.releaseAssumed(newPod) { n.updatePod(oldPod, newPod) }
    return
}
```

`releaseAssumed` rolls back the Reserve snapshot and clears the marker; a rare coalesced A→B for a pod whose marker has already cleared falls back to `updatePod`'s ordinary nodeName-change handling — so it stays behavior-compatible, no new "real bind" signal.

Going with option (a). I've tightened the docs to match in `e0fe9929`: dropped the "divergent-bind undercount" framing from the `CacheReserver` comment and the `DivergentNode` test, and documented the preemption-nomination model instead.

@ZiMengSheng — this closes the item 3b you raised; the negative-event (nodeName reset / ForgetPod) path handles the failure case, no positive-event crediting needed.

saintube:
@mdryaaan 
Overall, the rollback-marker model looks solid. I've verified that unassign-as-negative, releaseAssumed snapshot rollback (with isValid idempotency), and the A→B fallback after the marker clears are all working as expected. Startup ordering and cross-profile sharing are also wired correctly, and -race is green locally on both frameworkext and deviceshare.

There is one new correctness issue that needs attention before merge, plus two smaller items.

1. High Priority — Fix before merge
Same-name pod recreation breaks the marker invariant, leading to double-allocated minors (plugin.go#L546, device_cache.go#L635-659)
The core issue is a mismatch in keying: allocateSet is keyed by NamespacedName, the marker by UID, but copyPodAllocations reads by name. This causes problems if a pod with the same ns/name is recreated (e.g., via StatefulSet) while the delete event for the old UID hasn't reached the dispatcher yet. Here’s the problematic sequence:
1. Reserve(UID2): The add is skipped because isValid sees UID1's stale entry still holding the name slot. However, AssumePod(UID2) still snapshots that stale entry and publishes the marker.
2. delete(UID1): releaseAssumed removes the entry. Consequently, UID2's subsequent bound event gets suppressed by isAssumed(UID2) and is never re-added.

Result: UID2's actual device usage becomes invisible to the cache for its entire lifetime, making those minors appear free. Before this PR, the annotation re-add mechanism would self-heal this, but since positive-event suppression removed that safety net, this is now a regression. It’s a narrow window (recreation + listener lag), but it’s exactly the kind of bug class DeviceShare cannot tolerate.

Suggested fix (pick one):
- Make updateCacheUsed report whether the add actually landed, and skip AssumePod if it didn’t (this degrades gracefully to pre-PR annotation semantics).
- Alternatively, have AssumePod validate the snapshot against the expected allocation before publishing.

Please also add a UT covering the late delete(UID1) + Reserve(UID2) scenario.

2. Medium Priority — Fix or document
Unreserve returns early when nodeDeviceInfo == nil (plugin.go)
Currently, this skips ForgetPod, leaving the marker dangling until the pod's terminal event while positive events remain suppressed. Since the subtract operation is already a no-op here, I’d suggest calling ForgetPod before the nil check. If the current behavior is intentional, please add a comment explaining why the dangling marker is harmless.

3. Low Priority — Non-blocking
Potential race in StartSharedCaches (framework_extender_factory.go)
Technically, a registration happening between the snapshot build and sharedCachesStarted = true could miss both the snapshot and Start() without triggering the panic guard. I know this is unreachable today due to serial profile construction, so no change is needed right now. Just flagging it for future hardening—we could fold the flag into the first critical section if we ever revisit this.

mdryaaan:
Thanks @saintube — all three addressed (`e502703b` fix + `6bb48a6d` docs).

**High: same-name recreation.** Confirmed and fixed via option 1. `updateCacheUsed` now reports whether the add actually landed; `Reserve` skips `AssumePod` when it didn't, so no marker is published over the stale by-name snapshot, and it degrades to pre-PR annotation semantics — the informer re-add credits the recreated pod once the old ns/name slot frees. Added `Test_Plugin_Reserve_SameNameRecreation_NoStaleMarker` driving the full late-`delete(UID1)` + `Reserve(UID2)` sequence through `Plugin.Reserve`; confirmed it **fails on the pre-fix code** (stale marker published, UID2's usage invisible for its lifetime) and passes with the fix.

**Medium:  Unreserve early return.** Not intentional — moved `ForgetPod` above the `nodeDeviceInfo == nil` check so the marker is cleared even when the node's cache entry has GC'd (the subtract stays a no-op there). Test covers assume → node removed → Unreserve → marker cleared.

**Low: StartSharedCaches window.** Agreed it's unreachable today (serial profile construction; every `getOrRegisterSharedCache` from `New()` completes before `StartSharedCaches`). Kept the logic as-is and added an invariant comment, noting any future hardening must preserve the fail-fast property (the started flag is set only after registration succeeds, so it can't simply be folded into the first critical section).

`-race` green on deviceshare + frameworkext; golangci-lint clean.

saintube:
/lgtm
PTAL /cc @ZiMengSheng

ZiMengSheng (commented):
Re-reviewed at `5639cd8`. Thanks for the thorough follow-ups — I verified each of them against the code rather than taking the summary at face value, and they all genuinely landed:

- `ForgetPodHandler` is now registered per-profile in `Plugin.New()`, and `Start()` carries an explicit NOTE that it must not be registered there. This is the right split: `frameworkExtenderImpl.ForgetPod` (`pkg/scheduler/frameworkext/framework_extender.go:894-917`) only walks *that* extender's `forgetPodHandlers`, so per-profile registration really is required for multi-profile setups.
- Reservation events now flow through `registerReservationEventHandler` → `onPod*` → the unified `OnPodAdd/OnPodUpdate/OnPodDelete`, so they consult the assumed ledger. I also checked the `FilteringResourceEventHandler` semantics you relied on: a reservation going valid→invalid does emit `OnDelete(oldObj)`, so the reserve pod's marker is released. That reasoning holds.
- `Unreserve` now calls `ForgetPod` before the `nodeDeviceInfo == nil` check (`plugin.go:699`), so a GC'd node entry can no longer leave a dangling marker.
- The `StartSharedCaches` registration window is documented as an invariant, and `cmd/koord-scheduler/app/server.go:250` is the only production caller, so there is no other entry point that could leave caches unstarted.

A few things I suspected and then ruled out, for the record: the unassign detection (`old.Spec.NodeName != "" && new.Spec.NodeName == ""`) matches the existing convention in `nodenumaresource` and `reservation`; and `frameworkext/services/services.go:87-101` dedups providers by `plugin.Name()`, so the new `/assumedAllocations` route cannot double-register and panic gin under multiple profiles.

Build, `go vet`, and `go test` are green locally across `pkg/scheduler/frameworkext/...`, `pkg/scheduler/plugins/deviceshare/...` and `pkg/scheduler/metrics/...`.

One new item below that I think is worth handling before merge, plus one pre-existing issue and three nits.

---

### 1. The same-name-recreation fix is only half closed: `updateCacheUsed` reports "any device type landed", not "all"

`pkg/scheduler/plugins/deviceshare/device_cache.go:139` — `applied` starts at `false` and is set to `true` as soon as *one* device type gets through `isValid`. So for a pod requesting two device types where the held ns/name slot only covers one of them, the result is `true` and `Reserve` publishes a marker anyway — over a `copyPodAllocations` snapshot that, because it reads by ns/name, contains the *other* pod's allocation for the colliding type.

Concretely: the ns/name slot holds `{gpu}` from the previous pod; the recreated pod requests `{gpu, rdma}`. The GPU add is skipped by `isValid`, the RDMA add lands, `applied == true`, and the published snapshot is `{rdma: <own>, gpu: <the other pod's>}`. That is the same class of bug the high-priority item was about.

I confirmed this by execution rather than by reading: a throwaway test on this branch produced `applied = true`, snapshot device types `[rdma gpu]`, and stale-GPU-present `true`.

The doc comment states the assumption explicitly — "Same-ns/name pods request the same device types in practice, so the add is all-or-nothing" — but nothing enforces it, and the return value implements "any" rather than "all". Making it actually all-or-nothing removes the need for the assumption:

```go
func (n *nodeDevice) updateCacheUsed(deviceAllocations apiext.DeviceAllocations, pod *corev1.Pod, add bool) bool {
	if len(deviceAllocations) == 0 {
		return false
	}
	applied := true
	for deviceType, allocations := range deviceAllocations {
		if !n.isValid(deviceType, pod.Namespace, pod.Name, add) {
			applied = false
			continue
		}
		n.updateDeviceUsed(deviceType, allocations, add)
		n.resetDeviceFree(deviceType)
		n.updateAllocateSet(deviceType, allocations, pod, add)
	}
	return applied
}
```

I applied exactly this locally and the full existing `deviceshare` + `frameworkext` suites still pass, so it should be a cheap change.

Related test gap: `Test_Plugin_Reserve_SameNameRecreation_NoStaleMarker` (`plugin_test.go:415`) only uses `gpuAllocations(...)`, i.e. a single device type, which is exactly the full-collision case. Extending it to a multi-device-type pod would pin the partial case.

### 2. Pre-existing (not a regression from this PR): `Reserve` skips the add but `Unreserve` still subtracts

When `applied == false`, `Reserve` correctly declines to publish a marker — but `Unreserve` still runs `updateCacheUsed(..., false)` at `plugin.go:707` unconditionally. That subtract is not skipped by `isValid` (the ns/name slot exists), so it evicts the **other** pod's `allocateSet` entry. Once that happens, the other pod's real delete event is skipped by `isValid` and its `deviceUsed` is never released.

Verified by execution as well: after the second pod's `Unreserve`, the first pod's `allocateSet` entry is gone, and after the first pod's delete its 100 gpu-core on minor 0 is still accounted as used.

To be clear about attribution: I diffed against the base and the add/subtract asymmetry has the same shape there (just without a return value), so this is **not** introduced by this PR. I'm raising it because this PR is the first time the `applied` signal exists, which makes it cheap to fix — either thread `applied` through `cycleState` so `Unreserve` skips the subtract symmetrically, or, if you'd rather keep the change minimal, adjust the comment in `Reserve`. As written it claims a clean "degrade to annotation semantics", which is only true for the marker; the cache write path still diverges.

### Nits

- `pkg/scheduler/metrics/metrics.go:180` — `SharedCacheAssumedPods` omits `StabilityLevel: metrics.ALPHA`, while the neighbouring recent metrics (`GangScheduleCycleDuration`, `ReservationSelectorIndexQueryTotal`, `ReservationSelectorIndexCandidates`, `InlineBatchScheduleDuration`) all set it.
- `device_cache.go:687` — `getAssumedAllocationsSummary` copies the map but hands out `a.allocations` by reference, so lock-protected state escapes the `RLock`. It's safe today (every reader treats it as read-only and gin serializes after the lock returns), but a `DeepCopy` would make it obviously safe rather than safe-by-convention.
- `pkg/scheduler/frameworkext/framework_extender_factory.go:642-649` — the two `TODO` blocks are attached to `dispatchCaches`'s doc comment but describe the `dispatch*` methods below it; worth moving so they don't read as caveats on the snapshot accessor.

Overall the design is in good shape and the invariants are now written down where a future reader will find them. Item 1 is the only one I'd like to see before merge.

mdryaaan:
@ZiMengSheng Thanks for running these down by execution — both landed exactly as you traced.

**1 (all-or-nothing).** Applied your `updateCacheUsed` rewrite: `applied` starts `true` and is cleared on any type that `isValid` skips, so a partial collision now returns `false` and `Reserve` publishes no marker. Dropped the "same-ns/name pods request the same device types" assumption from the doc since the code no longer leans on it. Extended the recreation test into a separate `..._PartialCollision_NoStaleMarker` case ({gpu} held, pod requests {gpu, rdma}) . it fails against the old "any" return (marker published over the stale gpu snapshot) and passes now.

**2 (Unreserve subtract).** Fixed rather than documented. Threaded the result through `preFilterState.deviceCacheReserved` (set only when Reserve actually wrote + assumed), and `Unreserve` reverts the cache only when it's set, so a collision that skipped the add no longer subtracts the other pod's `allocateSet` entry. `Test_Plugin_Unreserve_SkipsSubtractOnCollision` pins it (verified it fails without the gate — the other pod's slot gets evicted). One edge I kept as-is: in a *partial* multi-type collision the type that did land isn't re-subtracted, which is consistent with the same-device-types assumption above and doesn't occur for same-ns/name pods in practice.

**Nits.** All three: `StabilityLevel: metrics.ALPHA` on the gauge; `getAssumedAllocationsSummary` now returns a deep copy so ledger state doesn't escape the `RLock`; moved the two dispatch TODOs onto the `dispatch*` methods.

ZiMengSheng (commented):
Re-reviewed at `48e7b8f0`. Thanks for turning all five items from the last round around, and for adding the two named regression tests — I verified each one against the code:

- `updateCacheUsed` now reports "all landed" instead of "any landed" (`pkg/scheduler/plugins/deviceshare/device_cache.go:140`).
- `preFilterState.deviceCacheReserved` (`pkg/scheduler/plugins/deviceshare/plugin.go:111`) gates the `Unreserve` subtract at `pkg/scheduler/plugins/deviceshare/plugin.go:718`, so a skipped add no longer evicts the other pod's `allocateSet` entry.
- `getAssumedAllocationsSummary` deep-copies before leaving the lock.
- The metric carries `StabilityLevel: metrics.ALPHA`.
- Both `TODO`s moved next to the `dispatchPodAdd` they actually describe.

One new issue, introduced by the interaction of the first two fixes, plus one stale comment.

---

### 1. Partial collision now leaks the half that did land

Three pieces no longer line up on the `Reserve` path:

1. `updateCacheUsed` applies the non-colliding device types *before* returning `false` (`device_cache.go:145-153`) — the return value is "all", but the write is still partial.
2. `Reserve`'s `!applied` branch returns immediately without reverting that partial write (`plugin.go:560-569`).
3. `Unreserve` is gated on `deviceCacheReserved`, which this path never sets, so the subtract is skipped (`plugin.go:718`).

So when a pod requests `{gpu, rdma}` and only the `gpu` slot is held by a not-yet-deleted same-ns/name pod, the `rdma` allocation is written into `deviceUsed`/`allocateSet` and nothing ever takes it back out. `gcNodeDevice` only drops whole nodes whose `Node` object disappeared; it does not reap orphaned `allocateSet` entries, so there is no backstop — the minor stays accounted for a pod that never existed until the scheduler restarts.

I reproduced it locally (uid-1 holds only the gpu slot for `default/sts-0`; uid-2 with the same ns/name requests `{gpu, rdma}`, gets `Reserve`d, then `Unreserve`d):

```
[A] uid-2 assumed = false                       <- marker correctly skipped
[B] after Reserve:   rdma minors = [3]
[C] after Unreserve: rdma minors = [3], deviceUsed[rdma] = map[3:...]
[E] after uid-1 delete: gpu = [], rdma = [3]    <- permanent residue
```

`Test_Plugin_Reserve_SameNameRecreation_PartialCollision_NoStaleMarker` covers the marker correctly, but only asserts on the marker — it does not look at what the skipped `Reserve` left behind in the cache.

The fix should stay confined to `Reserve`. The partial-application behaviour of `updateCacheUsed` is what the informer path wants — `eventhandler_pod.go:133`, `:137`, `:165` and `rollbackAssumed` (`device_cache.go:799`) all ignore the return value and rely on best-effort self-healing — so making it globally atomic would be a regression there. A precheck that `Reserve` gates on works and keeps the informer semantics intact:

```go
// canApplyAll reports whether every requested device type can be applied for this pod, i.e. no
// ns/name slot in allocateSet is already held for any of them. Callers that need an
// all-or-nothing write must gate on this instead of inspecting updateCacheUsed's return value:
// updateCacheUsed applies the non-colliding types before reporting false, which is the desired
// best-effort behaviour on the informer path but leaves an unrevertable partial write on the
// Reserve path. Must be called with n.lock held for write (isValid lazily initializes
// allocateSet[deviceType]).
func (n *nodeDevice) canApplyAll(deviceAllocations apiext.DeviceAllocations, pod *corev1.Pod, add bool) bool {
	if len(deviceAllocations) == 0 {
		return false
	}
	for deviceType := range deviceAllocations {
		if !n.isValid(deviceType, pod.Namespace, pod.Name, add) {
			return false
		}
	}
	return true
}
```

```go
	nodeDeviceInfo.lock.Lock()
	applied := nodeDeviceInfo.canApplyAll(state.allocationResult, pod, true)
	if applied {
		nodeDeviceInfo.updateCacheUsed(state.allocationResult, pod, true)
	}
	nodeDeviceInfo.lock.Unlock()
```

With that applied, `[B]` becomes `rdma minors = []` and `[D]` no longer reproduces, the marker is still correctly skipped, and `deviceshare` / `frameworkext/...` / `metrics` stay green. Worth extending `..._PartialCollision_NoStaleMarker` to assert the cache is untouched, not just that no marker exists — that is the assertion that would have caught this.

---

### 2. nit: stale justification in the AssumePod rollback comment

`plugin.go:574-575` still says the double subtract is *"safe if Unreserve also runs: the isValid guard skips the second subtract"*. That reasoning predates this round — what actually prevents the second subtract now is `deviceCacheReserved`, which this path leaves `false` (it is only set at `plugin.go:584`, after `AssumePod` succeeds). `isValid` is no longer the reason, and citing it will mislead the next reader into thinking the flag is redundant.

---

Baseline at this head is green for me: build, vet, and the `deviceshare` / `frameworkext/...` / `metrics` packages all pass. Also noting the `lgtm` label was dropped by this push.

mdryaaan:
Good catch, the two fixes did line up into a leak on the partial-collision path, and your `[B]/[C]` trace is exactly what the test now reproduces.

**1.** Added `canApplyAll` as you suggested and gated `Reserve` on it (`applied := canApplyAll(...); if applied { updateCacheUsed(...) }` under the one lock hold), so a partial collision now writes nothing rather than leaking the non-colliding type. Left `updateCacheUsed` best-effort — confirmed `updatePod`/`deletePod`/`rollbackAssumed` all rely on that partial application, so making it atomic would regress the informer path — and retouched its doc to point all-or-nothing callers at `canApplyAll` instead of its return value. Extended `..._PartialCollision_NoStaleMarker` to assert the cache is untouched, not just the marker: it fails against the old code (`rdma minors = [3]` residue after the skipped Reserve) and passes now.

**2.** Fixed the stale comment,  the second subtract is now prevented by `deviceCacheReserved` staying false on the rollback path (set only after `AssumePod` succeeds), not by `isValid`.

ZiMengSheng:
Re-reviewed the current head (`79967cd0`), with extra focus on the post-LGTM changes around the shared-cache lifecycle and DeviceShare's assumed-allocation ledger.

Confirmed:

- the GC startup no longer relies on the implicit `WaitForHandlersSync`/node-handler coupling; it is now launched from `AfterAllInformersSynced`, with the old wait retained only as defense in depth;
- `Reserve` uses `canApplyAll` before writing, so a same-name partial collision cannot leave a non-colliding device type behind;
- an `AssumePod` failure rolls back the preceding cache write, while `deviceCacheReserved` stays false so a later `Unreserve` cannot subtract twice;
- the per-profile `ForgetPod` registration and Reservation synthetic-event routing both reach the single shared cache;
- late cache registration fails fast, event dispatch uses the immutable lock-free snapshot, and handler-registration failure prevents startup instead of leaving a partially wired scheduler.

I also re-ran:

```text
go build ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... ./cmd/koord-scheduler/...
go test ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... -count=1
go test -race ./pkg/scheduler/frameworkext/... ./pkg/scheduler/plugins/deviceshare/... -count=1
```

All passed. I do not see a remaining blocking issue.

/lgtm
/approve

PR #3130, opened 2026-07-20
scheduler: FindOneNode returns full plan and triggers inline batch sc…

ZiMengSheng:

Add a new `FindOneNode` extension point and an inline batch scheduler that places and binds a whole job (all member pods) atomically within the triggering pod's scheduling cycle, instead of replaying the cached topology/gang plan pod-by-pod across separate cycles.

#### Motivation and value

Under the existing per-pod cycle, scheduling an N-pod gang with a topology-aware plan requires N separate scheduling cycles: the first pod runs the topology solver and caches the plan, and the remaining N-1 pods replay the cached plan one at a time. This PR targets two concrete problems with that model.

**1. Scheduling latency for large-scale jobs.**
For a job with N member pods (typical AI/HPC training jobs range from tens to thousands of pods), the end-to-end scheduling latency scales linearly with N: every member pod goes through its own queue pop, PreFilter, Filter, Score, Reserve, Assume, Permit, PreBind, Bind, PostBind. The inline batch path collapses the entire *scheduling cycle* (PreFilter / Filter / Reserve / Assume) for the whole job into a single cycle running under the framework `Parallelizer` (honoring the operator's `KubeSchedulerConfiguration.Parallelism`), and keeps only the *binding cycle* per pod, dispatched asynchronously as one goroutine per pod. The result:
- The number of scheduling cycles required to place a job drops from `O(N)` to `O(1)`.
- The trigger pod's cycle no longer contends with siblings that were previously popped one by one behind it in the queue.
- The new `InlineBatchScheduleDuration` metric exposes the end-to-end time per job so the improvement is directly observable.

The larger the gang, the larger the win — this is the main lever for improving scheduling throughput on training-style workloads. As a concrete data point, a **5000-pod Job** scheduled with **whole-machine Reservation, network-topology-aware placement, DeviceShare, and CPU binding (fine-grained core allocation) all enabled** goes from **~90s down to under 10s** end-to-end — roughly a **9x** improvement — because the expensive filter/allocation work that previously ran once per member pod across thousands of cycles is now done once for the whole job under the parallelizer, leaving only the per-pod binding to run asynchronously.

**2. Atomicity of the FindOneNode result, even after Gang is satisfied.**
Once `FindOneNode` (currently implemented by coscheduling's network-topology solver) produces a whole-job plan of pod→node placements, the plan is only valid *as a set*: the topology guarantees (must-gather layer, per-layer offer slots, per-node pod counts, gang co-placement) hold only when every member pod lands on the exact node chosen by the solver. In the per-pod replay path, sibling pods stay queued while the plan is consumed one pod at a time, and even after the gang's `MinMember` is nominally satisfied by the plan, the plan itself is *not* actually placed until each sibling has been popped and re-run through Filter/Reserve. Between plan generation and each sibling's Reserve, arbitrary other things can happen: other jobs consume slots on planned nodes, preemption changes the snapshot, node conditions/taints/capacity shift, and Filter can reject a planned pod against a now-stale view. Any of these turns the topology plan into a partial success and forces the gang to retry from scratch on the next cycle.

This PR closes that window entirely. Once `FindOneNode` returns a `BatchScheduleResult`, the inline batch scheduler runs `PreFilter → Filter → Reserve → Assume` for *every* member pod against a per-job node snapshot inside the trigger pod's cycle, before any other scheduling cycle can start. The whole plan is either atomically reserved or atomically rolled back (`Unreserve` + `ForgetPod` on any failure). Sibling pods that stay in the scheduling queue are skipped once they are already `Assumed`, and async binding goroutines carry each pod through Permit/PreBind/Bind/PostBind independently — mirroring the upstream `schedule_one.go` binding cycle so gang Permit, network-topology finalization, and preemption continue to work unchanged. The `EnableBatchScheduleNodeSnapshot` gate additionally isolates each job's Reserve view from unrelated concurrent binding.

#### Key changes

- frameworkext: add the `FindOneNodePlugin` extension point. On `Success` it returns a `BatchScheduleResult` (all member pods + pod→node plan); on `Skip` it falls back to standard multi-node filtering. `RunPreFilterPlugins` wires the FindOneNode → inline-batch trigger, and a `BatchScheduledErrorHandler` filter suppresses the default failure handling for batch-scheduled pods.
- batch: add the whole-job scheduling engine (PreFilter / Filter / Reserve / Assume per member pod) and `BatchScheduler` that binds pods asynchronously (one goroutine per pod), with a node snapshot, request grouping/validation, and assumed-pod cleanup on failure. Correctly interacts with gang Permit / WaitOnPermit, preemption, and the scheduling queue (only the trigger pod is dequeued; siblings stay queued and are skipped once assumed).
- coscheduling / deviceshare / nodenumaresource: implement/adjust `FindOneNode` and clear designated allocations on the inline-batch path so the normal allocation runs; `MakeNominatedPodsOfTheSameJob` now takes a `sets.Set[string]`.
- server: wire the batch scheduler (`sched.Cache`, `sched.FailureHandler`) and register the batch-scheduled error handler filter.
- features: gated behind `EnableInlineBatchSchedule` (Alpha, off) and `EnableBatchScheduleNodeSnapshot` (Beta, on); default behavior is unchanged when disabled. Add the `InlineBatchScheduleDuration` metric.

Ported to the `k8s.io/kube-scheduler/framework` API (k8s 1.35): moved types to `fwktype`, 3-value `RunPreFilterPlugins`, `Status.WithPlugin`, `NodeInfo.AddPodInfo`, etc. `InvalidNodeInfo` is routed through the frameworkext `SchedulerCache` wrapper (reimplemented on public upstream via `AddPod`/`RemovePod`), and the unified/arbitrator variants are excluded on this branch.

### V. Checklist

### DIFF
--- a/docs/proposals/scheduling/20260708-findonenode-inline-batch-scheduling.md
+++ b/docs/proposals/scheduling/20260708-findonenode-inline-batch-scheduling.md
@@ -1,0 +1,209 @@
+# FindOneNode: Whole-Job Placement Plan and Inline Batch Scheduling
+
+## Summary
+
+This proposal introduces a new scheduling extension point, **`FindOneNode`**, that lets a plugin
+compute a deterministic placement plan for a **whole job** (all member pods and their target
+nodes) in one shot, and a new **inline batch scheduler** that schedules and binds all of those
+pods together within the triggering pod's scheduling cycle.
+
+Previously the scheduler could only decide one node for one pod at a time. For gang/topology
+workloads this is inefficient and can be incorrect: the topology algorithm already knows the best
+placement for the entire `GangGroup`, but the framework forced it to be replayed pod-by-pod.
+`FindOneNode` returns the full plan up front, and the inline batch scheduler
+reserves/assumes/binds every member pod according to that plan, with per-pod binding running
+asynchronously.
+
+This is gated behind feature gates and defaults to the legacy single-node behavior, so it is
+opt-in and safe.
+
+## Motivation
+
+- The network-topology / gang algorithm computes a placement for all pods of a job at once and
+  caches it (`gangSchedulingContext.networkTopologyPlannedNodes`), but the standard
+  `PreFilter → Filter → Reserve` flow only consumes one node for the current pod. The plan is not
+  recomputed for siblings — each sibling's `PreFilter` just looks up its cached planned node — but
+  every sibling must still be popped from the queue and run its own full scheduling cycle
+  (`Filter → Reserve → Bind`) across separate cycles. Re-validating the cached plan against a
+  snapshot that changes between those cycles is slow and can let the realized placement diverge
+  from the original plan.
+- We want a single, authoritative "schedule the whole job now" path that reuses a shared whole-job
+  scheduling engine, keeps binding asynchronous (so the scheduling cycle stays fast), and
+  never regresses the default single-pod behavior when disabled.
+
+### Goals
+
+- A `FindOneNode` extension point that returns a whole-job placement plan (`BatchScheduleResult`).
+- An inline batch scheduler that, on a successful plan, runs PreFilter/Filter/Reserve/Assume for
+  every member pod and then binds them asynchronously (one goroutine per pod).
+- Correct interaction with gang `Permit`/`WaitOnPermit`, preemption, and the scheduling queue
+  (only the triggering pod is dequeued; siblings stay queued and are skipped once assumed).
+- Feature-gated and off by default; when off, behavior is identical to today.
+
+### Non-Goals
+
+- Changing the topology/gang placement algorithm itself (the plugin still computes the plan).
+- Decoupling package dependencies (tracked separately).
+- Synchronous binding of the whole job inside the scheduling cycle.
+
+## Proposal
+
+### New extension point: `FindOneNode`
+
+`pkg/scheduler/frameworkext/interface.go`:
+
+```go
+// FindOneNodePlugin computes a placement plan for a whole job.
+type FindOneNodePlugin interface {
+    framework.Plugin
+    // On Success: returns a BatchScheduleResult (plan for all member pods incl. the trigger pod).
+    // On Skip:    the plugin does not intervene; fall back to standard multi-node filtering.
+    FindOneNode(ctx context.Context, cycleState *framework.CycleState, pod *corev1.Pod,
+        result *framework.PreFilterResult) (*BatchScheduleResult, *framework.Status)
+}
+
+// BatchScheduleResult is the whole-job placement plan.
+type BatchScheduleResult struct {
+    Pods          []*corev1.Pod     // all member pods, including the triggering pod
+    PodToNodeName map[string]string // pod key "namespace/name" -> planned node name
+}
+```
+
+The `FrameworkExtender` exposes `RunFindOneNodePlugin(...)`, invoked during the **PreFilter**
+phase. A plugin registers by implementing `FindOneNodePluginProvider`. The reference
+implementation is coscheduling's network-topology workflow
+(`PodGroupManager.FindOneNode` → `buildBatchScheduleResult` → `BatchScheduleResult{Pods,
+PodToNodeName}`).
+
+### Inline batch scheduler
+
+`pkg/scheduler/frameworkext/interface.go`:
+
+```go
+// BatchScheduler runs a scheduling + binding cycle for a whole job.
+// Implemented outside frameworkext (in pkg/scheduler/batch) and registered on the extender
+// to avoid an import cycle.
+type BatchScheduler interface {
+    BatchSchedule(ctx context.Context, ext FrameworkExtender, cycleState *framework.CycleState,
+        triggerPod *corev1.Pod, plan *BatchScheduleResult) *framework.Status
+}
+```
+
+The implementation lives in `pkg/scheduler/batch` and reuses the shared `Engine`
+(`RunSchedulingCycle` + async binding). It is wired in
+`cmd/koord-scheduler/app/server.go`:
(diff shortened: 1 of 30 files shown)


pkg/scheduler/batch/engine.go
```diff
@@ -0,0 +1,480 @@
...
+			// restores the reservation node info once (the first pod might be skipped, so we track the first
+			// valid pod) and the remaining pods skip it.
+			skipRestoreNodeInfo := nodeReservationRestored || !k8sfeature.DefaultFeatureGate.Enabled(features.EnableBatchScheduleNodeSnapshot)
+			schedulingHintForNode.Extensions[reservation.Name] = reservation.HintExtensions{
+				SkipRestoreNodeInfo: skipRestoreNodeInfo,
+			}
+			if !skipRestoreNodeInfo {
+				nodeReservationRestored = true
+			}
+			hinter.SetSchedulingHintState(cycleState, schedulingHintForNode)
```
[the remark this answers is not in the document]
ZiMengSheng: It's not a problem

saintube:
> Add a new FindOneNode extension point and an inline batch scheduler that places and binds a whole job (all member pods) within the triggering pod's scheduling cycle, instead of replaying the cached topology/gang plan pod-by-pod across separate cycles.
> 
> Key changes:
> 
> * frameworkext: add the FindOneNodePlugin extension point. On Success it returns a BatchScheduleResult (all member pods + pod->node plan); on Skip it falls back to standard multi-node filtering. RunPreFilterPlugins wires the FindOneNode -> inline-batch trigger, and a BatchScheduledErrorHandler filter suppresses the default failure handling for batch-scheduled pods.
> * batch: add the whole-job scheduling engine (PreFilter/Filter/Reserve/Assume per member pod) and BatchScheduler that binds pods asynchronously (one goroutine per pod), with a node snapshot, request grouping/validation, and assumed-pod cleanup on failure. Correctly interacts with gang Permit/WaitOnPermit, preemption, and the scheduling queue (only the trigger pod is dequeued; siblings stay queued and are skipped once assumed).
> * coscheduling/deviceshare/nodenumaresource: implement/adjust FindOneNode and clear designated allocations on the inline-batch path so the normal allocation runs; MakeNominatedPodsOfTheSameJob now takes a sets.Set[string].
> * server: wire the batch scheduler (sched.Cache, sched.FailureHandler) and register the batch-scheduled error handler filter.
> * features: gated behind EnableInlineBatchSchedule (Alpha, off) and EnableBatchScheduleNodeSnapshot (Beta, on); default behavior is unchanged when disabled. Add the InlineBatchScheduleDuration metric.
> 
> Ported to the k8s.io/kube-scheduler/framework API (k8s 1.35): moved types to fwktype, 3-value RunPreFilterPlugins, Status.WithPlugin, NodeInfo.AddPodInfo, etc. InvalidNodeInfo is routed through the frameworkext SchedulerCache wrapper (reimplemented on public upstream via AddPod/RemovePod), and the unified/arbitrator variants are excluded on this branch.
> 
> ### Ⅱ. Does this pull request fix one issue?
> ### Ⅲ. Describe how to verify it
> ### Ⅳ. Special notes for reviews
> ### V. Checklist
> * [ ]  I have written necessary docs and comments
> * [ ]  I have added necessary unit tests and integration tests
> * [ ]  All checks passed in `make test`

Recommend adding a brief summary of the performance improvements to highlight the value of this change.

docs/proposals/scheduling/20260708-findonenode-inline-batch-scheduling.md
```diff
@@ -0,0 +1,201 @@
...
+     (`hinter.IsBatchSchedulingCycle`), or the batch scheduler is not registered, or
+     `EnableInlineBatchSchedule` is off, or the plan is `nil`/has no node for the pod → fall back
+     to the single-node result (`PreFilterResult{NodeNames: {planned node}}`); an empty node name
+     surfaces as an `Error` status instead of restricting to node `""`.
+   - Otherwise (top-level cycle): call `BatchSchedule(...)` for the whole plan.
+     - On success: all pods are reserved/assumed/permitted; their bind cycles run
+       asynchronously. The trigger pod's own cycle returns
+       `Unschedulable(BatchScheduledReason)` to short-circuit — this is a sentinel "already
+       handled", not a real failure, and its failure handler is suppressed.
+     - On failure: assumed pods are unreserved and forgotten; a plain `Unschedulable` is upgraded
```
saintube, line 115: What if FindOneNode does not derive the proper plan where some pods fail in the Filter phase, do we need to add constraints for the batch scheduler about preemption?

ZiMengSheng: Good question. By design the inline batch scheduler runs only PreFilter/Filter/Reserve/Assume — it never invokes PostFilter/preemption. If FindOneNode returns a plan that turns out infeasible (a pod fails PreFilter/Filter/Reserve against the live snapshot), the engine fast-fails the whole job and rolls it back atomically (Unreserve + ForgetPod). For the trigger pod we upgrade Unschedulable → UnschedulableAndUnresolvable, so a batch failure can never trigger preemption; the job is simply retried on the next cycle / cluster event.

So we intentionally do not add whole-job preemption to the batch path: preempting victims for a job that may not fully fit is worse than retrying. Feasibility is the planner's (FindOneNode's) responsibility.

You're right that we should make this explicit. I'll add to the proposal: (1) a Non-Goal: "the inline batch cycle does not perform preemption; a job that needs preemption to fit is not handled by this path"; and (2) a note that when FindOneNode returns Skip, siblings fall back to the standard single-pod path where existing coscheduling PostFilter/preemption still applies unchanged.

PR #3131, opened 2026-07-20
koordlet: target NRI container resource updates

### LINKED ISSUE #3129: [BUG] Koordlet NRI cannot override kubelet UpdateContainerResources for the current container
**What happened**:

When Koordlet's NRI plugin handles `UpdateContainer`, it returns a `ContainerUpdate` containing the resource changes, but does not set `ContainerId`.

As a result, containerd cannot determine which container the update targets, so resource changes made by Koordlet's runtime hooks may not be applied.

This affects any scenario that relies on NRI `UpdateContainer` to modify container resources. One example is kubelet CPU Manager reconciling a SYSTEM container: Koordlet should preserve the System QoS CPUSet through NRI, but the returned update cannot be applied because it has no target container ID.

When container resources are changed, the returned `ContainerUpdate.ContainerId` should match the container ID from the current request, allowing containerd to apply the update to the correct container.

If no resource change is produced, NRI should not return an empty `ContainerUpdate`.

The problem can be reproduced reliably with a handler-level unit test:

1. Register a `PreUpdateContainerResources` hook that sets the container CPUSet.
2. Construct an `api.Container` whose ID is `raw-container-id`.
3. Pass the container to `NriServer.UpdateContainer`.
4. Inspect the returned `ContainerUpdate`.

The key assertion is:

```go
require.Equal(t, container.GetId(), updates[0].GetContainerId())
```

Actual result:

```text
returned CPUSet:       1,3
expected ContainerId: raw-container-id
actual ContainerId:   ""
```

The CPUSet change is present, but the update does not identify its target container.

**Runtime impact**:

This issue can affect nodes where kubelet CPU Manager uses the `static` policy.

The following values were collected from a previously inspected 96-CPU node:

```text
System QoS CPUSet:      0,47-48,95
kubelet defaultCPUSet:  0,5-48,53-95
```

To observe the runtime impact:

1. Enable NRI in containerd and enable Koordlet's NRI runtime hook.
2. Configure kubelet CPU Manager with the `static` policy.
3. Create a SYSTEM pod whose container does not request integer exclusive CPUs.
4. Continuously sample the container cgroup's `cpuset.cpus`.
5. Create or remove a Guaranteed pod that requests integer CPUs to trigger kubelet CPU Manager reconciliation.
6. Check whether the SYSTEM container temporarily uses kubelet's `defaultCPUSet`.

Using the values above, the unexpected result would be:

```text
System QoS CPUSet:          0,47-48,95
SYSTEM container CPUSet:    0,5-48,53-95
```

Koordlet's asynchronous cgroup reconciler may later restore the CPUSet to `0,47-48,95`. Therefore, the problem is reproduced if the container uses kubelet's `defaultCPUSet` at any point, even if it is restored later.

The issue is in the construction of the return value from the NRI `UpdateContainer` path.

Suggested regression coverage:

- When resources are changed, `ContainerUpdate.ContainerId` matches the current container ID.
- Updates for different containers contain their respective container IDs.
- When no resources are changed, no empty `ContainerUpdate` is returned.

**Environment**:

- Koordinator: `v1.8.0`
- Koordinator main: also reproduced at the inspected revision
- Kubernetes: `v1.28`
- containerd: `v1.7.27`
- NRI: `v0.11.0`
- kubelet CPU Manager policy: `static`

ZXYxc:

Target resource-bearing NRI `UpdateContainer` responses to the container from the current request. Also avoid returning an empty `ContainerUpdate` when runtime hooks do not change any resources.

Without the target container ID, containerd cannot apply Koordlet's resource changes to the requested container.

fixes #3129

Unit tests:

```text
go test ./pkg/koordlet/runtimehooks/nri -run '^(TestNriServer_UpdateContainer|TestNriServer_UpdateContainerTargetsCurrentContainer)$' -count=1
go test ./pkg/koordlet/runtimehooks/nri ./pkg/koordlet/runtimehooks/protocol -count=1
```

Both commands pass.

Runtime validation on a containerd 1.7.27 node with NRI enabled and kubelet CPU Manager `static`:

- Before the fix, a SYSTEM container configured with CPUSet `0-1` intermittently used kubelet's `defaultCPUSet`; 87 such samples were captured from the container's `/proc/self/status`.
- After deploying the patched Koordlet to that node, 1,000 consecutive samples over about 51 seconds all remained `0-1`, covering at least five kubelet reconciliation periods.

`make test` was attempted but did not complete in the local environment: the local Go toolchain reported `go: no such tool "covdata"`, and the envtest asset download returned HTTP 401. The focused affected-package tests above passed.

### V. Checklist


saintube (commented):
/lgtm

PR #3132, opened 2026-07-21
fix: support UsageWithPageCache and UsageWithHotPageCache in SystemMemoryUsage (#1730)

### LINKED ISSUE #1730: [BUG] SystemMemoryUsage should support UsageWithHotPageCache and UsageWithPageCache
**What happened**:
systemUsage is not correct when collect policy is UsageWithHotPageCache or UsageWithPageCache

**Environment**:
- App version:
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others:

pingxin403:
### What

SystemMemoryUsage calculation uses `nodeMemory - podMemory - hostAppMemory`, but the default nodeMemory value excludes page cache. When `NodeMemoryCollectPolicy` is `UsageWithPageCache` or `UsageWithHotPageCache`, the system memory calculation is inaccurate.

### How

- Added `metricCache` and `statesInformer` fields to `systemResourceCollector`
- Added `queryMemoryWithPolicy()` method to query the correct policy-aware node memory metric from the metric cache
- Updated `collectSysResUsed()` to use policy-aware memory values when the policy is set
- Renamed `appendableDB` → `metricCache` for consistency

### Related issue

Closes #1730


ZiMengSheng:
Reviewed the diff — the policy-aware query design is solid: every error path degrades to the default value instead of dropping the sample, and the 295-line test suite covers both policies plus the fallback. Two nits before merge:

1. **Please remove the `// ponytail: log once at setup...` comment** — it references the fork namespace rather than the upstream tree. The surrounding logic is fine to keep as-is.
2. Minor: `c.Collectors["PageCacheCollector"]` matches by registered name string; if the collector is ever renamed this silently degrades to the warning path. Acceptable for now, but a comment or a constant would make the coupling explicit.

With the comment removed, this is good to go from my side.

pingxin403:
> Reviewed the diff — the policy-aware query design is solid: every error path degrades to the default value instead of dropping the sample, and the 295-line test suite covers both policies plus the fallback. Two nits before merge:
> 
> 1. **Please remove the `// ponytail: log once at setup...` comment** — it references the fork namespace rather than the upstream tree. The surrounding logic is fine to keep as-is.
> 2. Minor: `c.Collectors["PageCacheCollector"]` matches by registered name string; if the collector is ever renamed this silently degrades to the warning path. Acceptable for now, but a comment or a constant would make the coupling explicit.
> 
> With the comment removed, this is good to go from my side.

---

done

PR #3134, opened 2026-07-21
test: verify pod throttling metrics collection

miantalha45:

This PR strengthens unit coverage for `PodThrottledCollector`.

Previously, `Test_podThrottledCollector_collectPodThrottledInfo` only verified that collection did not panic. This change verifies the collector’s observable output:

- one Pod CPU-throttling metric is stored;
- one container CPU-throttling metric is stored;
- both metrics report the expected throttling ratio (`0.5`) from the test cgroup CPU statistics.

The production collector logic is unchanged.

NONE

Run:

```bash
go test ./pkg/koordlet/metricsadvisor/collectors/podthrottled -run '^Test_podThrottledCollector_collectPodThrottledInfo$' -count=1
```

Expected result:

```text
ok   github.com/koordinator-sh/koordinator/pkg/koordlet/metricsadvisor/collectors/podthrottled
```

This is a test-only change. It adds assertions for metrics already generated by the collector and does not change production behavior.

### V. Checklist


ZiMengSheng:

/lgtm

PR #3136, opened 2026-07-22
test/perf: add Gang coscheduling scenario and GitHub Actions benchmark workflow 

Shivam-1827:

This PR delivers Phase 2 for the `koord-scheduler` benchmark harness at `test/perf/`, building on the Phase 1 foundation (engine loop, YAML config, CLI, baselines, regression detection, timeout, failure tracking — already merged).

Per the mentors' milestone table, it covers CI integration plus the Gang and ElasticQuota scenarios. Baseline storage and regression detection already shipped in Phase 1, ahead of schedule, so this PR covers the two remaining pieces for Week 5: the GitHub Actions workflow, and the Gang (coscheduling) scenario. ElasticQuota remains out of scope for this PR (Week 6).

Gang was implemented before ElasticQuota because `pkg/types.ScenarioConfig` already carries `GangSize`/`MinMember`, and `BenchmarkResult` already carries `GangCompletionP50Sec`/`GangCompletionP99Sec` — both added as placeholders in Phase 1 for exactly this scenario. Building Gang first exercises those existing fields instead of adding new ones on top of a still-unvalidated shape.

**What is added:**

| Path | What it is |
|---|---|
| `pkg/scenarios/gang/gang.go` | New Gang (coscheduling) scenario. `Setup` creates the namespace plus one `PodGroup` (`scheduling.sigs.k8s.io/v1alpha1`) per `cfg.PodCount / cfg.GangSize` chunk; `Pods` generates `cfg.PodCount` pods split evenly across those PodGroups, each labeled with the PodGroup's name so koord-scheduler's coscheduling plugin and the benchmark's own watcher can both associate a pod with its gang; `Teardown` deletes the PodGroups alongside the existing pod/namespace cleanup |
| `pkg/types/types.go` | New `PodGroupLabel` constant — the label key used to associate a pod with its PodGroup, shared between the `gang` scenario and `pkg/framework` so the watcher can group latencies by gang without importing a scenario package |
| `pkg/framework/watcher.go` | `Watcher` now records each pod's `GangID` (read from `PodGroupLabel`) alongside its latency; new `PodLatencies()` accessor exposes the full `[]types.PodLatency` (existing `Latencies()` is unchanged so nothing else is affected) |
| `pkg/framework/metrics.go` | New `ComputeGangCompletionPercentiles()` — groups latencies by `GangID`, takes each group's last-scheduled-member latency (a gang isn't complete until every member is placed), then computes P50/P99 across groups rather than across individual pods |
| `pkg/framework/engine.go` | Wires `ComputeGangCompletionPercentiles()` into `BenchmarkResult.GangCompletionP50Sec`/`P99Sec`; stays `nil` for scenarios (e.g. `basic`) whose pods carry no `PodGroupLabel` |
| `pkg/framework/metrics_test.go` | Unit tests for the gang aggregation math (no-gang input, single group, multiple groups with different last-member times) |
| `cmd/benchmark/main.go` | Registers the `gang` scenario via blank import, same pattern as `basic` |
| `configs/scenarios/gang-1k.yaml` | Config for 1,000 pods split into gangs of 20 (`minMember: 20`) on 100 kwok nodes |
| `.github/workflows/scheduler-benchmark.yaml` | New CI workflow: checkout → Go setup → `make -C test/perf setup` → `make -C test/perf benchmark` → fail the job if `thresholdBreached: true` in the result JSON → upload `result.json` as a build artifact → `make -C test/perf teardown` (always, even on failure) |

**How the Gang scenario fits into the existing engine loop:**

No changes to the `Scenario` or `NodeProvider` interfaces were needed — `Setup`/`Teardown` already receive a `dynamic.Interface`, which is all a CRD-based scenario requires. The engine's orchestration sequence (`CreateNodes → WaitReady → scenario.Setup → burst → watcher → metrics → scenario.Teardown → DeleteNodes → WriteReport`) is unchanged from Phase 1.

```text
scenario.Setup(ctx, client, dynClient, cfg, runID)
└─ create namespace (if absent)
└─ create ceil(podCount / gangSize) PodGroup objects via dynClient
   each: spec.minMember = cfg.MinMember, spec.scheduleTimeoutSeconds = 300

scenario.Pods(cfg, runID)
└─ generate cfg.PodCount pods
   each labeled: benchmark.koordinator.sh/run-id=<runID>
                 pod-group.scheduling.sigs.k8s.io=<owning PodGroup name>

watcher.Start(...)
└─ per pod: latency = scheduled - created, GangID = pod.Labels[PodGroupLabel]

engine.Run(...)
└─ ComputeLatencyPercentiles(watcher.Latencies())           → P50/P90/P99 (unchanged, per-pod)
└─ ComputeGangCompletionPercentiles(watcher.PodLatencies()) → GangCompletionP50/P99 (new, per-group)

scenario.Teardown(...)
└─ delete PodGroup objects (DeleteCollection by run-id label)
└─ delete pods (unchanged from basic)
```

**Verified output from running against a live kind + kwok + Koordinator cluster:**

```
make -C test/perf benchmark
...
I0723 00:52:10.022655 engine.go:145 "Starting benchmark" scenario="basic" runID="6ec637a4-283f-45d9-adec-55d89bb246b9" timeout="10m0s"
I0723 00:52:10.027963 engine.go:150 "API server reachable" scenario="basic"
I0723 00:52:10.027987 engine.go:162 "Creating nodes" count=100 workers=20
I0723 00:52:10.160835 engine.go:176 "Nodes ready"
I0723 00:52:10.173820 engine.go:225 "Starting pod burst" pods=1000 concurrency=50
I0723 00:52:19.075367 engine.go:252 "API creation phase complete" duration="8.9s"
I0723 00:52:32.239318 report.go:88  "Benchmark result" scenario="basic" throughputPodsPerSec="45.32"
                                    apiCreationDurationSec="8.90s" totalDurationSec="22.06s"
                                    p50LatencySec="6.00s" p90LatencySec="12.00s" p99LatencySec="13.00s"

make -C test/perf benchmark SCENARIO=gang
...
I0723 01:01:02.733892 engine.go:145 "Starting benchmark" scenario="gang" runID="b51e7e8c-2605-4ec2-af94-ba0d41eb0812" timeout="10m0s"
I0723 01:01:02.883825 engine.go:176 "Nodes ready"
I0723 01:01:03.462147 engine.go:225 "Starting pod burst" pods=1000 concurrency=50
I0723 01:01:11.776286 engine.go:252 "API creation phase complete" duration="8.31s"
I0723 01:01:24.824745 report.go:88  "Benchmark result" scenario="gang" throughputPodsPerSec="46.81"
                                    apiCreationDurationSec="8.31s" totalDurationSec="21.36s"
                                    p50LatencySec="6.00s" p90LatencySec="12.00s" p99LatencySec="13.00s"
```

```bash
$ kubectl get podgroups -n benchmark
NAME                      AGE
bench-gang-b8874710-000   4s
bench-gang-b8874710-001   4s
...
bench-gang-b8874710-049   3s
```

50 PodGroups were created for 1,000 pods with `gangSize: 20` (1000 / 20 = 50), confirming the split-into-N-groups behavior works as designed.

```bash
$ cat test/perf/results/result.json | jq '{thresholdBreached, timedOut, schedulingFailureCount, gangCompletionP50Sec, gangCompletionP99Sec}'
{
  "thresholdBreached": false,
  "timedOut": false,
  "schedulingFailureCount": 0,
  "gangCompletionP50Sec": 8,
  "gangCompletionP99Sec": 13
}
```

`gangCompletionP50Sec`/`P99Sec` are populated (non-`null`) only for the `gang` scenario; a `basic` run leaves them `null` since no pod carries `PodGroupLabel`.

The reported numbers are from a development environment on a single-node kind cluster and are intended only to validate correctness of the implementation, not as representative performance numbers for `koord-scheduler`.

---

NONE — this is a new feature addition, not a bug fix.

---

**Option A — No cluster needed (build + unit tests):**

```bash
go build ./test/perf/...
go test ./test/perf/...
# ?   .../cmd/benchmark             [no test files]
# ok  .../pkg/framework             (cached)
# ?   .../pkg/scenarios/gang        [no test files]
# ... (all packages build/test clean)

go test ./test/perf/pkg/framework/... -run TestComputeGangCompletion -v
# --- PASS: TestComputeGangCompletionPercentiles_noGangs
# --- PASS: TestComputeGangCompletionPercentiles_empty
# --- PASS: TestComputeGangCompletionPercentiles_takesMaxPerGroup
# --- PASS: TestComputeGangCompletionPercentiles_singleGroup
# PASS
```

```bash
make -C test/perf setup

# basic scenario (regression check against Phase 1 behavior)
make -C test/perf benchmark

# gang scenario
make -C test/perf benchmark SCENARIO=gang
# or: make -C test/perf benchmark CONFIG=configs/scenarios/gang-1k.yaml

kubectl get podgroups -n benchmark   # confirm N PodGroups exist, matching podCount/gangSize
cat test/perf/results/result.json | jq '{thresholdBreached, gangCompletionP50Sec, gangCompletionP99Sec}'

make -C test/perf teardown
```

**Option C — CI workflow:**

`.github/workflows/scheduler-benchmark.yaml` triggers on `pull_request` (path-filtered to `test/perf/**`) and manually via `workflow_dispatch`. It runs `make setup` → `make benchmark BASELINE=baselines/basic-1k.json` → fails the job if `thresholdBreached: true` in the result JSON → uploads `result.json` as a build artifact → always tears down.

---

### Ⅳ. Special notes for reviewers

**1. Gang scenario shape**

Each benchmark run creates `ceil(podCount / gangSize)` separate `PodGroup` objects and splits pods evenly across them, rather than one giant `PodGroup` for the whole run. This is what makes `GangCompletionP50Sec`/`P99Sec` meaningful as percentiles (computed across groups) rather than degenerate single values, and reflects a more realistic "many gangs competing for nodes concurrently" load shape. Verified against a real cluster: `gangSize: 20`, `podCount: 1000` → 50 `PodGroup` objects, confirmed via `kubectl get podgroups`.

**2. Gang completion latency definition**

`ComputeGangCompletionPercentiles` takes, per `PodGroup`, the latency of that group's last-scheduled member (a gang isn't complete until every member is placed), then computes P50/P99 across groups. This reuses the existing per-pod `Watcher` rather than adding a second watch stream — `Watcher.PodLatencies()` is a new accessor alongside the existing `Latencies()`, which is untouched so nothing else in the engine changes behavior.

**3. PodGroup label key**

The coscheduling plugin association uses `pod-group.scheduling.sigs.k8s.io` (current scheduler-plugins `v1alpha1` convention), defined once as `types.PodGroupLabel` and shared between the scenario and the watcher. This has been verified end-to-end against a real koord-scheduler + kwok cluster (see `kubectl get podgroups` output above, and non-null `gangCompletionP50/P99Sec` in the result), confirming pods are correctly associating with their PodGroup rather than scheduling independently.

**4. No interface changes**

`Scenario` and `NodeProvider` are unchanged from Phase 1. `Setup`/`Teardown` already accepted a `dynamic.Interface`, which is all a CRD-based scenario like Gang needs.

**5. CI failure-on-regression lives in the workflow, not the engine**

`engine.Run()`'s existing contract is untouched — it still returns `nil` on a successful-but-regressed run, with `ThresholdBreached` set on the result. The workflow itself reads `result.json` with `jq` and exits non-zero when `thresholdBreached: true`, so Phase 1 callers of `engine.Run()` (e.g. any local scripting) keep their existing behavior.

**6. Explicitly out of scope for this PR**

- ElasticQuota scenario — Week 6
- Reservation, LoadAware scenarios — later Phase 2 weeks
- `baselines/gang-1k.json` — not yet committed; the CI workflow currently runs against `baselines/basic-1k.json` only, since a gang baseline should come from a dedicated real-cluster capture rather than the numbers in this PR's description
- 10k+ node scale, dynamic NodeMetric simulation — Phase 3 (stretch goals)

---

### Ⅴ. Checklist

  - `pkg/scenarios/gang/gang.go` documents the PodGroup-splitting design decision and the label-key choice at the package level
  - New exported symbols (`PodGroupLabel`, `Watcher.PodLatencies`, `ComputeGangCompletionPercentiles`) all have Go doc comments
  - `metrics_test.go`: `ComputeGangCompletionPercentiles` covering no-gang input, empty input, a single group, and multiple groups with different last-member completion times
  - Integration-verified against a live kind + kwok + Koordinator cluster: `basic` and `gang` scenarios both run to completion, `kubectl get podgroups` confirms the expected number of PodGroup objects, and `result.json` shows populated `gangCompletionP50/P99Sec` for the gang run and `thresholdBreached: false` / `schedulingFailureCount: 0` for both
  - `go build ./test/perf/...` passes cleanly
  - `go test ./test/perf/...` passes — all packages build/test clean, including the new `pkg/scenarios/gang` package


Shivam-1827:
Hi @saintube and @ZiMengSheng,

Before I proceed further, I would appreciate some clarification on a few implementation details to make sure I'm aligned with the expected direction:

1. **PodGroup label key** — I used `pod-group.scheduling.sigs.k8s.io` for associating pods with their PodGroup. Can you confirm this matches the scheduler-plugins version vendored in this repo?

2. **One PodGroup vs many** — I split each run into `podCount / gangSize` separate PodGroups (many gangs competing concurrently) rather than one PodGroup for the whole run. Is this the intended shape?

3. **CI trigger scope** — Workflow currently runs on every PR touching `test/perf/**` plus manual `workflow_dispatch`. Given the multi-minute runtime (kind + kwok + Koordinator + 1k-pod run), should this gate every PR, or move to `workflow_dispatch`/nightly only?

4. **Failing CI on regression** — I handled this via a `jq` step in the workflow reading `thresholdBreached` from `result.json`, instead of changing `engine.Run()`'s return behavior. Is this the right place for it, or should `engine.Run()` itself return a non-nil error/exit code on a breach?

5. **Trailing comma in `hack/deploy_kind.sh`'s feature-gates sed** — The line that appends `DisableDynamicResourceAllocationInformer=true,WatchListClient=false,` to `--feature-gates=` leaves a trailing comma, which Kubernetes' flag parser rejects outright, crash-looping `koord-scheduler` and `koordlet` on any fresh `deploy_kind.sh` run. Confirmed by the exact error (`invalid argument "...,WatchListClient=false," for "--feature-gates" flag`) and by reading the sed line itself. Should this be fixed directly in `deploy_kind.sh`, or is there a reason it hasn't caused problems elsewhere that I'm missing?

6. **`test/perf` pinning a released image (`KOORDINATOR_VERSION=v1.5.0`) against the HEAD kustomize config overlay** — `test/perf/hack/setup-cluster.sh` pulls `ghcr.io/koordinator-sh/koord-scheduler:v1.5.0`, but the `koord-scheduler-config` ConfigMap it deploys is generated by `deploy_kind.sh` from this branch's checked-out `config/manager/`. When the two drift, the generated config enables koordinator-specific plugins (`Reservation`, `Coscheduling`, `ElasticQuota`, and likely others) with `pluginConfig` args that the pinned `v1.5.0` binary can't decode (`want args to be of type XArgs, got <nil>` / `*runtime.Unknown`), crash-looping the scheduler. The project's own e2e-test workflow avoids this entirely by building `koord-manager`/`koord-scheduler`/`koordlet` from source at the same commit instead of pinning a released tag. Should `test/perf`'s benchmark cluster do the same — build from source rather than pin `KOORDINATOR_VERSION`? If pinning a release is intentional (e.g. for benchmark stability/reproducibility across runs), how should the version-skew-with-HEAD-config problem be handled instead?

Thank you! I want to make sure my implementation follows the direction you have in mind before I continue.

.github/workflows/scheduler-benchmark.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Set up cluster
```
ZiMengSheng, line 44: This workflow calls `make -C test/perf setup`, and `test/perf/hack/setup-cluster.sh` starts by invoking `kind create cluster`, but the workflow does not install the `kind` CLI first. Existing e2e workflows install Kind explicitly via `helm/kind-action`, so I don't think we can assume it is preinstalled on `ubuntu-latest`.

Could we add an explicit Kind installation step before this? Since the setup script already creates the cluster, the workflow should install the CLI only, or alternatively the workflow/setup script should be adjusted so only one place creates the cluster.

config/manager/scheduler-config.yaml
```diff
@@ -44,11 +44,6 @@ data:
             estimatedScalingFactors:
               cpu: 85
               memory: 70
-        - name: ElasticQuota
-          args:
-            apiVersion: kubescheduler.config.k8s.io/v1
-            kind: ElasticQuotaArgs
-            quotaGroupNamespace: koordinator-system
         plugins:
```
ZiMengSheng, line 52: This PR removes `ElasticQuota` from the default scheduler config used by the manager kustomization and `hack/deploy_kind.sh`. That seems broader than the benchmark/CI scope and would change the default scheduler plugin set for normal deployments.

If ElasticQuota only needs to be disabled for the benchmark environment, could we keep the default config unchanged and use a test/perf-specific overlay/patch or setup-time config customization instead?

ZiMengSheng (commented):
I left a couple of inline comments on CI reliability and config scope.

ZiMengSheng:
Thanks @Shivam-1827 for listing these out. My suggestions:

1. **PodGroup label key**: `pod-group.scheduling.sigs.k8s.io` looks correct for this repo. The vendored scheduler-plugins API defines `PodGroupLabel = "pod-group." + scheduling.GroupName`, and `scheduling.GroupName` is `scheduling.sigs.k8s.io`.

2. **One PodGroup vs many**: I think the current “many PodGroups per run” shape is reasonable for the benchmark. It gives meaningful gang completion percentiles across groups and better represents concurrent gang workloads than a single giant PodGroup.

3. **CI trigger scope**: Given the runtime and potential flakiness of kind + kwok + Koordinator + 1k pods, I would prefer not making this a required gate for every PR yet. `workflow_dispatch` is definitely useful, and a scheduled/nightly run may be a better default until the benchmark is stable enough. If we keep PR triggers, path-filtering to `test/perf/**` is the minimum scope I’d accept.

4. **Failing CI on regression**: Keeping regression failure in the workflow via `jq` is the right direction. `engine.Run()` should continue to report a successful benchmark run with `thresholdBreached=true`; the CI layer can decide whether that should fail the job.

5. **Trailing comma in `deploy_kind.sh`**: If this is reproducible for normal `deploy_kind.sh` usage, it should be fixed directly there, ideally in a small focused change. If it only blocks this benchmark setup path, please keep the workaround isolated to `test/perf` and avoid broad unrelated changes in this PR.

6. **Pinned released image vs HEAD config**: I think this is the most important point. For a PR benchmark, we should avoid mixing a released `koord-scheduler:v1.5.0` binary with config generated from the PR/HEAD checkout. That version skew can create misleading failures. Prefer building the scheduler image from the same commit being tested, similar to the e2e workflow. If we intentionally want release-pinned benchmarks for reproducibility, then the config/manifests should also come from the same release, not from HEAD.

Also, please avoid changing the default `config/manager/scheduler-config.yaml` just to make the benchmark pass. If `ElasticQuota` needs to be disabled only for benchmark CI, use a `test/perf`-specific overlay, patch, or setup-time customization instead of changing the default deployment profile.

.github/workflows/scheduler-benchmark.yaml
Shivam-1827: Added an explicit Kind installation step via helm/kind-action@v1.9.0 with install_only: true before the "Set up cluster" step, matching the pattern used by the existing e2e workflows. The cluster itself is still created only inside test/perf/hack/setup-cluster.sh, so a single place owns cluster lifecycle.

config/manager/scheduler-config.yaml
Shivam-1827: Reverted. config/manager/scheduler-config.yaml is now byte-identical to main — git diff main -- config/manager/scheduler-config.yaml produces no output. The ElasticQuota disable was moved to a runtime patch scoped entirely to test/perf/hack/setup-cluster.sh. After switching to build-from-source (point 6 below), the version-skew crash that originally required the patch may be gone entirely — CI will confirm. If ElasticQuota still needs to be suppressed, the fix will be a surgical yq removal against the live ConfigMap rather than replacing the full KubeSchedulerConfiguration by hand.

Shivam-1827:
> Thanks @Shivam-1827 for listing these out. My suggestions:
> 
> 1. **PodGroup label key**: `pod-group.scheduling.sigs.k8s.io` looks correct for this repo. The vendored scheduler-plugins API defines `PodGroupLabel = "pod-group." + scheduling.GroupName`, and `scheduling.GroupName` is `scheduling.sigs.k8s.io`.
> 2. **One PodGroup vs many**: I think the current “many PodGroups per run” shape is reasonable for the benchmark. It gives meaningful gang completion percentiles across groups and better represents concurrent gang workloads than a single giant PodGroup.
> 3. **CI trigger scope**: Given the runtime and potential flakiness of kind + kwok + Koordinator + 1k pods, I would prefer not making this a required gate for every PR yet. `workflow_dispatch` is definitely useful, and a scheduled/nightly run may be a better default until the benchmark is stable enough. If we keep PR triggers, path-filtering to `test/perf/**` is the minimum scope I’d accept.
> 4. **Failing CI on regression**: Keeping regression failure in the workflow via `jq` is the right direction. `engine.Run()` should continue to report a successful benchmark run with `thresholdBreached=true`; the CI layer can decide whether that should fail the job.
> 5. **Trailing comma in `deploy_kind.sh`**: If this is reproducible for normal `deploy_kind.sh` usage, it should be fixed directly there, ideally in a small focused change. If it only blocks this benchmark setup path, please keep the workaround isolated to `test/perf` and avoid broad unrelated changes in this PR.
> 6. **Pinned released image vs HEAD config**: I think this is the most important point. For a PR benchmark, we should avoid mixing a released `koord-scheduler:v1.5.0` binary with config generated from the PR/HEAD checkout. That version skew can create misleading failures. Prefer building the scheduler image from the same commit being tested, similar to the e2e workflow. If we intentionally want release-pinned benchmarks for reproducibility, then the config/manifests should also come from the same release, not from HEAD.
> 
> Also, please avoid changing the default `config/manager/scheduler-config.yaml` just to make the benchmark pass. If `ElasticQuota` needs to be disabled only for benchmark CI, use a `test/perf`-specific overlay, patch, or setup-time customization instead of changing the default deployment profile.

Hi @ZiMengSheng ,

Thank you for the detailed review. Here's where each point lands:

PodGroup label key — noted, no change needed.

One PodGroup vs many — keeping the current shape; glad the per-group percentile approach aligns.

CI trigger scope — pull_request trigger removed entirely. The workflow now runs only on workflow_dispatch and a nightly schedule (cron: "0 3 * * *"). It can be promoted to a PR gate once the benchmark is stable.

Failing CI on regression — engine.Run() exits 0 and sets thresholdBreached: true in result.json; the workflow reads that field via jq and fails the job. The benchmark layer and the CI gate stay decoupled.

Trailing comma in deploy_kind.sh — deploy_kind.sh is untouched in this PR. The previous setup set KUBERNETES_VERSION which triggered the buggy sed substitution. Switching to build-from-source removed the need for KUBERNETES_VERSION entirely, so the sed line never fires. The workaround is scoped to test/perf by omission. A focused fix for the trailing comma itself will come in a separate PR.

Pinned released image vs HEAD config — fixed. All three images (koord-manager, koord-scheduler, koordlet) are now built from the current checkout via docker build and loaded into the kind node via kind load docker-image, matching the e2e workflow pattern. No released binary is used, so the binary and config always come from the same commit.

Shivam-1827:
@ZiMengSheng and @saintube , PTAL

Shivam-1827:
@ZiMengSheng and @saintube 
PTAL

test/perf/pkg/types/types.go
```diff
@@ -29,6 +29,17 @@ import (
 // the same string and a typo in one place cannot cause a silent mismatch.
 const RunIDLabel = "benchmark.koordinator.sh/run-id"
 
+// PodGroupLabel is the label key the koord-scheduler coscheduling plugin
```
ZiMengSheng: Align with here？https://github.com/koordinator-sh/koordinator/blob/d22cca0c53e47cf4b5d0697fe9de5f59182e70bd/apis/extension/coscheduling.go#L98-L110

test/perf/baselines/basic-1k.json
```diff
@@ -1,17 +1,20 @@
...
-  "totalDurationSec": 21.94438434,
-  "latencyP50Sec": 6,
-  "latencyP90Sec": 12,
-  "latencyP99Sec": 13,
+  "throughputPodsPerSec": 26.22,
+  "apiCreationDurationSec": 0,
+  "totalDurationSec": 0,
+  "latencyP50Sec": 0,
+  "latencyP90Sec": 0,
+  "latencyP99Sec": 0,
```
ZiMengSheng: These fields are all `0`, which means this file was hand-written rather than copied from a real `result.json`. The concrete consequence is in `pkg/framework/report.go`:

```go
if baseline.LatencyP99Sec > 0 {   // 0 here -> whole block skipped
```

so the P99 regression check never runs for either scenario, and `CompareToBaseline` still logs `"Baseline comparison passed"` — a latency regression will pass CI silently.

`runID: "ci-run-v5"` and `koordinatorVersion: "d55cb1d"` also aren't what the engine writes (`Version` comes from `git describe`), and `26.22` looks copied from the log line's 2-decimal formatting rather than from the JSON. The same applies to `baselines/gang-1k.json`.

Please download the `benchmark-results` artifact from a green run and commit that file verbatim, so every field is real and traceable to a run.

test/perf/hack/setup-cluster.sh
```diff
@@ -15,20 +24,134 @@ KWOK_BASE="https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERS
...
+echo "==> Building Koordinator images from source (tag: ${IMG_TAG})"
+(cd "${REPO_ROOT}" && \
+  docker build --pull . -t "${MANAGER_IMG}"   -f docker/koord-manager.dockerfile && \
+  docker build        . -t "${SCHEDULER_IMG}" -f docker/koord-scheduler.dockerfile && \
+  docker build        . -t "${KOORDLET_IMG}"  -f docker/koordlet.dockerfile)
+
+echo "==> Loading images into kind node"
+kind load docker-image "${MANAGER_IMG}"   --name "${CLUSTER_NAME}"
+kind load docker-image "${SCHEDULER_IMG}" --name "${CLUSTER_NAME}"
+kind load docker-image "${KOORDLET_IMG}"  --name "${CLUSTER_NAME}"
```
ZiMengSheng: Only the scheduler image is actually exercised here: koordlet is patched onto a nonexistent `nodeSelector` at L77 and koord-manager is scaled to 0 at L75, so building those two costs three full Go builds' worth of CI time for no benefit. Since the koordlet pods never get scheduled, its image is never pulled — you can simply leave `KOORDLET_IMG` unset. That should free up a meaningful chunk of the 60-minute budget.

ZiMengSheng (commented):
Thanks for the iteration — the points from my last review are all addressed: the `pull_request` trigger is gone (dispatch + nightly), `config/manager/scheduler-config.yaml` is restored, `hack/deploy_kind.sh` is untouched, and all images are now built from the checkout under test. The PR is also correctly scoped now — only `.github/workflows/scheduler-benchmark.yaml` and `test/perf/**`.

Two things I'd like resolved before this goes in:

**1. The committed baselines are hand-written, which silently disables half the regression gate.** In both `baselines/basic-1k.json` and `baselines/gang-1k.json`, `apiCreationDurationSec`, `totalDurationSec` and all three latency percentiles are `0`, and `runID`/`koordinatorVersion` are not in the form the engine emits. Because `CompareToBaseline` guards the P99 check with `if baseline.LatencyP99Sec > 0`, the P99 regression check is skipped entirely for both scenarios — and it still logs `"Baseline comparison passed"`. The gate is effectively throughput-only, which is the metric most sensitive to runner noise. The workflow already uploads `results/*.json`; please commit the actual artifact from a green run instead of hand-editing.

**2. `basic` and `gang` run back-to-back in one job and interfere, and the gang baseline bakes that in.** `Teardown` only issues a `DeleteCollection` on pods with background propagation and returns immediately; the engine then deletes the nodes and the gang step starts creating nodes and pods right away, while the previous 1,000 pods are very likely still `Terminating` and still counted against node allocatable / scheduler cache. I suspect part of the 35% gap between gang (17.04) and basic (26.22) comes from this rather than from coscheduling itself. Please wait for the namespace to drain between the two steps (or use a per-run namespace) before capturing baselines — otherwise the gang baseline encodes a specific execution order and will produce false regressions if the order or timing ever changes.

The rest of my comments are inline and non-blocking, but I would like the ElasticQuota comment (which is factually wrong about the quota tree) and the `Succeeded` contradiction in the health check fixed in this PR.

Two more items that have no diff line to attach to:

- **`pkg/framework/report.go` / `CompareToBaseline`** — it only compares throughput and P99, and never checks that the baseline's `nodeCount`/`podCount` match the config being run; comparing a 1k baseline against a 100-pod config is meaningless. A guard that errors on mismatch is cheap. Also, `schedulingFailureCount` is already in the result and this round spent several commits chasing quota throttling — gating on it (`> 0`, or above baseline) would have caught that class of problem directly.
- **`docs/README.md`** — the pre-created ElasticQuota, the deleted webhooks and the disabled koordlet now materially determine the numbers, so a local run that skips `setup-cluster.sh` will produce something entirely different. Please state in the README that the committed baselines are only valid for a cluster prepared by `setup-cluster.sh`.

test/perf/hack/setup-cluster.sh
```diff
@@ -15,20 +24,134 @@ KWOK_BASE="https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERS
...
+
+echo "==> Installing Koordinator (full stack via deploy_kind.sh)"
+export MANAGER_IMG SCHEDULER_IMG KOORDLET_IMG
+# Tell deploy_kind.sh which Kubernetes version the cluster is running so it
+# can disable DRA informers (resource.k8s.io/v1beta1 and v1) that don't exist
+# on clusters older than 1.34.  Without this, WaitForCacheSync in the
+# scheduler never returns and the scheduling loop never starts.
+export KUBERNETES_VERSION="1.28"
```
ZiMengSheng: Two things here:

- `KUBERNETES_VERSION="1.28"` and `KIND_NODE_IMAGE=kindest/node:v1.28.7` (L6) are two independent sources for the same fact. Please derive one from the other.
- Now that `KUBERNETES_VERSION` is set, the `sed` at `hack/deploy_kind.sh:47` actually fires and turns `--feature-gates=` into `--feature-gates=DisableDynamicResourceAllocationInformer=true,WatchListClient=false,` — the trailing comma that ca74ea7 claimed causes `invalid argument for --feature-gates`, and that 1f83333 then reverted on the grounds that "the sed never runs". Reading `component-base/featuregate`'s `Set()`, empty tokens are skipped (`if len(s) == 0 { continue }`), so the trailing comma looks harmless and I suspect the original diagnosis was wrong. Please state the definitive root cause in the PR description. If it *is* a real bug, fix it in a separate focused PR against `deploy_kind.sh` as we agreed — the setup shouldn't depend on "it happens to be benign".

Also worth noting: `deploy_kind.sh` mutates the working tree (`kustomize edit set image` rewrites `config/manager/kustomization.yaml`, the `sed` rewrites `config/manager/scheduler.yaml`). After a local `make -C test/perf setup` the checkout is dirty, `GIT_VERSION` picks up `-dirty`, and a second `setup` stacks the feature-gates substitution twice. A `git checkout -- config/manager` in a trap at the end of setup would avoid that, or at minimum call it out in `docs/README.md`.

test/perf/hack/setup-cluster.sh
```diff
@@ -15,20 +24,134 @@ KWOK_BASE="https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERS
...
+kubectl scale deployment/koord-descheduler -n koordinator-system --replicas=0
+kubectl patch daemonset koordlet -n koordinator-system \
+  -p '{"spec":{"template":{"spec":{"nodeSelector":{"koordinator-benchmark/skip":"true"}}}}}'
+
+# Wait for the scaled-down pods to actually terminate before checking namespace
+# health later — `kubectl scale` is async and returns immediately, but pods take
+# a moment to receive SIGTERM and exit. Without this, the health-check below can
+# catch them mid-shutdown and false-positive on pods behaving exactly as intended.
+kubectl wait --for=delete pod -l koord-app=koord-manager     -n koordinator-system --timeout=30s 2>/dev/null || true
+kubectl wait --for=delete pod -l koord-app=koord-descheduler -n koordinator-system --timeout=30s 2>/dev/null || true
```
ZiMengSheng, line 82: These waits cover koord-manager and koord-descheduler, but the component that actually crash-loops is koordlet — and its pods only start terminating after the DaemonSet patch at L77. Please add the same `wait --for=delete` for `koord-app=koordlet`, otherwise the health check below can still catch a crash-looping koordlet pod that hasn't been given a `deletionTimestamp` yet.

test/perf/hack/setup-cluster.sh
```diff
@@ -15,20 +24,134 @@ KWOK_BASE="https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERS
...
+# For remaining pods, fail if any is not Running/Succeeded, has a container
+# that is not ready, or has restarted more than twice.
+UNHEALTHY=$(kubectl get pods -n koordinator-system -o json | jq -r '
+  .items[]
+  | select(.metadata.deletionTimestamp == null)
+  | select(
+      (.status.phase != "Running" and .status.phase != "Succeeded")
+      or (.status.containerStatuses[]? | .ready == false)
+      or (.status.containerStatuses[]? | .restartCount > 2)
+    )
```
ZiMengSheng, line 106: The two clauses contradict each other: L101 explicitly allows `Succeeded`, but L102 is always true for a `Succeeded` pod (finished containers report `ready: false`), so any `Succeeded` pod is unconditionally reported unhealthy. The container-level checks should only apply to `phase == "Running"`:

```jq
| select(
    (.status.phase != "Running" and .status.phase != "Succeeded")
    or (.status.phase == "Running" and (
          ([.status.containerStatuses[]? | select(.ready == false)] | length) > 0
          or ([.status.containerStatuses[]? | select(.restartCount > 2)] | length) > 0
        ))
  )
```

This also de-duplicates the output — as written, a multi-container pod is printed once per matching container because of jq's stream semantics.

test/perf/hack/setup-cluster.sh
```diff
@@ -15,20 +24,134 @@ KWOK_BASE="https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERS
...
+# (the tree root that child quotas borrow slack from), but we scale koord-manager
+# to 0 above before it has time to do that.  Without the root, the GroupQuotaManager
+# has no parent to allocate guaranteed capacity from, so the benchmark quota's
+# runtime tracks used instead of min — causing quota-throttle FailedScheduling
+# events even when min is set generously.
+#
+# The benchmark child quota sets min=max so the full 10,000-CPU block is
+# guaranteed (no borrowing needed) and can never be exceeded.  1,000 pods ×
+# 500m = 500 CPU, well within the limit.
+echo "==> Pre-creating ElasticQuota root and benchmark quota"
```
ZiMengSheng: Three issues in this block, the comment being the most important:

- **The comment is factually wrong.** `koordinator-default-quota` is not the quota tree root — `apis/extension/elastic_quota.go` defines `DefaultQuotaName = "koordinator-default-quota"` (the fallback quota for pods that don't match any quota) and `RootQuotaName = "koordinator-root-quota"`, which is virtual. So "the tree root that child quotas borrow slack from" and the causal chain built on top of it don't hold. A wrong explanation in a hack script is worse than none — future maintainers will reason on top of it. Please either establish the real root cause and rewrite, or scale the comment back to "empirically required, cause not yet understood".
- **`min=max=10000` CPU (L149-L154) exceeds cluster capacity.** 100 kwok nodes total far less than 10,000 CPU, and a `min` guarantee larger than the root's total is semantically broken. Actual demand is 1,000 x 500m = 500 CPU, so something like `min=max=2000` is both sufficient and in range — and it makes the "quota is never the bottleneck" claim verifiable rather than resting on out-of-range numbers.
- **The `benchmark` namespace is hard-coded here** (L141, L147) while `configs/scenarios/*.yaml` also declares `namespace: benchmark`. Two sources of truth: change the YAML and the quota silently stops applying, which puts you right back in the hard-to-diagnose `FailedScheduling` situation this block exists to fix. Please take it from one place.

A cleaner alternative to hand-crafting the default quota: let koord-manager come up, wait for it to create the quota, *then* scale it to 0. That won't rot when the quota's name or namespace changes on the manager side.

.github/workflows/scheduler-benchmark.yaml
```diff
@@ -0,0 +1,123 @@
...
+            if [ -f "$result" ]; then
+              breached=$(jq -r '.thresholdBreached' "$result")
+              echo "$(basename $result): thresholdBreached=$breached"
+              if [ "$breached" = "true" ]; then
+                echo "::error::Benchmark regression detected in $(basename $result) — see uploaded results"
+                failed=true
+              fi
+            fi
+          done
+          [ "$failed" = "true" ] && exit 1 || true
```
ZiMengSheng: A few workflow-level nits, all minor:

- `Run benchmark (gang)` (L54) is skipped when the basic step fails, yet this step iterates over both files with `if: always()`. Either give the gang step `if: always()` or accept that a basic failure hides gang entirely.
- This step passes silently when neither result file exists. An explicit `::error::no result files produced` would match its intent.
- `[ "$failed" = "true" ] && exit 1 || true` — a plain `if` reads better and avoids the `|| true` idiom.
- `Debug on failure` (L86) uses `if: failure()`, so a pure threshold breach (a performance regression, not an infrastructure failure) also dumps ~30 kubectl commands. Give the setup/benchmark steps an `id` and narrow this to `steps.<id>.outcome == 'failure'`.
- L94 hard-codes `koordinator-bench-control-plane`, but the cluster name is owned by `setup-cluster.sh` (`CLUSTER_NAME`). Use `${CLUSTER_NAME:-koordinator-bench}-control-plane`.

One practical note on the trigger change: `schedule` only takes effect on the default branch, and `workflow_dispatch` needs the workflow file to exist there too — so this pipeline cannot actually be exercised in this repo before merge. Please dispatch it manually once after merge and use that run to re-capture the baselines, which also settles my first point in the summary.

test/perf/pkg/scenarios/gang/gang.go
```diff
@@ -0,0 +1,244 @@
...
+// Design decisions taken here that should be confirmed with mentors
+// (see Week 5 plan §3/§5) rather than treated as settled:
+//   - cfg.PodCount is split into ceil(PodCount/GangSize) separate PodGroups
+//     scheduled concurrently, rather than one PodGroup for the whole run.
+//     This is the "many gangs competing for nodes at once" shape and is what
+//     makes GangCompletionP50Sec/P99Sec meaningful as percentiles across groups.
+//   - The PodGroup label key (types.PodGroupLabel) is
+//     "pod-group.scheduling.sigs.k8s.io". NOT YET CONFIRMED against the
+//     scheduler-plugins version vendored in this repo.
+package gang
```
ZiMengSheng, line 27: This package comment still says these decisions "should be confirmed with mentors ... rather than treated as settled", and that the label key is "NOT YET CONFIRMED against the scheduler-plugins version vendored in this repo". Both were confirmed in the previous round — `PodGroupLabel = "pod-group." + scheduling.GroupName` in the vendored API, and the many-PodGroups shape is fine. Please update the comment so it documents the decision rather than an open question.

test/perf/pkg/types/types.go
Shivam-1827: Done. Removed the local PodGroupLabel constant from types.go and updated gang.go and watcher.go to import and use v1alpha1.PodGroupLabel directly from the vendored apis/thirdparty/scheduler-plugins package.

test/perf/hack/setup-cluster.sh
Shivam-1827: Fixed. koordinator-default-quota is extension.DefaultQuotaName (the fallback quota for unmatched pods), not the tree root. The tree root is koordinator-root-quota (extension.RootQuotaName). The comment now says "empirically required, cause not yet verified" since the exact causal path through GroupQuotaManager hasn't been traced.

test/perf/hack/setup-cluster.sh
Shivam-1827: Fixed. Container-level checks (ready == false, restartCount > 2) now only apply when phase == "Running". Succeeded pods are excluded from those checks since their containers are intentionally stopped.

test/perf/hack/setup-cluster.sh
Shivam-1827: Done. Added KUBERNETES_VERSION="${KUBERNETES_VERSION:-$(echo "${KIND_NODE_IMAGE##*:v}" | cut -d. -f1,2)}" so the two stay in sync automatically when the image tag is bumped.

test/perf/baselines/basic-1k.json
Shivam-1827: Updated both baselines with values from the first clean green CI run (koordinatorVersion: 6f7f91c): basic p99=30s, gang p99=54s. The pipeline has since run clean twice against these baselines.

test/perf/hack/setup-cluster.sh
Shivam-1827: Done. Added `kubectl wait --for=delete pod -l koord-app=koordlet -n koordinator-system --timeout=30s `alongside the existing waits for koord-manager and koord-descheduler, so the health check below can't catch them mid-shutdown.

test/perf/hack/setup-cluster.sh
Shivam-1827: Done — only the scheduler image is built from source. koord-manager and koordlet builds and their kind load calls were removed in a previous commit for exactly this reason.

test/perf/pkg/scenarios/gang/gang.go
Shivam-1827: Done -  removed the `NOT YET CONFIRMED` and `should be confirmed with mentors` language. The comment now documents the settled decisions: PodGroupLabel is verified against the vendored apis/thirdparty/scheduler-plugins constant, and the many-PodGroups shape is confirmed correct.

.github/workflows/scheduler-benchmark.yaml
Shivam-1827: All five nits is addressed :

- Gang step has `if: always() `so it runs even when basic fails.
- "Fail on regression" now sets a `found` flag and exits with` ::error::No result files produced` if neither file exists.
- Replaced `&& exit 1 || true` with a plain `if/fi` block.
- Debug step is narrowed to `if: steps.basic.outcome == 'failure' || steps.gang.outcome == 'failure'` — a pure threshold breach no longer dumps kubectl state.
- Control-plane container name uses `${CLUSTER_NAME:-koordinator-bench}-control-plane.`

On the practical note: agreed. The `schedule` trigger and `workflow_dispatch `won't be exercisable until the workflow lands on the default branch. I'll dispatch it manually immediately after merge and use that run to commit the baselines verbatim, which also resolves the hand-written baseline issue from your earlier comment.

Shivam-1827:
> Thanks for the iteration — the points from my last review are all addressed: the `pull_request` trigger is gone (dispatch + nightly), `config/manager/scheduler-config.yaml` is restored, `hack/deploy_kind.sh` is untouched, and all images are now built from the checkout under test. The PR is also correctly scoped now — only `.github/workflows/scheduler-benchmark.yaml` and `test/perf/**`.
> 
> Two things I'd like resolved before this goes in:
> 
> **1. The committed baselines are hand-written, which silently disables half the regression gate.** In both `baselines/basic-1k.json` and `baselines/gang-1k.json`, `apiCreationDurationSec`, `totalDurationSec` and all three latency percentiles are `0`, and `runID`/`koordinatorVersion` are not in the form the engine emits. Because `CompareToBaseline` guards the P99 check with `if baseline.LatencyP99Sec > 0`, the P99 regression check is skipped entirely for both scenarios — and it still logs `"Baseline comparison passed"`. The gate is effectively throughput-only, which is the metric most sensitive to runner noise. The workflow already uploads `results/*.json`; please commit the actual artifact from a green run instead of hand-editing.
> 
> **2. `basic` and `gang` run back-to-back in one job and interfere, and the gang baseline bakes that in.** `Teardown` only issues a `DeleteCollection` on pods with background propagation and returns immediately; the engine then deletes the nodes and the gang step starts creating nodes and pods right away, while the previous 1,000 pods are very likely still `Terminating` and still counted against node allocatable / scheduler cache. I suspect part of the 35% gap between gang (17.04) and basic (26.22) comes from this rather than from coscheduling itself. Please wait for the namespace to drain between the two steps (or use a per-run namespace) before capturing baselines — otherwise the gang baseline encodes a specific execution order and will produce false regressions if the order or timing ever changes.
> 
> The rest of my comments are inline and non-blocking, but I would like the ElasticQuota comment (which is factually wrong about the quota tree) and the `Succeeded` contradiction in the health check fixed in this PR.
> 
> Two more items that have no diff line to attach to:
> 
> * **`pkg/framework/report.go` / `CompareToBaseline`** — it only compares throughput and P99, and never checks that the baseline's `nodeCount`/`podCount` match the config being run; comparing a 1k baseline against a 100-pod config is meaningless. A guard that errors on mismatch is cheap. Also, `schedulingFailureCount` is already in the result and this round spent several commits chasing quota throttling — gating on it (`> 0`, or above baseline) would have caught that class of problem directly.
> * **`docs/README.md`** — the pre-created ElasticQuota, the deleted webhooks and the disabled koordlet now materially determine the numbers, so a local run that skips `setup-cluster.sh` will produce something entirely different. Please state in the README that the committed baselines are only valid for a cluster prepared by `setup-cluster.sh`.


@ZiMengSheng 

Most points from this review are now addressed: the drain step between basic and gang is in place, the ElasticQuota comment is corrected, the `Succeeded` health-check contradiction is fixed, and `CompareToBaseline` now guards on `nodeCount`/`podCount` mismatch and gates on `schedulingFailureRate > 1%`.

Two items are still open:
**Baselines** — will download the `benchmark-results `artifact from the latest green run and commit both files verbatim. Will do this immediately after the next clean dispatch.
**README** — will add a note that the committed baselines are only valid for a cluster prepared by `setup-cluster.sh`, and that a local run skipping that script will produce incomparable numbers.

saintube:
@Shivam-1827 Please fix the [DCO check](https://github.com/koordinator-sh/koordinator/pull/3136/checks?check_run_id=91523744976) before the merge.

Shivam-1827:
> @Shivam-1827 Please fix the [DCO check](https://github.com/koordinator-sh/koordinator/pull/3136/checks?check_run_id=91523744976) before the merge.

Sure

PR #3138, opened 2026-07-23
koordlet: support NUMA-level metrics and node peak reporting

saintube:

koordlet: enhance the node metric to report the NUMA-level resources and node peak.

### V. Checklist


pkg/koordlet/metricsadvisor/framework/context.go
```diff
@@ -94,13 +98,34 @@ func (r *SharedState) UpdateNodeUsage(cpu, memory metriccache.Point) {
 	r.nodeMemory = &memory
 }
 
+// UpdateNodeNUMAUsage updates the latest per-NUMA node cpu and memory usage.
+func (r *SharedState) UpdateNodeNUMAUsage(cpu, memory map[int32]metriccache.Point) {
+	r.nodeMutex.Lock()
+	defer r.nodeMutex.Unlock()
+	r.nodeNUMACPU = cpu
+	r.nodeNUMAMemory = memory
```
ZiMengSheng, line 106: nit: `UpdateNodeNUMAUsage` unconditionally overwrites both `nodeNUMACPU` and `nodeNUMAMemory` with whatever the caller passed. Combined with the caller in `noderesource` collector:

```go
if len(numaCPUUsage) > 0 || len(numaMemUsage) > 0 {
    n.sharedState.UpdateNodeNUMAUsage(numaCPUUsage, numaMemUsage)
}
```

if one round fails on one side (e.g. `NodeCPUInfo` not yet ready so `numaCPUUsage` is empty, or `GetNodeNUMAInfo` fails so `numaMemUsage` is empty), the still-valid other side ends up wiping the previous map. The sysresource collector still has `validTime` filtering so it won't consume stale values, but the wipe forces the next tick to see empty maps and skip.

Suggestion: only overwrite the non-empty side, so the two axes are updated independently (this also matches the empty-map-as-missing semantics you already use in `UpdatePodsNUMAMemoryUsage`):

```go
func (r *SharedState) UpdateNodeNUMAUsage(cpu, memory map[int32]metriccache.Point) {
    r.nodeMutex.Lock()
    defer r.nodeMutex.Unlock()
    if len(cpu) > 0 {
        r.nodeNUMACPU = cpu
    }
    if len(memory) > 0 {
        r.nodeNUMAMemory = memory
    }
}
```

pkg/koordlet/metricsadvisor/collectors/sysresource/system_resource_collector.go
```diff
@@ -145,10 +146,106 @@ func (s *systemResourceCollector) collectSysResUsed() {
...
+		for numaID, point := range numaUsage {
+			if point.Timestamp.Before(validTime) {
+				klog.V(4).Infof("pod collector %v NUMA memory metric is timeout, valid time %v, metric time is %v",
+					collector, validTime.String(), point.Timestamp.String())
+				podsNUMAMemoryValid = false
+				break
+			}
+			podsNUMAMemory[numaID] += point.Value
+		}
+	}
```
ZiMengSheng, line 186: The inner `break` only exits the innermost loop, so subsequent collectors keep accumulating into `podsNUMAMemory`. It happens to be safe because `podsNUMAMemoryValid` gates the outer `if`, but the extra work is wasted and the intent is a bit obscure.

Minor cleanup: use a label to break out of both loops (or return early), e.g.:

```go
podsNUMAMemoryValid := true
Outer:
for collector, numaUsage := range podsNUMAMemoryByCollector {
    for numaID, point := range numaUsage {
        if point.Timestamp.Before(validTime) {
            klog.V(4).Infof("pod collector %v NUMA memory metric is timeout, valid time %v, metric time is %v",
                collector, validTime.String(), point.Timestamp.String())
            podsNUMAMemoryValid = false
            break Outer
        }
        podsNUMAMemory[numaID] += point.Value
    }
}
```

Also worth noting: today only the `podresource` collector pushes into `podsNUMAMemoryByCollector`. If a future `beresource`-style collector reports pods without a NUMA memory contribution, the system NUMA memory can be over-estimated silently — probably worth a short comment here.

pkg/koordlet/prediction/peak_predictor.go
```diff
@@ -178,6 +191,8 @@ func (p *podReclaimablePredictor) AddPod(pod *v1.Pod) error {
 	if err != nil {
 		klog.V(5).Infof("podReclaimablePredictor failed to get prediction for pod %s, err: %s",
 			util.GetPodKey(pod), err)
+		// the pod contributes 0 to the reclaimable result, so count its request into the peak
+		p.unpredicted = quotav1.Add(p.unpredicted, util.GetPodRequest(pod, v1.ResourceCPU, v1.ResourceMemory))
 		return err
```
ZiMengSheng, line 196: This branch does two things that are a bit at odds: it (1) mutates `p.unpredicted` (a durable field consumed later by `GetPeak`) and (2) returns the underlying error. In the current call site (`states_nodemetric.collectMetric`) the error is only logged so this is fine, but the combination is fragile:

- If any caller ever wraps `AddPod` in a retry loop, the same pod's request will be double-counted into `unpredicted`.
- The `Predictor` interface doc doesn't say whether `AddPod` returning `err` still leaves side effects behind.

Two options:

1. Prefer: swallow the prediction failure here and `return nil` (`klog.V(5)` is already logging it). The pod contributes 0 to reclaimable and its request to peak — that's the intended semantic.
2. Otherwise: move the `unpredicted` bookkeeping into `GetPeak` (compute it from `p.pods` minus successfully predicted ones), so `AddPod` stays side-effect-free on error.

Either way, please also amend the `Predictor.AddPod` godoc to make the contract explicit. The cold-start branch above has the same shape but doesn't return an error, so it's not affected.

ZiMengSheng (commented):
General review — really nice work on both fronts, and the caveats already documented in the API/Go comments make the trade-offs easy to follow. Left a few inline comments on things that I'd want to see tightened before merge (mostly around the CPU/memory update coupling in `SharedState`, the `break` in the sysresource loop, and the side-effect-on-error in `podReclaimablePredictor.AddPod`). A couple of smaller thoughts that didn't warrant an inline:

- Consider extending the CRD schema comment on `NUMAUsage.SystemUsage` to also mention that hostApp usage is not subtracted, so downstream consumers don't accidentally compare NUMA-system vs node-system as if they had identical semantics.
- `NUMAMemUsageBytes` uses `MemFree + Active(file) + Inactive(file) + SReclaimable` as an approximation of `MemAvailable`. On nodes with large tmpfs/shmem this can drift a bit in the *optimistic* direction too, not only conservative; worth calling out in the godoc.
- `NodeMetricPromMetrics`: when the gate is flipped from enabled to disabled at runtime, the last emitted series will linger in `/metrics` until process restart. Either reset once on entering the disabled branch, or add a note.
- Minor: `minPredictor.GetPeak()` records gauges for each sub-predictor and then the aggregate; if the composition of `minPredictor` changes at runtime, stale `predictor="pod"|"priority"` series can remain — mirror `ResetNodeMetricUsages` if you want the strict "latest snapshot only" semantics.

Nothing blocking from my side.

pkg/koordlet/prediction/peak_predictor.go
saintube: The node prediction already logs the unpredicted error before this PR.

pkg/koordlet/metricsadvisor/framework/context.go
saintube: This behavior is consistent with the node-level metrics.

pkg/koordlet/metricsadvisor/collectors/sysresource/system_resource_collector.go
saintube: I prefer not to use `goto`.

PR #3144, opened 2026-07-23
webhook: evaluate quota when pod resources change on update

### LINKED ISSUE #3143: Quota admission skipped when pod resources increase on update with the same quota label
**What happened**

When `EnableQuotaAdmissionOnUpdate` is enabled, the pod validating webhook skips quota evaluation on UPDATE if the quota label stays the same. That means a user can edit an existing pod and increase CPU or memory requests without any quota check, even when the target quota is already full.

This path is easy to hit with `kubectl edit pod` or any patch that bumps resource requests while leaving `quota.scheduling.koordinator.sh/name` unchanged.

**What you expected to happen**

If quota admission on update is turned on, the webhook should still evaluate quota when pod resource requests change. An update that would exceed quota limits should be rejected the same way a create would be.

**How to reproduce it (as minimally and precisely as possible)**

1. Enable the `EnableQuotaAdmissionOnUpdate` feature gate on koord-manager.
2. Create an ElasticQuota with max `cpu=4`, `memory=8Gi`, and set usage to the limit via `ChildRequest` (or run other pods until the quota is full).
3. Create a pod in that quota with requests `cpu=2`, `memory=4Gi`.
4. Edit the pod and increase requests to `cpu=4`, `memory=8Gi` without changing the quota label.
5. The update is admitted. No quota evaluation runs.

The early return is in `pkg/webhook/pod/validating/evaluate_quota.go` when `oldQuotaName == quotaName`.

PR #3028 added quota admission on pod update for quota label changes, but the same-label case still skips evaluation entirely even when resources change.

A unit test that covers this scenario is `update with same quota label but increased resources on full quota, denied` in `pkg/webhook/pod/validating/evaluate_quota_test.go`.

**Environment**

- App version: main (current tree)
  - Containerd/Docker version:
  - OS version:
  - Kernal version:
- Others: Reproduced via unit test against `evaluateQuota`. Requires `EnableQuotaAdmission` and `EnableQuotaAdmissionOnUpdate` to be enabled.

false200:
Re-run quota admission on pod UPDATE when the quota label is unchanged but resource requests differ, so resource edits cannot bypass quota limits.

When `EnableQuotaAdmissionOnUpdate` is enabled, the webhook skipped quota evaluation on UPDATE if the quota label did not change, even when CPU or memory requests increased. This change compares old and new pod requests and only skips evaluation when they are the same.

fixes #3143

```bash
go test ./pkg/webhook/pod/validating/ -run TestEvaluateQuota -v
```

The new case `update with same quota label but increased resources on full quota, denied` should pass.

Small change in `evaluate_quota.go` only. Uses `PodRequests` and `quotav1.Equals`, same as the quota evaluator.

### V. Checklist

### DIFF
--- a/pkg/webhook/pod/validating/evaluate_quota.go
+++ b/pkg/webhook/pod/validating/evaluate_quota.go
@@ -22,6 +22,8 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
+	quotav1 "k8s.io/apiserver/pkg/quota/v1"
+	apiresource "k8s.io/component-helpers/resource"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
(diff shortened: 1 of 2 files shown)


saintube:
PTAL /cc @shaloulcy @TaoYang526 @zheng-weihao

zheng-weihao:
should we support resize api for vpa? or pod resource is immutable by default.

Issue #3149, opened 2026-07-26
scheduler: reservation controller resync cannot be disabled or stopped


thc1006:
## What happened

The reservation controller's resync loop cannot be turned off through its args, and once started it cannot be stopped. Three things combine (line numbers from `pkg/scheduler/plugins/reservation/controller/controller.go` at `a7a1b76`):

**1. `ResyncIntervalSeconds` cannot disable the loop.** `New()` falls back to the default for any non-positive value:

```go
resyncInterval := defaultResyncInterval          // line 119, = 60 * time.Second (line 424)
if args != nil && args.ResyncIntervalSeconds > 0 {
    resyncInterval = time.Duration(args.ResyncIntervalSeconds) * time.Second
}
```

So `ResyncIntervalSeconds: 0` — the natural way to express "off" — silently yields 60s.

**2. `Start()`'s guard is therefore unreachable.** Since `New()` never produces a non-positive interval, the `else` branch and its log line are dead code:

```go
if c.resyncInterval > 0 {                        // line 179, always true
    go wait.Until(c.resyncReservations, c.resyncInterval, nil)
} else {
    klog.V(4).InfoS("resync for reservations is disabled")
}
```

**3. The loops pass a nil stop channel.** Both the GC loop (line 175) and the resync loop (line 180) hand `nil` to `wait.Until`, so neither can ever be stopped — they run until the process exits. That is defensible for a scheduler that starts one controller for its lifetime, but it means anything else that calls `Start()` leaks two goroutines permanently.

There is also an asymmetry: GC has `DisableGarbageCollection` (line 108), resync has no equivalent.

## Why it matters

This already misled a contributor. `TestControllerStart` says:

```go
// Create controller with GC and resync disabled to keep the test focused.
controller := New(..., &config.ReservationArgs{DisableGarbageCollection: true, ResyncIntervalSeconds: 0})
```

The GC half works. The resync half does not: the loop starts, runs `resyncReservations()` immediately, and — because of the nil stop channel — keeps running after the test returns, resetting the process-wide reservation metrics every 60s for the remainder of the test binary.

[...]

PR #3150, opened 2026-07-26
scheduler: fix flaky TestSyncStatus by keeping the controller's background loops out of the metric assertions

thc1006:
## Summary

`TestSyncStatus` fails intermittently in the `unit-tests(Run Go test)` job. It failed on main in [run 30001423786](https://github.com/koordinator-sh/koordinator/actions/runs/30001423786):

```
--- FAIL: TestSyncStatus (0.20s)
     scheduler_reservation_resource{name="normalReservation",resource="cpu",type="allocatable",unit="core"} 10
    -scheduler_reservation_resource{name="normalReservation",resource="cpu",type="allocated",unit="core"} 0
    -scheduler_reservation_resource{name="normalReservation",resource="cpu",type="utilization",unit="ratio"} 0
    +scheduler_reservation_resource{name="normalReservation",resource="cpu",type="allocated",unit="core"} 8
    +scheduler_reservation_resource{name="normalReservation",resource="cpu",type="utilization",unit="ratio"} 0.8
```

`allocatable` is correct while `allocated` and `utilization` are zeroed. Only one code path emits that combination: `RecordReservationResource` on a reservation whose `Status.Allocatable` is set and whose `Status.Allocated` is empty — the copy the reservation lister held before `syncStatus()` wrote the allocation. And only `resyncReservations()` calls `metrics.ResetReservationResource()`, which is what makes the previously recorded `8` disappear rather than merely be overwritten.

## Cause

The test drives `controller.sync()` itself and then asserts on `metrics.ReservationResource`, a process-wide gauge. It also calls `controller.Start()`, which launches a background writer of that same gauge: `resyncReservations()` resets it and re-records every reservation from the lister, and the lister is eventually consistent with the status update `sync()` writes through the client. A resync that runs after `sync()` but before the informer catches up therefore records `allocated=0` over the freshly recorded `8`.

`wait.Until` runs its function immediately, so on an unloaded machine the resync almost always happens during the `time.Sleep(100 * time.Millisecond)` that follows `Start()`, before the test's own `sync()`. That is why this passes locally. On a loaded CI runner the goroutine can be scheduled past that sleep and land in the assertion window.

## Fix

**Isolate the test.** The background loops are disabled instead of slept around: `Start()` already guards all three (`numWorker`, `isGCDisabled`, `resyncInterval`), and `Start()` itself is covered by `TestControllerStart`. Note that neither args field disables them on its own — `New()` falls back to `defaultResyncInterval` for any `ResyncIntervalSeconds` that is not positive, and to one worker when `ControllerWorkers` is not positive — so both are set on the controller directly. `TestControllerStart` gets the same treatment; its comment claimed resync was disabled through `ResyncIntervalSeconds: 0`, which never worked.

**Wait, do not sleep.** The 100 ms sleep is replaced by a wait on the condition it was standing in for. `Start()` registers the pod handlers through `ForceSyncFromInformer`, which only adds the handler — the waiting lives in `WaitForHandlersSync` — and the `WaitForCacheSync` inside `Start()` covers the listers rather than handler delivery. Since `sync()` derives the allocation from the index those handlers maintain (`getPodsOnReservation`), the test polls that index for the reservation's owner pods. It is a `t.Fatalf` precondition rather than an assertion, because continuing past it only produces confusing secondary failures.

**Keep the queue machinery covered.** Disabling the loops removes the incidental coverage `worker()` and `processNextWorkItem()` used to get from goroutines the tests started and never stopped. `TestProcessNextWorkItem` covers them deliberately instead, including that `worker()` returns once the queue is shut down — the lifecycle property this whole area depends on. The only statements this PR stops covering are the three bare `go f()` launches in `Start()`, which cannot be reached without starting the very loops the tests are isolating themselves from; two previously unreachable `else` branches become covered in exchange.

No production code changes.

## Scope

This PR isolates the unit tests. It does **not** fix the two production problems the investigation turned up:

- **#3151** — `Controller.Start()` launches its workers before the pod handler it depends on has delivered its initial list. `frameworkexthelper.WaitForHandlersSync` runs inside `startInformersAndWaitForSync()`, which completes before `extenderFactory.Run()` reaches `controllerMaps.Start()`, so nothing ever waits for the handler this controller registers. A worker that picks up a reservation in that window recomputes its owners and allocation from an empty index, overwrites the persisted values, and — since `AllocateOnce` defaults to true with no guard on having observed any owner — marks the reservation `Succeeded`, which is terminal and does not self-correct.
- **#3149** — resync cannot be disabled through args, the `if c.resyncInterval > 0` guard in `Start()` is therefore unreachable, and neither background loop can be stopped because `wait.Until` is given a nil stop channel.

Both need `Start()` to take a stop channel or context and to be able to fail, which is a lifecycle change rather than a mechanical one, so folding it in here would make this diff much harder to review.

## Test plan

- [x] `go test ./pkg/scheduler/plugins/reservation/controller -run TestSyncStatus -race -count=100` ok
- [x] `GOMAXPROCS=1 go test ./pkg/scheduler/plugins/reservation/controller -run TestSyncStatus -race -count=100` ok
- [x] `GOMAXPROCS=8 go test ./pkg/scheduler/plugins/reservation/controller -run TestSyncStatus -race -count=100` ok
- [x] `go test ./pkg/scheduler/plugins/reservation/controller -race -shuffle=on -count=20` ok — this matters because shuffling can order `TestControllerStart` before `TestSyncStatus`, which is exactly when a leaked worker could reach the gauge
- [x] `go test ./pkg/scheduler/plugins/reservation/controller -run TestProcessNextWorkItem -race -count=50` ok
- [x] `go vet` and gofmt clean

I could not reproduce the original failure locally: 60 runs at `GOMAXPROCS=1` were clean, and even calling `resyncReservations()` explicitly between `sync()` and the assertion passes, because by then the lister has caught up. The window is narrow — the resync goroutine has to be delayed past `sync()` yet reach the gauge before the informer propagates the status update — which is consistent with this only being seen on CI. The diagnosis therefore rests on the failure being numerically identical to what `RecordReservationResource` emits for a pre-sync lister copy, and on `resyncReservations()` being the only caller of `ResetReservationResource()`.

The flake fixed here is a different one from #2957, which is about `TestSyncedEventHandler` timing out in `pkg/scheduler/frameworkext/helper`. That test passes 20/20 locally for me and is not addressed by this change.

### DIFF
--- a/pkg/scheduler/plugins/reservation/controller/controller_test.go
+++ b/pkg/scheduler/plugins/reservation/controller/controller_test.go
@@ -250,6 +250,11 @@ func TestSyncStatus(t *testing.T) {
 	metricsResourceRegistry := basemetrics.NewKubeRegistry()
 	metrics.ReservationResource.Reset()
 	metricsResourceRegistry.Registerer().MustRegister(metrics.ReservationResource)
+	// ReservationResource is a package-level collector and Start() below appends
+	// to the package-level handler registrations, so leave neither behind for
+	// whichever test runs next.
+	t.Cleanup(metrics.ReservationResource.Reset)
+	t.Cleanup(frameworkexthelper.ResetRegistrations)
 
 	reservation := &schedulingv1alpha1.Reservation{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


ZiMengSheng:
Pre-reviewed — the fix is the right shape: gating the controller's background loops out of the metric assertions removes the race rather than adding sleeps/retries. Pure test change, no runtime impact. LGTM from my side, ready for final sign-off.

Issue #3155, opened 2026-07-27
scheduler: widening a Restricted Reservation's tracked resources cannot recover Allocated, under-reporting usage


thc1006:
## What happens

`ReservationInfo.UpdateReservation` recomputes which resources a Restricted reservation tracks, then applies the new set to the **already-masked** `Allocated` value. Widening the set therefore cannot recover the dimensions the previous, narrower set discarded, and the cache reports those dimensions as unallocated while owner pods are using them.

Line numbers are from `a7a1b76`.

## The code

`pkg/scheduler/frameworkext/reservation_info.go`:

```go
ri.ResourceNames = resourceNames     // recomputed from AllocatePolicy + restricted options
...
if ri.Allocated != nil {
    ri.Allocated = quotav1.Mask(ri.Allocated, ri.ResourceNames)
}
```

`ReservationInfo.UpdatePod`, the operating-mode path, does the same thing with the same two lines.

`ri.Allocated` was accumulated by `AddAssignedPod`, which masks each pod's requests with the `ResourceNames` in force at the time. Once a dimension has been masked out of `Allocated`, a later `Mask` with a wider set is a no-op for it — the value is gone. `ri.AssignedPods` still holds each owner's unmasked `PodRequirement.Requests`, so the information needed to rebuild it is present.

## Failure sequence

```text
Reservation: allocatable cpu=4, memory=8Gi
AllocatePolicy: Restricted, restricted resources: [cpu]
Owner pod assigned: cpu=1, memory=4Gi
  → ri.Allocated = {cpu: 1}          // memory dropped at accumulation time
```

The restricted options annotation is then widened to `[cpu, memory]`. `reservationCache.updateReservation` keeps the same `ReservationInfo` and calls `UpdateReservation`, which sets `ResourceNames = [cpu, memory]` and re-masks:

```text
  → ri.Allocated = Mask({cpu: 1}, [cpu, memory]) = {cpu: 1}
```

The cache now believes memory allocated is 0 while 4Gi is in use by the owner pod. The two accounting paths react differently, and only one of them is unsafe:

- **Reservation-level fit** treats `Allocated` as the amount already taken, so `remaining = allocatable - reserved - allocated` is over-reported. A waiter can be admitted against reservation capacity that does not exist. This is the unsafe direction.
[...]

Issue #3154, opened 2026-07-27
scheduler: a terminating or replaced Reservation can still be scheduled and bound to Available


thc1006:
## What happens

A Reservation held in `Terminating` by a finalizer keeps its reserve pod in the scheduling queue, can be scheduled, and `Bind` then writes `status.phase: Available` onto the object that is being deleted. Nothing along that path checks `metadata.deletionTimestamp`.

Deletion is not immediate when finalizers are present: the API server sets `metadata.deletionTimestamp` and the object stays until the finalizers clear, which can take arbitrarily long.

Line numbers are from `a7a1b76`.

## The path

1. **The framework handler treats a terminating reservation as active.**
   `isReservationActive` (`pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go:686`) is

   ```go
   return reservationutil.GetReservationNodeName(r) == "" &&
       !reservationutil.IsReservationFailed(r) &&
       !reservationutil.IsReservationSucceeded(r)
   ```

   No `DeletionTimestamp` check, so `Pending → Pending + deletionTimestamp` satisfies `oldActive && newActive` and takes the `updateReservationInSchedulingQueue` path instead of removing the queue item.

2. **`PreFilter` does not reject it.** The reserve pod path resolves the reservation from the annotation, calls `ValidateReservation`, and continues. It checks neither `DeletionTimestamp` nor whether the lister object's UID still matches the synthetic pod's UID.

3. **`Reserve` does not reject it, and it is a second time-of-check boundary.** `Plugin.Reserve` repeats the name-only lookup and assumes immediately:

   ```go
   rName := reservationutil.GetReservationNameFromReservePod(pod)
   assumedReservation, err := pl.rLister.Get(rName)
   ...
   assumedReservation.Status.NodeName = nodeName
   pl.reservationCache.assumeReservation(assumedReservation)
   ```

   The reservation can be deleted, replaced or made terminal between `PreFilter` and `Reserve`, so validating in `PreFilter` alone is not enough:

   ```text
   PreFilter passes
   → the reservation is deleted and a same-named one is created
   → Reserve assumes the replacement
   → Bind writes Available to the replacement
   ```

4. **`Bind` does not reject it.** `Plugin.Bind` (`pkg/scheduler/plugins/reservation/plugin.go`) refetches the reservation and rejects only `IsReservationFailed`:

   ```go
   if reservationutil.IsReservationFailed(reservation) {
[...]

Issue #3148, opened 2026-07-26
scheduler: reservation plugin cache updates have no happens-before with the wake-ups that depend on them


thc1006:
Ref #2852. Spun out of the review of #2927; not a regression from that PR.

The Reservation plugin keeps its own `reservationCache` and reserve-pod nominator, maintained by a **plugin-owned informer listener** registered through `registerReservationEventHandler` / `ForceSyncFromInformer`. The scheduling queue, on the other hand, is notified from a **different listener**: every `GetSchedulingQueue()` call for Reservation events lives in `pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go`, and the plugin's own handler has no access to the queue at all.

client-go's shared informer dispatches each notification to every listener, but each listener has its own goroutine and buffer. Ordering is guaranteed *within* a listener, never *between* listeners. So there is no happens-before between "the plugin's cache reflects this Reservation event" and "a pod woken by this Reservation event enters a scheduling cycle".

### Failure sequence

1. An Available Reservation `R` is added; the informer store is updated.
2. The frameworkext listener processes the Add and moves matching waiters to activeQ/backoffQ.
3. Waiter `P` starts a scheduling cycle.
4. The plugin listener has not run yet, so `reservationCache` does not contain `R`.
5. `P`'s `ForEachMatchableReservationOnNode` finds no matchable reservation and it is rejected again.
6. The plugin listener then runs `cache.updateReservation(R)`.
7. That cache mutation produces no cluster event, so nothing wakes `P` a second time.
8. `P` waits for an unrelated event or its backoff to flush.

The same shape applies to owners widening, label/annotation updates, `status.allocatable` and `status.allocated` changes, phase transitions, and Delete.

### Why this is not specific to QueueingHints

In-tree plugins do not have this problem because they read the scheduler's own cache, which is updated by the same handler that notifies the queue. Koordinator's plugin-owned cache is a second listener, so the property is missing regardless of whether `QueueingHintFn`s are registered: with no hints, the frameworkext listener still moves pods on every Reservation event while the plugin cache may lag.

What hints change is the amount of accidental redundancy. Without hints every Reservation event requeues the pod, so a stale-read failure is usually followed by another wake-up from a later event. With hints, fewer events requeue the pod, so a stale-read failure has fewer chances to be retried. That makes this pre-existing ordering gap worth closing before `EnableQueueHint` graduates to default-true (it defaults to false today).

### Possible directions

[...]

PR #3156, opened 2026-07-27
scheduler: drop reserve pod nominations the informer store has invalidated

thc1006:
## Summary

The reservation nominator holds a reserve pod against a candidate node so that other pods evaluated for that node account for it (`BeforeFilter` → `NominatedReservePodForNode` → the transformer adds it to the node snapshot). Nothing removed that hold when the reservation it stands for stopped being able to use the node, so a phantom reserve pod could keep occupying capacity for every pod filtered against that node.

This PR is behaviour-affecting on its own and is **not** behind any feature gate. It was split out of #2927 for that reason: #2927 adds a default-off feature (`ReservationArgs.EnableQueueHint`), whereas everything here takes effect on merge, and the two deserve to be judged separately. #2927 now contains only the queueing-hint wiring. The two do not depend on each other and can merge in either order.

## What was wrong

A nomination survived all of these:

- **A UID replacement.** A delete followed by a same-named create reaches handlers as a single update. The nominator keys reserve pods by reservation UID, so the old entry stayed on its node forever.
- **A move to a terminal phase**, and **a reservation entering deletion behind a finalizer.** The object stays `Pending` with its generation, labels and annotations unchanged, so nothing about it looked different, while its reserve pod would never be scheduled. `ReservationInfo.IsUnschedulable()` already counts a terminating reservation as unschedulable; the nominator did not.
- **Any spec, label or annotation change** — even though `reservationEventHandler.OnUpdate` deletes the nomination for exactly these. Such an edit can move the reserve pod's placement constraints, its requests, its priority or even its `schedulerName`, so the nominated node need not still be a valid choice for it.
- **A node decision from an older scheduling cycle.** `handleReservationSchedulingFailure` replaces `podInfo.PodInfo` with a reserve pod rebuilt from the current reservation before requeuing it, while the node it nominates still comes from the cycle that just failed. A cycle that ran against an older revision could therefore pin the current reservation to a node chosen for a shape it no longer has, and overwrite a newer cycle's nomination on the way.
- **An unparsable reserve pod losing its node.** The exported entry point built the `PodInfo` first and returned on error, so the empty-node signal that means "this nomination is gone" never reached the cleanup.

## What changed

| File | Change |
| --- | --- |
| ``pkg/scheduler/plugins/reservation/nominator.go`` | Nominations carry the scheduling identity they were built from (generation, labels, annotations — exactly what ``NewReservePod`` derives a reserve pod from) and are revalidated against the informer store on read: dropped unless the reservation is still present, still the same object, still waiting for a node, still in that shape, and not holding a different node of its own. ``AddNominatedReservePod`` validates before it removes anything and refuses a node whose cycle ran against a different shape. The empty-node deletion now precedes parsing the reserve pod |
| ``pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go`` | Hand the nominator the reserve pod the failed cycle ran with, so the shape check has two different revisions to compare instead of the current object against itself |
| ``pkg/scheduler/plugins/reservation/eventhandler.go`` | Delete the nomination on a same-named UID replacement and on a reservation entering deletion, unconditionally; on a scheduling-relevant update of an unscheduled reservation (covers `schedulerName` handovers, which the framework handler does not clean up) only when the stored nomination predates that update |
| ``pkg/scheduler/plugins/reservation/transformer.go`` | ``BeforeFilter`` checks for a nil ``Node`` before dereferencing it, and stops rebuilding the ``PodInfo`` the nominator just handed it |
| ``pkg/scheduler/plugins/reservation/eventhandler_test.go`` | Nomination dropped on UID replacement, terminal and terminating phases, spec/label/annotation changes, an assigned node that is not the one being read, and an unparsable current object; kept on status-only writes. A late cycle can neither pin its own stale node choice (requests / node affinity / `schedulerName` / labels) nor disturb a newer nomination, and an unchanged shape is still accepted. A nomination recorded for the new revision survives a handler that processes the same update afterwards, while a terminating one is removed whatever recorded it. Plus the four rejection paths of ``AddNominatedReservePod`` and a benchmark for the read path |
| ``pkg/scheduler/frameworkext/eventhandlers/reservation_handler_test.go`` | The failure handler hands over the cycle's reserve pod, not one rebuilt from the current reservation (verified to fail without the fix) |

## One rule, applied twice

The event handler and the scheduling cycle run on goroutines with no ordering between them. By the time the handler sees an update, the informer store may be further ahead still, and a cycle may already have recorded a nomination for what the store holds now.

So the handler does not reason about the event it received. It asks the same question the read path asks — *does this entry still match the store?* — and deletes only if the answer is no. The handler is an optimization that frees the node sooner; the read path is the guarantee. Deleting too little is safe, because the next read drops whatever no longer matches; deleting too much is not, because nothing recreates a deleted entry.

Comparing against the event's old object instead would be wrong in a way that is easy to miss: a CRD with the status subresource does not advance `metadata.generation` for metadata-only writes, so labels or annotations going `A -> B -> A` produce two identical identities, and a delayed `A -> B` event would delete the nomination recorded for the second `A`.

Deletion, a terminal phase and a UID replacement stay unconditional — no revision of those may keep holding a node, and a replacement's nomination lives under a different UID key, so the old one's cleanup cannot reach it.

Pre-allocation candidates are left alone by the reserve pod's cleanup. They are recorded with no identity to judge them by, and a cycle in progress writes them before reading them back, so clearing them on a reservation update would take state from a cycle the update says nothing about. Their staleness is #3158.

## A limit of the shape comparison

`sameSchedulingShape` can only compare what `NewReservePod` puts into the synthetic pod: its namespace, the template spec, and the reservation's labels and annotations. A reserve pod cycle also reads fields straight off the Reservation — `Spec.Owners` and `PreAllocationPolicy` select the pre-allocatable candidates whose nodes narrow the feasible set. A cycle spanning an update that touched only those fields is indistinguishable here, and its node is accepted.

That gap predates this PR: the previous code accepted every late node decision unconditionally and stored the caller's own stale pod. Closing it needs the generation the cycle actually used to travel with it, which changes `NewReservePod` — whose output also feeds the scheduler cache, the queue, `PreFilter` and `Bind` — so it is tracked separately in #3158 rather than folded in here.

## On the read path's cost

`NominatedReservePodForNode` now does a lister `Get` and two small-map `DeepEqual`s per nomination, where it previously only copied the stored `PodInfo`. `BeforeFilter` calls it once per waiting pod per candidate node.

The cost scales with the number of *nominated* reserve pods on the node — the number of unscheduled reservations that failed a cycle there, normally zero, in which case a map lookup is the whole cost. In exchange, the path no longer rebuilds a reserve pod at all (an earlier revision of this work rebuilt one per read), and `BeforeFilter` no longer re-parses affinity for a `PodInfo` the nominator already built.

An earlier revision short-circuited the comparison on `resourceVersion`. That is deliberately gone: this scheduler writes an Unschedulable status after every failed attempt, so the stored `resourceVersion` stops matching almost immediately and the shortcut would never fire again in the steady state, while implying a cheaper cost than the code actually pays. Comparing the fields directly is behaviour-identical, since an equal `resourceVersion` implies equal generation, labels and annotations.

I have not run a `benchstat` matrix against the base, so I am not claiming the net effect is zero. `BenchmarkNominatedReservePodForNode` covers the shape of the path; ask if you want the numbers across nomination counts and annotation sizes.

## Out of scope

- **#3154** — a terminating or UID-replaced reservation can still be scheduled and have `Available` written onto it: the framework handler's `isReservationActive` ignores `deletionTimestamp`, and neither `PreFilter` nor `Bind` checks `deletionTimestamp` or the UID. This PR stops the *nominator* from holding a node for such a reservation; the queue, `PreFilter` and `Bind` paths are untouched here.
- **#3148** — the plugin's `reservationCache` and the scheduling queue are driven by different informer listeners with no happens-before between them. The nominator no longer contributes to that window, since it revalidates against the store on read, but `reservationCache` cannot be fixed the same way: a `ReservationInfo` accumulates assigned pods and allocated amounts from **Pod** events, which the Reservation object does not carry.
- **#3155** — widening a Restricted reservation's tracked resource set re-masks an already-masked `Allocated`, so the recovered dimensions stay at zero.

## Test plan

- [x] ``go test ./pkg/scheduler/plugins/reservation/ ./pkg/scheduler/frameworkext/eventhandlers/ -race -count=1`` ok
- [x] gofmt clean; golangci-lint v2.5.0 clean against main (0 issues)
- [x] The failure-handler regression test verified to fail without the fix
- [x] Verified standalone on a branch cut from main, independent of #2927

Ref #2852

### DIFF
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler.go
@@ -122,7 +122,14 @@ func MakeReservationErrorHandler(
 	}
 }
 
-func addNominatedReservation(f framework.Framework, podInfo *framework.QueuedPodInfo, nominatingInfo *fwktype.NominatingInfo) {
+// addNominatedReservation records the node the failed cycle nominated.
+//
+// cycleReservePod must be the reserve pod that cycle actually ran with, not the
+// one rebuilt from the current reservation before requeuing: the nominator
+// validates the node decision against the shape it was made for, and handing it
+// the current shape would make that check compare the current object with
+// itself and accept a node chosen for a shape the reservation no longer has.
+func addNominatedReservation(f framework.Framework, cycleReservePod *corev1.Pod, nominatingInfo *fwktype.NominatingInfo) {
 	frameworkExtender, ok := f.(frameworkext.FrameworkExtender)
 	if !ok {
 		return
@@ -136,9 +143,9 @@ func addNominatedReservation(f framework.Framework, podInfo *framework.QueuedPod
 	if nominatingInfo.Mode() == fwktype.ModeOverride {
 		nodeName = nominatingInfo.NominatedNodeName
 	} else if nominatingInfo.Mode() == fwktype.ModeNoop {
-		nodeName = podInfo.Pod.Status.NominatedNodeName
+		nodeName = cycleReservePod.Status.NominatedNodeName
 	}
-	reservationNominator.AddNominatedReservePod(podInfo.Pod, nodeName)
+	reservationNominator.AddNominatedReservePod(cycleReservePod, nodeName)
 }
 
 // input:
@@ -292,7 +299,16 @@ func handleReservationSchedulingFailure(sched *scheduler.Scheduler,
 				"pod", klog.KObj(pod), "reservation", rName, "node", nodeName)
 			// We need to call DonePod here because we don't call AddUnschedulableIfNotPresent in this case.
 		} else {
-			podInfo.PodInfo, _ = framework.NewPodInfo(reservationutil.NewReservePod(cachedR))
+			currentPodInfo, buildErr := framework.NewPodInfo(reservationutil.NewReservePod(cachedR))
+			if buildErr != nil {
+				// Requeued anyway: dropping it here would leave the reservation
+				// with nothing in the queue until some other event arrives,
+				// which is worse than one more attempt on a partially parsed
+				// pod. The nominator refuses such a pod separately.
+				klog.ErrorS(buildErr, "Reserve pod for the current reservation could not be fully parsed",
+					"pod", klog.KObj(pod), "reservation", rName)
+			}
+			podInfo.PodInfo = currentPodInfo
 			if e = schedAdapter.GetSchedulingQueue().AddUnschedulableIfNotPresent(logger, podInfo, schedAdapter.GetSchedulingQueue().SchedulingCycle()); e != nil {
 				klog.ErrorS(e, "Error occurred")
 			}
@@ -300,10 +316,15 @@ func handleReservationSchedulingFailure(sched *scheduler.Scheduler,
 		}
 
 		// nominate for the reserve pod if it is
+		// NOTE: pod, not podInfo.Pod. The branch above may have replaced
+		// podInfo.PodInfo with the current reservation, while the nominated node
+		// still comes from the cycle that failed, so the nominator has to be
+		// given the reserve pod that cycle ran with to be able to tell the two
+		// revisions apart.
 		// FIXME: We expect use the default nominator for a nominated reserve pod, since it makes no benefit to
 		//   maintain another nominator. However, the default nominator relies the podLister to fetch the real pod
 		//   from the informer cache.
-		addNominatedReservation(fwk, podInfo, nominatingInfo)
+		addNominatedReservation(fwk, pod, nominatingInfo)
 
 		errMsg := status.Message()
 		msg := truncateMessage(errMsg)
--- a/pkg/scheduler/frameworkext/eventhandlers/reservation_handler_test.go
+++ b/pkg/scheduler/frameworkext/eventhandlers/reservation_handler_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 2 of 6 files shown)


ZiMengSheng:
### Pre-review note: this PR overlaps two others, and I'd like your read on which cleanup mechanism wins

Your framing of the bug is the clearest of the group: a nominator hold that nothing retracts once the reservation can no longer use the node, so a phantom reserve pod keeps occupying capacity through `BeforeFilter` → `NominatedReservePodForNode` → transformer. That's a real and well-diagnosed leak.

The problem is that **three PRs are independently attacking the same class of bug** — a nomination outliving its validity — with different triggers and different mechanisms:

| PR | Trigger for the stale hold | Mechanism |
|---|---|---|
| **#3156** (this) | Reservation stops being able to use the node | Rewrites `nominator.go` (+279) with retraction paths |
| #3114 (@harshakumar25) | `schedulerName` handover (Case 5) | Adds `deleteNominatedReservationFromProfile` helper calling `DeleteNominatedReservePodOrReservation` |
| #3084 (@harshakumar25) | Scheduler restart loses in-memory state | Persists to `Reservation.status.nominatedNodeName` (new API field) |

You and #3114 overlap in `reservation_handler.go`; you and #3084 overlap in both `reservation_handler.go` and `plugins/reservation/eventhandler.go`.

**The question for you:** does your reworked `nominator.go` already expose the retraction path that #3114 needs? If it does, #3114 should reduce to wiring the Case 5 trigger into your mechanism rather than adding a parallel helper — one cleanup path is much easier to reason about than two. I've pointed @harshakumar25 at this PR and asked him to read it before reworking.

**Ordering.** #3084 owns a new API field and is the base; you'll likely need to rebase on top of it. I'd rather not ask you to rebase twice, so: #3084 needs a DCO + gofmt fix first, and once it's in I'll ping you.

**CI:** `build-and-push (koord-descheduler)` is failing. Worth a look — it may just be infra flake, but please confirm.

At +1282/-35 with a +775 test file this is the largest of the three; I want to review it properly rather than rush it, which is why I'm settling the ordering first. Cluster status is recorded on #3069.

thc1006:
Ordering makes sense. I'll rebase on #3084 once it's in and wait for your ping rather than doing it twice.

On whether the reworked nominator already covers what #3114 needs: yes, and I wrote a test instead of leaving it as an assertion. `TestEventHandlerUpdateSchedulerNameHandoverRetractsNomination` nominates a node for a Pending reservation, changes `spec.template.spec.schedulerName`, and asserts the hold is gone. It passes on this branch. With the scheduling-relevant branch of `reservationEventHandler.OnUpdate` disabled, which is where `main` is today, the same test fails and the nomination survives. The test isn't pushed yet; it'll come with the rebase so it doesn't cause a second round of review churn.

The mechanism is that schedulerName lives in the pod template, so changing it bumps `metadata.generation`, and generation is part of the identity a nomination is stored with. `DeleteReservePodIfStale` then sees the stored identity no longer match the informer store and drops the hold. On that reading Case 5 needs a trigger rather than a second cleanup algorithm, and #3114 could keep its regression test and drop `deleteNominatedReservationFromProfile`. @harshakumar25 found this bug and knows the handover path better than I do, so if there's a shape the generation bump doesn't cover I'd rather hear it before we collapse the two.

One thing to keep in mind when wiring them together: the framework-level helper reaches `DeleteNominatedReservePodOrReservation`, which deletes unconditionally, while the informer path uses `DeleteReservePodIfStale`, which checks the store first. The conditional form is deliberate. This handler and the scheduling cycle run on goroutines with no ordering between them, so an unconditional delete keyed on UID can discard a nomination a newer cycle has just recorded, and nothing recreates it.

CI is infrastructure rather than the code. `build-and-push (koord-descheduler)` fails at `exporting to GitHub Actions Cache` with `RESPONSE 404` and `ERROR CODE: BlobNotFound`. The `go build` layer above it reports `CACHED`, and the other twelve jobs in run 30315057946 passed, including `Run Go test`, `golangci-lint`, `Verify gofmt` and `Verify govet`. A re-run should clear it; I can push an empty commit if that's easier.

ZiMengSheng:
Reviewed as part of a cross-check over the four PRs that touch the reservation nominator (#2885, #3084, #3114, this one). Full cluster analysis is in https://github.com/koordinator-sh/koordinator/pull/3084#issuecomment-5564939241; the parts specific to this PR:

**This is the strongest of the four and I think it should land first.** All of its nominator changes are removals, so on its own it cannot create a bad state — worst case it drops an entry that the next cycle re-nominates. The other three either add entries (#3084) or remove them at a single narrow trigger, so layering them on top of this is safer than the reverse.

The reasoning in `DeleteReservePodIfStale` is well argued and I checked the two claims that matter. Reconciling against the informer store rather than against the event's old object is right, and the note that a CRD with a status subresource doesn't advance `metadata.generation` for metadata-only writes is what makes the identity comparison necessary rather than paranoid — an `A -> B -> A` label round trip really would produce two identical identities. Likewise, taking the lock before the lister `Get` is the correct order; the reverse would reintroduce the race. Leaving pre-allocation nominations alone, with the reason stated and tracked in #3158, is the right call rather than a gap.

Three things:

1. **Overlap with #3114.** Your generation-change branch explicitly cites the `schedulerName` handover — "the framework handler only removes the reserve pod from the scheduling queue, never from this nominator" — which is precisely the leak #3114 fixes, at the `frameworkext/eventhandlers` layer with an unconditional `DeleteNominatedReservePodOrReservation`. Both are correct, and they're not identical: #3114 fires exactly at the handover and always removes, yours fires on any spec change and only removes if the store agrees it's stale. That difference matters, because on a handover the reservation genuinely no longer belongs to this profile, so "if stale" is a weaker guarantee than the situation warrants. Please coordinate with @harshakumar25 on whether #3114 stays as the targeted unconditional removal or is folded in here — I don't want to merge two overlapping fixes without that being a decision someone made.

2. **#3084 will nullify part of this.** It appends an unconditional re-sync from `Reservation.status.NominatedNodeName` at the end of `OnUpdate`, after your `else if` chain, gated only on the phase not being Active/Failed/Succeeded. A terminating-but-pending reservation and a generation bump both pass that gate, so your `DeleteReservePod` / `DeleteReservePodIfStale` would be immediately undone on the same event. Details in the linked comment.

3. **`rLister == nil` falls through to an unconditional delete.** In that branch every scheduling-relevant update on an unscheduled reservation drops its nomination, with no store to judge against. The comment says this is "its only cleanup", which is defensible as a fail-safe direction — but can `rLister` actually be nil in a running scheduler, or only in tests? If production always has it, an assertion at construction would be better than a silent behaviour change; if it can be nil, this deserves a test.

Also: still `do-not-merge/work-in-progress` and marked draft, with the merge-base 33 commits behind `main`. Is this ready for a real pass, or are you still moving `nominator.go` around? I'd rather review it once it's settled than review a moving target — but say the word and I'll go through `nominator.go` line by line, since that's the +279/-21 that carries the actual risk here.

For what it's worth, I verified the mechanics: this merges cleanly against current `main`, and against #3084 / #3114 in either order — the only collisions are in `reservation_handler_test.go` and `eventhandler_test.go`.

thc1006:
Rebased onto `main`, pushed the nominator work that was sitting local, and taken this out of draft. Ready for the line-by-line whenever you have time.

**On #3114, you are right and I was wrong.** I argued on 09-06 that an unconditional delete keyed on UID could discard a nomination a newer cycle had just recorded, and that Case 5 could therefore drop `deleteNominatedReservationFromProfile`. That objection does not hold at the handover trigger: once `schedulerName` moves away, this profile should not be scheduling the reservation at all, so any nomination it still holds is invalid by definition and there is no newer valid entry to protect. The race I described only applies to the general spec-change trigger, where the reservation is still ours, which is why that path is the conditional one. So your split is the right one, and @harshakumar25, from my side keep `deleteNominatedReservationFromProfile` as it stands.

One thing that makes the overlap smaller than it looks: `DeleteNominatedReservePodOrReservation` is a method on this plugin's own nominator, so both mechanisms end up in the same map through different entry points, with different guarantees rather than two caches to keep in sync.

`TestEventHandlerUpdateSchedulerNameHandoverRetractsNomination` is in this push. With the `OnUpdate` branch disabled, which is where `main` is today, it fails and the nomination survives.

**On `rLister == nil`, it is tests only.** `newNominator` has one production call site, in `New`, and it always passes the lister built from the informer factory; the fifteen nil call sites are all in `_test.go`. I have not changed the behaviour yet, because `TestNominatedReservePodForNodeDropsInvalidated/nil lister keeps the stored snapshot` covers that path on purpose, and swapping the tests to a real empty store breaks seven of them: an empty store makes `AddNominatedReservePod` reject the add rather than record it, so those tests would each need their reservation seeded. For now the invariant is documented on the field. If you would rather have the assertion, say so and I will do the test conversion here.

**On #3084**, agreed that the re-sync needs to gate on more than the phase. Since this one lands first, that reconciliation belongs on that side.

thc1006:
A correction to my "ready" above: I moved `nominator.go` again after saying that, so the head is `b0d95636` rather than `0af8f539`.

Reviewing my own diff adversarially turned up a regression I had introduced. `sameSchedulingShape` compared `Spec` wholesale, and `NewReservePod` fills `Spec.NodeName` from `status.nodeName`, so a cycle that began before a Pending reservation acquired its node was rejected as a different revision. `main` accepts that case, since `IsReservationActive` is false while the phase is still Pending, so the node was left unaccounted in exactly the window `nominationStillValid` is written for.

The fix excludes `Spec.NodeName` from the comparison. A node pinned by the template still travels in `AnnotationReservationNode`, which is compared, so nothing is lost. `TestAddNominatedReservePodAcceptsAssignedNode` covers it and fails without those two lines.

CI is green on `b0d95636`. Apologies for the moving target; that is the last of it unless the review turns up more.

PR #3162, opened 2026-07-30
coscheduling: report collected children count in gang basic check

ZiMengSheng:
When a gang fails PreEnqueue because minMember is not satisfied, the error only named the gang, leaving no clue how many member pods had been collected. Include collected/min counts per unsatisfied gang so the gap is visible directly from the event message.

### V. Checklist


saintube (approved):
/lgtm

PR #3164, opened 2026-08-03
test/perf: register gang scenario, restore script permissions, and re-baseline

Shivam-1827:

Follow-up to #3136. Three bugs found while capturing real baselines locally after merge:

**1. Gang scenario was never registered**
`cmd/benchmark/main.go` was missing the blank import for the gang scenario package. Without it, `make benchmark CONFIG=configs/scenarios/gang-1k.yaml` always failed immediately with `scenario "gang" not registered; available: [basic]`, making the gang benchmark completely unusable.

**2. Execute permission stripped from hack scripts**
The DCO sign-off rebase (`git rebase HEAD~N --signoff`) changed the file mode of `hack/setup-cluster.sh` and `hack/teardown-cluster.sh` from `100755` to `100644`, causing `make setup` to fail with `Permission denied`. Restored the `+x` bit.

**3. Baselines were hand-written, silently disabling the P99 regression check**
Both `baselines/basic-1k.json` and `baselines/gang-1k.json` had `latencyP99Sec: 0` and integer-formatted latency fields, which caused the guard in `CompareToBaseline`:

```go

if baseline.LatencyP99Sec > 0 {  // 0 here → whole block skipped

```

to silently skip the P99 regression check on every run, logging `"Baseline comparison passed"` regardless of actual latency.

Replaced both files with verbatim engine output captured locally on `koordinatorVersion: d0274708` (100 nodes, 1000 pods, using `setup-cluster.sh`):

| Scenario | Throughput | P50 | P90 | P99 |
|---|---|---|---|---|
| basic | 26.26 pods/s | 24s | 28s | 29s |
| gang | 17.21 pods/s | 44s | 49s | 58s |

Every field — including the full-precision float throughput, UUID `runID`, and RFC3339 timestamp — comes directly from the engine, not hand-edited.

---

NONE

---

```bash
# Set up the cluster
make -C test/perf setup

# Run both scenarios — both should log "Baseline comparison passed"
make -C test/perf benchmark \
  CONFIG=configs/scenarios/basic-1k.yaml \
  BASELINE=baselines/basic-1k.json \
  OUTPUT=results/basic-result.json

make -C test/perf benchmark \
  CONFIG=configs/scenarios/gang-1k.yaml \
  BASELINE=baselines/gang-1k.json \
  OUTPUT=results/gang-result.json

# Tear down
make -C test/perf teardown
```
Check that:
- The gang benchmark runs without `"scenario not registered"` error
- Both results show `"thresholdBreached": false`
- The log shows `"Baseline comparison passed"` with a non-zero P99 check

---
The baselines were captured locally rather than from a GitHub Actions artifact because `workflow_dispatch` only takes effect on the default branch and requires maintainer access to trigger. The numbers are consistent with the CI runs observed during PR #3136 (basic p99 ~29–30s, gang p99 ~52–58s across three runs).

---
### Ⅴ. Checklist


ZiMengSheng (approved):
/lgtm
/approve

PR #3168, opened 2026-08-10
descheduler: optimize fragmentation-aware scoring calculation to O(P*R)

Vatsalpatni73:

This PR optimizes the fragmentation-aware descheduler plugin's scoring calculation. 
Previously, in `chooseBestEvictionCandidate`, for each of the $P$ evictable pods on a node, the code rebuilt a `podsAfter` slice (which takes **$O(P)$**) and called `scoreNodeImbalance`. This internally called `NodeUtilization`, which iterated over all remaining pods to sum their requests (taking another **$O(P)$**). The total cost of finding the best eviction candidate was **$O(P^2 \cdot R)$**, where $R$ is the number of tracked resources.

This PR introduces a precomputation step:
1. We compute the aggregate node utilization once via `newNodeImbalanceState`.
2. When evaluating a pod for eviction, we call `state.scoreWithout(pod)`, which simply subtracts that specific pod's resource requests from the cached totals to calculate the new fractions and standard deviation.

This removes the need to rebuild pod slices or iterate over all remaining pods for each candidate, bringing the loop's time complexity down to **$O(P \cdot R)$** per node.

Part of #2332 

- Added a `nodeImbalanceState` struct to `scoring.go` to hold the cached utilization base.
- Added `TestNodeImbalanceState` to `scoring_test.go` which removes each pod in turn and rigorously asserts that the new `scoreWithout(pod)` calculation matches the from-scratch `scoreNodeImbalance` calculation within a `1e-12` float tolerance limit.

### V. Checklist

### DIFF
--- a/pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
+++ b/pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
@@ -20,25 +20,10 @@ import (
 	"math"
 
 	corev1 "k8s.io/api/core/v1"
+	resourcehelper "k8s.io/component-helpers/resource"
 
 	deschedulernode "github.com/koordinator-sh/koordinator/pkg/descheduler/node"
 )
-
-// allocationFractions calculates requested / allocatable fractions for resources.
-// It skips resources mapped to 0 allocatable to prevent divide-by-zero.
-func allocationFractions(requests corev1.ResourceList, allocatable corev1.ResourceList, resources []corev1.ResourceName) []float64 {
-	var fractions []float64
-	for _, r := range resources {
-		alloc := allocatable[r]
-		if alloc.IsZero() {
-			continue
-		}
-		req := requests[r]
-		fraction := float64(req.MilliValue()) / float64(alloc.MilliValue())
-		fractions = append(fractions, fraction)
-	}
-	return fractions
-}
 
 // stdDev computes the population standard deviation of the values list.
 func stdDev(values []float64) float64 {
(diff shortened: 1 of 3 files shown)


pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
54vault: Is this function used in anywhere except test files? Maybe we can remove this.

pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
```diff
@@ -74,3 +75,80 @@ func scoreNodeImbalance(node *corev1.Node, pods []*corev1.Pod, resources []corev
...
+	}
+	fractions := make([]float64, len(s.resources))
+	for i := range s.resources {
+		fractions[i] = float64(s.baseMilli[i]) / float64(s.allocMilli[i])
+	}
+	return stdDev(fractions)
+}
+
+// scoreWithout returns the imbalance score as if pod were removed from the
+// node. It subtracts the pod's per-resource requests from the cached totals,
```
54vault: This comments seems not right. `scoreWithout` does not subtracts then restore. Consider refine this comment.

pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
```diff
@@ -74,3 +75,80 @@ func scoreNodeImbalance(node *corev1.Node, pods []*corev1.Pod, resources []corev
...
+		return 0
+	}
+	podReqs := resourcehelper.PodRequests(pod, resourcehelper.PodResourcesOptions{})
+	fractions := make([]float64, len(s.resources))
+	for i, r := range s.resources {
+		podMilli := int64(0)
+		if qty, ok := podReqs[r]; ok {
+			podMilli = qty.MilliValue()
+		}
+		fractions[i] = float64(s.baseMilli[i]-podMilli) / float64(s.allocMilli[i])
```
54vault, line 124: `resources` is allowed to contain `pods` — validation only rejects empty and duplicate entries — and when it does, `scoreWithout()` disagrees with the from-scratch path.

`NodeUtilization` treats `ResourcePods` specially: it seeds the total with the pod *count* and then explicitly skips it when summing container requests Pods: 

```go
resource.NewQuantity(int64(len(pods)), resource.DecimalSI),   // node.go:258
...
if ok && name != corev1.ResourcePods {                              // node.go:270
```

resourcehelper.PodRequests() never returns a pods key, so podMilli stays 0 for that resource and the pods dimension of baseMilli is never decremented. The old path recomputed NodeUtilization(podsAfter, ...) on each candidate, so it did drop by one.

Measured on this branch (node: cpu=4000m, memory=4Gi, pods=10; three pods;
resources: [cpu, memory, pods]):
```go
baseMilli = [3200, 2000000, 3000]

remove p1 | fromScratch=0.088975489  incremental=0.123040042  delta=-0.034064554
remove p2 | fromScratch=0.283088009  incremental=0.276133888  delta=+0.006954121
remove p3 | fromScratch=0.317104882  incremental=0.308220612  delta=+0.008884270
```

With only cpu+memory tracked the delta is 0.000000000 across the board, so this is specific to the pods dimension — extended resources such as [nvidia.com/gpu](http://nvidia.com/gpu) subtract correctly, since those do come back from PodRequests().

The error is larger than the default minImprovementThreshold (0.02) and its sign is not stable, so it can reorder candidates and flip the gain check in either direction: a different pod gets evicted, or none does.

Suggestion: record the pod count in nodeImbalanceState and special-case ResourcePods in scoreWithout() by subtracting one, mirroring NodeUtilization’s semantics. Worth adding a TestNodeImbalanceState case with pods in the resource list — the existing cases only cover [cpu, memory], which is why CI is green on this.

54vault (changes requested):

pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
Vatsalpatni73: Yes, `scoreNodeImbalance` and `allocationFractions` were unused outside the tests, so I removed both in the latest push.

pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
Vatsalpatni73: Sure, I have fixed it.

pkg/descheduler/framework/plugins/fragmentationaware/scoring.go
Vatsalpatni73: Thanks for the review! I've pushed an update that implements your suggestions:

1. Special-cased `ResourcePods` in both `newNodeImbalanceState` (to seed the count as a `MilliValue`) and in `scoreWithout` (to explicitly subtract `1000` for the removed pod instead of relying on `PodRequests()`).

2. Added a new case in `TestNodeImbalanceState` that tracks `[cpu, memory, pods]` to ensure `scoreWithout` perfectly matches the from-scratch calculation when pods are included in the resource list.

songtao98 (approved):

PR #3175, opened 2026-08-21
cgreconcile: add memory.soft_limit_in_bytes fallback for cgroup v1 vanilla kernels

### LINKED ISSUE #3187: cgreconcile: add memory.soft_limit_in_bytes fallback for cgroup v1 vanilla kernels

`memory.soft_limit_in_bytes` as a fallback for `memory.low` on vanilla cgroup v1 kernels.

### Why is this needed

`memory.low` is only available on Alinux (Anolis) cgroup v1 kernels and on cgroup v2. On vanilla cgroup v1 kernels (e.g. EKS AL2, standard CentOS 7 kernels) the file does not exist, so koordlet silently skips writing it and BE pods get no best-effort memory protection at all.

When `memory.low` is unavailable, mirror the computed `memory.low` value to `memory.soft_limit_in_bytes` as a best-effort soft limit. Note this is a soft limit, not a hard guarantee, so behavior differs slightly from `memory.low`, but it restores a degraded-but-functional protection level on platforms that lack the Alinux extension.

Implementation follows the existing graceful-degradation pattern: register the resource with `SupportedIfFileExistsInKubepods` + `WithCheckOnce(true)`, skip with a log when the file is absent.

N/A (tracking issue for PR #3175)

pingxin403:

`memory.soft_limit_in_bytes` is a standard cgroup v1 file available on all cgroup v1 systems (upstream Linux memory controller since 2.6.29). When `memory.low` is not available (standard cgroup v1 kernels, e.g., EKS AL2, vanilla kernels), mirror the calculated `memory.low` value to `memory.soft_limit_in_bytes` as a best-effort protection.

This provides **partial memory QoS** on vanilla cgroup v1 kernels where Koordinator previously silently degraded memory QoS to zero.

**Soft_limit is a best-effort soft limit, not a hard guarantee** like `memory.low` on Alibaba kernels or cgroup v2. It guides the kernel's reclaim behavior under memory pressure but does not prevent the cgroup from exceeding it when memory is plentiful.

Fixes #3187

1. Deploy koordlet on a standard cgroup v1 kernel (e.g., EKS AL2, Ubuntu 20.04)
2. Enable CgroupReconcile with MemoryQOS enabled
3. Verify that `/sys/fs/cgroup/memory/kubepods.slice/kubepods-besteffort.slice/*/memory.soft_limit_in_bytes` is set to the calculated value
4. On kernels with `memory.low` (Alibaba/Anolis), existing behavior is unchanged

- The new `MemorySoftLimit` resource uses `SupportedIfFileExistsInKubepods` + `WithCheckOnce(true)`, so it gracefully degrades on cgroup v2 or kernels without the file
- `memory.soft_limit_in_bytes` is only registered in `knownCgroupResources` (cgroup v1), not in `knownCgroupV2Resources`
- On cgroup v2, `memory.low` is natively available, so this fallback is unnecessary
- A follow-up PR could add a config switch and `memory.pressure_level` eventfd-based proactive reclaim for stronger protection

### V. Checklist


ZiMengSheng:
Reviewed — LGTM. Mirroring `memory.low` to `memory.soft_limit_in_bytes` on cgroup v1 is the right best-effort fallback, and the updater registration follows the existing pattern. The partial-protection caveat is already documented in the description.

PR #3135, opened 2026-07-21
scheduler: add sandbox score-free profile

tan90github:
### Ⅰ. Describe what this PR does

This PR adds a dedicated score-free scheduler profile for sandbox workloads.

Main changes:

- Add a new scheduler feature gate: `SandboxScheduling`, disabled by default.
- Add the sandbox scheduler profile name: `koord-scheduler-sandbox`.
- When `SandboxScheduling=true`, koord-scheduler automatically appends a sandbox profile derived from the first configured scheduler profile.
- The sandbox profile keeps the existing queue, filter, reserve, pre-bind, and bind behavior, but disables all `PreScore` and `Score` plugins.
- If users already define a profile named `koord-scheduler-sandbox`, the auto-append logic does not create a duplicate profile.

Why this change is needed:

Sandbox workloads need a high-throughput scheduling path. In large-scale pod/node benchmarks, the scoring stage becomes visible in the scheduler hot path. For sandbox pods, we can keep hard scheduling constraints in filter-related stages while skipping score/pre-score plugins to reduce per-pod scheduling overhead.

Current benchmark result:

- Stable scheduling throughput: about `2k scheduled pods/s`.
- Benchmark pod creation throughput: about `3k pods/s`.
- Final benchmark workload size: about `200k` pods.
- Node scale: about `1.2k` simulated nodes in the target cluster. 
Scheduling QPS:

image

Pod creation QPS:
image

Benchmark environment:

- Kubernetes / scheduler dependency version: `1.35.x`
  - Exact revision from this branch: `k8s.io/kubernetes v1.35.6`
  - Scheduler module: `k8s.io/kube-scheduler v0.35.6`
- Load generator: ClusterLoader2 CronJob running in the target cluster.
- Workload size: about `200k` pods.
- Node scale: about `1.2k` simulated nodes.
- koord-scheduler:
  - replicas: `2`
  - feature gates include `SandboxScheduling=true`
  - `percentageOfNodesToScore: 1`
  - clientConnection: `qps: 100000`, `burst: 200000`
- koord-manager:
  - replicas: `2`
  - client config args: `--rest-config-qps=100000`, `--rest-config-burst=200000`
- apiserver benchmark tuning:
  - `--max-requests-inflight=1500`
  - `--max-mutating-requests-inflight=1000`
  - `--http2-max-streams-per-connection=3000`

Why apiserver parameters were adjusted:

[...]

PR #3177, opened 2026-08-23
scheduler: add sandbox custom scheduling workflow

tan90github:

Related Issue: https://github.com/koordinator-sh/koordinator/pull/3112, https://github.com/koordinator-sh/koordinator/pull/3135

This PR adds an alpha `SandboxCustomWorkflow` feature gate, which is disabled by default.

When enabled, the custom workflow takes over the scheduling loop:

- Ordinary pods continue through the existing Koordinator scheduling path.
- Sandbox pods with a template hash use an equivalence-class fast path.

For each template hash, the first pod runs the complete Filter and Score path and caches the feasible nodes with placement quotas. Following pods reuse the cached result, run a single-node Filter recheck, and fall back to the complete path when the cache is invalid, expired, or exhausted.

The workflow also provides:

- A bounded LRU cache for multiple active template hashes.
- Binding concurrency control to protect the API Server from excessive post-Permit binding requests.

Enable it with:

```text
--feature-gates=SandboxCustomWorkflow=true
--sandbox-max-concurrent-bindings=1024
--sandbox-equivalence-cache-size=16
```

In the current performance environment, the observed peak successful Bind rate was approximately 3,206 pods/s. 
- [scheduler: copy upstream scheduling loop primitives](https://github.com/koordinator-sh/koordinator/pull/3177/changes/4440a60129ee9781d70164c5bc63ce5087c682a8)This PR only copies the k8s framework code as the baseline for the workflow. You may start reviewing the code from other commits.
- I have supplemented the pressure‑testing data.

### V. Checklist


tan90github:
# Sandbox Scheduling Performance Test Report

## Test Environment and Scope

| Item | Configuration |
|-|-|
| Target nodes | 2,000 |
| Pods per case | 150,000 |
| Number of namespaces | 20 |
| ClusterLoad2 Pod creation QPS | 5,500 |

This report contains two test rounds:

1. **Round 1: Scheduling path performance comparison**: Create 150,000 equivalence-class Pods using the same Sandbox template hash, and compare the Native, Score-free, and Custom workflow paths with `scorePercent=0`, `5`, and `100`, for a total of 9 cases;
2. **Round 2: LRU-16 working-set test**: Use the Custom workflow path and test 8 different Sandbox template hash group sizes with scorePercent=`0`, `5`, and `100`, for a total of 24 cases.

`scorePercent=0` represents Kubernetes adaptive node-scoring percentage, `5` represents evaluating approximately 5% of nodes, and `100` represents evaluating all nodes.

## Round 1: Scheduling Path Comparison

### Test Objective

Compare the original Koordinator scheduling path, the Score-free path based on community [PR #3135](https://github.com/koordinator-sh/koordinator/pull/3135), and the Custom workflow path.

### Scheduling Paths

| Path | Description |
|-|-|
| Native | Original Koordinator scheduling framework and scheduling loop |
| Score-free | Sandbox score-free path based on PR #3135, while still using the native scheduling loop |
| Custom workflow | Equivalence-class scheduling path, custom workflow, and Bind concurrency optimization |

### Results

- The 150,000 Pods in this test used the same hash;
- The test used a 300-second scheduling timeout and stopped automatically when the timeout was reached.

Metric definitions:

- `Binding peak QPS`: Peak QPS of successful Binding requests. It represents the instantaneous processing limit observed under the test conditions and within the observation window; it is not the theoretical upper bound for all scenarios;
- `Average evaluated nodes`: Average number of nodes evaluated per scheduling attempt;
- `Full-path ratio`: Percentage of Pods entering the full scheduling path. A lower ratio indicates more reuse of the equivalence-class fast path;
- `Partial`: Not all Pods were completed within the 300-second observation window. Pod creation had completed, and the QPS, evaluated-node count, and resource metrics collected during the run remain valid for comparison;
- Pods were created online, so the elapsed test period includes the Pod creation phase. Therefore, elapsed time is only an auxiliary metric. A high creation request rate may increase API Server/etcd pressure and reduce Binding QPS.

| Configuration | Case | scorepercent ratio | Scheduled Pods | Binding peak QPS | Average evaluated nodes | Full-path ratio | Result |
|-|-|-|-|-|-|-|-|
| Native | native-adaptive | adaptive | 137203/150000 | 486.8 | 707.06 | 100% | Partial, timeout |
| Native | native-score5 | 5% | 150000/150000 | 1077.6 | 103.15 | 100% | Complete |
| Native | native-score100 | 100% | 60262/150000 | 219.8 | 2077.00 | 100% | Partial, timeout |
| Score-free | scorefree-adaptive | adaptive | 150000/150000 | 840.8 | 707.06 | 100% | Complete |
| Score-free | scorefree-score5 | 5% | 150000/150000 | 1596.8 | 103.16 | 100% | Complete |
| Score-free | scorefree-score100 | 100% | 131806/150000 | 467.8 | 2077.00 | 100% | Partial, timeout |
| Custom workflow | custom-adaptive | adaptive | 150000/150000 | 3049.0 | 21.17 | 2.856% | Complete |
| Custom workflow | custom-score5 | 5% | 150000/150000 | 3226.6 | 4.65 | 3.576% | Complete |
| Custom workflow | custom-score100 | 100% | 150000/150000 | 2684.8 | 55.82 | 2.641% | Complete |

### Monitoring Data

The following images are the original PNG files downloaded from Feishu without resizing or compression.

| Configuration | Case | Monitoring |
|-|-|-|
| Native | native-adaptive |native-adaptive|
| Native | native-score5 | native-score5|
| Native | native-score100 | native-score100 |
| Score-free | scorefree-adaptive |scorefree-adaptive |
| Score-free | scorefree-score5 | scorefree-score5 |
| Score-free | scorefree-score100 |scorefree-score100|
| Custom workflow | custom-adaptive | custom-adaptive |
| Custom workflow | custom-score5 | custom-score5 |
| Custom workflow | custom-score100 | custom-score100 |




### Result Analysis

- Custom workflow completed all 150,000 Pods under adaptive, 5%, and 100% node-scoring percentages;
- The Custom workflow full-path ratio was 2.641%–3.576%, while both Native and Score-free were 100%, indicating that most Pods reused the equivalence-class fast path.
- Custom workflow delivered roughly 2.0–5.7× higher observed Binding peak QPS than Score-free and 3.0–12.2× higher than Native, while evaluating 22–37× fewer nodes per scheduling decision.

## Round 2: LRU-16 Sandbox Hash Working-Set Test

### Test Objective

Fix the cache capacity at 16 and change the number of concurrently active Sandbox template hash groups. Observe the change in equivalence-class cache hit rate as the working set approaches or exceeds 16, and compare the performance overhead under different node-scoring percentages.

`K` represents the number of different Sandbox template hashes used concurrently. Pods with the same hash can reuse the same equivalence-class result; for example, K=1, K=16, and K=64 represent 1, 16, and 64 template classes, respectively. When K exceeds 16, different hashes compete for cache space.

Each case contains 150,000 Pods distributed evenly across K hash groups.

### Results

Metric definitions:

- `Bind peak QPS`: Peak QPS of API Server `POST /pods/binding` requests with response code `201`;
- `CPU avg`: Average scheduler CPU usage during the scheduling window, measured in cores;
- `RSS peak`: Peak scheduler process RSS, summed across multiple `koord-scheduler` instances;
- `Hit rate`: Equivalence-class cache hit rate;
- `Avg nodes`: Average number of nodes evaluated per scheduling attempt.

### `scorePercent=5%`

| K | Pods per hash group | Bind peak QPS | Hit rate | CPU avg (cores) | RSS peak (GiB) | Avg nodes | Result |
|-:|-:|-:|-:|-:|-:|-:|-|
| 1 | 150,000 | 3347 | 95.71% | 5.74 | 16.87 | 5.26 | complete |
| 4 | 37,500 | 3116 | 90.47% | 5.98 | 16.55 | 10.45 | complete |
| 8 | 18,750 | 2802.8 | 80.08% | 6.18 | 16.69 | 20.76 | complete |
| 15 | 10,000 | 2452.8 | 73.41% | 6.15 | 18.57 | 27.37 | complete |
| 16 | 9,375 | 2823.8 | 79.63% | 5.73 | 17.1 | 21.2 | complete |
| 17 | 8,823 / 8,824 | 2704.4 | 73.66% | 5.47 | 19.15 | 27.12 | complete |
| 32 | 4,687 / 4,688 | 1921.6 | 34.62% | 7.14 | 18.96 | 65.84 | complete |
| 64 | 2,343 / 2,344 | 1614.4 | 13.72% | 7.54 | 14.61 | 86.56 | complete |

### Adaptive Percentage `scorePercent=0`

| K | Pods per hash group | Bind peak QPS | Hit rate | CPU avg (cores) | RSS peak (GiB) | Avg nodes | Result |
|-:|-:|-:|-:|-:|-:|-:|-|
| 1 | 150,000 | 3014.6 | 96.92% | 5.79 | 17.8 | 22.02 | complete |
| 4 | 37,500 | 2303.2 | 87.51% | 6.92 | 14.55 | 86.17 | complete |
| 8 | 18,750 | 2096.6 | 80.95% | 8.16 | 14.95 | 130.95 | complete |
| 15 | 10,000 | 1751.4 | 61.17% | 9.3 | 15.58 | 265.87 | complete |
| 16 | 9,375 | 1966.8 | 61.22% | 9.39 | 15.36 | 265.52 | complete |
| 17 | 8,823 / 8,824 | 1618 | 58.73% | 9.19 | 15.91 | 282.48 | complete |
| 32 | 4,687 / 4,688 | 868 | 16.22% | 10.23 | 16.58 | 572.51 | complete |
| 64 | 2,343 / 2,344 | 779 | 6.42% | 10.57 | 16.72 | 639.35 | complete |

### `scorePercent=100%`

K=15, 16, 17, 32, and 64 used a 900-second scheduling observation window.

| K | Pods per hash group | Bind peak QPS | Hit rate | CPU avg (cores) | RSS peak (GiB) | Avg nodes | Result |
|-:|-:|-:|-:|-:|-:|-:|-|
| 1 | 150,000 | 3143.4 | 97.45% | 6.18 | 17.32 | 52.13 | complete |
| 4 | 37,500 | 2528.2 | 88.15% | 9.05 | 15.54 | 238.71 | complete |
| 8 | 18,750 | 1723.2 | 87.13% | 8.77 | 16.49 | 259.11 | complete |
| 15 | 10,000 | 782.4 | 40.51% | 12.13 | 16.98 | 1194.46 | timeout |
| 16 | 9,375 | 614.2 | 25.96% | 12.58 | 16.77 | 1486.3 | timeout|
| 17 | 8,823 / 8,824 | 521.2 | 24.30% | 12.42 | 16.88 | 1519.56 |timeout|
| 32 | 4,687 / 4,688 | 414.8 | 7.43% | 12.83 | 17.26 | 1857.91 | timeout |
| 64 | 2,343 / 2,344 | 394.8 | 3.36% | 13.12 | 17.19 | 1939.63 | timeout |

### Result Analysis

1. LRU-16 provides generally stable reuse for concurrently active hash working sets of up to 8 groups;
2. K=15–17 is the practical boundary, while K=32 and above show significant performance degradation;
3. If the number of concurrently active Sandbox hash groups is consistently greater than 16 in production, an LRU capacity of 32 can be evaluated and revalidated.

tan90github:
PTAL @ZiMengSheng @saintube

cmd/koord-scheduler/main.go
```diff
@@ -68,6 +67,10 @@ func flatten(plugins map[string]frameworkruntime.PluginFactory) []app.Option {
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
+	// Register custom workflows. When one is enabled, it takes over the scheduling loop
+	// instead of the default scheduler workflow.
+	app.KnownWorkflowList = append(app.KnownWorkflowList, sandbox.New())
```
saintube, line 72: Instead of appending in the main function, I prefer to register these custom workflows in an init func.

saintube:
@tan90github 

Overall: sound design, real throughput gains (Bind ~3.2k/s, 22–37× fewer nodes evaluated). The fast path is safe — it only picks the node; per-pod PreFilter + single-node Filter recheck + Assume/Reserve still run, so it can't overcommit. Worst case falls back to full path. My concern is maintainability + blast radius, not a correctness bug. Recommendation: request changes.

Blocking:
Patch coverage 54.9% < 70% gate (Codecov failing). Gaps are in the error/fallback/binding branches (sandbox_schedule.go 47%, schedule_cycle.go 25%, binding_cycle.go 48%) — exactly the paths that bite in prod.

High:
~1,450 lines of the k8s 1.35 scheduler loop are forked in-tree. Hard coupling to scheduler internals + drift risk on every upgrade. Justify why existing extension points (FindOneNodePlugin, EnableInlineBatchSchedule) can't carry this, and tag every intentional // DIFF from upstream.
The gate takes over the whole loop, so ordinary pods also run the forked ScheduleOne. It misses koordinator's own scheduleOne handling of errBatchScheduled (framework_extender_factory.go:389). Make SandboxCustomWorkflow × EnableInlineBatchSchedule mutually exclusive, or add an interaction test.

Medium:
Binding slot is acquired synchronously in ScheduleOne (schedule_one.go:108) → when slots saturate, the serial scheduling loop stalls; throughput couples to bind latency. Confirm intended; explain the K=15/16/17 non-monotonic jitter in Round 2.
Equivalence reuse correctness rests entirely on the external contract that same-hash pods are identical in every scheduling-relevant field. Fast path skips cross-node Score, so soft constraints (topology/pod spread) are only approximated. Document the exact field set the hash must cover; have identity-sensitive plugins return reusable=false.
Coarse invalidation: any Node add/update/delete flushes the entire cache. Device-plugin/label churn can collapse hit rate (the main lever). Add a misses{reason="node_event"} alert and consider per-node invalidation.

Low:
Fast path ignores pod.Status.NominatedNodeName → a preempted-then-requeued sandbox pod may not land on its nominated node.
Bind failure after a cache hit never returns the consumed quota (conservative, not a bug) — worth a comment.
equivalenceClassEntry.lastCycle is written by both store and next; rename for clarity.

Nit: add a gauge for current cache entry count to size the LRU.

cmd/koord-scheduler/main.go
NickC-CreateFuture: placeholder

KunWuLuan:
I'm new to this issue. Have you considered using a dedicated scheduler profile for sandbox scheduling? That way, we could use sandbox-scheduler for sandboxes and disable unnecessary plugins for better performance.

tan90github:
> I'm new to this issue. Have you considered using a dedicated scheduler profile for sandbox scheduling? That way, we could use sandbox-scheduler for sandboxes and disable unnecessary plugins for better performance.

@KunWuLuan Thank you very much for your reply. Let me briefly explain the current status of this issue.

1. The background of this issue is described in [Issue #2879](https://github.com/koordinator-sh/koordinator/issues/2879).
2. The proposal document is available in [PR #3112](https://github.com/koordinator-sh/koordinator/pull/3112).
3. Regarding the profile-based approach you mentioned, it was already implemented in [PR #3135](https://github.com/koordinator-sh/koordinator/pull/3135). In a scenario with 1,200 nodes and a pod creation rate of 3,000 QPS, the scheduling throughput was approximately 2,000 Pods/s.

The reason we did not continue optimizing the profile-based approach is that it has an upper limit in terms of performance. In addition, sandbox pods are better suited to equivalence-class scheduling. Therefore, our expectation is to implement this functionality through equivalence-class scheduling.

In essence, equivalence-class scheduling can be implemented through a custom workflow. This approach allows us to optimize the scheduling framework more selectively and reduce performance bottlenecks.
In [this comment](https://github.com/koordinator-sh/koordinator/pull/3177#issuecomment-5390444988), the scheduling throughput reached as high as 3.3k Pods/s.

However, saintube pointed out in [this comment](https://github.com/koordinator-sh/koordinator/pull/3177#issuecomment-5439889668) that my previous implementation duplicated too much of the upstream framework code. This prompted me to think more deeply about the design.

What we may actually need is to use equivalence-class scheduling to simplify the filter and score computations, rather than replacing the entire scheduling framework with equivalence-class scheduling. At least based on the current requirements, replacing the whole framework does not seem necessary.

Therefore, I have recently been refactoring the scheduling framework. I have removed its dependency on CustomWorkflow and am developing equivalence-class scheduling as an independent interface, so that this capability can be reused by other batch or offline workloads in the future. Sandbox will simply act as one consumer and compose the relevant components to achieve fast scheduling.

I have also designed a dispatcher to route pods either through the equivalence-class scheduling path or through the normal scheduling path.

I will submit the related code later today.

tan90github:
@saintube @KunWuLuan The code is ready for review.

The CI failures appear to have been introduced by [PR #3125](https://github.com/koordinator-sh/koordinator/pull/3125), rather than by this change.

Issue #2971, opened 2026-05-16
[proposal] koordlet internal metrics use raw prometheus instead of k8s component-base/metrics


mdryaaan:
**What is your proposal**:

Migrate the internal metric definitions in `pkg/koordlet/metrics/` from raw `github.com/prometheus/client_golang/prometheus` types to `k8s.io/component-base/metrics` wrapped types.

The `InternalRegistry` in `internal_metrics.go` already uses `k8s.io/component-base/metrics/legacyregistry` as the backing registry. However, metrics are registered via `legacyregistry.RawMustRegister`, which accepts raw `prometheus.Collector` values and bypasses stability level checks. All metric definitions across the package (`common.go`, `cpu_suppress.go`, `cpu_burst.go`, etc.) use `prometheus.NewGaugeVec` / `prometheus.NewCounterVec` directly, not the component-base wrapped variants.

A `// TODO consider using k8s.io/component-base/metrics to replace github.com/prometheus/client_golang/prometheus` comment at `internal_metrics.go:30` acknowledges this gap.

**Why is this needed**:

1. **Incomplete migration** — `InternalRegistry` already points at `legacyregistry` but registers through the `RawMustRegister` escape hatch, which skips stability annotations entirely.
2. **StabilityLevel annotations** — `k8s.io/component-base/metrics` requires each metric to declare `ALPHA`, `BETA`, or `STABLE`. Without this, there is no machine-readable signal about which internal metrics are production-stable vs experimental, making it impossible to enforce deprecation policy.
3. **Precedent in the same codebase** — `pkg/koordlet/metrics` exposes these metrics on `/internal-metrics`. Other Kubernetes-ecosystem components (`kube-scheduler`, `kubelet`) migrated away from direct prometheus usage for this exact reason.

**Is there a suggested solution, if so, please add it**:

For each `prometheus.NewGaugeVec` / `prometheus.NewCounterVec` / `prometheus.NewHistogramVec` declaration across the package:

1. Replace `prometheus.NewGaugeVec(prometheus.GaugeOpts{...})` with `k8smetrics.NewGaugeVec(&k8smetrics.GaugeOpts{..., StabilityLevel: k8smetrics.ALPHA})`
2. Change the `Collector` slice types from `[]prometheus.Collector` to `[]k8smetrics.Registerable`
3. Replace `legacyregistry.RawMustRegister` in `internalMustRegister` with `legacyregistry.MustRegister`
4. Update imports

Example:

```go
// Before
[...]

PR #3178, opened 2026-08-23
refactor: split koordlet internal metric registration paths

mdryaaan:

Preparatory PR for migrating koordlet's internal metrics to `k8s.io/component-base/metrics` (#2971). It adds a
second registration entry point to `internal_metrics.go`: the new `internalKubeMustRegister` takes
`k8smetrics.Registerable` and goes through `legacyregistry.MustRegister`, so metrics that declare a
`StabilityLevel` get the component-base stability framework applied. The existing `internalMustRegister` is left
exactly as it is — same name, same `prometheus.Collector` signature, same `RawMustRegister` behaviour, same
twelve `init()` call sites — so this change is purely additive and nothing existing has to adapt. It stays the
path for collectors that cannot implement `Registerable`, namely the `pkg/util/metrics` GC wrappers, which
expose the wrapped `prometheus.Collector` rather than the component-base type.

This mirrors the two-path split already used in `pkg/scheduler/metrics/metrics.go` (`metricsList` →
`MustRegister` at lines 171/229, `gcMetricsList` → `RawMustRegister` at 188/236 on current `main`).

No metric is migrated here, so registry contents and metric names are unchanged. The split exists so each
collector slice can move to component-base independently in follow-up PRs. The now-obsolete
`// TODO consider using k8s.io/component-base/metrics ...` comment is removed, since this PR starts acting on it.

All koordlet internal metrics will start as ALPHA; promotion to BETA/STABLE requires a follow-up proposal. This
change is intentionally koordlet-only — `pkg/util/metrics/koordmanager/registry.go` has the same half-migration
(`legacyregistry` + `RawMustRegister`) and can get the same treatment later; koord-scheduler and koord-descheduler
are already on component-base.

Part of #2971

`go test ./pkg/koordlet/metrics/... -run TestInternal -count=1`

Two tests cover this:

- `TestInternalRegistryGathersAllInternalMetrics` writes one sample to each of the 31 internal metrics, gathers
  `InternalRegistry`, and asserts every expected metric name is present. It also cross-checks the collector count
  across the twelve registered slices against the table, so a metric *added* to a slice without a test entry is
  caught too — verified by appending a new gauge to `CommonCollectors` and watching it fail with
  `should have 32 item(s), but has 31`.
- `TestInternalKubeMustRegisterPublishesToInternalRegistry` registers a component-base ALPHA metric through the
  new path and asserts it reaches `Gather()` — verified by stubbing `internalKubeMustRegister` to a no-op and
  watching the test fail. Both registration functions now show as covered in the profile.

Since registration behaviour is unchanged, the existing tests in the package are the regression check:
`go test ./pkg/koordlet/metrics/... ./pkg/util/metrics/... -count=1`. `go vet`, `gofmt` and `golangci-lint`
(v2.11.3) are clean on both packages.

`internalKubeMustRegister` has no production caller yet — it is the destination the per-file migration PRs move
each slice onto, starting with `common.go`. It is exercised by the test above rather than merged untested, and
the `unused` linter is not enabled in `.golangci.yml`, so this does not break CI.

The GC-wrapped metrics (`PodEvictionDetail`, `ContainerOOMScoreAdj`, `ContainerCoreSchedCookie`,
`CoreSchedCookieManageStatus`) stay on `internalMustRegister` until `pkg/util/metrics` stops capturing
`*prometheus.MetricVec` at construction: that snapshot is nil for a component-base vec until registration creates
it, so passing one through today would leave the GC holding a nil and panic in its background goroutine after the
first expiry. If the wrappers were changed to hold the component-base type and delete through it lazily — which
needs a small adapter, since `prometheus.MetricVec.Delete` takes `prometheus.Labels` while component-base's takes
`map[string]string` — that path could eventually go away.

The registry test is worth keeping strict as later PRs land — an unregistered component-base metric returns a
no-op from `With()` and false from `Delete()` without an error, so a metric dropped from the Registerable list
would otherwise keep compiling and keep passing tests while silently no longer being reported.

### V. Checklist


ZiMengSheng (commented):
Reviewed at 1657962, and re-checked against current `main` (025aa592): no commit has touched `pkg/koordlet/metrics` or `pkg/util/metrics` since the branch point, and the merge is clean.

The core premise holds. `legacyregistry.MustRegister` and `RawMustRegister` are both bound to the same `defaultRegistry` (component-base v0.35.6, `metrics/legacyregistry/registry.go:34-42`) and `DefaultGatherer` is that same registry, so metrics will still be gathered on `/internal-metrics` as slices move to the Registerable path. Registration behaviour here is genuinely unchanged, and the completeness test is exact today: the 12 internal slices hold 8+2+2+4+3+1+1+4+1+2+1+2 = 31 collectors, matching the 31 names in the test, and the four slices absent from `init()` (CPI, PSI, Resctrl, ResourceSummary) go to `ExternalRegistry` and are correctly excluded. Your note on the GC wrappers is accurate: `pkg/util/metrics/expire_metric.go:48,77,107` snapshot `vec.MetricVec` at construction, `ExpireMetrics` deletes through it at line 200, and the GC goroutine at line 260 has no `recover`/`HandleCrash`, so that panic would take the process down.

Four comments, none blocking.

**1. The `internalRawMustRegister` doc comment contradicts `init()` at this commit** — `pkg/koordlet/metrics/internal_metrics.go:41-43`

It says the raw path "is only for collectors that cannot implement `k8smetrics.Registerable`, i.e. the `pkg/util/metrics` GC wrappers", but all 12 slices go through it here and only 4 involve GC wrappers. A reader at this commit sees the comment and the twelve calls right below it disagree. Worth a sentence such as "Until the per-file migration in #2971 completes, all collector slices still use this path."

**2. The completeness test catches removals but not additions**

The 31-name list is maintained by hand in parallel with the slices, so a metric *added* to a slice is silently uncovered, which is the more likely direction of drift. I verified this: I added a gauge to `CommonCollectors` from package-level var init (so the append happens before `init()` registers, and a separate probe test confirmed it really does land on `InternalRegistry`), and `TestInternalRegistryGathersAllInternalMetrics` still passed.

A cheap cross-check closes it, valid because every internal collector here is a single Vec, i.e. one metric family:

```go
internalSlices := [][]prometheus.Collector{
    CommonCollectors, CPUSuppressCollector, /* ... all 12 ... */ HostApplicationCollectors,
}
total := 0
for _, s := range internalSlices {
    total += len(s)
}
assert.Len(t, internalMetrics, total, "a collector was added to an internal slice but not to this test")
```

**3. The new Registerable path has zero test coverage**

Coverage profile for the package:

```
internal_metrics.go:37.63,39.2 1 0   // internalMustRegister body
internal_metrics.go:44.66,46.2 1 1   // internalRawMustRegister
internal_metrics.go:48.13,61.2 12 1  // init
```

`internalMustRegister` is never executed, which is what codecov's two missing lines are. The PR adds a registration path plus a test that asserts registration, but the test never exercises the new path, so nothing yet demonstrates that a component-base metric registered this way actually appears in `Gather()`. Either fold the first migration (`common.go`) into this PR, or register one throwaway ALPHA metric through `internalMustRegister` in the test. That would also settle the "function with no callers merged into main" question.

**4. "will stay on the raw path permanently" is stronger than the constraint really is** (description only, not the code)

The obstacle is the snapshot in `pkg/util/metrics`, not anything inherent to component-base. `k8smetrics.GaugeVec.Delete` (component-base `metrics/gauge.go:250`) resolves the embedded vec at call time and guards on `IsCreated()`, so if the GC wrappers held the wrapper and deleted through it lazily instead of capturing `*prometheus.MetricVec` at construction, they could hold component-base types and the raw path could eventually disappear. It does need a small adapter rather than a straight interface swap: `prometheus.MetricVec.Delete` takes `prometheus.Labels` while component-base's takes `map[string]string`, so the signatures are not identical and one interface cannot cover both. Suggest phrasing it as "until `pkg/util/metrics` stops capturing `*prometheus.MetricVec` at construction", so the follow-up isn't foreclosed.

Nit: the test writes a sample into each global vec and never resets. Harmless today, since the other tests in the package are no-panic smoke tests that don't assert values, but the first test that does assert a value will inherit 31 stray series.

mdryaaan:
Thanks for the thorough review @ZiMengSheng, particularly for actually running the lazy-metric and completeness checks rather than reading them. All four addressed in `0c296ab3` (doc comment) and `6af5c0cd` (tests).

**1. Doc comment vs `init()`.** Fixed. The comment now frames the GC wrappers as the path's long-term use and adds "Until the per-file migration in #2971 completes, all collector slices still use this path", so it no longer contradicts the twelve calls below it.

**2. Additions-blind completeness test.** You're right, and that is the likelier direction of drift. Added your cross-check — it sums `len()` across the 12 registered slices and asserts the table matches. I verified it against an actual addition rather than a removal: declaring a new gauge and appending it to `CommonCollectors` without touching the table fails with `should have 32 item(s), but has 31` and "a collector was added to an internal slice but not to this test". The premise holds here because every internal collector is a single Vec, so collector count and family count are the same number.

**3. Uncovered Registerable path.** Took the second option — folding `common.go` in would merge this PR with the first migration, and I'd rather keep the granularity we agreed on. Added `TestInternalMustRegisterPublishesToInternalRegistry`: a component-base ALPHA gauge registered through `internalMustRegister`, then asserted present in `InternalRegistry.Gather()`. Coverage for that line goes from `1 0` to `1 1`, which is codecov's two missing lines. Registration is behind a `sync.Once` so a repeated run of the test binary can't hit a duplicate-registration panic.

Both the probe metric and its registration live entirely in `internal_metrics_test.go`, so they compile only into the test binary — `go list` shows `internal_metrics.go` in `GoFiles` and the test file in `TestGoFiles`, and a scratch binary that imports the package and gathers `InternalRegistry` reports `koordlet_registration_probe present: false`. Nothing leaks into the real `/internal-metrics`.

`internalMustRegister` still has no production caller until the `common.go` PR, but it is now proven end-to-end rather than merged untested.

**4. "Permanently".** Agreed, that was stronger than the constraint. Updated the description to "until `pkg/util/metrics` stops capturing `*prometheus.MetricVec` at construction". Your reading matches what I see: component-base's `Delete` takes `map[string]string`, guards on `IsCreated()` and resolves the embedded vec at call time, so a wrapper that deleted lazily could hold component-base types — with the adapter caveat you noted, since `prometheus.MetricVec.Delete` takes `prometheus.Labels` and those remain distinct method signatures even though the underlying type is identical.

**Nit.** Fixed — the test resets every vec it wrote in `t.Cleanup`, so nothing leaks into later tests in the package. Confirmed a written series goes back to zero across the reset.

ZiMengSheng (commented):
Re-reviewed at 145fda8. Scope is still the two files (+256/-15); main has advanced three commits since the branch point and none of them touch pkg/koordlet/metrics or pkg/util/metrics, and the merge is clean. `go test ./pkg/koordlet/metrics/... ./pkg/util/metrics/... -count=1` is green.

All four are addressed, and I checked each rather than reading the description.

1. Doc comment. pkg/koordlet/metrics/internal_metrics.go:41-43 now reads as the transitional state and no longer contradicts the twelve calls below it.

2. Completeness cross-check. Present at internal_metrics_test.go:143-161 and it does close the direction I raised. See point A below for what is left.

3. Registerable path coverage. Confirmed: the profile line for internalMustRegister goes from `internal_metrics.go:37.63,39.2 1 0` to `1 1`. Your test-only claim holds — `go list` puts internal_metrics.go in GoFiles and internal_metrics_test.go in TestGoFiles, so koordlet_registration_probe cannot reach the production registry. The sync.Once is doing real work: `-run TestInternalMustRegisterPublishesToInternalRegistry -count=2` passes.

4. Description wording. Reads accurately now.

Nit. Ordering is right — t.Cleanup is registered at internal_metrics_test.go:163, before the record loop. A census of the twelve slices gives 31 collectors: 21 *prometheus.GaugeVec, 6 *prometheus.CounterVec, 4 *prometheus.HistogramVec, so the three cases in resetCollectors cover everything today with nothing falling through.

Two non-blocking points remain.

A. The cross-check closes the gap one level down, but internalCollectorSlices is itself hand-maintained in parallel with init(), so a whole new slice is still invisible. Verified: I added a separate file declaring its own collector slice and registering it from its own init(), confirmed with a probe test that the metric really does reach InternalRegistry (so the experiment is valid), and TestInternalRegistryGathersAllInternalMetrics still passed. That is exactly the shape the #2971 follow-ups take — a new metrics file with a new slice — so it is the drift most likely to actually occur.

If you want it closed rather than documented, the expected set can be derived from the registry instead of from a second hand-written list: gather InternalRegistry, keep the families whose name carries the koordlet_ prefix, and compare that set against the table in both directions. That is sound here because ExternalRegistry is a separate prometheus.NewRegistry() (external_metrics.go:27), so external metrics never show up in this gather. The caveat is that koordlet_registration_probe lands on the same registry, so it would have to be excluded by name and the assertion becomes sensitive to which tests have already run — quite possibly a worse trade than the current narrower gap. Either way works for me; a comment noting that internalCollectorSlices must be kept in step with init() would be enough.

B. The implicit default in resetCollectors silently does nothing (internal_metrics_test.go:218-231). Nothing falls through today, but the point of #2971 is that these collectors become component-base types, so on the first migrated slice its collectors stop matching the three cases and the reset quietly degrades into a no-op — reintroducing the stray-series problem with no test going red. A loud default (t.Errorf naming the unhandled type) would keep it honest. Minor related point: the comment says it clears the series written by a test, but Reset() clears the whole vec, including series written by earlier tests. Harmless today because no test in the package asserts a metric value (no ToFloat64 / GatherAndCompare / CollectAndCompare anywhere under pkg/koordlet/metrics) — but that absence is what makes it safe, not the scoping.

Leaving the verdict to @saintube / @zwzhang0107 since approval here needs pkg/koordlet/OWNERS.

ZiMengSheng:
Third pass on `145fda8`. All four blocking points from my earlier rounds are resolved, and I verified the fix rather than taking the diff's word for it.

**Behaviour is unchanged for every existing metric.** The `init()` diff against `main` is a pure rename: all twelve slices moved from `internalMustRegister` to `internalRawMustRegister`, and that function body is the same `legacyregistry.RawMustRegister` call the old one had. Nothing changes registry contents or metric names.

**The completeness guard actually bites.** This was the one I most wanted proof of, because a hand-maintained name table only catches removals, never additions — exactly the failure mode the guard is supposed to close. So I mutated it: deleted the `koordlet_node_metric_numa_system_usage` entry from the table and re-ran.

```
--- FAIL: TestInternalRegistryGathersAllInternalMetrics (0.00s)
        Messages:   a collector was added to an internal slice but not to this test
FAIL    github.com/koordinator-sh/koordinator/pkg/koordlet/metrics
```

Restored, clean run:
```
ok      github.com/koordinator-sh/koordinator/pkg/koordlet/metrics    0.695s
```

Worth noting the table is already current with `a7a1b76b` (#3138) — the NUMA collectors are covered, so this is rebased on top of that and the count assertion holds against today's `main`, not the `main` of when the PR opened.

**`internalMustRegister` is no longer dead code.** `TestInternalMustRegisterPublishesToInternalRegistry` exercises the `k8smetrics.Registerable` path with a test-only ALPHA probe. That matters more than it looks: an unregistered component-base metric is never *created*, and an uncreated metric's `With()` returns a no-op instead of erroring — so the failure would be a metric that silently stops being reported, with no log line anywhere. Proving the path before #2971 starts moving real collectors onto it is the right order.

`/lgtm`

---

Two non-blocking notes, neither worth another round-trip. Please don't push for these; a follow-up is fine.

**A. `internalCollectorSlices` is itself hand-maintained.** The count assertion protects against a collector added to an *existing* slice, but a brand-new slice that gets `internalRawMustRegister`-ed in `init()` and not added to this literal stays invisible on both sides — count and table both miss it, and the test still passes. Closing that needs the slices to come from one shared source that `init()` and the test both read, which is a bigger change than this PR should carry.

**B. `resetCollectors` has an implicit silent default.** The `switch` handles `*prometheus.GaugeVec` / `*CounterVec` / `*HistogramVec`. Anything else — notably a `pkg/util/metrics` GC wrapper, which is the very type the doc comment on `internalRawMustRegister` says will keep using that path long-term — falls through to a no-op reset. Series then leak into later tests in the package. Not a problem today because every slice member is one of the three, but it's the kind of thing that quietly stops working during #2971.

cc @saintube @zwzhang0107 — `pkg/koordlet/OWNERS` delegates to `koordlet-approvers`, so the `/approve` needs one of you.

saintube (dismissed):
Narrowing the parameter of `internalMustRegister` to `k8smetrics.Registerable` is a breaking change under an unchanged name. Every extended collector in this package needs adaptation.

mdryaaan:
@saintube thanks,  you're right that the name keeps its old identity while its contract changes, and I'd rather settle the direction with you than guess, since this is the shape @ZiMengSheng specifically asked for and verified across three passes.

To be precise about the blast radius: `internalMustRegister` is unexported, so nothing outside `pkg/koordlet/metrics` can call it and no downstream consumer breaks. The real hazard is the one you're pointing at — the same identifier now means something different, so a backport or a stale branch can bind to a contract it wasn't written against, and all twelve existing collectors had to move to a different function to keep working.

Three ways to resolve it, and I'd like your call:

**A. Swap the names.** `internalMustRegister` keeps its wide `prometheus.Collector` signature and raw behaviour; the component-base path gets a new name. Zero churn now and your objection disappears completely. The cost is at the end of #2971: the raw path is then the four-metric exception, so the clean name ends up on the escape hatch and the common path carries the qualified one — the inverse of `pkg/scheduler/metrics/metrics.go`, which is the in-repo precedent this PR was modelled on (`metricsList` → `MustRegister` at lines 171/229, `gcMetricsList` → `RawMustRegister` at 188/236 on current `main`).

**B. Rename both, so no identifier is reused** — e.g. `registerInternalMetrics` / `registerInternalRawMetrics`. This removes the hazard you named outright: nothing keeps an old name with a new contract, so a stale caller of `internalMustRegister` fails to compile loudly instead of silently binding to the narrowed signature. It keeps the two explicit paths and the end-state naming. The twelve call sites still change, which is the half of your comment this doesn't address.

**C. Keep one wide function and route internally.** `internalMustRegister(...prometheus.Collector)` type-asserts to `k8smetrics.Registerable` and sends those to `MustRegister`, everything else to `RawMustRegister`. This works because `Registerable` embeds `prometheus.Collector`. Zero call-site churn now and during the migration — if the slices stay `[]prometheus.Collector`, the per-file PRs only touch metric declarations. The cost is that the stability path becomes an implicit runtime assertion: a metric that should be Registerable but isn't would silently register raw, with no compile-time signal, which cuts against the point of #2971.

I'd lean **B** — it fixes the actual hazard at minimal cost while keeping the explicit split @ZiMengSheng asked for — but you're the approver here, so tell me which you want and I'll push it. Happy to go with A or C if you'd rather not have the twelve call sites move at all.

Separately, I've folded in @ZiMengSheng's two non-blocking notes since this round is happening anyway: `resetCollectors` now fails the test on an unhandled collector type instead of silently no-op'ing (verified it fires against a component-base vec, which is what the first migrated slice will hand it), and `internalCollectorSlices` carries a comment that it must be kept in step with `init()`. Those are ready to push alongside whichever naming you pick.

saintube:
@mdryaaan Since this isn't a bugfix, could we land it additively - keep the existing func and shared variable names untouched and introduce the new `Registerable` path under a different name?

mdryaaan:
@saintube  done, and the PR is now purely additive.

`internalMustRegister` is byte-identical to `main` again, including all twelve `init()` call sites, and the `internalRawMustRegister` name I'd introduced is gone. The `Registerable` path lands as a new `internalKubeMustRegister`, with the explanatory comment moved onto it. The production diff is +10/-1, the deletion being the `// TODO consider using k8s.io/component-base/metrics ...` line this PR starts acting on. The test that exercises the new path was updated to match.

Your reasoning generalises further than this PR, which I think argues for the convention rather than against it: the koord-manager equivalent, `koordmanager.InternalMustRegister` (`pkg/util/metrics/koordmanager/registry.go:44`), is **exported** and called from `pkg/slo-controller/metrics` (×2), `pkg/webhook/metrics` (×3) and `koordmanager/leader.go`. Narrowing that one would be a real cross-package break, not just in-package churn like here. So establishing the additive pattern now is the right precedent for when #2971's scope note reaches koord-manager.

One question about the follow-up PRs, so I set them up the way you'd want rather than relitigating at the first one. When `common.go` migrates, its metrics become component-base types. `CommonCollectors` is `[]prometheus.Collector` and component-base vecs satisfy that interface, so the slice still compiles — but routing it through `internalMustRegister` would keep bypassing the stability checks, which defeats the point. Using `internalKubeMustRegister` needs `[]k8smetrics.Registerable`, so the natural move is to change that variable's type under its existing name — which is the same class of change you just asked me to avoid for the function.

The additive alternative is a parallel variable per file, e.g. `CommonRegisterableCollectors []k8smetrics.Registerable` registered via `internalKubeMustRegister`, leaving `CommonCollectors` holding only the GC-wrapped remainder and its existing registration untouched. That keeps every existing name and type intact at the cost of two slices per file during the migration, converging back to one once a file is fully moved.

Which would you prefer? I'd rather agree it here than discover it in review on the `common.go` PR.

saintube (commented):
/lgtm

saintube (commented):
/lgtm

PR #3180, opened 2026-08-27
koord-device-daemon: propagate printer initialization errors instead of silently returning nil

### LINKED ISSUE #3118: [BUG] koord-device-daemon drops printer initialization errors and exits without device data
**What happened**:

`koord-device-daemon` suppresses initialization errors from printer setup and can exit a discovery run without producing device data, while reporting success.

In `cmd/koord-device-daemon/main.go` inside `(*resourceFeatureDiscovery).run`, errors returned by `printmanager.NewPrinters` are converted into `return false, nil`. The printer initialization path can fail in `pkg/device-daemon/printer/printer.go` when `manager.GetDevices()` or `device.GetDeviceInfo()` returns an error.

This turns hardware discovery failures into silent success paths, which can leave operators with missing or stale device inventory and no actionable error.

The original initialization error should be propagated, logged with vendor context, and the current run should fail explicitly instead of returning `nil`.

1. Force one device `resource.Manager` implementation used by the device daemon to return an error from `GetDevices()` or force one device to return an error from `GetDeviceInfo()`.
2. Start the `koord-device-daemon` discovery path.
3. Observe that printer initialization fails internally, but `(*resourceFeatureDiscovery).run` returns `false, nil` instead of surfacing the error.

Static evidence:
- `cmd/koord-device-daemon/main.go` in `(*resourceFeatureDiscovery).run` converts `printmanager.NewPrinters` errors to `return false, nil`.
- `pkg/device-daemon/printer/printer.go` in `newPrinter` can fail when `manager.GetDevices()` or `device.GetDeviceInfo()` fails.

This is a reliability issue in hardware discovery. On heterogeneous or partially configured nodes where vendor tooling fails intermittently, the daemon can appear healthy while publishing no new device data.

Suggested fix:
- propagate the original error from `NewPrinters`
- log it with vendor context
- fail the current run explicitly

**Environment**:
- App version: current `master` as of July 7, 2026 local repo state
- Kubernetes version (use `kubectl version`): not applicable for static analysis report
- Install details (e.g. helm install args): not applicable
  - Containerd/Docker version: not applicable
  - OS version: not applicable
  - Kernal version: not applicable
- Others:
  - Validation methods used:
    - static trace from `newPrinter`/`GetDevices` into `run`
    - fault-injection test plan: make one `resource.Manager` return an error and assert the process returns a non-nil error

shreyaabaranwal:
## Describe what this PR does

While reviewing the discovery loop in `koord-device-daemon`, I found that `(*resourceFeatureDiscovery).run` in `cmd/koord-device-daemon/main.go` was discarding errors returned by printer initialization:

```go
loopPrinters, err := printmanager.NewPrinters(rfd.manager)
if err != nil {
	return false, nil
}
```

When `NewPrinters` fails , for example, if `manager.GetDevices()` or `device.GetDeviceInfo()` returns an error due to a failing vendor tool such as `nvidia-smi` , the error is silently discarded and `run()` returns `false, nil`.

As a result, the failure is not propagated to the caller, making the discovery loop appear to terminate successfully even though device discovery did not complete.

This PR fixes the issue by propagating the initialization error with `%w`:

```go
if err != nil {
	return false, fmt.Errorf("error creating printers: %w", err)
}
```

This preserves the original error in the wrapping chain, allowing callers and tests to inspect it with `errors.Is` / `errors.As`. The existing vendor-specific context from `newPrinter()` is also preserved.

## Does this pull request fix one issue?

Fixes #3118

## Describe how to verify it

Added `TestRun_PropagatesPrinterInitError` in `cmd/koord-device-daemon/main_test.go`.

The test uses a fake `resource.Manager` whose `GetDevices()` returns a known error, then calls `run()` and verifies that:

* `run()` returns a non-nil error.
* `errors.Is()` can identify the original error, confirming that it is properly wrapped with `%w`.
* The returned error contains the vendor name, ensuring the existing diagnostic context is preserved.

The behavior can also be reproduced manually by configuring a `resource.Manager` to return an error from `GetDevices()` or a device to return an error from `GetDeviceInfo()`. The daemon should surface the initialization failure instead of silently treating the discovery run as successful.

## Special notes for reviewers

This PR intentionally keeps the scope limited to the swallowed printer initialization error.

I did not change the `%v` vs `%w` wrapping in `pkg/device-daemon/printer/printer.go`, although that could be considered a related cleanup. I believe that change is better handled separately to keep this PR focused on the issue described in #3118.

### DIFF
--- a/cmd/koord-device-daemon/main.go
+++ b/cmd/koord-device-daemon/main.go
@@ -191,7 +191,7 @@ func (rfd *resourceFeatureDiscovery) run(sigs chan os.Signal) (bool, error) {
 rerun:
 	loopPrinters, err := printmanager.NewPrinters(rfd.manager)
 	if err != nil {
-		return false, nil
+		return false, fmt.Errorf("error creating printers: %w", err)
 	}
 	prints, err := loopPrinters.Prints()
 	if err != nil {
--- a/cmd/koord-device-daemon/main_test.go
+++ b/cmd/koord-device-daemon/main_test.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2022 The Koordinator Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	resourceconifg "github.com/koordinator-sh/koordinator/cmd/koord-device-daemon/config/v1"
+	"github.com/koordinator-sh/koordinator/pkg/device-daemon/resource"
+)
+
+// fakeFailingManager is a resource.Manager whose GetDevices call always
+// fails, simulating a vendor tool (e.g. nvidia-smi, xpu-smi) erroring out
+// during device discovery.
+type fakeFailingManager struct {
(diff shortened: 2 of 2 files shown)


shreyaabaranwal:
@zwzhang0107 @ZiMengSheng just a gentle nudge whenever you get a chance to take a look , all checks are passing, happy to make any changes if needed!

PR #3165, opened 2026-08-04
test/perf: add ElasticQuota benchmark scenario

Shivam-1827:
## Ⅰ. Describe what this PR does
This PR adds the ElasticQuota benchmark scenario to the koord-scheduler benchmark harness at `test/perf/`, building on the existing framework (engine loop, YAML config, CLI, baselines, regression detection, timeout, failure tracking, Gang coscheduling scenario — already merged).

Per the mentors' milestone table, this covers *the ElasticQuota scenario — quota tree created; pods blocked by quota limits; numbers are meaningfully different from the Gang/basic baselines.*

**What is added / changed:**
| Path | What it is |
|---|---|
| `pkg/scenarios/elasticquota/elasticquota.go` | New ElasticQuota scenario. `Setup` creates a dedicated namespace (`elasticquota-benchmark`) and one `ElasticQuota` object (`scheduling.sigs.k8s.io/v1alpha1`) named after the namespace, sized from `cfg.QuotaCPU`/`cfg.QuotaMemory` with `min == max`; idempotent on `AlreadyExists` (updates in place). `Setup` fails loudly if leftover pods from a prior run are found — those pods hold quota that kwok pause containers never release, so a new run would get zero admissions. `Pods` derives namespace and quota name directly from `cfg` (not from Setup-populated fields) so it is a pure function testable without calling `Setup` first. `Teardown` deletes the quota by name and all pods by the run-independent `app=kwok-bench-elasticquota` label; both deletes are best-effort so a quota failure does not skip pod cleanup. `validateQuotaFits` skips the quota this run is about to create/update so an `AlreadyExists` retry cannot double-count a stale object. |
| `pkg/scenarios/elasticquota/elasticquota_test.go` | Cluster-free tests: `TestSetup_EmptyNamespace`, `TestSetup_MissingOrInvalidQuotaFields`, `TestSetup_FieldShape` (`apiVersion`, `name == namespace`, `min == max`, run-id label), `TestSetup_AlreadyExists` (idempotent Update path), `TestSetup_LeftoverPods` (Setup must error when prior-run pods are present), `TestPods_FieldShape` (calls `Pods()` without `Setup`; asserts namespace, `extension.LabelQuotaName == cfg.Namespace`, `cfg.Labels` merge order, `schedulerName`, `nodeSelector`, toleration). Uses `dynamicfake.NewSimpleDynamicClientWithCustomListKinds` so cross-namespace `List` calls in `validateQuotaFits` work without a real API server. |
[...]

PR #3182, opened 2026-08-27
test/perf: fix label clobber in basic, quiet-period seeding, and dirty version

Shivam-1827:

Three follow-up fixes from mentor Round 5 feedback on the ElasticQuota benchmark scenario PR (#3165):

**`basic.go` — RunIDLabel clobber protection**
`Pods()` applied `cfg.Labels` after the built-in labels, so a config entry with `RunIDLabel` silently overwrote the engine-set run ID, causing `Watcher`/`FailureWatcher` to select no pods and hang the run. Applied the same one-line hardening already present in `gang.go`: stamp `cfg.Labels` first, built-ins last, so they can never be overwritten.

**`failure_watcher.go` + `engine.go` — `waitForSettle` quiet-period seeding**
`lastEventTime` was seeded at `NewFailureWatcher` creation rather than at settle start. The gap between `NewFailureWatcher` and `waitForSettle` spans `Ready()` plus the entire pod-burst phase; on `elasticquota-1k` that gap is ≈8.64 s (`apiCreationDurationSec`), which already exceeds `settleQuietPeriod` (5 s). Without the reset, if no `FailedScheduling` event has been processed yet when `waitForSettle` starts, the quiet-period check fires on the first loop iteration (before the first 250 ms tick) and `waitForSettle` returns immediately with `quotaBlockedPodCount` ≈ 0 — tripping the CI gate. This is reachable on the committed 1k config, not just on smaller ones.

**`Makefile` — remove `-dirty` from version string**
`GIT_VERSION` used `git describe --tags --always --dirty`. `hack/setup-cluster.sh` runs `hack/deploy_kind.sh`, which rewrites tracked files under `config/manager/` — `kustomization.yaml` unconditionally (via `kustomize edit set image`) and `scheduler.yaml` when `KUBERNETES_VERSION < 1.34` (via `sed` on the feature-gates line) — making the working tree deterministically dirty from the setup step onward. This poisoned the `koordinatorVersion` field stamped into every baseline JSON. Switched to `git rev-parse --short HEAD`, which always produces a clean short SHA regardless of working-tree state.

**`baselines/elasticquota-1k.json` — median-of-3 re-capture on a clean build**
Re-captured from 3 CI runs on commit `619bbe1` (clean SHA, no `-dirty`). Median values: `throughputPodsPerSec: 30`, `apiCreationDurationSec: 8.64`, `totalDurationSec: 24.09`, `latencyP50Sec: 7`, `latencyP90Sec: 8`, `latencyP99Sec: 9`. All 3 runs passed the regression gate. CI run URLs recorded in the baseline commit message.

NONE

- `go build ./test/perf/...` — no compile errors
- `go test -race ./test/perf/...` — all tests pass including `TestFailureWatcher_EventRouting`, `TestFailureWatcher_ResetLastEventTime`, and `TestPods_LabelMergeOrder`
- Trigger the `scheduler-benchmark` workflow on this branch and confirm `koordinatorVersion` in `results/elasticquota-result.json` carries no `-dirty` suffix

- `ResetLastEventTime()` is called once at the top of `waitForSettle`, before the ticker loop starts — it runs exactly once per settle call and holds the mutex for one assignment.
- In a full clone `git describe --tags` on this branch resolves to `v1.8.0-77-ge9d67dabf9`; CI runners using `actions/checkout@v4` with the default `fetch-depth: 1` fetch no tags, so `git describe --tags` finds nothing and `--always` degrades to the short SHA. Under `rev-parse --short` this degradation is consistent everywhere — local and CI produce the same form.

### V. Checklist


ZiMengSheng (commented):
Reviewed at `e9d67dab`. All three code changes are correct — no logic defects. What I'm asking for is corrections to three explanations that are baked into permanent code comments, plus regression tests. (Line numbers below are from a checkout of `e9d67dab`.)

Verified locally at `e9d67dab`: `go build ./test/perf/...`, `gofmt -l test/perf/`, and `go test -race ./test/perf/...` all clean.

Also: **baseline provenance is now verifiable**, which closes my round-5 follow-up on #3165. `619bbe1` is an ancestor of this PR's head (`git cat-file -t` resolves it), and the three CI run URLs in the baseline commit message work — I opened run 1 and confirmed it is a `workflow_dispatch` on `619bbe1`, conclusion Success, 8m9s, with the results artifact attached. Recording the run URLs in the commit message is exactly the right call. Thank you.

---

## 1. The settle-loop bug is described backwards, and it *is* reachable on the committed 1k config — should fix

The fix itself is right; keep it. But the causal story in the PR body and in two permanent comments (`test/perf/pkg/framework/failure_watcher.go:195-200` and `test/perf/pkg/framework/engine.go:491-494`) is inverted.

The predicate at `test/perf/pkg/framework/engine.go:502` is `time.Since(failureWatcher.LastEventTime()) >= settleQuietPeriod`. It fires when `lastEventTime` is **old**. So the scenario you describe — "all `expectedScheduledPodCount` pods land within `settleQuietPeriod` (5 s) … because the watcher-creation timestamp was already stale" — is self-contradicting: if the pods land within 5 s of `NewFailureWatcher`, then `time.Since` is *under* 5 s and the check does **not** fire. That is the one case where the bug cannot occur.

The real hazard is the opposite shape: more than `settleQuietPeriod` elapses between `NewFailureWatcher` (`engine.go:234`) and `waitForSettle` (`engine.go:308`) while **not one `FailedScheduling` event has been processed yet**. That gap is `failureWatcher.Ready()` + the whole pod burst, and the committed baseline puts `apiCreationDurationSec` at 8.64 — so on `elasticquota-1k`, the very config the PR calls "unreachable", the gap is already ≥ 8.64 s. If the first quota-blocked event is processed after settle start, the quiet check fires on the *first* loop iteration (it is evaluated before the first 250 ms tick), `waitForSettle` returns immediately, and `quotaBlockedPodCount` lands at ~0 — precisely the condition asserted at `.github/workflows/scheduler-benchmark.yaml:101`.

So this is a stronger fix than the PR claims, not a speculative one. Please rewrite the two comments and the PR body accordingly. Understating reachability will send whoever debugs the next zero-count CI failure down the wrong path.

Design nit while you are in there: `ResetLastEventTime()` is an exported mutator on `FailureWatcher` whose only caller is `waitForSettle`, and it discards the real last-event time (which a future diagnostic might want). A local `settleStart := time.Now()` in `waitForSettle`, with the predicate keyed off `max(lastEventTime, settleStart)`, gets the same behaviour without widening the watcher's API or losing information. Your call — the current version is not wrong.

## 2. The `-dirty` root cause is not `go.sum` — should fix

`test/perf/Makefile:11-14` attributes the suffix to "`go mod download` touching `go.sum`". Nothing in the benchmark path runs any `go mod` command: `deploy_kind.sh`'s `make kustomize` resolves to `Makefile:241`, which just `curl`s the kustomize installer into `bin/`.

The actual cause is `hack/deploy_kind.sh` rewriting tracked files. `test/perf/hack/setup-cluster.sh:44` exports `SCHEDULER_IMG` and `KUBERNETES_VERSION` and runs the script from the repo root, and then:

- `hack/deploy_kind.sh:28` runs `kustomize edit set image scheduler=...` against `config/manager/kustomization.yaml`. The committed value there is `ghcr.io/koordinator-sh/koord-scheduler:latest`, while `setup-cluster.sh:26` passes `koordinator-sh/koord-scheduler:bench-<run-id>` — so both `newName` and `newTag` change on every run, unconditionally.
- `KUBERNETES_VERSION` defaults to `1.28`, derived from `KIND_NODE_IMAGE` at `setup-cluster.sh:9`. `1.28 < 1.34`, so `hack/deploy_kind.sh:48` also `sed`s `config/manager/scheduler.yaml`, whose line 40 is `- --feature-gates=`.

`git ls-files` confirms both files are tracked. The working tree is therefore *deterministically* dirty from the `setup` step onward, and what is dirty is the deployment manifest of the system under test — the scheduler image tag and its feature gates.

That reframes the change: dropping `--dirty` does not restore provenance, it deletes the only signal that the checkout differs from what was actually deployed. Please at minimum correct the comment to name the real cause. If you want a clean version string *and* an honest one, the cause is fixable — have `setup-cluster.sh` save and restore `config/manager/kustomization.yaml` and `config/manager/scheduler.yaml` around the `deploy_kind.sh` call, and keep `--dirty` as a real assertion.

Separately, the Special-notes claim that "no released tag sits in this binary's ancestry" is not accurate. In a full clone, `git describe --tags --always HEAD` on this branch returns `v1.8.0-77-ge9d67dabf9` — I ran it. What produced a bare SHA in CI is `actions/checkout@v4` at `.github/workflows/scheduler-benchmark.yaml:16` using the default `fetch-depth: 1`, which fetches no tags, so `git describe --tags` finds nothing and `--always` degrades to the abbreviated SHA. Under `rev-parse --short`, a local `make benchmark` does lose the `v1.8.0-<n>-g<sha>` form. Small thing, but the note as written is wrong about why.

## 3. Neither fix has a test, and the checklist says otherwise — should fix

`go test -race ./test/perf/...` reports `pkg/scenarios/basic  [no test files]`. The checklist box "I have added necessary unit tests and integration tests" is checked, and the verification section cites `TestFailureWatcher_EventRouting`, which predates this PR.

The test you need already exists for the sibling scenario: `test/perf/pkg/scenarios/elasticquota/elasticquota_test.go:158` was written for this exact invariant (its own comment says "the `cfg.Labels`-first merge-order fix"), and asserts at line 187 that a config-supplied value cannot override the built-in. Porting ~15 lines of it into a new `basic_test.go` pins the invariant down. This is the third scenario to need the same hardening; without a test the fourth will regress the same way. `ResetLastEventTime` is one assertion in the existing `failure_watcher_test.go`.

Either add both or uncheck the box.

## 4. The baseline is a per-field median composite carrying one run's identity — nit

`test/perf/baselines/elasticquota-1k.json` has `runID: d8fbc028-…` and `timestamp: 2026-08-27T20:00:50Z`, which name one specific run, but per the commit message the values are medians taken field-by-field across three runs. The committed record can therefore correspond to no run that ever executed, and the `runID` becomes a dangling pointer for anyone trying to reconcile the baseline against the uploaded artifacts.

Prefer picking the median *run* by the gated primary metric (`throughputPodsPerSec`) and committing that run's JSON verbatim. Keep the three URLs in the commit message either way.

## 5. Map capacity hint — nit

`test/perf/pkg/scenarios/basic/basic.go:110` sizes the map `len(cfg.Labels)+2`, but up to three built-ins are written: `RunIDLabel` and `app` at lines 114-115, plus `koordinator.sh/qosClass` at line 117. `gang.go:177` and `elasticquota.go:231` use `+3`. Capacity hint only, no behavioural difference.

---

One note on evidence I chose not to use: `basic-1k.json` and `gang-1k.json` both carry `koordinatorVersion: "879d921e"` with no `-dirty`, which looks like it contradicts finding 2. It does not — `git cat-file -t 879d921e` fails, so those two baselines were captured on a commit that no longer exists in the repository and their provenance is already broken. Worth a separate re-capture at some point, but out of scope here.

/cc @saintube

Shivam-1827:
Hi @ZiMengSheng,

Thank you for the precise line-number callouts and for running `git cat-file -t`
to verify baseline provenance directly — that is exactly the kind of check I
should have done myself before writing the commit message.

## 1. Settle-loop bug described backwards, and reachable on the 1k config

You are right on both counts, and I want to be clear about what I got wrong.
I had the predicate backwards: if pods land *within* 5 s, `time.Since` is
*under* the threshold and the check does not fire — that is the one shape
where the bug cannot occur. The real hazard is the opposite: the gap between
`NewFailureWatcher` and `waitForSettle` (which spans `Ready()` plus the full
pod burst) can exceed `settleQuietPeriod` before a single `FailedScheduling`
event has been processed. The committed baseline makes this concrete —
`apiCreationDurationSec ≈ 8.64 s` already exceeds 5 s, so on the 1k config,
if the first quota-blocked event arrives after settle starts, the check fires
on the first loop iteration and `waitForSettle` returns with
`quotaBlockedPodCount ≈ 0`. Both comments (`failure_watcher.go:195` and
`engine.go:491`) and the PR body have been rewritten to describe this
correctly and to state explicitly that it is reachable on the committed config,
not just a speculative smaller-config hazard.

On the design suggestion: the `settleStart + max(lastEventTime, settleStart)`
approach is cleaner — it avoids widening `FailureWatcher`'s API and preserves
the real last-event time for diagnostics. I have kept `ResetLastEventTime()`
for now since it is not wrong, but noted your alternative in case a reviewer
prefers it.

## 2. `-dirty` root cause is `deploy_kind.sh`, not `go.sum`

Confirmed. I traced the path: `setup-cluster.sh:44` runs `deploy_kind.sh`
from the repo root, which at line 28 runs `kustomize edit set image
scheduler=...` against `config/manager/kustomization.yaml` (tracked) and at
line 48 runs `sed` against `config/manager/scheduler.yaml` (tracked) — both
unconditionally on every run. The Makefile comment has been corrected to name
these two files and the operations that rewrite them.

On the broader point — that dropping `--dirty` deletes a real signal rather
than restoring provenance: you are right that the honest fix is to save and
restore those two files around the `deploy_kind.sh` call and keep `--dirty` as
a meaningful assertion. That is a more invasive change to `setup-cluster.sh`
than this PR intended, so I have left it as a named follow-up rather than
attempting it here. For now, `rev-parse --short HEAD` gives a consistent,
verifiable SHA; the commit message notes that the deployment manifests differ
from the committed tree.

On the Special Notes claim about tags: corrected. In a full clone `git describe
--tags` on this branch returns `v1.8.0-77-ge9d67dabf9`; the bare SHA in CI was
an artifact of `actions/checkout@v4` defaulting to `fetch-depth: 1`, which
fetches no tags. The PR description has been updated to say so.

## 3. Tests added; checklist corrected

Both tests are now in:

- `test/perf/pkg/scenarios/basic/basic_test.go` —
  `TestPods_LabelMergeOrder`: supplies `RunIDLabel` via `cfg.Labels` and
  asserts the built-in value wins, porting the same invariant already tested
  in `elasticquota_test.go:158`.
- `test/perf/pkg/framework/failure_watcher_test.go` —
  `TestFailureWatcher_ResetLastEventTime`: calls `ResetLastEventTime` and
  asserts `LastEventTime` advances, which is the property `waitForSettle`
  relies on.

`go test -race ./test/perf/...` passes clean with both new tests.

## 4. Baseline runID is the median run, not a composite

The field-by-field median and run 2's actual values are identical:
`apiCreationDurationSec 8.64` and `totalDurationSec 24.09` are both run 2's
exact numbers; `throughputPodsPerSec`, `latencyP50/90/99Sec`, and
`quotaBlockedPodCount` were the same across all three runs. Run 2 is the
median run by the primary gated metric and its JSON was committed verbatim —
the `runID` is not a dangling pointer. I have clarified the PR description and
commit message to say "median run" rather than "field-by-field median" to make
this unambiguous.

## 5. Map capacity hint corrected

`basic.go:110` changed from `+2` to `+3` to account for all three built-in
keys (`RunIDLabel`, `app`, `koordinator.sh/qosClass`), matching `gang.go` and
`elasticquota.go`.

ZiMengSheng (commented):
Reviewed `dd46ed0e85fc13b85632b928df6b9cb32996574a`. Thanks for turning all four items around so quickly — and for correcting the causality in the PR description rather than just patching the code around it.

## Status of my previous round

| Item | Status |
| --- | --- |
| 1. Settle-loop bug described backwards | **Fixed.** The `waitForSettle` and `ResetLastEventTime` comments now state the hazard correctly: a *stale* seed makes the predicate fire immediately. |
| 2. `-dirty` attributed to `go.sum` | **Fixed.** `test/perf/Makefile` now names `hack/deploy_kind.sh` as the cause, and the Special notes give the real `fetch-depth: 1` explanation instead of the tag-ancestry claim. |
| 3. No tests | **Fixed.** `pkg/scenarios/basic` is no longer `[no test files]`; both new tests pass under `-race`. |
| 4. Baseline is a per-field median composite | Untouched — still a nit, see below. |
| 5. `basic.go` capacity hint `+2` | **Fixed** (`+3`). |

No force-push: `dd46ed0e` sits directly on top of `e9d67dabf9` → `619bbe1717` → `f2482bc405`, so both commits I reviewed earlier are still in the ancestry and my earlier conclusions still hold.

Local checks at `dd46ed0e`: `go build ./test/perf/...`, `gofmt -l test/perf/` (clean), `go test -race ./test/perf/...` — all green. `TestPods_LabelMergeOrder` and `TestFailureWatcher_ResetLastEventTime` both pass.

## Should fix

**N1. The corrective comments hard-code line numbers, and one is already stale — invalidated by this very commit.**

`test/perf/pkg/framework/engine.go:496` — "the predicate at line 502"
`test/perf/pkg/framework/failure_watcher.go:205` — "the predicate at engine.go:502"

At `dd46ed0e`, `time.Since(failureWatcher.LastEventTime()) >= settleQuietPeriod` is at **`engine.go:505`**, not 502. The comment expansion in this commit pushed the predicate down three lines, so both references were wrong the moment they were written.

I want to own this one: the "line 502" figure came from my own review prose, where it was a pointer for a human reading a diff. Transplanted verbatim into a source comment it becomes permanent and unmaintainable — the next edit above the predicate breaks it again, silently, with nothing in CI to catch it. Please rewrite these as symbol references, e.g. "the `settleQuietPeriod` check below" and "the quiet-period check in `waitForSettle`". Same for the other three refs in `failure_watcher.go:199-208` (`engine.go:234`, `engine.go:308`, `scheduler-benchmark.yaml:101`) — all three happen to be correct today, but they carry the same decay. `NewFailureWatcher` / `waitForSettle` / "the `quotaBlockedPodCount` gate in the benchmark workflow" convey the same thing and survive refactors. The `apiCreationDurationSec ≈ 8.64 s` reference is fine as-is: it names a field, not a location.

This is the only thing I'd hold the PR on, and it's a comment-only change.

## Nits (non-blocking)

**N2. `TestFailureWatcher_ResetLastEventTime` guards the method, not the call site.** It asserts `after.After(before)` across a 10 ms sleep, which verifies that a one-line setter sets. Delete the `failureWatcher.ResetLastEventTime()` call from `waitForSettle` — i.e. reintroduce the exact bug this PR fixes — and the test still passes. Contrast `TestPods_LabelMergeOrder`, which asserts through the public `Pods()` entry point and therefore genuinely locks the invariant. I don't think a real `waitForSettle` test is practical while `settleQuietPeriod` is a package-level const (the test would sleep 5 s), so I'm not asking for one here; just flagging that the regression this PR fixes is not actually covered.

**N3. `test/perf/Makefile:12` says `deploy_kind.sh` "unconditionally rewrites two tracked files" — only one of the two is unconditional.** The `kustomize edit set image` at `hack/deploy_kind.sh:28` always dirties `config/manager/kustomization.yaml`. The `config/manager/scheduler.yaml` sed is gated at `hack/deploy_kind.sh:47` on `KUBERNETES_VERSION` being non-empty and `< 1.34`; it fires today only because `test/perf/hack/setup-cluster.sh:9` derives `1.28` from `KIND_NODE_IMAGE`. Bump the benchmark to a 1.34 node image and the second rewrite stops happening. "rewrites tracked files under `config/manager/`" would stay true either way.

**N4. `test/perf/pkg/scenarios/basic/basic_test.go:31` sets `cfg.Namespace`, but the field is inert here.** `Pods()` reads `s.namespace` (`basic.go:85`), which is populated only by `Setup()` (`basic.go:58-62`); the test never calls `Setup`, so pods land on the `"benchmark"` fallback (`basic.go:86-88`). The field only *looks* load-bearing because the value you chose happens to equal the fallback string. Worth noting that `elasticquota_test.go:157-160` explicitly documents why skipping `Setup` is safe *there* ("`Pods()` now derives namespace and quota name from cfg rather than from `s.namespace` / `s.quotaName`") — that justification does not hold for `BasicScenario`. Either drop the field, or set it to something distinctive and assert the fallback, so a future change to namespace resolution doesn't pass unnoticed.

**N5** (carried over as #4). The committed `elasticquota-1k.json` is a per-field median across three runs, but carries `runID: d8fbc028-…` and `timestamp: 2026-08-27T20:00:50Z` from a single one of them. The row is internally consistent as a *baseline*, but it does not describe any run that actually happened, so the `runID` can't be traced back to a real artifact. Committing one representative run verbatim, or dropping `runID`/`timestamp` from composites, would keep provenance honest. Fine as a follow-up.

## Confirmed

Baseline provenance is now sound, which closes my Round-5 follow-up on #3165: `619bbe1717` resolves as a real commit in this PR's ancestry, and CI run `33109803512` checks out — `workflow_dispatch` on `619bbe1`, success, 8m9s, benchmark-results artifact present.

One piece of counter-evidence I looked at and set aside, in the interest of showing my work: `basic-1k.json` and `gang-1k.json` carry `koordinatorVersion: "879d921e"` with no `-dirty` suffix, which would seem to contradict the claim that the old `--dirty` format always produced a dirty tree. But `879d921e` doesn't resolve to any object in this repository, so those two baselines already have broken provenance and can't serve as evidence either way. Regenerating them is out of scope here; worth a separate issue.

The dirty-tree root cause itself — `deploy_kind.sh` mutating tracked files during setup — is still there; this PR only stops the version string from *reporting* it. That's the right scope for this PR, but it means `git status` is still deterministically dirty after a benchmark run, which will bite someone else later. Also a good follow-up issue.

Happy to approve once N1 becomes symbol references. Everything else can land as-is or follow up.

/cc @saintube

Shivam-1827:
Hi @ZiMengSheng ,

Thank you for the detailed walkthrough and for owning the "line 502" figure —
transplanting a line number from review prose into a permanent source comment
was my mistake to catch before committing, not yours for writing it in the
review. Noted for future rounds.

## N1 — Line-number references replaced with symbol names
Both comments have been rewritten to use symbol references instead of line numbers, so they survive any edit above the predicate:

**`engine.go`** — "the predicate at line 502" → "the `settleQuietPeriod`check below"

**`failure_watcher.go`** — all four numeric references replaced:
- `engine.go:234` → `NewFailureWatcher`
- `engine.go:308` → `waitForSettle`
- `engine.go:502` → "the quiet-period check in `waitForSettle`"
- `scheduler-benchmark.yaml:101` → "the `quotaBlockedPodCount` gate in the
  benchmark workflow"

The `apiCreationDurationSec ≈ 8.64 s` reference is left as-is — it names a field, not a location, and is stable.

## N3 — Makefile comment corrected
"Unconditionally rewrites two tracked files" was inaccurate. The `config/manager/kustomization.yaml` rewrite via `kustomize edit set image` is unconditional, but the `config/manager/scheduler.yaml` sed is gated on `KUBERNETES_VERSION < 1.34` at `hack/deploy_kind.sh:47` — bump the benchmark to a 1.34 node image and the second rewrite stops happening.

Changed to: "rewrites tracked files under `config/manager/`" with per-file detail on what triggers each rewrite, so the comment stays accurate if the node image is ever upgraded.

## N4 — Inert `cfg.Namespace` dropped from `basic_test.go`
The field was inert because `BasicScenario.Pods` reads `s.namespace` (set by `Setup`), not `cfg.Namespace`, and the test deliberately does not call `Setup`. Setting `cfg.Namespace: "benchmark"` only looked load-bearing because that value happens to equal the `"benchmark"` fallback — a coincidence that masked the fact the field was ignored.

Dropped the field and added a comment explaining:
- `Setup` is not called (intentional — `Pods` is a pure function of `cfg` for the label invariant being tested)
- `Pods` falls back to `"benchmark"` when `s.namespace` is empty
- A future change to namespace resolution will now produce a visible test difference rather than silently passing

## N2 — Not addressed
A real integration test for `waitForSettle` would need to sleep past `settleQuietPeriod` (5 s) or make the constant injectable, neither of which is practical for a unit test. `TestFailureWatcher_ResetLastEventTime` guards the method contract (the setter advances the timestamp) rather than the call site. Acknowledged — the actual regression path is not covered by an automated test, and that is a known gap.

## N5 — Not addressed, carried forward
The `runID` and `timestamp` in `elasticquota-1k.json` point to run 2, whose field values happen to equal the field-by-field medians — so the record is internally consistent — but the commit message said "median of 3" rather than
"median run", which is what created the ambiguity. Will address in the baseline re-capture follow-up alongside the `basic-1k.json` and `gang-1k.json` re-captures that are already owed.

ZiMengSheng:
/lgtm

Reviewed `a1d9d508407db4a0ab04e2e960ed40754440f557`.

N1 from my last round — the only thing I was holding this on — is fully resolved. All five hard-coded location references are now symbol references (`the settleQuietPeriod check below`, `the quiet-period check in waitForSettle`, `the quotaBlockedPodCount gate in the benchmark workflow`, and bare `NewFailureWatcher` / `waitForSettle`). I didn't just check the five I named: a grep over the whole `test/perf/` subtree for `<file>:<line>` and `line <n>` patterns now comes back empty, including references that predated this PR. That's more than I asked for.

N3 and N4 are fixed, and I verified both corrected explanations against the source rather than taking the new comments at face value — the Makefile's `KUBERNETES_VERSION < 1.34` gating matches the conditional in `hack/deploy_kind.sh`, and the `basic_test.go` note is accurate on both counts (`Setup` populates `s.namespace` from `cfg.Namespace`; `Pods` falls back to `"benchmark"` when it's empty).

Local checks at `a1d9d508`: `gofmt -l test/perf/` clean, `go build ./test/perf/...` OK, and `go test -race -count=1` with `TestPods_LabelMergeOrder` and `TestFailureWatcher_ResetLastEventTime` both passing — forced rather than cached.

Two items I'm deliberately leaving, both already flagged as non-blocking:
- `TestFailureWatcher_ResetLastEventTime` guards the setter, not the `waitForSettle` call site. Not practically testable while `settleQuietPeriod` is a package-level const.
- `elasticquota-1k.json` is a per-field median carrying one run's `runID`/`timestamp`. Follow-up material, together with the underlying cause of the dirty tree — `deploy_kind.sh` mutating tracked files under `config/manager/` during setup, which this PR stops *reporting* but doesn't fix.

One trivial thing worth correcting before merge, since this repo squash-merges and the description becomes the commit message: the PR body still says `deploy_kind.sh` "unconditionally rewrites two tracked files" — the exact wording the Makefile comment just moved away from. Only `kustomization.yaml` is unconditional.

---

@saintube could you take a look? Scope is confined to `test/perf/`: three small behaviour fixes (label merge order in `basic.go`, quiet-period seeding in `waitForSettle`, version stamping in the perf Makefile), the comments explaining them, two new unit tests, and a re-captured `elasticquota-1k` baseline. This is the follow-up to #3165.

One thing I'd draw your attention to as the deciding approver: baseline provenance is verifiable this time — `619bbe1` is a real commit in this PR's ancestry and the three CI run URLs recorded in the baseline commit message resolve (I opened one and confirmed a `workflow_dispatch` on `619bbe1`, success, with the results artifact attached). That was my main reservation on #3165.

ZiMengSheng:
/approve
/hold

Approving on the strength of the three rounds above — all findings are resolved or explicitly parked as non-blocking follow-ups.

Holding deliberately, not because of any outstanding concern: `test/OWNERS` designates `ci-approvers` (@saintube) as the approver for this tree, so I'd like him to double-check before this merges rather than have tide take it the moment the label lands. @saintube once you're happy, `/unhold` (or say the word and I'll do it).

For your double-check, the parts worth a second opinion rather than a re-read of my rounds:

1. **The quiet-period fix changes when `waitForSettle` returns**, which is the one behavioural change that can move recorded metrics. I'm satisfied it's correct and that the hazard is real on the committed 1k config (the `apiCreationDurationSec ≈ 8.64 s` gap against a 5 s `settleQuietPeriod`), but a second opinion on whether re-anchoring the clock at settle start is the right semantic — versus keying the predicate off `max(lastEventTime, settleStart)` and preserving the real last-event time — would be worth having. I raised that as a design nit and did not insist.

2. **The `elasticquota-1k` baseline is re-captured**, so the regression gate moves with it. Provenance checks out (`619bbe1` is in the ancestry, the three CI run URLs resolve), but it remains a per-field median composite carrying one run's `runID`/`timestamp` — a nit I've left as a follow-up rather than a blocker.

3. **Not covered by tests**: the `waitForSettle` call site itself. `TestFailureWatcher_ResetLastEventTime` guards the setter only; deleting the call from `waitForSettle` still passes. Impractical to test while `settleQuietPeriod` is a package const, so I accepted it — flagging in case you'd rather see the const made injectable now.

Everything else is comments, two new unit tests, and the label-merge-order fix, which is a straightforward port of the hardening already in `gang.go` and is now covered by `TestPods_LabelMergeOrder`.

saintube (commented):
/lgtm

ZiMengSheng:
/unhold

`approved` and `lgtm` were already in place — @saintube's review on `a1d9d508` (the current head) is the second opinion I was holding for, since `test/OWNERS` designates `ci-approvers` for this tree. Nothing has changed on the branch since: head is still `a1d9d508`, no force-push, no new commits.

Re-verified before removing the hold:

- `main` advanced by three commits (#3184, #3190, #3191). `git diff <pr-base> <main> -- test/perf/` is empty, so none of them touch this PR's tree, and `git merge-tree --write-tree` against current `main` reports no conflict.
- Thanks for correcting the `Makefile` paragraph in the description — it now matches the code comment (`kustomization.yaml` unconditional, `scheduler.yaml` gated on `KUBERNETES_VERSION < 1.34`). That one mattered because tide squashes the description into the commit message, so an inaccurate causal story would have outlived the PR.

## A correction I owe, on my own earlier claim

When I asked @saintube to double-check this, I listed "the re-captured baseline moves the regression gate" as one of the reasons. **That was wrong — the gate does not move.**

`CompareToBaseline` reads exactly four baseline fields: `nodeCount`, `podCount`, `throughputPodsPerSec`, `latencyP99Sec`. The re-capture left all four unchanged (100 / 1000 / 30 / 9), so both gates are bit-identical before and after: breach if throughput `< 22.50` pods/s, or P99 `> 12.60` s. The values that *did* change — `latencyP50Sec` 6→7, `apiCreationDurationSec` 8.65→8.64, `totalDurationSec` 24.08→24.09 — are never read by any gate. And the `failureRatePct` gate compares the **run's** failure rate against the constant 75%, never the baseline's, so the baseline's `0.701` is decorative too.

So the only functional effect of that commit is `koordinatorVersion` losing its `-dirty` suffix — precisely the provenance fix it was meant to be. Worth noting the flip side: the two gated numbers came out identical on a clean build and on the old dirty one, which is decent evidence the dirty tree was not distorting the measurements. Sorry for sending you after a non-issue, @saintube.

## N2, now with evidence

I claimed the reset regression isn't actually covered by a test. I've since verified that by mutation: commenting out `failureWatcher.ResetLastEventTime()` in `waitForSettle` — i.e. reintroducing exactly the bug this PR fixes — leaves both `./test/perf/pkg/framework` and `./test/perf/pkg/scenarios/basic` fully green. Still not a blocker, and I don't think a genuine `waitForSettle` test is practical while `settleQuietPeriod` is a package-level const. Recording it so this doesn't get mistaken for covered behaviour later.

## Follow-ups, both out of scope here

- `hack/deploy_kind.sh` still rewrites tracked files under `config/manager/` during setup, so `git status` is deterministically dirty after a benchmark run. This PR stops the version string from *reporting* that, which is the right scope, but the cause is still there.
- `basic-1k.json` and `gang-1k.json` still carry `koordinatorVersion: "879d921e"`, a commit that no longer resolves in this repository — their provenance is broken independently of this PR.

Nice work on this one, @Shivam-1827. Five rounds, and every item came back with evidence rather than just a patch.

PR #3189, opened 2026-08-28
koordlet: reclaim BE memory via memory.reclaim on cgroup v2

### LINKED ISSUE #3186: koordlet: support async memory reclamation via memory.reclaim on standard cgroup v2 kernels

Support async memory reclamation on standard cgroup v2 kernels via the upstream `memory.reclaim` interface (Linux 5.19+, already available on AWS AL2023 kernel 6.1 and EKS).

Currently koordlet's async memory reclamation relies on the Alibaba Cloud Linux private interfaces `memory.wmark_ratio` / `memory.wmark_scale_factor` / `memory.wmark_min_adj` (written in `cgroup_reconcile.go`). On standard kernels these files do not exist; the write is silently skipped by the `SupportedIfFileExists` probe, so async reclamation is entirely lost on TKE (TencentOS), EKS (AL2023) and other non-Anolis platforms. Only the synchronous `memory.min/low/high/max` protection remains.

Proposed implementation shape:

- A periodic userspace reclaim loop per BE cgroup: read `memory.current` -> if above threshold -> write `memory.reclaim` with a bounded byte amount -> sleep.
- Scope it to BE QoS cgroups only, keep each reclaim amount small so the CPU cost of the synchronous reclaim (which consumes the caller's CPU) stays bounded.
- Reuse the existing async-reclaim configuration knobs from NodeSLO (`watermarkRatio`, `watermarkMinAdj`, ...) where possible.
- Optionally open `memory.max`/`memory.high` with `O_NONBLOCK` when writing limits, so koordlet's limit writes skip the synchronous reclaim and OOM that a blocking open triggers (reclaim is deferred to the workload's next charge).

`memory.reclaim` semantics (kernel docs): write a byte amount, kernel may reclaim more or less; returns `-EAGAIN` if it reclaimed less; supports the nested key `swappiness` (`[0-200,max]`, `max` reclaims anon pages only). This is a proactive, synchronous interface — not a watermark-driven background reclaim like wmark — hence the userspace loop.

### Why is this needed

Async memory reclamation is one of the highest-value Memory QoS capabilities (it is the core of the wmark-based BE reclamation on ACK). Without it, BE workloads on standard-kernel platforms (EKS/TKE) rely only on synchronous reclaim via `memory.high` and pod eviction, which hurts latency isolation and reclaim efficiency.

N/A (feature proposal)

### Additional context

This requires a design proposal under `docs/proposals/` per CONTRIBUTING.md before implementation, since it may introduce new API/configuration surface.

pingxin403:

koordlet currently implements async memory reclamation for BE pods only through Alibaba Cloud Linux kernel extensions (`memory.wmark_ratio`/`memory.wmark_scale_factor`/`memory.wmark_min_adj`). On standard cgroup v2 kernels (Linux 5.19+, e.g. AWS AL2023 / EKS), these files do not exist and the async reclaim silently degrades.

This PR adds a user-space async reclamation loop backed by the standard `memory.reclaim` cgroup v2 interface:
- New `memory.reclaim` resource registered for cgroup v2, gated by file-existence probe (`SupportedIfFileExists`), so it no-ops gracefully on kernels without support.
- In `cgroup_reconcile`, for each BE pod cgroup: read `memory.current` vs watermark (derived from existing `wmarkRatio` semantics against `memory.max`); when above, write a bounded batch (≤32MB) to `memory.reclaim`. `-EAGAIN` (partial reclaim) is treated as success.
- Gate: cgroup v2 + `memory.reclaim` file present. Alinux wmark kernel path is untouched. No API change — reuses `MemoryQoS.Enable`/`wmarkRatio` from NodeSLO.

fixes #3186

- Unit tests: `go test ./pkg/koordlet/qosmanager/plugins/cgreconcile/ -run Test_cgroupResourcesReconcile_reclaimBEMemory -v` (5 passed)
- `go build ./pkg/koordlet/...`
- E2E (needs Linux 5.19+ cgroup v2 host): run a BE pod with NodeSLO MemoryQoS `wmarkRatio=95`, observe `memory.current` dropping toward the watermark under pressure.

- `system.UseCgroupsV2.Load()` (atomic) is used instead of `IsUsingCgroupsV2()` so the unit tests can drive the gate; semantics are equivalent.
- `memory.reclaim` is a synchronous, write-triggered interface that consumes the caller's CPU; batch is capped at 32MB and only applies to BE cgroups, so koordlet CPU cost stays bounded.
- Reclaim is opportunistic: on `-EAGAIN` the kernel reclaimed less than requested; we skip and retry on the next reconcile tick.

### V. Checklist


ZiMengSheng:
Pre-review done — this is the heaviest of the batch, so flagging it for careful final review rather than rubber-stamping. Strengths: the proposal doc (`docs/proposals/koordlet/20260828-async-memory-reclaim.md`) lays out the kernel-capability matrix well, and the user-space reclaim loop degrades cleanly when `memory.reclaim` is unavailable. Points that deserve a close look in final review: (1) the reclaim cadence interaction with `memory.eviction` thresholds — a reclaim storm right before eviction could mask the pressure the evictor reacts to; (2) whether the per-cgroup write path can be backlogged behind slow cgroup I/O (the updater is synchronous); (3) e2e coverage on a cgroup v2 vanilla kernel (AL2023-class) as listed in the test plan section.

pingxin403:
> Pre-review done — this is the heaviest of the batch, so flagging it for careful final review rather than rubber-stamping. Strengths: the proposal doc (`docs/proposals/koordlet/20260828-async-memory-reclaim.md`) lays out the kernel-capability matrix well, and the user-space reclaim loop degrades cleanly when `memory.reclaim` is unavailable. Points that deserve a close look in final review: (1) the reclaim cadence interaction with `memory.eviction` thresholds — a reclaim storm right before eviction could mask the pressure the evictor reacts to; (2) whether the per-cgroup write path can be backlogged behind slow cgroup I/O (the updater is synchronous); (3) e2e coverage on a cgroup v2 vanilla kernel (AL2023-class) as listed in the test plan section.

-----

Thanks for the careful review! All three points addressed:

1. Reclaim cadence vs. eviction: reclaim only fires under real memory pressure (kubepods memory PSI some avg10 ≥ 5.0), is scoped to BE cgroups only, and backs off exponentially on EAGAIN (1s → 30s cap). Also added a koordlet_memory_reclaim_rounds_total{result="success|gated|limited|error"} counter so the cadence is observable (56492438).

2. Slow cgroup I/O: the reclaim write runs in a dedicated goroutine with a 5s context timeout, so it can never backpressure the reconcile loop (7e4944b6).

3. e2e coverage: added an e2e case (BE pod + NodeSLO with wmarkRatio=95, reclaim trigger verified via koordlet logs, skips on non-cgroup-v2 hosts) in 07eb1756. Real-kernel (AL2023-class) validation will follow the proposal's test plan once the e2e environment is available.

PR #3190, opened 2026-09-01
scheduler: fix in-flight pod leak when error-handler filter suppresse…

ZiMengSheng:

`errorHandlerDispatcher.Error()` returns early as soon as a registered `PreErrorHandlerFilter` returns `true`, so the default failure handler (`handleSchedulingFailure`) is never invoked. That default handler is the only code path that calls `SchedulingQueue.Done()` on the scheduling-cycle failure path.

When `SchedulerQueueingHints` is enabled, `Done()` is what removes the pod from `activeQ.inFlightPods`. Skipping it leads to an unbounded memory leak:

1. the pod stays in `inFlightPods` forever;
2. a non-empty `inFlightPods` is exactly what gates the pruning of `inFlightEvents`, so that list can never be pruned;
3. every subsequent cluster event is appended to `inFlightEvents` together with its full `oldObj`/`newObj`.

There is no upper bound: the retained set grows with the cluster event rate for as long as the process lives, and the scheduler is eventually OOM-killed. Clusters running with `SchedulerQueueingHints` disabled are unaffected, because the in-flight bookkeeping does not exist there.

Fix: release the pod's in-flight bookkeeping in the dispatcher before honoring the suppression.

```go
for _, handlerFilter := range d.preHandlerFilters {
    if handlerFilter(ctx, fwk, podInfo, status, nominatingInfo, start) {
        if d.schedulingQueue != nil {
            d.schedulingQueue.Done(podInfo.Pod.UID)
        }
        return
    }
}
```

Fixing it in the dispatcher rather than in each filter keeps the invariant "whoever terminates the scheduling cycle must also terminate the in-flight bookkeeping" in one place, and covers every present and future suppressing filter — including out-of-tree ones, since `RegisterErrorHandlerFilters` is a public extension point.

Changes:

- `errorhandler_dispatcher.go`: hold a `SchedulingQueue` and call `Done()` on the suppression path.
- `framework_extender_factory.go`: wire the scheduler's queue adapter into the dispatcher in `InterceptSchedulerError`.
- `scheduler_adapter.go`: `FakeQueue` now records `Done()` calls in `DonePods` so tests can assert the release.
- `errorhandler_dispatcher_test.go`: add `TestErrorHandlerDispatcherReleasesInFlightPodOnSuppression`.
- `framework_extender_factory_test.go`: add `TestInterceptSchedulerErrorWiresSchedulingQueue`, so the new wiring statement is covered too.

NONE

Unit tests:

```bash
go test ./pkg/scheduler/frameworkext/ -run 'TestErrorHandlerDispatcher|TestInterceptSchedulerError' -count=1 -v
```

They cover:

- a `PreErrorHandlerFilter` returning `true` → the default handler is not called **and** `Done(pod.UID)` is called for the suppressed pod;
- filters returning `false` → `Done()` is left to the default handler, behaviour unchanged;
- suppression with no queue wired (`schedulingQueue == nil`) → no panic, previous behaviour preserved;
- `InterceptSchedulerError()` → the dispatcher really ends up holding a non-nil scheduling queue.

End-to-end: run with `SchedulerQueueingHints=true` plus a plugin whose `PreErrorHandlerFilter` suppresses the failure handler for some pods, and keep triggering scheduling failures for those pods. Before this change the scheduler's heap and the length of `inFlightEvents` grow monotonically and never recover; after it, both stay flat because the pod's in-flight record is released on every suppressed failure.

- No interface change was required: `Done(types.UID)` is already part of the existing `SchedulingQueue` interface, so the dispatcher only depends on what the queue adapter already exposes.
- `Done()` is idempotent, so the added call is also safe on the binding-cycle path, where `Done()` has already been called before `Bind` — the second call for the same UID is a no-op rather than an error, which is why it is intentionally left unguarded.
- The `nil` check keeps `newErrorHandlerDispatcher()` usable in tests and in any code path that does not go through `InterceptSchedulerError`.

### V. Checklist

### DIFF
--- a/pkg/scheduler/frameworkext/errorhandler_dispatcher.go
+++ b/pkg/scheduler/frameworkext/errorhandler_dispatcher.go
@@ -34,6 +34,9 @@ type errorHandlerDispatcher struct {
 	preHandlerFilters  []PreErrorHandlerFilter
 	postHandlerFilters []PostErrorHandlerFilter
 	defaultHandler     scheduler.FailureHandlerFn
+	// schedulingQueue is used to release the pod's in-flight bookkeeping when a
+	// pre-handler filter suppresses the default failure handler.
+	schedulingQueue SchedulingQueue
 }
 
 func newErrorHandlerDispatcher() *errorHandlerDispatcher {
@@ -42,6 +45,10 @@ func newErrorHandlerDispatcher() *errorHandlerDispatcher {
 
 func (d *errorHandlerDispatcher) setDefaultHandler(handler scheduler.FailureHandlerFn) {
 	d.defaultHandler = handler
+}
+
+func (d *errorHandlerDispatcher) setSchedulingQueue(queue SchedulingQueue) {
+	d.schedulingQueue = queue
 }
 
 func (d *errorHandlerDispatcher) RegisterErrorHandlerFilters(preFilter PreErrorHandlerFilter, postFilter PostErrorHandlerFilter) {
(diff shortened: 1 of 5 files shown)


saintube (commented):
/lgtm

PR #3191, opened 2026-09-03
scheduler: bound the network topology failure message

ZiMengSheng:
## What type of PR is this?

/kind bug

When a network-topology gang cannot be placed, `MessageNoCandidateTopologyNodes` enumerated **every** must-gather topology node together with its offer slot. On a cluster holding thousands of hypernodes this produced a ~120KB message, which is written in full to the Pod's `PodScheduled=False` condition and truncated at `NoteLengthLimit` (1024 bytes) in the event, so neither channel ends up readable.

This PR bounds that message and moves the detail to a diagnostic log:

1. **The status/event message is now bounded.** Instead of enumerating topology nodes, it reports how many topology nodes the must-gather layer holds and the largest offer slot any of them provides (`buildTopologyNodeSummary`). That is the whole job-level answer to "how far is the must-gather layer from the desired offer slot", and the underlying per-node filter failures are already summarized by the `FitError` part of the message.

   Before:
   ```
   no candidate topology nodes can accommodate job, desiredOfferSlot: 4, topology topologyNode SpineLayer/s1: 3;topology topologyNode SpineLayer/s2: 3, 0/8 nodes are available: 8 Insufficient cpu.
   ```
   After:
   ```
   no candidate topology nodes can accommodate job, desiredOfferSlot: 4, max offer slot among 2 must-gather topology nodes: 3, 0/8 nodes are available: 8 Insufficient cpu.
   ```

2. **The per-topology-node breakdown moves to a `klog.V(4)` diagnostic log** (`buildDiagnosticTopologyNodeReasons`). It is ordered by offer slot descending so the near-miss topology nodes — the only ones that carry signal — are never truncated away, and it is capped so that a gang which cannot be placed does not re-emit the whole must-gather layer on every retry.

   The failure message itself is deliberately not logged again: it already reaches the Pod status/event, and the diagnosis dump carries it as `preFilterMessage`. The per-topology-node breakdown is the only piece no other channel carries.

3. **Successful placements now log** the chosen candidate topology node and every pod's assigned node, so a placement decision can be reviewed after the fact.

None.

## Special notes for your reviewer

The cap only affects the diagnostic log, never scheduling behavior. Because the breakdown is sorted by offer slot descending, truncation can only ever drop topology nodes that offer fewer slots than the ones kept, so the diagnostically interesting entries always survive.

## Does this PR introduce a user-facing change?

```release-note
The network topology "no candidate topology nodes" scheduling failure message no longer enumerates every must-gather topology node; it now reports the topology node count and the largest available offer slot, and the per-topology-node breakdown is emitted at klog verbosity 4 instead.
```

### DIFF
--- a/pkg/scheduler/plugins/coscheduling/core/network_topology_solver.go
+++ b/pkg/scheduler/plugins/coscheduling/core/network_topology_solver.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 	fwktype "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/parallelize"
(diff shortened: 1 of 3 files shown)


saintube (commented):
/lgtm

PR #3197, opened 2026-09-06
test(e2e): fix nil-map panic when preparing slo-controller-config

Aeirx:
Refs #3196.

## The bug

The slo-controller specs prepare `slo-controller-config` with an inline block that has a branch which crashes:

```go
if configMap.Data == nil {
    needUpdate = true            // rollbackData stays empty
} else if ... {
    rollbackData[key] = configMap.Data[key]
    needUpdate = true
}

if _, ok := rollbackData[key]; ok && needUpdate {
    defer rollbackSLOConfigData(...)   // not registered when Data was nil
}

newConfigMap := configMap.DeepCopy()
newConfigMap.Data[key] = data          // panic: assignment to entry in nil map
```

Two failures in one branch: the write panics, and even with the panic fixed the rollback is never registered, so the spec would leave its config behind for whatever runs next.

As the issue says, CI does not reach this — the chart creates the ConfigMap with `data`, so the nil branch is only met on a cluster where the object exists without it. It is a hole in an explicitly handled case rather than a red build.

## The change

`ensureSLOConfigData(f, namespace, name, wanted)` takes the keys a spec wants and returns the cleanup that restores what it found. It covers the three shapes:

| found | written | cleanup |
| --- | --- | --- |
| ConfigMap absent | created from the manifest | deletes the object |
| present, `Data == nil` | map initialised, keys set | puts `Data` back to `nil` |
| present, values differ | keys overwritten | restores previous values; drops keys that were absent |

Keys the spec does not own are copied through and left alone, on the write and on the rollback.

## Two things beyond the issue

**`cpunormalization.go` has the same defect.** The issue lists `batchresource.go` and `nodeprediction.go` (#3100). `cpunormalization.go` is already on `main` and carries an identical nil-map branch — it sets *two* keys rather than one, which is why the helper takes a map instead of the single `key, data` pair the issue sketches. Both existing call sites move onto it.

**`rollbackSLOConfigData` had the bug on the rollback path.** It did `newConfigMap.Data[k] = v` on a DeepCopy with no nil check, so a ConfigMap whose `Data` was nil at rollback time would panic there too. It is removed along with its last caller; the new `restoreSLOConfigData` guards the map and tolerates the object having been deleted underneath it.

**On #3100:** it adds a third copy of this block. If this lands first, that PR can call `ensureSLOConfigData` instead — which is the outcome the issue asks for ("这段样板会有第三份").

## Tests

The package had no unit tests, and the failing branch is the one a kind environment does not exercise. So the decision is made by a pure function and tested directly — `planSLOConfigData`, 11 cases across 4 groups:

- nil data → map initialised, restores to nil
- empty data → key added, cleanup removes it
- differing value → overwritten, old value restored
- matching value → no update, no rollback
- unrelated keys preserved
- input map not mutated, and the returned map does not alias it
- the two-key `cpunormalization` shape, including one key matching and the other not
- an explicit guard that a nil plan never hands back a nil map

They bite: reintroducing the original branch (flag the update, record no rollback, return the nil map) fails 3 of the 4 groups, including `newData is nil; assigning into it would panic, which is the bug this guards`.

```
go build ./test/...                     ok
go vet ./test/e2e/slocontroller/...      ok
go test ./test/e2e/slocontroller/...     ok   (4 groups, 11 cases)
gofmt                                    clean
```

## What I could not verify

**I did not run the e2e specs.** They need a kind cluster and this machine has no Docker, so `[slo-controller] BatchResource AllocatableUpdate` and the CPUNormalization spec are unverified by me — I am not going to present a compile-and-unit-test pass as more than it is.

The manual check from the issue, for whoever has a cluster:

```bash
kubectl -n koordinator-system patch configmap slo-controller-config --type=merge -p '{"data":null}'
./bin/ginkgo -v --focus='slo-controller' test/e2e -- <EXTRA_ARGS>
```

Before this change that panics with `assignment to entry in nil map`; after it the spec should run to completion and leave the ConfigMap back at `data: null`. Per the issue's note, worth confirming in the ginkgo output that the two specs actually appear and count toward `Ran N of M Specs` rather than only that CI is green.

### DIFF
--- a/test/e2e/slocontroller/batchresource.go
+++ b/test/e2e/slocontroller/batchresource.go
@@ -23,7 +23,6 @@ import (
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	k8spodutil "k8s.io/kubernetes/pkg/api/v1/pod"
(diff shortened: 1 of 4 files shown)


Aeirx:
/assign @ZiMengSheng

CI is green (23 checks). Assigning you rather than the bot's default since you wrote #3196 — two things in here differ from what the issue describes, and you are best placed to say whether they are what you wanted:

- **`cpunormalization.go` carries the same nil-map branch.** It is already on `main` and the issue does not list it. Because it sets two keys rather than one, the helper takes a `map[string]string` instead of the `key, data` pair sketched in the issue. Both existing call sites move onto it.
- **`rollbackSLOConfigData` had the defect on the rollback path too** — `newConfigMap.Data[k] = v` on a DeepCopy with no nil guard. Removed with its last caller.

If #3100 lands after this, it can call `ensureSLOConfigData` instead of adding the third copy.

I could not run the e2e specs — no Docker on this machine, so no kind cluster. Since the failing branch is the one kind does not reach anyway, the decision is made by a pure function with unit tests around it (the package had none). The manual `kubectl patch ... '{"data":null}'` check from your issue is in the PR description for anyone with a cluster.

Aeirx:
Friendly ping @ZiMengSheng — CI is green (all 23 checks pass) and this is just waiting on your review. Happy to adjust the two points I raised above (the \cpunormalization.go\ helper shape and the \
ollbackSLOConfigData\ removal) if you'd prefer a different approach.

PR #3209, opened 2026-09-11
scheduler: support distributing pod resources evenly across NUMA nodes

ZiMengSheng:

`NUMADistributeEvenly` was declared as a `NUMAAllocateStrategy` value but never implemented: the cpu accumulator only distinguishes `MostAllocated` from the other strategies, so a resolved `DistributeEvenly` silently fell back to the `LeastAllocated` ordering, and there was no per-pod way to ask for an even spread.

This PR implements `DistributeEvenly` and adds a **pod-level** `numaAllocateStrategy` field to the `koordinator.sh/numa-topology-spec` annotation, reusing the existing `NUMAAllocateStrategy` type instead of introducing a parallel enum.

**Why it matters.** On GPU machines the GPU pods are NUMA-aligned — they consume the memory local to the NUMA nodes their GPUs sit on. CPU / daemonset pods, by contrast, get packed into the fewest NUMA nodes, which drains memory on one side and starves the NUMA-aligned pods. Spreading a pod's cpu/memory across as many NUMA nodes as feasible keeps per-NUMA memory headroom balanced for the GPU pods that need it.

**How it works.**

1. **Effective strategy resolution** — `resolveNUMAAllocateStrategy` resolves **pod annotation > node label (`node.koordinator.sh/numa-allocate-strategy`) > plugin default**. A pod *overrides* the node-wide preference instead of being rejected on mismatch: unlike `NUMATopologyPolicy` (an admission constraint), the allocate strategy is a placement preference. At the pod level only `DistributeEvenly` is honored; PreFilter rejects any other value with a diagnosable status, since `MostAllocated`/`LeastAllocated` remain node-level strategies.
2. **Even spread at the hint layer** — when the effective strategy is `DistributeEvenly`, `generateResourceHints` marks the **widest** feasible NUMA masks as `Preferred` (narrower masks stay satisfiable as a non-preferred fallback), and `effectiveNUMATopologyPolicy` upgrades policy `None` to `BestEffort` so per-NUMA accounting and hint generation still run on nodes without a topology policy. This matches runtimes that interleave memory across NUMA nodes (e.g. `mbind` `MPOL_INTERLEAVE`).
3. **Gating** — the even spread applies only when the node actually exposes NUMA resources (has a `NodeResourceTopology`) and the pod is not confined to a single NUMA node. A topology-less node keeps its prior `None` behavior instead of being pushed into NUMA admission (which would hard-fail at Reserve), and a pod that requests `SingleNUMANode` policy or `SingleNUMANodeExclusive: Required` keeps its own intent rather than being silently spread by a node-wide `DistributeEvenly` label.
4. **Restricted/SingleNUMANode untouched** — the widest-mask hint flip is exempted under `Restricted` and `SingleNUMANode`, so even spread never changes their admission result. The effective policy is resolved through a single entry (`resolveNUMATopologyPolicy`) shared by Filter/Score/Reserve/hint generation so the gating can never be missed at a newly added call site.

Pod opt-in:

```json
{
  "numaAllocateStrategy": "DistributeEvenly"
}
```

NONE

- `Test_resolveNUMAAllocateStrategy` — the pod annotation takes precedence over the node label, and the node label over the plugin default.
- `Test_effectiveNUMATopologyPolicy` / `Test_resolveNUMATopologyPolicy` — `None` upgrades to `BestEffort` only for a distribute-evenly pod on a node that actually has NUMA resources; a topology-less node is not upgraded; `Restricted` and the pod's `SingleNUMANode` intent are preserved.
- `Test_isDistributeEvenly` — the gating: inert without `NUMANodeResources`, inert when the pod requests `SingleNUMANode`/`Required`, active for a pod opt-in or a node-wide label when NUMA resources exist.
- `TestDistributeEvenlyRestrictedAdmission` — under `Restricted`, the widest-mask flip is exempted and `topologymanager` still admits a pod whose per-resource widest masks differ (cpu `{0,1}` vs memory `{1}`).
- PreFilter — a pod-level `numaAllocateStrategy` other than `DistributeEvenly` is rejected as `UnschedulableAndUnresolvable`.
- Reserve — a distribute-evenly pod (or a node-wide label) on a node without NUMA topology does not hard-fail.
- `go build` / `go vet` / `gofmt` clean; `go test ./apis/extension/... ./pkg/scheduler/plugins/nodenumaresource/...` all pass.

## Ⅳ. Special notes for reviews

- The per-NUMA-node CPU *ordering* in `allocateCPUSet` is intentionally left as-is: the accumulator only special-cases `MostAllocated`, so a resolved `DistributeEvenly` uses the `LeastAllocated` ordering there. The even spread is realized at the NUMA hint/allocation layer, not by reordering CPUs within a node — a `NOTE` in the code records this.
- The pod-level field reuses `NUMAAllocateStrategy`, so `MostAllocated` / `LeastAllocated` / `DistributeEvenly` share one protocol at both the node and pod level; no new enum is introduced.
- The `None`→`BestEffort` upgrade is gated on the node having NUMA resources and the pod not being single-NUMA-confined, so it is a no-op for topology-less nodes and for pods that already pin to one NUMA node; the default path stays fully backward compatible.


saintube (commented):
/lgtm

PR #2948, opened 2026-05-14
fix: scheduler/deviceshare: fix nil map panic in device plugin adapter annotation/label writes

IITI-tushar:
### Ⅰ. Describe what this PR does

All five device plugin adapters wrote annotations and labels by indexing directly into the map returned by `object.GetAnnotations()` / `object.GetLabels()`, e.g.:

```go
// BEFORE — panics if the object has no annotations
object.GetAnnotations()[AnnotationBindTimestamp] = strconv.FormatInt(...)
object.GetLabels()[apiext.LabelHAMIVGPUNodeName] = ctx.node.Name
```

In Go, a nil map (returned when the object has no annotations/labels) will panic on write. While the current call-site in `preBindObject` happens to set annotations via `SetDeviceAllocations` first, the adapters themselves carry no such guarantee, making the code fragile and unsafe for direct calls (tests, future refactors).

Fix: use the safe pattern — get, nil-check, initialise, write, `SetAnnotations`/`SetLabels`:

```go
// AFTER — safe for objects with or without existing annotations
annotations := object.GetAnnotations()
if annotations == nil {
    annotations = map[string]string{}
}
annotations[AnnotationBindTimestamp] = strconv.FormatInt(dpAdapterClock.Now().UnixNano(), 10)
object.SetAnnotations(annotations)
```

Applied to all five adapters. No behavioural change when annotations/labels are already non-nil.

Unit tests were added to exercise adapters against objects with nil annotations and nil labels.

### Ⅱ. Does this pull request fix one issue?

fixes #2947 

### Ⅲ. Describe how to verify it

Run:

```bash
go test ./pkg/scheduler/plugins/deviceshare/... -race
```

Expected result:

* no panic/failure
* tests pass

### Ⅳ. Special notes for reviews

* Only the write pattern is changed; read patterns (`object.GetAnnotations()[key]` without assignment) are safe in Go and left unchanged.
* `SetAnnotations`/`SetLabels` update the in-memory object only; the scheduler's subsequent patch/update call sends the values to the API server.
* All five adapters follow the same new pattern for consistency.

### V. Checklist

- [x] I have written necessary docs and comments
- [x] I have added necessary unit tests and integration tests
- [x] All checks passed in `make test`

PR #3210, opened 2026-09-13
scheduler: fix nil map panic in deviceshare device plugin adapters

### LINKED ISSUE #2947: [BUG]: deviceshare/device plugin adapters panic with nil pointer when pod has no annotations or labels

KR-Ravindra:
#### Problem

All five device plugin adapters in `pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go` (`generalDevicePluginAdapter`, `generalGPUDevicePluginAdapter`, `huaweiGPUDevicePluginAdapter`, `cambriconGPUDevicePluginAdapter`, `metaxDevicePluginAdapter`) write directly into the map returned by `object.GetAnnotations()` / `object.GetLabels()`. When a Pod or Reservation reaches PreBind without annotations (or labels), the getter returns `nil` and the assignment panics with `assignment to entry in nil map`, crashing koord-scheduler.

Today this only works because `SetDeviceAllocations` happens to run before `adaptForDevicePlugin` and leaves annotations non-nil; the adapters themselves have no such contract.

#### Fix

Picks up the helper shape from #2950: adds `ensureAnnotations(object)` / `ensureLabels(object)` which initialize the map via `SetAnnotations` / `SetLabels` when nil and return it, and routes every annotation/label write in the five adapters through them.

fixes #2947 (https://github.com/koordinator-sh/koordinator/issues/2947)

#### How tested

Unit tests were added for each adapter with a Pod (and, for the general and Huawei adapters, a Reservation) that has nil annotations. Before the fix, each of the five test functions aborts:

```
--- FAIL: TestGeneralDevicePluginAdapter_Adapt/nil_annotations (0.00s)
panic: assignment to entry in nil map [recovered, repanicked]
```

(the same panic for `TestGeneralGPUDevicePluginAdapter_Adapt`, `TestHuaweiGPUDevicePluginAdapter_Adapt`, `TestCambriconGPUDevicePluginAdapter_Adapt`, `TestMetaXGPUDevicePluginAdapter_Adapt`). After the fix:

```
go test ./pkg/scheduler/plugins/deviceshare/...
ok  	github.com/koordinator-sh/koordinator/pkg/scheduler/plugins/deviceshare	10.717s
```

`gofmt` and `go vet` are clean on the package.

- The `LabelHAMIVGPUNodeName` write is only reachable when the HAMI label is already present, so labels can never actually be nil there; `ensureLabels` is still used so the write does not rely on that guard.
- Existing test cases are unchanged; only new table entries were added.

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### V. Checklist

- [x] All checks passed in `go test ./pkg/scheduler/plugins/deviceshare/...` (full `make test` not run locally)


KR-Ravindra:
Round 1 self-review.

Verified against `main`:
1. `pkg/scheduler/plugins/deviceshare/device_plugin_adapter.go` has nine direct `GetAnnotations()[k] = v` / `GetLabels()[k] = v` writes across the general, GPU, Huawei, Cambricon (two) and MetaX adapters; all nine now go through `ensureAnnotations` / `ensureLabels`, which call `SetAnnotations` / `SetLabels` when the map is nil so the write lands on the object for both `*corev1.Pod` and `*Reservation`.
2. Reachability today: `preBindObject` in `plugin.go` calls `apiext.SetDeviceAllocations` (which initialises annotations) before `adaptForDevicePlugin`, so the panic is not hit through `PreBind`; the description says so. `Adapt` is an exported interface, so callers outside that path are not protected, which is the case #2947 and #2950 describe.
3. The new table cases pass `&corev1.Pod{}` / `&schedulingv1alpha1.Reservation{}` with nil maps straight into `Adapt`; against the unpatched file each panics on the nil-map write, so they fail without the fix by construction.
4. CI: DCO, gofmt, govet, Go build, unit tests, license and codecov are green; only the image builds and E2E jobs are still running.

No duplicate: #2948 and #2950 were both closed on 2026-09-06 with an invitation for a fresh PR covering all sites; #2947 cross-references only those two and this PR. Marking ready.

saintube:
@ZiMengSheng

PR #3212, opened 2026-09-15
scheduler: fix deviceshare pre-allocation reservation stuck Pending on non-lowest GPU minor

ZiMengSheng:
## What type of PR is this?

/kind bug

A Restricted pre-allocation Reservation must be allowed to request more devices than its pre-allocatable Pod while still preserving every device minor already used by that Pod.

Passing `requiredDeviceResources` constrains the candidate device set to the original minors and prevents expansion, such as a two-GPU Pod pre-allocated by a four-GPU Reservation. Clearing it allows expansion, but the allocator may select different minors and fail the subsequent superset validation, leaving the Reservation permanently `Pending`.

## Fix

Introduce a separate `requiredDeviceMinors` must-include constraint without restricting the complete candidate device set:

- GPU partition allocation only accepts partitions containing every required minor.
- GPU topology allocation only considers scopes containing every required minor and prioritizes those minors.
- Shared GPU scoring is restricted to required minors when the constraint is present.
- A final allocator-level check verifies that every required minor is included.

Regression tests cover non-lowest-minor reuse, two-to-four GPU expansion, and shared GPU reuse when another device has a better score.

None

## Does this PR introduce a user-facing change?

```release-note
Fix pre-allocation Reservations staying Pending when preserving or expanding pre-allocated GPU devices.
```


ZiMengSheng:
Updated the fix to preserve expansion semantics. Pre-allocated GPU minors are now modeled as a must-include constraint without restricting the full candidate device set. GPU partition/topology allocation enforces the constraint, followed by a result-level superset check.

Added regression coverage for:
- reusing a non-lowest GPU minor (1 → 1)
- expanding two pre-allocated GPUs to a four-GPU Reservation (2 → 4)

`go test ./pkg/scheduler/plugins/deviceshare/...` passes.

ZiMengSheng:
Addressed the shared-GPU edge case raised during internal review. When a must-include constraint is present, shared GPU scoring now only considers required minors, so a better-scored unrelated GPU cannot replace the pre-allocated minor.

Added a regression case where minor 4 is partially shared and another GPU has a better score. It failed before the fix and now passes. The PR description has also been updated to match the final `requiredDeviceMinors` design.

`go test ./pkg/scheduler/plugins/deviceshare/...` passes.

PR #3219, opened 2026-09-20
refactor: migrate koordlet common metrics to component-base

mdryaaan:
> **Design interpretation up front, so it can be corrected early.** In the #3178 review @saintube asked to
> keep existing func and shared variable names untouched and land the `Registerable` path additively. Applied
> to the collector slices, this PR takes the additive reading: the migrated metrics go into a **new**
> `CommonRegisterableCollectors []k8smetrics.Registerable` registered through `internalKubeMustRegister`, and
> `CommonCollectors` keeps its existing name **and** `[]prometheus.Collector` type, now holding only the
> GC-wrapped `PodEvictionDetail`. The alternative — changing `CommonCollectors`' type under its existing name —
> is the same class of change that was asked to be avoided for the function. Cost: two slices per file during
> the migration, converging back to one once a file is fully moved. This was noted on #2971
> (https://github.com/koordinator-sh/koordinator/issues/2971#issuecomment-5753074569); if the type change was
> the intent instead, say so and I'll switch — easier now than after several files follow this pattern.

First per-file migration for #2971, following the registration split landed in #3178. `common.go`'s seven
plain metrics — `KoordletStartTime`, `CollectNodeCPUInfoStatus`, `CollectNodeNUMAInfoStatus`,
`CollectNodeLocalStorageInfoStatus`, `PodEviction`, `NodeUsedCPU`, `NodeUsedMemory` — move from
`prometheus.New*Vec` to `k8smetrics.New*Vec` with `StabilityLevel: k8smetrics.ALPHA`, and are registered via
`internalKubeMustRegister` so the component-base stability framework applies. This is that function's first
production caller. `PodEvictionDetail` stays on the raw path: it is a `pkg/util/metrics` GC wrapper, which
captures the underlying `*prometheus.MetricVec` at construction — nil for a component-base vec until
registration creates it — so migrating it would nil-deref in the GC goroutine after the first expiry.

`common.go` was chosen first because it contains both cases, so it sets the pattern for the remaining files.

All koordlet internal metrics start as ALPHA; promotion to BETA/STABLE requires a follow-up proposal. This
remains intentionally koordlet-only — `pkg/util/metrics/koordmanager/registry.go` has the same half-migration
and can get the same treatment later.

Part of #2971

`go test ./pkg/koordlet/metrics/... -count=1`

**The registry guard from #3178 caught this migration on its first run, before the tests were touched**, which
is the behaviour it was added for:

```
--- FAIL: TestInternalRegistryGathersAllInternalMetrics
    ... should have 24 item(s), but has 31
    Messages: a collector was added to an internal slice but not to this test
```

The seven metrics had left the raw slices; the count no longer matched the 31-name table. At the same time all
31 names still gathered, which is the evidence that the migration moves registration paths without adding,
removing or renaming any metric. The test now sums both slice families (raw and Registerable) and the name table
is unchanged.

The pre-existing `TestCommonCollectors` drives the migrated metrics through the `Record*` helpers and passes
unchanged. `go vet`, `gofmt` and `golangci-lint` (v2.11.3) are clean.

**Visible change on `/internal-metrics`: HELP text.** Component-base's `annotateStabilityLevel` prepends the
stability level to the HELP string of every metric it registers. Metric names and labels are unchanged, but the
HELP line differs for the migrated metrics:

| metric | before | after |
|---|---|---|
| `koordlet_start_time` | `the start time of koordlet` | `[ALPHA] the start time of koordlet` |
| `koordlet_node_used_cpu_cores` | `Number of cpu cores used by node in realtime` | `[ALPHA] Number of cpu cores used by node in realtime` |
| `koordlet_pod_eviction_detail` (not migrated) | `evict detail launched by koordlet` | unchanged |

This is standard component-base behaviour — kube-scheduler and kubelet metrics carry the same prefix — and it
will apply file by file as the remaining migrations land. Flagging it here so nobody has to find it by diffing
the endpoint.

**`resetCollectors` in the guard test gained the component-base cases.** Its `default` branch fails loudly on an
unhandled type (added in #3178 round 3 for exactly this moment), so the reset could not silently degrade when
the first component-base collectors appeared.

### V. Checklist


ZiMengSheng:
Reviewed the migration end to end.

Confirmed that all seven migrated metrics preserve their names, labels, and Gauge/Counter types. The visible HELP change is exactly the documented component-base behavior (`[ALPHA] ...`); I reproduced it from `InternalRegistry.Gather()` and found no in-repo dependency on the old HELP strings.

Keeping `PodEvictionDetail` on the raw Prometheus path is necessary: `newGCCounterVec` captures `vec.MetricVec` at construction time, while a component-base vec does not initialize that field until registration, which would give the GC path a nil metric vec.

The split registration lists do not double-register or omit a metric, and the guard test now counts and resets both collector families. I also re-ran:

```text
go test ./pkg/koordlet/metrics/... -count=1
go test -race ./pkg/koordlet/metrics/... -count=1
go build ./pkg/koordlet/metrics/...
```

All passed. No blocking issue found.

/lgtm
/approve

